PR #5, opened 2021-12-02
WIP: Try to fix a bug in erase

utaal:
This,

```
fn if_spec_cond(#[spec] i: int) -> u64 {
    let mut a: u64 = 2;
    if i == 3 {
        assert(i == 3);
    }
    a = a + 1;
    assert(a == 3);
    a
}
```

currently panics with 

```
Verifying root module
thread 'rustc' panicked at 'assertion failed: expect != Mode::Exec', rust_verify/src/erase.rs:641:17
```

This is an in-progress attempt at fixing the issue.

This test case still panics with `--compile`, with:

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', compiler/rustc_middle/src/hir/map/mod.rs:180:30
```


utaal:
Closing in favor of 07e0d687af7f8c653c4e0a8b6b49e6b444359f0a

PR #8, opened 2021-12-28
pcell

tjhance:
This is my first attempt at porting [Seagull's `LinearCell`](https://github.com/vmware-labs/verified-betrfs/blob/concurrency-experiments/concurrency/framework/Cells.s.dfy).

First of all, Rust already has a thing called `Cell`, so I needed a new name. I decided on `PCell` (for "permissioned cell").

`PCell` is implemented externally. There are two more types, `PermissionToken` ("proof") and `Permission` ("spec"; serves as a `view()` of `PermissionToken`). This feels a little tedious, but I wasn't sure what would happen if I tried to combine them into a single `proof` type:

```rust
#[proof]
pub struct PermissionToken<V> {
  pub pcell: int,
  pub value: option::Option<V>,
}
```

Are users allowed to modify the fields of a #[proof] struct at will? For this design to be sane, that needs to not be the case, but I wasn't sure.

In this PR, I implemented the `write` method. I wanted to make a constructor as well, but ran into a couple of issues:

 * First of all, it seems Verus doesn't yet support `impl` methods that don't take a `self` argument. This is easy enough to work around, though, since I can just put the constructor outside of `impl`.
 * The constructor needs to return a pair `(PCell<V>, PermissionToken<V>)`, where the left argument is #[exec] and the right argument is #[proof]. This also seemed to be unsupported, and is much harder or impossible to work around.

Finally, one last issue I had was that the implementation of `PCell` requires the use of `MaybeUninit`. Rust apparently implements `MaybeUninit` via a `union`, which Verus complained about... I put a fix in `rust_to_vir_base.rs`, but I don't know if that change was sensible.

### DIFF
--- a/source/pervasive/cell.rs
+++ b/source/pervasive/cell.rs
@@ -1,0 +1,113 @@
+use std::cell::UnsafeCell;
+use std::mem::MaybeUninit;
+
+#[allow(unused_imports)] use builtin::*;
+#[allow(unused_imports)] use crate::pervasive::*;
+
+//type Identifier = int;
+
+#[verifier(external_body)]
+pub struct PCell<V> {
+  ucell: UnsafeCell<MaybeUninit<V>>,
+}
+
+// TODO find a way to consolidate Permission and PermissionToken
+
+#[spec]
+pub struct Permission<V> {
+  pub pcell: int,
+  pub value: option::Option<V>,
+}
+
+#[verifier(external_body)]
+#[proof]
+pub struct PermissionToken<V> {
+  dummy: std::marker::PhantomData<V>,
+}
+
+pub struct PCellWithToken<V> {
+  pub pcell: PCell<V>,
+  #[proof] pub token: PermissionToken<V>,
+}
+
+impl<V> PermissionToken<V> {
+  #[verifier(pub_abstract)]
+  #[spec]
+  pub fn view(&self) -> Permission<V> {
+    arbitrary()
+  }
+}
+
+// TODO put these in impl:
+
+//// new_empty
+#[inline(always)]
+#[verifier(external_body)]
+fn new_empty_external<V>() -> PCell<V> {
+  ensures(|p: PCell<V>| false);
+  return PCell { ucell: UnsafeCell::new(MaybeUninit::uninit()) };
+}
+
+#[inline(always)]
+pub fn new_empty<V>() -> PCellWithToken<V> {
+  ensures(|pt : PCellWithToken<V>|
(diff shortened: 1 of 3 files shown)


Chris-Hawblitzel:
For the last issue, I just pushed a commit that should remove the need to change `rust_to_vir_base.rs`.

utaal:
This is cool!

> First of all, it seems Verus doesn't yet support impl methods that don't take a self argument. This is easy enough to work around, though, since I can just put the constructor outside of impl.

Yes, this is somewhere between an incomplete feature and a bug. I have a TODO for that.

> The constructor needs to return a pair (PCell<V>, PermissionToken<V>), where the left argument is #[exec] and the right argument is #[proof]. This also seemed to be unsupported, and is much harder or impossible to work around.

The `struct` you used seems like an ok workaround, but possibly tedious in the long run (although it's not uncommon to use structs for multi-returns in rust). We should discuss with @Chris-Hawblitzel whether we want multi-mode tuples.

utaal:
> Are users allowed to modify the fields of a #[proof] struct at will?

If the fields are pub and the user has ownership, they are!

```
impl<V> PermissionToken<V> {
   #[verifier(pub_abstract)] #[spec]
   pub fn view(&self) -> Permission<V> { arbitrary() }
 }
```

This is exactly what I'd have suggested for this issue!

---

```
pub struct Permission<V> {
   pub pcell: int,
   pub value: option::Option<V>,
 }
```

I was wondering if we should try and leverage the type system to pair a `Permission` and its `PCell`, by using a marker type or a branded lifetime (a-la-GhostCell), for example. We should be able to make fresh types by introducing a closure (each closure is its own type); or we may be able to use branded lifetimes but I'm unclear how/if GhostCell does that without having to enclose the code in one closure per lifetime, I need to read that code again.

Chris-Hawblitzel (approved):

tjhance:
> I was wondering if we should try and leverage the type system to pair a Permission and its PCell, by using a marker type or a branded lifetime (a-la-GhostCell), for example. We should be able to make fresh types by introducing a closure (each closure is its own type); or we may be able to use branded lifetimes but I'm unclear how/if GhostCell does that without having to enclose the code in one closure per lifetime, I need to read that code again.

What additional value would we get from this, given that we already use the unique IDs and SMT equality check?

utaal:
> What additional value would we get from this

Actually I take it back. I was thinking it may be good to get a type error, instead of a verification error (and wouldn't even have to run z3), but it would likely make vectors (or similar) of PCells painful (as they'd all have a different type). So, nevermind.

PR #9, opened 2022-01-03
Support named lifetimes in function declarations

utaal:
The borrow-checker is in charge of guaranteeing that the reference's lifetime is tied to an immutable borrow of the referenced object. Thus we can treat the reference as an immutable copy of the current value. I believe this is safe. @Chris-Hawblitzel can you think of anything that this would allow and that's unsound?

cc @tjhance


utaal:
Discussed offline with @Chris-Hawblitzel .

PR #10, opened 2022-01-03
Revert accidental change in d54340188

utaal:
Spotted by @tjhance , confirmed to be accidental by @yizhou7 

@Chris-Hawblitzel I'm unfamiliar with the assigned sets: can you think of a test that would have caught this? (I'd like to add a regression test if we can)


Chris-Hawblitzel (approved):
Yes, it would be good to test variable assignment inside while loops:
- variables that get assigned should be havoc'd
- variables that don't get assigned should have their original value
- variables that are declared inside the while loop should not cause any changes in behavior to variables outside the while loop

PR #16, opened 2022-01-05
initial work to support `assert_bit_vector` primitive

yizhou7:
`assert_bit_vector` is added as a primitive: a separate SMT query is made using bitvector theory, and a assume with integer theory is put in place of the original assertion. Currently only supports very limited bitvector operations, but can and will be expanded to a larger set.

### DIFF
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -52,6 +52,8 @@ pub enum BinaryOp {
     EuclideanDiv,
     EuclideanMod,
 
+    UintXor,
+
     BitXor,
     BitAnd,
     BitOr,
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -33,7 +33,7 @@ pub enum TypX {
 pub enum Constant {
     Bool(bool),
     Nat(Arc<String>),
-    // BitVec(Arc<String>, u32),
+    BitVec(Arc<String>, u32),
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -53,12 +53,11 @@ pub enum BinaryOp {
     EuclideanMod,
 
     UintXor,
-
     BitXor,
+
     BitAnd,
     BitOr,
     BitAdd,
-
     Shr,
     Shl,
 }
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -33,6 +33,7 @@ pub enum TypX {
 pub enum Constant {
     Bool(bool),
     Nat(Arc<String>),
+    // BitVec(Arc<String>, u32),
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
@@ -50,9 +51,12 @@ pub enum BinaryOp {
     Gt,
     EuclideanDiv,
     EuclideanMod,
+
     BitXor,
     BitAnd,
     BitOr,
+    BitAdd,
+
     Shr,
     Shl,
 }
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -49,6 +49,11 @@ pub enum BinaryOp {
     Gt,
     EuclideanDiv,
     EuclideanMod,
+    BitXor,
+    BitAnd,
+    BitOr,
+    Shr,
+    Shl,
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -26,6 +26,7 @@ pub enum TypX {
     // Lambda deliberately omits argument, return types to make box/unbox for generics easier
     Lambda,
     Named(Ident),
+    BitVec(u32),
 }
 
 #[derive(Clone, PartialEq, Eq, Hash)] // for Debug, see ast_util
--- a/source/air/src/ast.rs
+++ b/source/air/src/ast.rs
@@ -52,7 +52,6 @@ pub enum BinaryOp {
     EuclideanDiv,
     EuclideanMod,
 
-    UintXor,
     BitXor,
 
     BitAnd,
--- a/source/docs/bv_expriments/bv_examples.rs
+++ b/source/docs/bv_expriments/bv_examples.rs
@@ -2,48 +2,52 @@ use builtin::*;
 mod pervasive;
 use pervasive::*;
 
-pub fn concat64() ->  {
-    unimplemented!();
-}
+// pub fn concat64() ->  {
+//     unimplemented!();
+// }
 
-#[spec]
-fn modulo_8(i: int) -> int {
-	i % 8
-}
+// #[spec]
+// fn modulo_8(i: int) -> int {
+// 	i % 8
+// }
 
 #[exec]
+// #[verifier(bit_vector)]
 fn and_7(b: u32) -> u32 {
-	ensures(|ret: u32| ret == modulo_8(b));
-	b & 7
-}
-
-#[exec]
-fn bv128_to_u64(b: bv128) -> (u64, u64){
-   let lower_mask:bv128 = 0x00000000000000001111111111111111;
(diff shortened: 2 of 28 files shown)


source/air/src/ast.rs
```diff
@@ -49,6 +51,14 @@ pub enum BinaryOp {
     Gt,
     EuclideanDiv,
     EuclideanMod,
+
+    BitXor,
```
parno, line 55: Is `BitXor` being especially highlighted with extra space, or is this just an unintended artifact?  Same thing seems to happen in a few places below, e.g., in `source/air/src/printer.rs`.

source/docs/bv_expriments/bv_examples.rs
```diff
@@ -0,0 +1,59 @@
+use builtin::*;
```
parno, line 1: Eventually this file should migrate into some combination of `source/rust_verify/example/` and `source/rust_verify/tests`.

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -207,6 +207,8 @@ pub(crate) enum Attr {
     Trigger(Option<Vec<u64>>),
     // custom error string to report for precondition failures
     CustomReqErr(String),
+    // verify using bitvector theory, not converting back and forth with integer
```
parno, line 210: I'm not sure that this bit:

> , not converting back and forth with integer

will be clear to future readers.  Can we improve or remove it?

source/vir/src/ast.rs
```diff
@@ -340,6 +348,8 @@ pub struct FunctionAttrsX {
     pub custom_req_err: Option<String>,
     /// coerce f(e, ...) to f(e.view(), ...)
     pub autoview: bool,
+    /// Verify using bit vector theory, not converting back and forth with integer
```
parno, line 351: Same comment as above.

source/vir/src/func_to_air.rs
```diff
@@ -369,6 +369,9 @@ pub fn func_decl_to_air(
                 let axiom = Arc::new(DeclX::Axiom(expr));
                 decl_commands.push(Arc::new(CommandX::Global(axiom)));
             }
+            if function.x.attrs.bit_vector {
+                // TODO: function level bitvector mode
```
parno, line 373: Should we have an `unimplemented!()` here?

source/vir/src/prelude.rs
```diff
@@ -275,6 +277,10 @@ pub(crate) fn prelude_nodes() -> Vec<Node> {
             (<= 0 ([height] x))
             :pattern (([height] x))
         )))
+
+        // BitVector
+        // these are the uninterpreted integer versions
+        (declare-fun [uint_xor] (Int Int) Int)
```
parno, line 283: Should we only include these when the program contains bit-vector operations, so we don't bloat the prelude when the program doesn't?

source/vir/src/sst_to_air.rs
```diff
@@ -285,6 +289,8 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp) -> Expr {
                         BinaryOp::Mul => panic!("internal error"),
                         BinaryOp::EuclideanDiv => air::ast::BinaryOp::EuclideanDiv,
                         BinaryOp::EuclideanMod => air::ast::BinaryOp::EuclideanMod,
+                        BinaryOp::BitXor => panic!("internal error"),
```
parno, line 292: Can we expand the "internal error" into something more detailed, so the person who sees it has a bit more context about what went wrong?

source/vir/src/sst_to_air.rs
```diff
@@ -452,6 +445,60 @@ fn one_stmt(stmts: Vec<Stmt>) -> Stmt {
...
 
+// convert the sst expression into bv air expression
+fn exp_to_bv_expr(state: &State, exp: &Exp, parent_width: u32) -> (Expr, u32) {
+    let mut parent_width = parent_width;
+
+    match &exp.x {
+        ExpX::Const(crate::ast::Constant::Nat(s)) => {
+            assert!(parent_width != 0);
+            // Nat constant will get an inferred bitwidth from the parent
+            // the width is needed when printing bv constants
```
parno, line 456: Just to make this clearer, can we add a period at the end of the last line and capitalize "The width..."?

tools/update-rust.sh
```diff
@@ -14,7 +14,7 @@ fi
 
 (
     cd rust
-    current_branch=`git branch --show-current`
+    current_branch=`git name-rev --name-only HEAD`
```
parno, line 17: Is this related to the BV changes?  If not, perhaps it should be handled separately?

parno (commented):
Some mostly minor questions/suggestions.

source/air/src/ast.rs
yizhou7: Oh it was somewhat intended to highlight the added operations due to this change. We could remove it if needed.

source/docs/bv_expriments/bv_examples.rs
yizhou7: Yes, currently it is still in a primitive state, but we should do that once we gather more use cases.

source/rust_verify/src/rust_to_vir_base.rs
yizhou7: Yes, I will just remove it. We will make sure the bit vector encoding is documented somewhere, along with a paper pencil proof of soundness.

source/vir/src/prelude.rs
yizhou7: Oh yeah one related point, we might need access to some none rust native bv operations. For example we might need bit selection, bv concatenation. We might want to group them into a separate module? Then if that module is imported, we can add these declarations into the prelude, and in the normal case it just won't be there. Thoughts?

source/vir/src/sst_to_air.rs
yizhou7: Oh this one is should be an unreachable error, because we just had a clause that handles the translation. 
https://github.com/secure-foundations/verus/blob/e8058d8fe32c79c83e172722fddd338fa4627f24/source/vir/src/sst_to_air.rs#L273

source/vir/src/sst_to_air.rs
yizhou7: Sure.

tools/update-rust.sh
yizhou7: Ah yes, this is not related to the bv change. I will remove this from the pull request. I have been running these on the server, but I don't have the privilege to upgrade the git version. The old git does not recognize the command there.

source/vir/src/prelude.rs
Chris-Hawblitzel: If you decide to implement this no-bloat optimization (no hurry), you can do it on a module-by-module basis, since Verus verifies one module at a time.  In this case, you could just move the code above to a different function in `prelude.rs` that gets called on demand for the modules that need it.

source/vir/src/sst_to_air.rs
Chris-Hawblitzel: I think you don't need `parent_width` at all, since `exp.typ` will give you the width.

source/vir/src/func_to_air.rs
utaal: Is this code path reachable? If so, and this is a placeholder for an upcoming feature, I'd much rather return error instead of crashing, with something like:
```
pub(crate) fn unsupported_err_span<A>(span: Span, msg: String) -> Result<A, VirErr> {
```
( https://github.com/secure-foundations/verus/blob/98d4c5bee9db2dfbe7e1a02bafac4258f51bdd28/source/rust_verify/src/util.rs#L34-L39 )

which would need to be moved/replicated in this crate.

yizhou7: I think this code is only reachable when `#[verifier(bit_vector)]` attribute is specified. Currently we don't have test cases developed for that just yet.

source/vir/src/sst_to_air.rs
yizhou7: Ah good to know, that would actually simplify things a lot here.

yizhou7:
The changes here are Incorporated in #24 .

PR #19, opened 2022-01-07
introduce open_invariant

tjhance:
This introduces the `Invariant<V>` proof type and the `open_invariant` feature.

`Invariant<V>` (defined in `pervasive/invariants.rs`) is defined to have two methods:

```rust
#[proof]
impl<V> Invariant<V> {
    #[verifier(pub_abstract)]
    #[spec]
    pub fn inv(&self, _v: V) -> bool

    #[verifier(pub_abstract)]
    #[spec]
    pub fn namespace(&self) -> int
}
```

The `inv`, of course, specifies what the invariant actually is, while `namespace` helps us ensure we don't open more than one invariant at once.

The `open_invariant` feature is used like this:

```rust
  #[proof] let i : Invariant<V> = ...

  open_invariant!(&i => inner => {
    // assume that i.inv(inner) holds at the beginning
    // ...
    // assert that i.inv(inner) holds at the end
  });
```

### Rust -> VIR

The `open_invariant!` macro is defined as follows:

```rust
#[macro_export]
macro_rules! open_invariant {
    ($eexpr:expr => $iident:ident => $bblock:block) => {
        #[verifier(invariant_block)] {
            #[allow(unused_mut)] let (guard, mut $iident) = crate::open_invariant_begin($eexpr);
            $bblock
            crate::open_invariant_end(guard, $iident);
        }
    }
}
```

When Verus sees the `#[verifier(invariant_block)]` attribute, it checks that the AST is EXACTLY the code that would be generated by the macro, to make sure the user doesn't cheat. It then converts it into an `OpenInvariant` node in VIR.

### Modes

The `Invariant<V>` must of course be a `proof` variable. Right now, all code inside the invariant block is expected to be `proof` code, although later we will add the ability to have a single atomic exec instruction. We check the control flow to make sure it is not possible to early-exit out of an invariant block.

### Verification conditions

Generating asserts and assumes for `i.inv(inner)` is the easy part.

The more interesting part is to make sure the user does not open the same invariant twice. It _is_ occasionally useful to open more than one invariant at once. In LinearDafny, this ability is very restricted: nested invariants must be right next to each other. To enforce this, we also require that invariants may only be opened in a non-ghost method, and only ghost code can be inside the invariant (so you can't have nested invariant across method boundaries).

To support nested invariants more generally, the user can declare on a function what invariants it may open. The default is that an `exec` fn can open any invariant, while a `proof` fn cannot open any invariant. However, the user can override this with `opens_invariants_any()` or `opens_invariants_none()` in the header of a function.

I also added headers for `opens_invariants(..)` and `opens_invariants_except(..)` so the user can specify an arbitrary set. However, these aren't implemented right now.

See `vir/src/inv_masks.rs` for details on the VC generation. I aimed to not have it generate any extra VCs in the common case.

### Error messages

Here are some example error messages (rust_verify/example/invariants.rs):

```
Verifying root module
error: possible invariant collision
  --> rust_verify/example/invariants.rs:37:20
   |
37 |   open_invariant!(&i => inner => {
   |                    ^ this invariant
38 |     open_invariant!(&i => inner2 => {
   |                      ^ might be the same as this invariant

error: Cannot show invariant holds at end of block
  --> rust_verify/example/invariants.rs:57:34
   |
57 |     open_invariant!(&i => inner => {
   |  __________________________________^
58 | |     inner = 1;
59 | |   });
   | |___^

error: callee may open invariants disallowed at call-site
  --> rust_verify/example/invariants.rs:94:20
   |
94 |   open_invariant!(&i => inner => {
   |                    ^ invariant opened here
95 |     callee_mask_full(); // ERROR
   |     ^^^^^^^^^^^^^^^^^^ might be opened again in this call

error: Cannot show invariant namespace is in the mask given by the function signature
   --> rust_verify/example/invariants.rs:100:20
    |
100 |   open_invariant!(&i => inner => { // ERROR
    |                    ^ invariant opened here

error: aborting due to 4 previous errors; 1 warning emitted
```

### Outstanding issues

 - The way that the `Invariant<V>` type, as well as the `inv` and `namespace` functions, is kind of jank.
   - The SST -> AIR conversion actually hardcode the names of the `inv` and `namespace` functions (which are in `pervasive`), but arguably `Invariant` should be a special type in VIR, with `inv` and `namespace` being special nodes, so that the VIR is self-contained, and so that all name-hardcoding happens in the rust -> VIR conversion.
   - ~There's a bug where the AIR function declarations for the `inv` and `namespace` don't even get generated if the user doesn't explicitly reference them.~ This is now fixed.

Things to handle in later PRs:

 - Need to implement the more general headers for specifying which invariants a function might open.
 - Need to add a notion of an atomic instruction.


source/pervasive/invariants.rs
```diff
@@ -0,0 +1,80 @@
...
+    unimplemented!();
+}
+
+#[macro_export]
+macro_rules! open_invariant {
+    ($eexpr:expr => $iident:ident => $bblock:block) => {
+        #[verifier(invariant_block)] {
+            #[allow(unused_mut)] let (guard, mut $iident) = crate::open_invariant_begin($eexpr);
+            $bblock
+            crate::open_invariant_end(guard, $iident);
```
utaal, line 77: Why aren't these `crate::pervasive::invariants::open_invariant_begin` / `crate::pervasive::invariants::open_invariant_end` ? Is it possible that you are relying on this behaviour https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3fe821824832fae9cec28092d6d2c05e in the tests and examples? I.e. importing `invariants::*` at the crate root makes them appear to be at the crate root.

source/pervasive/mod.rs
```diff
@@ -59,3 +52,11 @@ pub fn unreached<A>() -> A {
...
     println!("{}", i);
 }
+
+/// In spec, all types are inhabited
+#[spec]
+#[verifier(external_body)]
+#[allow(dead_code)]
+pub fn arbitrary<A>() -> A {
+    unimplemented!()
+}
```
utaal, line 62: This seems like a spurious change. Can you revert it?

source/rust_verify/example/invariants.rs
```diff
@@ -0,0 +1,169 @@
+// rust_verify/tests/example.rs expect-failures
+
+#[allow(unused_imports)]
+use builtin::*;
+mod pervasive;
+use pervasive::*;
+use crate::pervasive::{invariants::*};
+
+/*
```
utaal, line 9: I'd suggest adding a note about why this code and some of the code later on is commented out. If it's not relevant anymore, consider removing it.

source/rust_verify/example/invariants.rs
```diff
@@ -0,0 +1,169 @@
+// rust_verify/tests/example.rs expect-failures
```
utaal, line 1: This may not make sense here (e.g. if it's important to showcase the errors), but consider moving the failing cases to separate tests: with `expect-failures` we're less likely to notice if the examples drift and become incorrect, becausse the unexpected errors will not cause a test failure.

source/rust_verify/src/erase.rs
```diff
@@ -826,6 +832,96 @@ fn erase_stmt(
...
+    let mut result_stmts = Vec::new();
+
+    // First statement:
+    // let (guard, mut ident) = open_invariant_begin(arg);
+
+    // first, pattern match to get at the arg
+
+    let open_stmt = &block.stmts[0];
+    match &open_stmt.kind {
+        StmtKind::Local(local) => {
```
utaal, line 851: When you have only one pattern, you can also use

```
if let StmtKind::Local(local) = &open_stmt.kind {
  ...
} else {
  panic!(...)
}
```

No need to change this, just a heads up.

source/rust_verify/src/erase.rs
```diff
@@ -826,6 +832,96 @@ fn erase_stmt(
...
+            match &**local.init.as_ref().expect("erase_inv_block") {
```
utaal, line 852: I'd recommend a slightly more descriptive error message for `expect`.

source/rust_verify/src/erase.rs
```diff
@@ -826,6 +832,96 @@ fn erase_stmt(
...
+                        }));
+                        let id = open_stmt.id;
+                        let span = open_stmt.span.clone();
+                        result_stmts.push(Stmt { id, kind, span });
+                    } else {
+                        match expr_opt {
+                            None => {
+                                // expr was erased; just erase the whole statement
+                            }
+                            Some(expr) => {
```
utaal, line 882:
```
if let Some(expr) = expr_opt {
```
is a common pattern, fyi (again no need to change it here).

source/rust_verify/src/erase.rs
```diff
@@ -826,6 +832,96 @@ fn erase_stmt(
...
+        }
+        _ => {
+            panic!(
+                "erase_inv_block: the checks in invariant_block_to_vir should ensure this match statement succeeds"
+            );
+        }
+    }
+
+    // Middle statement: the user-supplied block
+    match erase_stmt(ctxt, mctxt, expect, &block.stmts[1], false) {
```
utaal, line 907: But please consider using `if let Some(stmt) = erase_stmt(...)` here!

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -294,6 +297,10 @@ fn fn_call_to_vir<'tcx>(
     let is_ensures = f_name == "builtin::ensures";
     let is_invariant = f_name == "builtin::invariant";
     let is_decreases = f_name == "builtin::decreases";
+    let is_opens_invariants_none = f_name == "builtin::opens_invariants_none";
+    let is_opens_invariants_any = f_name == "builtin::opens_invariants_any";
+    let is_opens_invariants = f_name == "builtin::opens_invariants";
+    let is_opens_invariants_except = f_name == "builtin::opens_invariants_except";
```
utaal, line 303: Fwiw, there is a more robust way of doing this that doesn't rely on path strings (https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/lang_items/index.html), I've just never gotten around to it yet.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -312,12 +319,27 @@ fn fn_call_to_vir<'tcx>(
...
+        || is_opens_invariants_none
+        || is_opens_invariants_any
+        || is_opens_invariants
+        || is_opens_invariants_except;
     let is_quant = is_forall || is_exists;
     let is_directive = is_hide || is_reveal || is_reveal_fuel;
     let is_cmp = is_equal || is_eq || is_ne || is_le || is_ge || is_lt || is_gt;
     let is_arith_binary = is_add || is_sub || is_mul;
 
+    if f_name == BUILTIN_INV_BEGIN || f_name == BUILTIN_INV_END {
```
utaal, line 336: Is this relying on the fact that the macro uses these functions when imported at the root of the crate? If that's the case, that may be brittle.

Edit: it's not. This code path isn't reached when using it correctly because the calls are matched by the `invariant_block_to_vir` directly.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -363,6 +385,39 @@ fn fn_call_to_vir<'tcx>(
         let header = Arc::new(HeaderExprX::Requires(Arc::new(vir_args)));
         return Ok(mk_expr(ExprX::Header(header)));
     }
+    if is_opens_invariants || is_opens_invariants_except {
+        return err_span_str(
+            expr.span,
+            "'is_opens_invariants' and 'is_opens_invariants_except' are not yet implemented",
+        );
+
+        /*
```
utaal, line 394: If you're planning to merge this soon (which I recommend to avoid it diverging from `main`, which would make conflict resolution more complex), can you please remove this block comment?

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -780,6 +835,199 @@ pub(crate) fn pattern_to_vir<'tcx>(
     Ok(pattern)
 }
 
+/// Check for the #[verifier(invariant_block)] attribute
+
```
utaal, line 839: Very minor: usually there isn't a newline between the docstring and the function.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -780,6 +835,199 @@ pub(crate) fn pattern_to_vir<'tcx>(
     Ok(pattern)
 }
```
utaal, line 837: Please consider moving the following (till 1030) to a new file, e.g. `rust_to_vir_invariant.rs`: it's already factored out, I believe, so it should mostly be a matter of copying over the imports (if it's more cumbersome than I think, let me know!).

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -780,6 +835,199 @@ pub(crate) fn pattern_to_vir<'tcx>(
...
+    // `match` statements below are for.
+    //
+    // We also need to "recover" the $inner, $eexpr, and $bblock for converstion to VIR.
+    //
+    // If the AST doesn't look exactly like we expect, print an error asking the user
+    // to use the open_invariant! macro.
+
+    let body = match &expr.kind {
+        ExprKind::Block(body, _) => body,
+        _ => {
```
utaal, line 888: Very minor, but you _should_ just be able to do
```
_ => panic!("invariant_block_to_vir called with non-Body expression"),
```
without the braces.

source/rust_verify/tests/open_invariant.rs
```diff
@@ -0,0 +1,328 @@
...
+          let mut idx = 0;
+          while idx < 5 {
+            open_invariant!(&i => inner => {
+              break;
+            });
+          }
+        }
+
+    } => Err(err) => assert_vir_error(err)
+}
```
utaal, line 328: Great test suite!

source/vir/src/early_exit_cf.rs
```diff
@@ -0,0 +1,161 @@
+use crate::ast::{ArmX, CallTarget, Expr, ExprX, Stmt, StmtX, VirErr};
+use air::ast::Span;
+use air::errors::error;
+
+#[derive(Copy, Clone, Debug)]
+enum StatementType {
+    Return,
+    //Break,
```
utaal, line 8: Are these commented out because they're currently unsupported? Can you add a TODO explaining that?

source/vir/src/early_exit_cf.rs
```diff
@@ -0,0 +1,161 @@
...
+    let mut v = Vec::new();
+    expr_get_early_exits_rec(expr, false, &mut v);
+    return v;
+}
+
+/// While recursing, we keep track of whether we entered a loop or not; then we can know
+/// if a break/continue would cause an exit at the high-level expr.
+/// (Well, it will be useful when we implement break/continue, anyway.)
+
+fn expr_get_early_exits_rec(expr: &Expr, in_loop: bool, results: &mut Vec<EarlyExitInst>) {
```
utaal, line 43: In a separate PR, I'm introducing a generic expr visitor that should allow to write these without having to manually recurse. If we think it's worthwhile, we can port this to that later.

source/vir/src/modes.rs
```diff
@@ -340,6 +341,25 @@ fn check_expr(typing: &mut Typing, outer_mode: Mode, expr: &Expr) -> Result<Mode
             }
             Ok(mode)
         }
+        ExprX::OpenInvariant(inv, binder, body) => {
+            assert_no_early_exit_in_inv_block(&body.span, body)?;
```
utaal, line 345: This may belong to `well_formed`, unless this check is required for the following mode code.

utaal:
This is very cool! Thanks for the extensive PR description too, it's very helpful! Things looks reasonable to me, and you're the expert on the VCs here. Most of my comments are minor.

> The SST -> AIR conversion actually hardcode the names of the inv and namespace functions (which are in pervasive), but arguably Invariant should be a special type in VIR, with inv and namespace being special nodes, so that the VIR is self-contained, and so that all name-hardcoding happens in the rust -> VIR conversion.

Are you planning to address this in this PR, or later?

I still need to wrap my head around the `open_invariants_*` headers: I'll get back to those in a bit. Maybe @Chris-Hawblitzel has more context for those and may be able to review them more effectively.

tjhance:
> > The SST -> AIR conversion actually hardcode the names of the inv and namespace functions (which are in pervasive), but arguably Invariant should be a special type in VIR, with inv and namespace being special nodes, so that the VIR is self-contained, and so that all name-hardcoding happens in the rust -> VIR conversion.
>
> Are you planning to address this in this PR, or later?

I was looking for input on what the right thing to do is, & whether it's important to fix.

source/pervasive/invariants.rs
tjhance: are you saying that another module importing this one won't work unless they import `invariants::*`? I thought macros were supposed to do the right thing with identifier lookups. Else how could the macro possibly work across crates?

(i don't know what i'm talking about, please set me straight)

source/vir/src/early_exit_cf.rs
tjhance: i wanted to make sure that if somebody adds a feature (e.g., Continue) they'll be forced to update the `match` statement.

source/rust_verify/example/invariants.rs
tjhance: FWIW, all these examples are in the `tests/` directory as well, which is obviously more robust. I made the examples file here because it was easier to experiment & mess around.

Is there an easy way to run all the tests in a test file and get their output (even when the tests pass)? If so, it would be much easier to experiment in the main test file, instead. Then I'd leave the `examples` file for proper, working examples.

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: i'll add a comment as explanation

source/vir/src/early_exit_cf.rs
utaal: You may be able to do this with the new visitor, but that's an important point that I'll keep in mind. (And to be clear, I'm not opposed to keep this as separate code if it helps, was just looking for opportunities to reduce the number of visitor implementations.)

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: i don't mind too much, but I do feel like it belongs in this file (it is part of the large mutual recursion on expressions & statements)

tjhance:
thanks to chris pointing out that spec functions are actually supported as parameters, I added a constructor for `Invariant` (see the example in `main` in `rust_verify/example/invariants.rs`)

source/vir/src/early_exit_cf.rs
Chris-Hawblitzel: A good balance would be to use the generic expr/stmt visitor infrastructure, and when using this for the early-exits visitor, implement a minimal visit that just skims each case and does nothing:
```
  ExprX::Tuple(..) => {}
  ExprX::Ctor(..) => {}
...
```
This way we catch any new features that get added, but we don't have to manually implement the recursion into the subexpressions for each case.

utaal: Yep, that’s what I had in mind too!

source/rust_verify/src/rust_to_vir_expr.rs
utaal: Makes sense!

source/rust_verify/example/invariants.rs
utaal: I generally work in the test file, selecting a specific test with `tools/cargo.s test -p rust_verify --test invariants -- --nocapture substring_match_of_the_test_fn`. You can also ask for the full backtrace and to log IRs (which only makes sense for a single test), with `VERIFY_LOG_IR_PATH=log/ RUST_BACKTRACE=1 ./tools/cargo.sh test -p rust_verify --test refs -- --nocapture arg_spec` for example.

> Then I'd leave the examples file for proper, working examples.

This would be nice!

source/rust_verify/example/invariants.rs
utaal: Tests instead of the example for failure cases, as discussed in the other comment, if you can. Otherwise just add a comment to say why they're commented out. Not a big deal, though.

source/pervasive/invariants.rs
utaal: Those functions' canonical paths are `crate::pervasive::invariants::open_invariant_begin` and `crate::pervasive::invariants::open_invariant_end` I believe, and _I believe_ this macro only works if you're importing `invariants::*` **and you are at the crate root** (which is the case for all examples and tests, I think). Try the canonical path instead? It should just work.

> Else how could the macro possibly work across crates?

The way `pervasive` is included is a hack currently. It's imported as if it was source in the current crate's `crate::pervasive` module. If you were writing a macro to be used outside the crate, you'd have to use `crate_name::canonical::path` instead, I think. The issue is that macros are expanded before resolution, so they cannot perform identifier lookups. Ping me on Slack if this doesn't make sense yet.

utaal:
Regarding the path hardcoding issue:

> but arguably Invariant should be a special type in VIR, with inv and namespace being special nodes, so that the VIR is self-contained, and so that all name-hardcoding happens in the rust -> VIR conversion.

I may be misunderstanding something here, but you'd still need a name for these in AIR, right? It could be something that doesn't necessarily match the path of the functions in builtin, though. Does @Chris-Hawblitzel have a preference?

source/vir/src/early_exit_cf.rs
tjhance: oh, one problem with this though is that we track extra info in the recursion with the `in_loop` variables (this variable is currently unused, but I wrote this with the expectation we'll eventually need to handle `break` and `continue`). So it can't be done with the generic visitor :\

tjhance:
I tried cleaning up `examples.rs` but ended up getting a proof failure. @Chris-Hawblitzel i could use your help figuring out why this is happening.

The parts relevant to the problem are:

`pervasive/invariants.rs`
```rust
#[proof]
#[verifier(external_body)]
#[verifier(returns(proof))]
pub fn invariant_new<V, F: Fn(V) -> bool>(#[proof] v: V, #[spec] inv: F, #[spec] ns: int) -> Invariant<V> {
  requires([
    inv(v),
  ]);
  ensures(|i: Invariant<V>|
    forall(|v: V| i.inv(v) == inv(v))
    && equal(i.namespace(), ns)
  );

  unimplemented!();
}
```

and

`rust_verify/examples/invariants.rs`
```rust
pub fn main() {
  #[proof] let u: u32 = 5;

  #[proof] let i = invariant_new(u, |u: u32| u % 2 == 1, 0);

  open_invariant!(&i => inner => {
    assert(i.inv(inner)); // this passes
    assert(inner % 2 == 1); // this FAILS (incorrectly)
  });
}
```

Here's a pastebin with the AIR file, with the relevant assert marked: https://pastebin.com/mrBgcKVR

It seems like the assert ought to follow from the previous one, and from the ensures, `ens%pervasive.invariants.invariant_new.`. Is it a problem with triggering or something?

tjhance:
this might help:

```rust
pub fn main() {
  #[proof] let u: u32 = 5;

  #[spec] let the_fn = |u: u32| u % 2 == 1;
  #[proof] let i = invariant_new(u, the_fn, 0);

  open_invariant!(&i => inner => {
    assert(i.inv(inner));
    assert(the_fn(inner)); // this one fails
    assert(inner % 2 == 1);
  });
}
```

Chris-Hawblitzel:
It looks like the AIR code for this example is missing an `(assume (uInv 32 inner@))`.  This should come from `typ_invariant`.

tjhance:
thanks, the typ_invariant issue should now be fixed

source/vir/src/early_exit_cf.rs
Chris-Hawblitzel: The current generic visitor uses client-defined state (`env`) to handle this; the new generic visitor probably should, too.

Chris-Hawblitzel (approved):

source/vir/src/ast_visitor.rs
```diff
@@ -240,6 +240,10 @@ where
                         }
                     }
                 }
+                ExprX::OpenInvariant(inv, _binder, body) => {
+                    expr_visitor_control_flow!(expr_visitor_dfs(inv, map, mf));
+                    expr_visitor_control_flow!(expr_visitor_dfs(body, map, mf));
```
utaal, line 245: My understanding is that the `binder` is relevant to the `body` (it's the "inner" introduced by `open_invariant!(&i => inner => {`), so I think this should be:

```
ExprX::OpenInvariant(inv, binder, body) => {
    expr_visitor_control_flow!(expr_visitor_dfs(inv, map, mf));
    map.push_scope(true);
    let _ = map.insert(binder.name.clone(), binder.a.clone());
    expr_visitor_control_flow!(expr_visitor_dfs(body, map, mf));
    map.pop_scope();
}
```

utaal: Similarly to https://github.com/secure-foundations/verus/blob/cc89012c5c73be6cb2493259e83854f088efd30f/source/vir/src/ast_visitor.rs#L434-L442

PR #20, opened 2022-01-07
tweaks to Error builder functions

tjhance:
Just some small changes to the Error builder functions for consistency, and a bit more documentation. Also, use `Into<String>` (I was started to get annoyed having to type `to_string()` all the time or handle multiple method names.)

### DIFF
--- a/source/air/src/errors.rs
+++ b/source/air/src/errors.rs
@@ -41,7 +41,7 @@ pub type Error = Arc<ErrorX>;
 // To build an error, use one of the constructors below:
 
 /// Basic error, with a message and a single span to be highlighted with ^^^^^^
-pub fn error_basic<S: Into<String>>(msg: S, span: &Span) -> Error {
+pub fn error<S: Into<String>>(msg: S, span: &Span) -> Error {
     return Arc::new(ErrorX { msg: msg.into(), spans: vec![span.clone()], labels: Vec::new() });
 }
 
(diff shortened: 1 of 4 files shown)


Chris-Hawblitzel (approved):
Looks good to me.  A couple comments:
- Can we make the span the first argument, since it's usually shorter than the string argument?
- Maybe consider just `error` instead of `error_basic`?

tjhance:
> Can we make the span the first argument, since it's usually shorter than the string argument?

Ah, I actually moved it to the second argument intentionally, to be consistent with `error_with_label`.

And I think `error_with_label` needs to have it as the second argument, since the span should logically go next to the label string argument.

Chris-Hawblitzel:
Ok, that makes sense.  It's inconsistent with `err_str` and `err_string` at the moment, but is the idea that we'll replace those with the new functions?

tjhance:
I was kind of thinking of fazing them out, since for any error message more complicated than a msg+span you'd want to use the Error builder methods. They have a lot of call sites, though, so it seemed like a lot of work to fix them all right now (especially since we'd want to do so with an eye towards improving them).

tjhance:
it'd kinda be nice to be able to do

```
error(msg)
    .primary_span(&span, label)
```

but then it would be too easy to accidentally make a stub error message without any spans

tjhance:
what if we had `error(msg)` return an `ErrorStub` object? And then you get a proper `Error` object after you add your first span

Chris-Hawblitzel:
I think what you have already is fine.

PR #24, opened 2022-01-10
Bit-vector basic operators

channy412:
This pull request implements `assert_bit_vector`, supporting basic operators such as shift/and/xor/add/less than/etc. Every bit-vector variable/constant/operation are treated as unsigned.`asssert_bit_vector` makes a separate z3 query, and the main query assumes its corresponding integer properties. When making the corresponding property, bit operators that are not existing at the integer level (namely <<, >>, &, ^, | ) are translated as uninterpreted functions.

Function level bit vector reasoning is not supported yet, although it can be annotated with
#[verifier(bit_vector)]


source/air/src/ast.rs
```diff
@@ -48,6 +50,21 @@ pub enum BinaryOp {
...
+    BitSub,
+    BitMul,
+    BitDiv, // unsigned
+    BitLt,  // unsigned
+    BitGt,  // unsigned
+    BitLe,  // unsigned
+    BitGe,  // unsigned
+    BitMod, // unsigned
+    Shr,    // logical
+    Shl,
```
tjhance, line 67: could the signedness be incorporated into the name? e.g., LShr, or UDiv?

tjhance, line 67: could the signedness be incorporated into the name? e.g., LShr, or UDiv?

source/vir/src/ast_util.rs
```diff
@@ -57,6 +58,16 @@ pub fn types_equal(typ1: &Typ, typ2: &Typ) -> bool {
     }
 }
 
+pub fn bitwidth_from_type(et: &Typ) -> Option<u32> {
+    if let TypX::Int(IntRange::U(size)) = &**et {
+        return Some(*size);
+    }
+    if let TypX::Int(IntRange::I(size)) = &**et {
```
yizhou7, line 65: We should probably not have this case, since we don't yet have support for singed  bitvector.

source/air/src/printer.rs
```diff
@@ -139,6 +154,21 @@ impl Printer {
...
                     BinaryOp::EuclideanDiv => "div",
                     BinaryOp::EuclideanMod => "mod",
+
+                    BinaryOp::BitXor => "bvxor",
+                    BinaryOp::BitAnd => "bvand",
+                    BinaryOp::BitOr => "bvor",
+                    BinaryOp::BitAdd => "bvadd",
+                    BinaryOp::BitSub => "bvsub",
+                    BinaryOp::BitMul => "bvmul",
+                    BinaryOp::BitDiv => "bvudiv", // unsigned
```
yizhou7, line 164: In case we forget, we should to make sure these operations in the original context have unsigned operands.

source/air/src/ast.rs
channy412: Sure, that sounds good

source/air/src/printer.rs
channy412: I learned that Rust blocks mixed types(signed and unsigned) for %, /, <, >. One strange thing is that Rust allows having mixed types for shifting. In case that both operands are signed integers, % and / are blocked in 'main' branch.
I will make a commit that forces bitvector reasoning to only have unsigned integers on a separate branch.

source/rust_verify/tests/bitvector.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+test_verify_one_file! {
+    #[test] test4_fails code! {
+        #[proof]
+        fn test4(b: u32) {
+            assert_bit_vector( (b << 2) >> 2 == b); // FAILS
+        }
+    } => Err(err) => assert_one_fails(err)
+}
```
Chris-Hawblitzel, line 95: It would be good to add a failing test where the `assert_bit_vector` succeeds but the subsequent `assert` about integers fails (because it doesn't match the `assert_bit_vector`).

It might also be good to add a failing test to check that signed integers don't accidentally use the unsigned integer reasoning.

Chris-Hawblitzel (approved):

source/air/src/typecheck.rs
```diff
@@ -116,10 +118,37 @@ fn check_exprs(
...
+
+    let w0 = get_bv_width(&t0)?;
+    let w1 = get_bv_width(&t1)?;
+
+    if w0 != w1 {
+        return Err(format!("in call to {}, argument should have the same width", f_name));
+    }
+
+    // return bool type if it is comparision op
+    if f_name.eq("bvlt") || f_name.eq("bvgt") || f_name.eq("bvle") || f_name.eq("bvge") {
```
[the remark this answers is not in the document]
channy412: Sure, I will make that change.

source/rust_verify/tests/bitvector.rs
channy412: I agree that those tests are great to have. 
About banning signed integers, I will make a few more commits to make sure of that before merging this into main.

channy412:
Made a new pull request ( #32 ) with a few commits addressing Chris and Yi's comment.

PR #30, opened 2022-01-11
introduce atomic operations

tjhance:
This PR depends on https://github.com/secure-foundations/verus/pull/19 and https://github.com/secure-foundations/verus/pull/22.

Here, I introduce the notion of an _atomic operation_ and ensure that any invariant block can have at most 1 atomic operation. I also introduce a stub Atomic library, and pull together a nontrivial example of a concurrent counter.

### Atomic operations

A function can be declared atomic with `#[verifier(atomic)]`. This is a trusted annotation, and it can only be used with `external_body`.

To check that any block only has a single atomic operation inside it, I added additional checks to `vir/src/modes.rs`. There's more information documented in the file. The gist is that, while processing an invariant block, we collect all the executable calls, then at the end of the block, we count up the atomic instructions we found and error if we found anything not right. Note that any "local" executable code (local variables or field accesses) count as "free" even if they are executable.

### Examples

`pervasive/atomics.rs` has a sample library. I'll fill in all the types and method later; right now we just have `PAtomicU32` and a couple of methods. This library is structured similar to `PCell`, the main difference being that its methods are marked atomic.

`rust_verify/examples/og_counter.rs` gives a nontrivial example of an atomic counter, incremented twice. It's incomplete, mostly because it has a system of unforgeable tokens which is currently trusted because we haven't yet built the state machine infrastructure. The tokens are based off the [example from this doc](https://docs.google.com/document/d/1Y6sigqVg1eWRcj8-j9tdJPG11pQHjardASNhIARO4-g/edit).

It does, however, showcase the atomics. One thing of note is how they do work slightly differently than they did in LinearDafny, where atomic blocks were tied to the atomic statements; here, we create an atomic along with a `Permission` object, and put the `Permission` in an invariant. Then we open up the invariant with an `open_invariant!` call, access the `Permission` token, and perform the atomic operation inside.

### Error messages

```
error: open_invariant cannot contain more than 1 atomic operation
  --> test.rs:24:36
   |
24 |       open_invariant!(&i => inner => {
   |  ____________________________________^
25 | |         atomic_op();
   | |         ----------- atomic here
26 | |         atomic_op();
   | |         ----------- atomic here
27 | |     });
   | |_____^


error: open_invariant cannot contain non-atomic operations
  --> test.rs:24:36
   |
24 |       open_invariant!(&i => inner => {
   |  ____________________________________^
25 | |         non_atomic_op();
   | |         --------------- non-atomic here
26 | |     });
   | |_____^


error: open_invariant cannot contain an 'exec' loop
  --> test.rs:25:36
   |
25 |        open_invariant!(&i => inner => {
   |   ____________________________________^
26 |  |         while x < 10 {
   |  |_________-
27 | ||             x = x + 1;
28 | ||         }
   | ||_________-
29 |  |     });
   |  |_____^ this invariant block contains a loop


error: 'atomic' is a trusted annotation: it may not be used on a verified function
  --> test.rs:24:1
   |
24 | fn untrusted_atomic_op() { }
   | ^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -5,7 +5,9 @@ use crate::ast::{
 use crate::ast_util::{err_str, err_string, get_field};
 use crate::util::vec_map_result;
 use air::ast::Span;
+use air::errors::{error, error_with_label};
 use air::scope_map::ScopeMap;
+use std::cmp::min;
 use std::collections::HashMap;
 
 // Exec <= Proof <= Spec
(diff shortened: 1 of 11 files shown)


tjhance:
cc @parno for the example with the unforgeable tokens

source/vir/src/modes.rs
```diff
@@ -98,7 +177,12 @@ fn add_pattern(typing: &mut Typing, mode: Mode, pattern: &Pattern) -> Result<(),
     }
 }
 
-fn check_expr(typing: &mut Typing, outer_mode: Mode, expr: &Expr) -> Result<Mode, VirErr> {
+fn check_expr(
```
Chris-Hawblitzel, line 180: I think it would be cleaner to make atomic_insts a field of Typing rather than an additional parameter to check_expr and check_stmt.

source/vir/src/modes.rs
```diff
@@ -214,14 +317,15 @@ fn check_expr(typing: &mut Typing, outer_mode: Mode, expr: &Expr) -> Result<Mode
             for binder in params.iter() {
                 typing.insert(&expr.span, &binder.name, false, Mode::Spec);
             }
-            check_expr_has_mode(typing, Mode::Spec, body, Mode::Spec)?;
+            let mut atomic_insts = None;
```
Chris-Hawblitzel, line 320: `None` is fine here, but you could also just pass atomic_insts as-is (since the closure is a spec).

Chris-Hawblitzel (approved):

source/vir/src/modes.rs
tjhance: I imagined in the future we would have exec closures too. Regardless, nothing in the closure actually gets executed at this point control-flow-wise, so I thought None made the most sense.

PR #31, opened 2022-01-12
support impl methods without a self argument

tjhance:
Trying to support this so I can write proper `new` methods for the atomics library. This task _almost_ seemed like it would be really easy: rustc has a function called `qpath_res` which can look up any `QPath` and return a `DefId`, so for any call, I could just look up a `DefId` and we're off to the races; unless the `DefIf` is a Local, in which case we fall back to a dynamically computed function.

The trouble was converting the DefId to a VIR path. Naively, this results in paths that includes something like `impl#0` in it. I then realized the rest of the code does this conversion by taking the path to the type and using that path (the `self_path`) instead.

So, I still had to determine the `self_path` when handling the call expression. So to do that, I had to check if the `qpath` was type relative, get the type, and call `qpath_res` _again_ on the type to get the def id of the type, which was all slightly awkward.

It seems to me that currently, the rustc DefId -> VIR Path conversion is fairly brittle. Currently every place we do this conversion (including this new place) seems like it's using redundant information, whereas we really ought to be able to get all the information we need from the DefId. But I couldn't figure out how to use rustc's library to get from the DefId directly to the type ... which might be why it's written this way to begin with?

### DIFF
--- a/source/rust_verify/example/impl_basic.rs
+++ b/source/rust_verify/example/impl_basic.rs
@@ -9,9 +9,40 @@ struct Car {
 }
 
 impl Car {
+    fn new() -> Car {
+        ensures(|result: Car| equal(result.passengers, 201));
+        Car { four_doors: false, passengers: 201 }
+    }
+
     fn get_passengers(&self) -> u64 {
         ensures(|result: u64| result == self.passengers);
         self.passengers
+    }
+}
+
+#[derive(PartialEq, Eq)]
+struct TemplateCar<V> {
+    four_doors: bool,
+    passengers: u64,
+    the_v: V,
+}
+
+impl<V> TemplateCar<V> {
+    fn template_new(v: V) -> TemplateCar<V> {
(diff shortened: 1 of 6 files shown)


tjhance:
OK, i came up with a way to clean this up better. I figured out how I can go from a DefId to the `impl` to the type name.

So now `def_id_to_vir_path` will always give you the right `Path` when the `DefId` points to a function, without the caller having to do extra work to construct the type themselves.

This cleans up the call sites a lot, (although `fn_call_to_vir` still has to do a little extra work to compute paths for the autoview cases).

utaal (approved):
This looks good!

Thanks for addressing the DefId -> VIR Path issue. That seems a much cleaner to me.

I had a couple stylistic comments, but it was just easier to send you a patch. So, if it looks good to you, please merge #36 into this (you can just click merge there), and then feel free to merge this into main. Thanks!

PR #32, opened 2022-01-14
`assert_bit_vector` and `as` for bit-vector

channy412:
Compared to the previous pull request, this pull request has a better way to block signed integers. 
Also, it implements "as" for bit-vector via 'extract' and 'concat' in Z3.


source/vir/src/sst_to_air.rs
```diff
@@ -490,6 +499,121 @@ fn one_stmt(stmts: Vec<Stmt>) -> Stmt {
...
+            return Arc::new(ExprX::Binary(bop, lh, rh));
+        }
+        ExpX::Unary(op, exp) => {
+            // bv type casting by 'as' keyword
+            // convert Clip into concat/extract
+            if let TypX::Int(old_range) = &*exp.typ {
+                if let UnaryOp::Clip(new_range) = op {
+                    match new_range {
+                        IntRange::U(new_n) | IntRange::I(new_n) => match old_range {
+                            IntRange::U(old_n) | IntRange::I(old_n) => {
```
Chris-Hawblitzel, line 566: Minor style note: you can avoid nesting if/if/match/match by doing nested matching on a pair: `match (&*exp.typ, op) {
                (TypX::Int(IntRange::U(old_n) | IntRange::I(old_n)), UnaryOp::Clip(IntRange::U(new_n) | IntRange::I(new_n))) => ...`

Chris-Hawblitzel (approved):

source/air/src/ast.rs
```diff
@@ -58,6 +76,7 @@ pub enum MultiOp {
     Sub,
     Mul,
     Distinct,
+    Extract,
```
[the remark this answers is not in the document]
channy412: Thanks for the comments! All of them look great to me.

source/vir/src/sst_to_air.rs
channy412: It looks much much cleaner this way! Thanks!

Chris-Hawblitzel:
You can go ahead and merge this into the `main` branch.

PR #39, opened 2022-01-20
flesh out the atomic library to the extent possible

tjhance:
Flesh out the trusted wrapper around [std::sync::Atomic](https://doc.rust-lang.org/std/sync/atomic/)

There are a few things we can't yet support:

 * `compare_exchange(_weak)` because Verus doesn't yet support `Result`
 * `fetch_nand` because (as far as I can tell) Verus doesn't support bitwise negation


utaal:
I'll take an in-depth look later, but for

> `compare_exchange(_weak)` because Verus doesn't yet support Result

can we make our own `Result` alternative in `pervasive`, for now? (like `pervasive::Option`)

parno:
> fetch_nand because (as far as I can tell) Verus doesn't support bitwise negation

@yizhou7 @channy412 : Is this something you two could address?

channy412:
Yes, I will address that. Thanks for letting me know.

channy412:
46f92022542aec8e01edd72f3bf7b81d12ba73a4  I added bit-wise negation support, but I will try some testing before I make a pull request.

utaal:
> compare_exchange(_weak) because Verus doesn't yet support Result

It looks like https://github.com/secure-foundations/verus/pull/43 will land soon.

> fetch_nand because (as far as I can tell) Verus doesn't support bitwise negation

I added a `pervasive::result::Result` that supports the new accessor feature (`#[is_variant]`), https://github.com/secure-foundations/verus/blob/b439b9bdbacc7f5c00bc27a5cc8ed1d255b20d49/source/rust_verify/example/adts.rs#L63-L69

Would this help?

source/pervasive/atomics.rs
```diff
@@ -1,89 +1,464 @@
-use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::atomic::{
```
utaal, line 1: Sorry I only spotted this now, but can you rename `pervasive::atomics` to `pervasive::atomic` (singular), to match the rust std naming convention?

(We should also probably rename `invariants` to `invariant` and maybe move both plus `cell` to a new `sync` module, but not in this PR.)

source/rust_verify/tests/atomic_lib.rs
```diff
@@ -0,0 +1,637 @@
...
+            assert(perm.value == 4);
+
+            let l = at.fetch_min(&mut perm, 7);
+            assert(l == 4);
+            assert(perm.value == 4);
+
+            let l = at.into_inner(perm);
+            assert(l == 4);
+
+            assert(false); // FAILS
```
utaal, line 68: What is the purpose of these `assert(false)`s?

Unfortunately, due to the fact that verus currently doesn't have the same error reporting behavior as dafny (which tries its best to report the earliest assertion failure), this may hide an assertion elsewhere in the SMT query (for one of the other `assert`s). This is why we generally separate failing cases into separate functions, just to make sure they may not hide other failures in the postive tests.

utaal (changes requested):

utaal:
I need to make another pass when I have a bit more time later, but otherwise this all looks great. Thanks for figuring out how to do this with macros!

source/rust_verify/tests/atomic_lib.rs
tjhance: It's a "smoke test" to make sure there aren't contradictory post-conditions in the trusted methods. It's not a substitute for careful review ... but it's also cheap.

tjhance: anyway, given what you say, it sounds better to remove it

tjhance:
i added support for the remaining features & addressed your other comments

source/rust_verify/tests/atomic_lib.rs
utaal: > It's a "smoke test" to make sure there aren't contradictory post-conditions in the trusted methods. It's not a substitute for careful review ... but it's also cheap.

Ah, makes sense, thanks!
I think I maybe have a solution that addresses my concerns and also preserves the smoke test: https://github.com/secure-foundations/verus/pull/46
If that looks good, please merge it before merging this.

source/pervasive/atomic.rs
```diff
@@ -0,0 +1,468 @@
...
+make_integer_atomic!(PAtomicU16, PermissionU16, AtomicU16, u16, wrapping_add_u16, wrapping_sub_u16, 0, 0xffff);
+make_integer_atomic!(PAtomicU32, PermissionU32, AtomicU32, u32, wrapping_add_u32, wrapping_sub_u32, 0, 0xffff_ffff);
+make_integer_atomic!(PAtomicU64, PermissionU64, AtomicU64, u64, wrapping_add_u64, wrapping_sub_u64, 0, 0xffff_ffff_ffff_ffff);
+
+make_integer_atomic!(PAtomicI8, PermissionI8, AtomicI8, i8, wrapping_add_i8, wrapping_sub_i8, -0x80, 0x7f);
+make_integer_atomic!(PAtomicI16, PermissionI16, AtomicI16, i16, wrapping_add_i16, wrapping_sub_i16, -0x8000, 0x7fff);
+make_integer_atomic!(PAtomicI32, PermissionI32, AtomicI32, i32, wrapping_add_i32, wrapping_sub_i32, -0x8000_0000, 0x7fff_ffff);
+make_integer_atomic!(PAtomicI64, PermissionI64, AtomicI64, i64, wrapping_add_i64, wrapping_sub_i64, -0x8000_0000_0000_0000, 0x7fff_ffff_ffff_ffff);
+
+// TODO usize and isize
```
utaal, line 468: Are you missing something for these, or just planning to implement them later?

utaal (approved):
Thank you!

source/pervasive/atomic.rs
tjhance: I can't implement them until we have a way to use constants like `usize::MAX`. I'll add a note

Issue #23, opened 2022-01-10
Use a visitor that doesn't require constructing an `Expr` when performing checks (and not mapping the AST)


utaal:
Depends on #22 which contains an expression visitor that isn't also a mapper.

Especially in `well_formed` we're abusing `map_expr_visitor` (and similar): they require the mapper function to produce a new Expr, but when just performing checks by walking the AST this new Expr is just discarded.

PR #42, opened 2022-01-21
Use the generic expr_visitor_dfs for early_exit_cf (see #23)

utaal:
This continues the refactor for #23, using the generic `expr_visitor_dfs` (extended to allow the user to ask to skip an AST subtree). @tjhance correctly raised this concern using a generic visitor:

> i wanted to make sure that if somebody adds a feature (e.g., Continue) they'll be forced to update the match statement.

So I'm using the approach suggested by @Chris-Hawblitzel:

> A good balance would be to use the generic expr/stmt visitor infrastructure, and when using this for the early-exits visitor, implement a minimal visit that just skims each case and does nothing:
```
  ExprX::Tuple(..) => {}
  ExprX::Ctor(..) => {}
```
> This way we catch any new features that get added, but we don't have to manually implement the recursion into the subexpressions for each case.


tjhance (approved):

utaal:
Thank you!

PR #43, opened 2022-01-21
Bitvector bit-wise negation, boolean and/or

channy412:
This is to add bit-wise negation and boolean operators for bit-vector. Since rust hir shares the`!` operator for both boolean and bit operator, I checked operand's type to choose what operator to use. 

One example I need help with is the below one.

`assert_bit_vector( (b1 == b2) || (b1 == b3) >>= (b1 | b2 | b3  == b2 | b3) );`

Currently, this example timeouts for 60 seconds. In the smt file, however, when I comment out the following option, Z3 returns `unsat` instantly. 
`(set-option :smt.case_split 3)`

I tried manually changing the `bvor` encoding from binaryOp style to multiOp, (  using (bvor b1 b2 b3) instead of (bvor (bvor b1 b2) b3) ) , but it still makes 60 seconds timeout.

### DIFF
--- a/source/air/src/typecheck.rs
+++ b/source/air/src/typecheck.rs
@@ -141,6 +141,13 @@ fn check_bv_unary_exprs(
             }
             Ok(Arc::new(TypX::BitVec(w_new)))
         }
+        UnaryOp::BitNot => {
+            let t0 = check_expr(typing, &exprs[0])?;
+            match get_bv_width(&t0) {
+                Ok(_) => Ok(t0.clone()),
+                Err(..) => panic!("Interner Error: not a bv type inside a bvnot"),
+            }
+        }
         _ => panic!("Interner Error: not a bv unary op, got {}", f_name),
     }
 }
@@ -205,6 +212,9 @@ fn check_expr(typing: &mut Typing, expr: &Expr) -> Result<Typ, TypeError> {
             }
         }
         ExprX::Unary(UnaryOp::Not, e1) => check_exprs(typing, "not", &[bt()], &bt(), &[e1.clone()]),
+        ExprX::Unary(UnaryOp::BitNot, e1) => {
+            check_bv_unary_exprs(typing, UnaryOp::BitNot, "bvnot", &[e1.clone()])
+        }
         ExprX::Unary(UnaryOp::BitExtract(high, low), e1) => {
             check_bv_unary_exprs(typing, UnaryOp::BitExtract(*high, *low), "extract", &[e1.clone()])
         }
--- a/source/vir/src/sst_to_air.rs
+++ b/source/vir/src/sst_to_air.rs
@@ -654,6 +654,7 @@ fn exp_to_bv_expr(state: &State, exp: &Exp) -> Expr {
                 BinaryOp::BitOr => air::ast::BinaryOp::BitOr,
                 BinaryOp::Shl => air::ast::BinaryOp::Shl,
                 BinaryOp::Shr => air::ast::BinaryOp::LShr,
+                BinaryOp::Implies => air::ast::BinaryOp::Implies,
                 _ => panic!("unhandled bv binary operation {:?}", op),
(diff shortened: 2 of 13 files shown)


channy412:
Copying from here (https://github.com/Z3Prover/z3/blob/master/src/smt/params/smt_params_helper.pyg), it says, 
('case_split', UINT, 1, '0 - case split based on variable activity, 1 - similar to 0, but delay case splits created during the search, 2 - similar to 0, but cache the relevancy, 3 - case split based on relevancy (structural splitting), 4 - case split on relevancy and activity, 5 - case split on relevancy and current goal, 6 - activity-based case split with theory-aware branching activity')
But I am not sure what these mean. 
Verus' default value for this option is 3. The above example works for 0,1,2,4,6 but not for 3,5.

channy412:
When I run the below SMT query with older Z3 (4.4.1 and 4.8.5), it works. 
However, on versions 4.8.14(latest) and 4.8.6, this SMT file timeout 60 seconds. Also, when I remove `push` on line 3, it suddenly works again.

[log3.txt](https://github.com/secure-foundations/verus/files/7929884/log3.txt)

source/air/src/typecheck.rs
```diff
@@ -141,6 +141,13 @@ fn check_bv_unary_exprs(
             }
             Ok(Arc::new(TypX::BitVec(w_new)))
         }
```
Chris-Hawblitzel, line 143: It seems like `check_bv_unary_exprs` should in general return an `Err` when the exprs have the wrong type, rather than panicking.  Also, it doesn't look like you need multiple exprs here, since you only use `exprs[0]`.

source/vir/src/sst_to_air.rs
```diff
@@ -637,8 +644,32 @@ fn exp_to_bv_expr(state: &State, exp: &Exp) -> Expr {
...
+                _ => {
+                    assert_unsigned(&lhs);
+                    assert_unsigned(&rhs);
+                }
+            };
+            let lh = exp_to_bv_expr(state, lhs);
+            let rh = exp_to_bv_expr(state, rhs);
+            let _ = match op {
+                BinaryOp::And => {
+                    return Arc::new(ExprX::Multi(MultiOp::And, Arc::new(vec![lh, rh])));
```
Chris-Hawblitzel, line 660: You can use `mk_and` here.

source/vir/src/sst_to_air.rs
```diff
@@ -637,8 +644,32 @@ fn exp_to_bv_expr(state: &State, exp: &Exp) -> Expr {
...
+                }
+                BinaryOp::Or => return Arc::new(ExprX::Multi(MultiOp::Or, Arc::new(vec![lh, rh]))),
```
Chris-Hawblitzel, line 662: You can use `mk_or` here.

Chris-Hawblitzel (approved):

channy412:
I ran an experiment on `random-seed` from 0 to 99, and the above behavior is preserved regardless of the value of `random-seed`. 
Older Z3 (4.4.1 and 4.8.5) gave me `UNSAT` instantly for any `random-seed`, and newer Z3(4.8.14, 4.8.6) timed out 30 seconds for any `random-seed`(0~99).

source/air/src/typecheck.rs
channy412: Thanks for the comments :)

parno:
Interesting!  Maybe we should move additional discussion of the bit-vector weirdness to an issue?  You could summarize your experiments when you open it.  I'd also be curious if anything changes if you use u32 or u8 instead of u64.

tjhance:
By the way, in my experience, Z3's random-seed often has little to no effect (to the point where randomizing variable names can have a larger effect). If you're running some kind of experiment, be careful to make sure whatever you're doing is meaningful

PR #44, opened 2022-01-25
Demarcate VIR phases by checking that simplified VIR does not use nodes that should have been simplified

utaal:
So that we are reminded if we make changes to the phases that use different nodes, and as a template for phases after simplify.


utaal:
This stemmed from me working on simplify and realizing I didn’t have clear what goes away after that phase. I’d suggest (in the fullness of time) we make one such function for each pass that removes/adds an allowed node (`Expr` or `Typ`, here), as a form of checked documentation.

Chris-Hawblitzel (approved):
Looks fine to me.  I'd recommend creating a new file to hold the check_*_simplified functions.

PR #57, opened 2022-02-07
Add the fndef macro that defines an uninterpreted spec function (with no body)

utaal:
Following up from @tjhance and @Chris-Hawblitzel's conversation about these functions.

### DIFF
--- a/source/builtin_macros/src/fndecl.rs
+++ b/source/builtin_macros/src/fndecl.rs
@@ -1,9 +1,8 @@
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::{braced, parse_macro_input, Error, Expr, Field, FieldsNamed, Ident, Token, Type};
 
 #[inline(always)]
-pub fn fndef(input: TokenStream) -> TokenStream {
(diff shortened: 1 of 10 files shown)


tjhance:
I like the idea! Maybe we should do something with the word `trusted` in it? maybe like `trusted_fn!`

tjhance:
also, is there a reason it's a proc_macro?

utaal:
The idea is to use this for pure uninterpreted functions, which I don’t think count as trusted per se. What’s trusted are the axioms that give them meaning, potentially.

proc macro because there’s no capture in macro_rules that matches just a function decl apparently?

tjhance:
did you try the `tt` type? I think it supports arbitrary token streams

utaal:
Tried, it apparently supports _some_ token streams (I actually failed at finding an authoritative answer to what exactly it matches). :/

tjhance:
Ah well, proc macro it is then

utaal:
> pure uninterpreted functions

I'm now thinking this should be called `fndecl!`

tjhance (approved):

Chris-Hawblitzel (approved):

PR #60, opened 2022-02-11
Add pervasive::map::Map::remove

utaal:


source/pervasive/map.rs
```diff
@@ -27,6 +27,12 @@ impl<K, V> Map<K, V> {
 
     fndecl!(pub fn insert(self, key: K, value: V) -> Map<K, V>);
 
+    #[spec]
+    #[verifier(pub_abstract)]
+    pub fn remove(self, key: K) -> Map<K, V> {
+        arbitrary()
+    }
+
```
Chris-Hawblitzel, line 35: We could use `fndecl` here

source/pervasive/map.rs
```diff
@@ -68,6 +74,13 @@ pub fn axiom_map_insert_different<K, V>(m: Map<K, V>, key1: K, key2: K, value: V
     ensures(equal(m.insert(key2, value).index(key1), m.index(key1)));
 }
 
+#[proof]
+#[verifier(external_body)]
+#[verifier(broadcast_forall)]
+pub fn axiom_map_remove_domain<K, V>(m: Map<K, V>, key: K) {
+    ensures(equal(#[trigger] m.remove(key).dom(), m.dom().remove(key)));
+}
+
```
Chris-Hawblitzel, line 83: This axiom looks fine.  We should also add something like `axiom_map_insert_different` for `remove`.

Chris-Hawblitzel (changes requested):

source/pervasive/map.rs
```diff
@@ -68,6 +70,24 @@ pub fn axiom_map_insert_different<K, V>(m: Map<K, V>, key1: K, key2: K, value: V
...
+#[proof]
+#[verifier(external_body)]
+#[verifier(broadcast_forall)]
+pub fn axiom_map_remove_different<K, V>(m: Map<K, V>, key1: K, key2: K) {
+    requires([
+        m.dom().contains(key1),
```
tjhance, line 85: strictly speaking, we don't actually need this condition, do we?

do we have a philosophy on such things?

Chris-Hawblitzel (approved):

source/pervasive/map.rs
utaal: I think we don't strictly need it for soundness.

I'm matching `axiom_map_insert_different` here; I believe the strategy has been to keep axioms just as powerful as they need to, and not more (at least until we have the option of multiple/a-la-carte `broadcast_forall` import sets). @Chris-Hawblitzel may have more to say.

PR #67, opened 2022-02-18
state_machine! macro

tjhance:


tjhance:
I posted this as a fairly clean checkpoint for review on the parts of the state machine code that are not related to concurrency. However, as I've worked more & more on concurrency i've been making small changes to the other parts of the code. I think it probably does not make much sense anymore to review this PR.

PR #74, opened 2022-02-25
support lifetime bounds

tjhance:
by "support" I of course just mean "ignore them instead of panicking"


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -498,7 +500,7 @@ pub(crate) fn check_generic_bound<'tcx>(
 pub(crate) fn check_generics_bounds<'tcx>(
     tcx: TyCtxt<'tcx>,
     generics: &'tcx Generics<'tcx>,
-    function_decl: bool,
+    _function_decl: bool,
```
utaal, line 503: If we don't need this parameter anymore, can you please remove it?

utaal (approved):
Looks good to me. And I cannot see ways this may introduce unsoundness (@Chris-Hawblitzel maybe have a peek to double-check?).

Chris-Hawblitzel (approved):

PR #81, opened 2022-03-10
clean up PCell, add PPtr

tjhance:
I cleaned up `PCell`, using `&mut` parameters now that they're available. I also added a companion `PPtr`. I'm doing this now, because these were necessary for the `Rc` example over on the state machine branch.

`PPtr` is similar to `PCell`, but has a few key differences:
 * In `PCell<T>`, the type `T` is placed internally to the `PCell`, whereas with `PPtr`, the type `T` is placed at some location on the heap.
 * Since `PPtr` is just a pointer (represented by an integer), it can be `Copy`
 * The `ptr::Permission` token represents not just the permission to read/write the contents, but also to deallocate the memory.

Each has the following trusted methods `empty`, `put`, `take`, `replace`, and `as_ref`. To be complete, we still need `as_mut`, but of course we need the ability to return `&mut` first.

I wasn't sure about the name of `as_ref` (vs. `borrow`). Actually, I'm not really sure if _either_ is appropriate (we can't actually implement the `AsRef` or `Borrow` traits, since the function takes two arguments ...) but I still felt like `as_ref` was more appropriate.

### DIFF
--- a/source/pervasive/ptr.rs
+++ b/source/pervasive/ptr.rs
@@ -1,0 +1,188 @@
+use std::ptr::NonNull;
+use std::mem::MaybeUninit;
+use std::alloc::{Layout};
+use std::alloc::{dealloc};
+
+#[allow(unused_imports)] use builtin::*;
+#[allow(unused_imports)] use builtin_macros::*;
+#[allow(unused_imports)] use crate::pervasive::*;
+#[allow(unused_imports)] use crate::pervasive::modes::*;
+
+/// PPtr ("i.e., permissioned pointer").
+///
+/// This is similar to PCell, but has a few key differences:
+///  * In PCell<T>, the type T is placed internally to the PCell, whereas with PPtr,
+///    the type T is placed at some location on the heap.
+///  * Since PPtr is just a pointer (represented by an integer), it can be `Copy`
+///  * The `ptr::Permission` token represents not just the permission to read/write
+///    the contents, but also to deallocate.
+
+
+// TODO implement: borrow_mut, Drop
+
+// TODO Identifier should be some opaque type, not necessarily an int
+//type Identifier = int;
+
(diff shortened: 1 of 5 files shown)


source/pervasive/ptr.rs
```diff
@@ -0,0 +1,188 @@
...
+///
+/// This is similar to PCell, but has a few key differences:
+///  * In PCell<T>, the type T is placed internally to the PCell, whereas with PPtr,
+///    the type T is placed at some location on the heap.
+///  * Since PPtr is just a pointer (represented by an integer), it can be `Copy`
+///  * The `ptr::Permission` token represents not just the permission to read/write
+///    the contents, but also to deallocate.
+
+
+// TODO implement: borrow_mut, Drop
```
utaal, line 21: Can one really `Drop` this? The dropper would need to have the permission, right? (if `drop` of PPtr disposes the allocation)

utaal:
I have a small preference for `borrow` (instead of `as_ref`). `as_ref` generally turns a `Container<T>` into a `Container<&T>`, whereas `borrow` would match `std::cell::RefCell::borrow`. It's okay if these don't match the traits.
Also the `Borrow` trait is rarely used, I think, and generally only for types where the borrowed version is a different type (e.g. `String` and `&str`). For example, `RefCell` has a `borrow` function but doesn't implement `Borrow`.

source/pervasive/ptr.rs
tjhance: ah the TODO was more, "figure out what to do" 😅 

well, dropping the PPtr should do nothing anyway (it's not an owning pointer, nor is it a smart pointer). I thought maybe we would want to disallow dropping the `Permission` token, if that ever becomes possible (to prevent leaks, I mean)

PR #85, opened 2022-03-14
Support let (non-mut) decl without initialization, and later init of non-mut let

### LINKED ISSUE #29: Verus demands `let mut` where Rust does not
If you create a variable and leave off the initializer, and declare it later,

```
let x: type;
// ...
x = ...;
```

then Rust doesn't require a `mut` on the declaration. However, Verus currently requires it (see the `"variable must be declared 'mut' to allow assignment"` error in `vir/src/modes.rs`)

utaal:
Fixes #29

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -317,7 +317,7 @@ pub enum ExprX {
     /// Choose specification values satisfying a condition, compute body
     Choose { params: Binders<Typ>, cond: Expr, body: Expr },
     /// Assign to local variable
-    Assign { not_mut_init: bool, lhs: Expr, rhs: Expr },
+    Assign { init_not_mut: bool, lhs: Expr, rhs: Expr },
     /// Reveal definition of an opaque function with some integer fuel amount
     Fuel(Fun, u32),
     /// Header, which must appear at the beginning of a function or while loop.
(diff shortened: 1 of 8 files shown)


Chris-Hawblitzel:
What about this:

```
fn test() {
    #[spec] let x: int;
    x = 1;
    x = 2;
    assert(false);
}
```

source/vir/src/ast.rs
```diff
@@ -315,7 +315,7 @@ pub enum ExprX {
     /// Choose specification values satisfying a condition, compute body
     Choose { params: Binders<Typ>, cond: Expr, body: Expr },
     /// Assign to local variable
-    Assign(Expr, Expr),
+    Assign { is_init: bool, lhs: Expr, rhs: Expr }, // (is_init, lhs, rhs)
```
tjhance, line 318: i think you can remove this comment here

utaal:
> What about this:
> 
> ```
> fn test() {
>     #[spec] let x: int;
>     x = 1;
>     x = 2;
>     assert(false);
> }
> ```

---

I forgot about this, thank you. It should now be fixed (by disallowing delayed initialization for spec variables).

Chris-Hawblitzel (approved):
Minor comment: `init_non_mut` might be clearer than `not_mut_init` to emphasize that it is an initialization

PR #96, opened 2022-03-18
handle case where conditional returns on one branch but not the other

tjhance:
Add support for the Rust idiom,

```rust
let x = if { expr } else { return; }
```

and similarly for match statements (which we internally lower into if-else).

Also: handle code that makes use of blocks implicitly returning units. Not that I expect that to be very common, but the two cases look similarly syntactically, so I wanted to make sure handling one case didn't screw up the other one. The fact that it's hard to tell them apart made some of the corner cases a little weird, but I documented the assumptions in `ast_to_sst` and wrote tests.

### DIFF
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -687,6 +687,8 @@ fn expr_to_stm_opt(
             // If e1 evaluates to `proceed_on`, then evaluate and
             // return e2; otherwise, return the value `other`
             // (without evaluating `e2`).
+            // Also note: if `e2` is a pure expression, we don't need to do the
+            // special handling.
             let short_circuit = match op {
                 BinaryOp::And => Some((true, false)),
                 BinaryOp::Implies => Some((true, true)),
@@ -716,7 +718,7 @@ fn expr_to_stm_opt(
                     let e1 = unwrap_or_return_never!(e1, stms1);
                     stms1.append(&mut stms2);
                     let e2 = unwrap_or_return_never!(e2, stms1);
-                    let bin = mk_exp(ExpX::Binary(*op, e1, e2));
+                    let bin = mk_exp(ExpX::Binary(*op, e1, e2.clone()));
 
                     if let BinaryOp::Arith(arith, inferred_mode) = op {
                         // Insert bounds check
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -533,13 +533,12 @@ fn if_to_stm(
             } else {
                 // We have `if ( stms0; e0 ) { stms1; e1 } else { stms2; e2 }`.
                 // We turn this into:
-                //  stms0; if e0 { stms1 } else { stms2 };
-                //  if e0 { e1 } else { e2 }
+                //  stms0;
+                //  if e0 { stms1; temp = e1; } else { stms2; temp = e2; };
+                //  temp
 
-                // If statement, put results from e1/e2 in a temp variable, return temp variable
                 let (temp, temp_var) = state.next_temp(&expr.span, &expr.typ);
                 let temp_id = state.declare_new_var(&temp, &expr.typ, false, false);
-                // if e0 { stms1; temp = e1; } else { stms2; temp = e2; }
                 stms1.push(init_var(&expr.span, &temp_id, &e1));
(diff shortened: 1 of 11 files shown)


tjhance:
It's possible that it would be cleaner if we just do the control-flow analysis to determine if an expression will definitely not return, and skip generating all VCs after those points.

Chris-Hawblitzel:
It might be that `expr_to_stm_opt` needs a better return type.  Instead of Some(Exp) | None, maybe it needs to return Some(Exp) | Unit | Never.

Here's another (silly) test case:
```
fn test() -> u32 {
    ensures(|r: u32| r < 3);
    let x: u32 = {return 1; 3};
    let y: u32 = {return 2;};
    x + y
}
```

tjhance:
This change ended up getting pretty big, as I kept running into miscellaneous bugs or missing things ...

 * Updated to have `ast_to_sst` explicitly determine whether an expression is a never-return or an implicit unit.
 * General clean-up and documentation in `ast_to_sst`.
 * Allow the case where a fn returns early and doesn't give a final expression
 * Implement `loop { ... }` statements (i know this one seems unrelated, but these are relevant for never-return logic, and they were pretty easy to implement)

Now, what I _wanted_ to do, was move the post-condition emitting from sst_to_air into ast_to_sst, because all the early `return` statements are handled in ast_to_sst, and I wanted to be consistent. Unfortunately, that revealed a handful of bugs that I didn't want to fold into this diff:

 * we need to apply the trait_typ_substs to the post-conditions emitted in ast_to_sst
 * we need to apply the bitvector mode to post-conditions emitted in ast_to_sst

Right now, these are both latent bugs for functions that use early `return` statements.

source/vir/src/ast_to_sst.rs
```diff
@@ -161,11 +196,54 @@ fn function_can_be_exp(ctx: &Ctx, expr: &Expr, path: &Fun) -> Result<bool, VirEr
...
+/// the code after this will be unreachable, which means the user might be allowed
+/// to leave off a return expression. We need to detect that case, or else we might
+/// wrongly determine that it returns a 'unit' and we'll create malformed AIR code.
+///
+/// So when does Rust do this? As far as I can tell, it only does this if:
+///
+///   (i) it's a 'loop' statement, and
+///   (ii) it doesn't have ANY 'break' statement in it
+///        (It is possible that a 'break' statement in a while loop might itself be
+///        unreachable, but Rust doesn't seem to take that into account for this purpose.)
```
utaal, line 221: Could we ask rustc for this information? I think this may be constructed for MIR, which would be potentially inconvenient for us, but it may be worth a look (e.g. for `break`, or for returns inside a loop). IIRC, the control flow graph is stored as https://docs.rs/rustc-cfg/0.4.0/rustc_cfg/struct.Cfg.html

Chris-Hawblitzel (approved):

source/vir/src/ast_to_sst.rs
```diff
@@ -339,102 +468,181 @@ fn stm_call(
...
+            if stms1.len() == 0 && stms2.len() == 0 {
+                // In this case, we can construct a pure expression.
+                let expx = ExpX::If(e0.clone(), e1.clone(), e2.clone());
+                let exp = SpannedTyped::new(&expr.span, &expr.typ, expx);
+                (vec![], ReturnValue::Some(exp))
+            } else {
+                // We have `if ( stms0; e0 ) { stms1; e1 } else { stms2; e2 }`.
+                // We turn this into:
+                //  stms0; if e0 { stms1 } else { stms2 };
+                //  if e0 { e1 } else { e2 }
```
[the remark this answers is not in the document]
tjhance: thanks, I don't know what I was thinking when I wrote this, it's totally wrong

utaal:
> Could we ask rustc for this information? I think this may be constructed for MIR, which would be potentially inconvenient for us, but it may be worth a look (e.g. for break, or for returns inside a loop).

I kinda wanted to know about this!

source/vir/src/ast_to_sst.rs
tjhance: Ah, sorry, well I don't actually know that much about MIR, so I don't really know much what it would take to extract the information. The relevant control flow analysis, even with the breaks, on the other hand, probably isn't that hard (and we'll need that information anyway).

Yesterday I looked through rustc_typeck and found the place where it computes this, although I don't know how to get that information out either.

PR #97, opened 2022-03-19
Datatype field updates (including as &mut arguments)

utaal:
This adds support for `a.b.c = 3;` and `call(&mut a.b.c)` using the same strategy of havoc-ing the base variable (`a`), and assuming all the unchanged fields are the same before and after a snapshot.

I'd like to add more tests for this, but I'd also like to land it as it touches a few critical places and would prefer it doesn't diverge too much. Some of this logic may have to change with a proper fix to https://github.com/secure-foundations/verus/issues/115 (which is next up for me).

### DIFF
--- a/source/vir/src/ast_simplify.rs
+++ b/source/vir/src/ast_simplify.rs
@@ -155,6 +155,50 @@ fn pattern_to_exprs(
     }
 }
 
+fn simplify_assign_lhs(
+    ctx: &GlobalCtx,
+    state: &mut State,
+    datatype_path: &Path,
+    variant: &Ident,
+    field: &Ident,
+    base: &Expr,
+    rhs: &Expr,
+) -> Result<(Expr, Expr), VirErr> {
+    let datatype = &ctx.datatypes[datatype_path];
+    assert_eq!(datatype.len(), 1);
+    let fields = &datatype[0].a;
+    let binders = fields
+        .iter()
+        .map(|f| {
+            let field_op;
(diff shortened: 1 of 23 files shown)


source/vir/src/ast_simplify.rs
```diff
@@ -330,6 +330,27 @@ fn simplify_one_expr(ctx: &GlobalCtx, state: &mut State, expr: &Expr) -> Result<
             let exprx = ExprX::Ctor(datatype, variant, binders, None);
             Ok(SpannedTyped::new(&expr.span, &expr.typ, exprx))
         }
+        ExprX::Assign { init_not_mut, lhs, rhs } => match &lhs.x {
+            ExprX::VarLoc(_) => Ok(expr.clone()),
+            ExprX::UnaryOpr(UnaryOpr::Field(_), _) => {
+                let (temp_decls, rhs) = small_or_temp(state, rhs);
```
Chris-Hawblitzel, line 336: It looks like this evaluates `rhs` before `lhs`.  In the bigger picture, though, I think this code belongs in `ast_to_sst`, which is generally in charge of order of evaluation and introducing temporary variables.

utaal: This should be address in https://github.com/secure-foundations/verus/pull/97/commits/bc7d2f3081e9305d25ff6a5281f61d896cbfda03

Chris-Hawblitzel (approved):

PR #98, opened 2022-03-21
Bit-vector Updates: fix issues, add function mode, trigger selection, examples, etc

channy412:
1. Fixed problems that cause unsoundness and failure to active triggers:  
In the `prelude`, the (integer version of) bit-wise operator's argument type is changed from `int` to `poly`. Also, it additionally takes in the bit-width of input arguments. For example, uninterpreted bitwise `and` is declared as `(declare-fun [uint_and] (Int Poly Poly) Int)`. To be specific, the normal `sst_to_air` translation of bitwise operators is done by the following steps: 1) box the arguments, 2) apply the uninterpreted bit-wise operator, 3) apply uClip/iClip according to the bit-width. 
  
2. Since the last PR, `let`, quantifier binding, `if and else` is added in the bit-vector support. The supported functionalities for bit-vector can be found in `exp_to_bv_expr` in `sst_to_air.rs`. 

3. Bit-vector function mode is added, and it only suports empty body function. It will interpret `requires` and `ensures` as bit-vector, prove in bit-vector reasoning, and port it as integer axioms.

4. Examples are added. 
`example/bitvector_garbage.rs`  contains the example that describes a 32-bit integer as an array of 16 elements.
`docs/bv_experiments/bitmaps.rs` contains the bitmap example, which is about the vector of u64, to represent arbitrary long bit-vectors. This example currently hangs on the default setting. It will work when the following option value is set as 0. `(set-option :smt.case_split 3)`
`example/bitvector_basic.rs` contains basic properties for bit-wise operators, such as commutativity, associativity, etc.

5. Automatic trigger selection for bitwise operators. A trigger containing a bitwise operator will have the lowest precedence in the selection process.


utaal:
Are you planning to merge by squashing? If not, can you please clean up the history? (There are _a lot_ of commits and reverts of merge commits, which doesn't look right.)
If instead, this is not ready for review, please mark it as draft.
Thanks!

source/rust_verify/tests/atomic_lib.rs
```diff
@@ -424,17 +424,17 @@ const ATOMIC_I64: &str = code_str! {
 
     let l = at.fetch_or(&mut perm, 2);
     assert(l == 9);
-    assert_bit_vector((9 as u64 | 2 as u64) == 11 as u64);
+    assume((9 as i64 | 2 as i64) == 11 as i64);
```
utaal, line 427: Why are these assumes now?

source/vir/src/well_formed.rs
```diff
@@ -89,8 +89,34 @@ fn check_function(ctxt: &Ctxt, function: &Function) -> Result<(), VirErr> {
...
+            return err_str(&function.span, "bit-vector function mode must be declared as proof");
+        }
+        if let Some(body) = &function.x.body {
+            crate::ast_visitor::expr_visitor_check(body, &mut |expr| {
+                match &expr.x {
+                    ExprX::Block(_, _) => {}
+                    _ => {
+                        return err_str(
+                            &function.span,
+                            "bit-vector function mode cannot have a body exression",
```
utaal, line 102: "expression"

source/vir/src/sst_to_air.rs
```diff
@@ -356,11 +356,29 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: ExprCtxt) -> Expr {
...
                 let expr = exp_to_expr(ctx, exp, expr_ctxt);
-                Arc::new(ExprX::Apply(
+                let expr = try_box(ctx, expr, &exp.typ).expect("Box");
+                let bit_expr = ExprX::Apply(
                     Arc::new(crate::def::UINT_NOT.to_string()),
-                    Arc::new(vec![expr]),
-                ))
+                    Arc::new(vec![width_exp, expr]),
+                );
+                let uclip_name = crate::def::U_CLIP;
```
utaal, line 367: I think you can inline these.

source/docs/bv_expriments/bitmap.rs
```diff
@@ -0,0 +1,178 @@
+#[allow(unused_imports)]
```
utaal, line 1: Is this something that would verify with the current implementation? If so, consider moving it to examples/ so it is run by CI too.

utaal: (Also, minor bv_expriments -> bv_experiments)

source/vir/src/triggers_auto.rs
```diff
@@ -227,54 +244,87 @@ fn gather_terms(ctxt: &mut Ctxt, ctx: &Ctx, exp: &Exp, depth: u64) -> (bool, Ter
...
+                let mut terms = vec![];
+                for e in args {
+                    let (b, u, t) = gather_terms(ctxt, ctx, &e.a, depth + 1);
+                    is_pures.push(b);
+                    num_ops.push(u);
+                    terms.push(t);
+                }
+                (is_pures, num_ops, terms)
+            };
+            // let (is_pures, terms): (Vec<bool>, Vec<Term>) =
```
utaal, line 286: Please clean up commented out code.

utaal (changes requested):
Some initial comments.
Thank you!

source/rust_verify/tests/atomic_lib.rs
tjhance: I suggested chanhee change this because these assertions aren't fundamental to the test, and verus apparently lacks the signed int reasoning necessary to make the assertion go through.

source/docs/bv_expriments/bitmap.rs
channy412: This example currently hangs on the default setting. It works fine with `(set-option :smt.case_split 0)`. (Default value is 3)
I will make sure to move this to examples when I work on multiple backend supports.

channy412:
> Some initial comments. Thank you!

Thanks a lot for the comments! I updated it in the latest commit

source/vir/src/sst_to_air.rs
```diff
@@ -742,10 +804,57 @@ fn exp_to_bv_expr(state: &State, exp: &Exp) -> Expr {
...
+                return exp_to_bv_expr(exp);
+            }
+            _ => panic!("unsupported for bit-vector: UnaryOpr {:?}", op),
+        },
+        ExpX::If(cond, tbranch, fbranch) => {
+            let cond = exp_to_bv_expr(&cond);
+            let bv_tbranch = exp_to_bv_expr(&tbranch);
+            let bv_fbranch = exp_to_bv_expr(&fbranch);
+            return Arc::new(ExprX::IfElse(cond, bv_tbranch, bv_fbranch));
+        }
```
Chris-Hawblitzel, line 821: The `ExpX::If` and `ExpX::Bind` look like duplicates of what's in `exp_to_expr`.  Can we merge `exp_to_bv_expr` and `exp_to_expr` together so that we don't duplicate these cases?  Here's what I'd suggest.  `exp_to_expr` already has a special `ExprCtxt` that gets passed around.  You can redefine `ExprCtxt` so that it is a struct with a boolean value saying whether we're processing a bit vector expression or not.  Then the match arms can test this boolean value, as in `ExpX::Unary(op, exp) if expr_ctxt.is_bit_vector_mode => { ... }`.

source/vir/src/triggers_auto.rs
```diff
@@ -39,6 +39,7 @@ enum App {
     // u64 is an id, assigned via a simple counter
     Other(u64),
     VarAt(UniqueIdent, VarAt),
+    BitOp,
```
Chris-Hawblitzel, line 42: This needs to track which bit vector operator is represented (`BitOp(...something...)`).  For example, xor can't be the same as shr here.

source/vir/src/triggers_auto.rs
```diff
@@ -195,23 +200,35 @@ fn trigger_var_depth(ctxt: &Ctxt, term: &Term, depth: u64) -> Option<u64> {
...
-    Score { depth, size: term_size(term) }
+fn make_score(term: &Term, depth: u64, num_operators: u64) -> Score {
+    Score { depth, size: term_size(term), num_operators }
 }
 
-fn gather_terms(ctxt: &mut Ctxt, ctx: &Ctx, exp: &Exp, depth: u64) -> (bool, Term) {
-    let (is_pure, term) = match &exp.x {
-        ExpX::Const(c) => (true, Arc::new(TermX::App(App::Const(c.clone()), Arc::new(vec![])))),
-        ExpX::Var(x) => (true, Arc::new(TermX::Var(x.clone()))),
+fn gather_terms(ctxt: &mut Ctxt, ctx: &Ctx, exp: &Exp, depth: u64) -> (bool, u64, Term) {
```
Chris-Hawblitzel, line 207: Rather than computing `num_operators` inside `gather_terms`, I would write a separate function on the `Term` like `term_size`.  Then most of the changes below won't be needed.

Chris-Hawblitzel (changes requested):

source/vir/src/sst_to_air.rs
```diff
@@ -356,11 +356,35 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: ExprCtxt) -> Expr {
...
                 let expr = exp_to_expr(ctx, exp, expr_ctxt);
-                Arc::new(ExprX::Apply(
+                let expr = try_box(ctx, expr, &exp.typ).expect("Box");
+                let bit_expr = ExprX::Apply(
                     Arc::new(crate::def::UINT_NOT.to_string()),
-                    Arc::new(vec![expr]),
-                ))
+                    Arc::new(vec![width_exp, expr]),
+                );
+                if let TypX::Int(range) = &*exp.typ {
```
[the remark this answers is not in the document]
channy412: Sure, I will do that

source/vir/src/sst_to_air.rs
channy412: I see, I agree that this change would make codes look much better. Thanks!

source/vir/src/triggers_auto.rs
channy412: Oh, you're right. I will make changes.

source/vir/src/triggers_auto.rs
channy412: I will add information on that. Thanks!

source/vir/src/sst_to_air.rs
```diff
@@ -312,26 +312,83 @@ pub(crate) fn constant_to_expr(_ctx: &Ctx, constant: &crate::ast::Constant) -> E
...
+                _ => panic!("unsupported for bit-vector: bind conversion, {:?} ", exp.x),
+            };
+        }
+        _ => panic!("unsupported for bit-vector: expression conversion {:?}", exp.x),
+    }
+}
+
 pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: ExprCtxt) -> Expr {
+    if expr_ctxt.is_bit_vector {
+        assert_bitvec_exp_supported(exp)
```
Chris-Hawblitzel, line 371: `exp_to_expr` is already recursing `exp`, and the call to `assert_bitvec_exp_supported` introduces a nested recursion over `exp`, which is asymptotically inefficient.  I think this only needs one recursion.

source/vir/src/triggers_auto.rs
```diff
@@ -86,7 +98,6 @@ First, we prefer triggers containing the fewest number of terms:
 - {f(x, y)} (1 term) is better (safer) than {g(x), h(y)} (2 terms)
 We choose this because a smaller number of terms leads to fewer quantifier instantiations,
 meaning less chance of an SMT timeout.
-
```
Chris-Hawblitzel, line 89: I don't know why these three blank lines got removed.  I'd prefer for them to stay.

source/vir/src/triggers_auto.rs
```diff
@@ -195,8 +205,63 @@ fn trigger_var_depth(ctxt: &Ctxt, term: &Term, depth: u64) -> Option<u64> {
     }
 }
 
-fn make_score(term: &Term, depth: u64) -> Score {
-    Score { depth, size: term_size(term) }
+fn make_score(term: &Term, depth: u64, num_operators: u64) -> Score {
+    Score { depth, size: term_size(term), num_operators }
+}
+
+fn count_bit_operators(ctx: &Ctx, exp: &Exp) -> u64 {
```
Chris-Hawblitzel, line 212: This will be much simpler if it just counts the `BitOp` terms in `Term` (similar to how `term_size` works) rather than traversing the `Exp`.

Chris-Hawblitzel (changes requested):

source/vir/src/triggers_auto.rs
channy412: Will do, sorry about that

source/vir/src/triggers_auto.rs
channy412: I see, that's a really better way. Thanks!

source/vir/src/sst_to_air.rs
Chris-Hawblitzel: I see the new commit with `assert_bitvec_exp_supported` lifted out of the recursion, which solves the performance problem.  But does `assert_bitvec_exp_supported` need to be recursive at all?  Since `exp_to_expr` is already recursing over the whole syntax tree, can we just do a nonrecursive check inside `exp_to_expr` instead having a separate recursive traversal?

channy412: Yes, of course. I will merge these two. 
(I was thinking I might lift `assert_bitvec_exp_supported` to incorporate a function that checks if an expression can flow into Singular, but on second thought, merging these two makes more sense. - traversing only once, not having to call the checking function)

Chris-Hawblitzel (approved):
I think you can go ahead and squash and merge

PR #104, opened 2022-03-23
add some well-formedness checks in pub / requires / ensures

tjhance:
A handful of well-formedness checks:

 * Check requires/ensures/decreases for private items, if the function is public (addresses https://github.com/secure-foundations/verus/issues/101)
 * Check requires/ensures/decreases for invalid uses of external_body datatypes
 * Error if a function is declared `publish` but not `pub`

I wasn't actually sure if we need to to the privacy check for `decreases` clauses though. Are `decreases` clauses exported?Can strongly connected components span module boundaries?

### DIFF
--- a/source/pervasive/atomic.rs
+++ b/source/pervasive/atomic.rs
@@ -100,12 +100,12 @@ macro_rules! atomic_types {
 
         impl $p_ident {
             #[spec] #[verifier(publish)]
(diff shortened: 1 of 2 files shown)


tjhance:
removed the changes that were made independently

utaal (approved):
Thank you!

PR #106, opened 2022-03-25
support boolean xor

tjhance:
Also, give a more informative error message for the 'not supported' cases


Chris-Hawblitzel (approved):
Interesting, I didn't realize that both Rust and Z3 have xor for the bool type.  One thing to consider: it appears that xor in Z3 is n-ary, like and/or, rather than binary.  We might want to treat Xor the same as MultiOp And/Or in AIR for consistency with Z3.

PR #111, opened 2022-03-27
Treat arithmetic overflow in non-spec code as an error rather than as uninterpreted

Chris-Hawblitzel:
This should make it easier to diagnose verification failures caused by potential integer overflow.  See https://github.com/secure-foundations/verus/issues/45 .

### DIFF
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -22,7 +22,7 @@ type Args = Arc<Vec<Arg>>;
 
 pub(crate) struct State {
     // View exec/proof code as spec
-    // (used for is_const functions, where we let Rust --compile check for overflow)
+    // (used for is_const functions, which are viewable both as spec and exec)
     view_as_spec: bool,
     // Counter to generate temporary variables
     next_var: u64,
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -1,6 +1,6 @@
 use crate::ast::{
-    BinaryOp, CallTarget, Constant, Expr, ExprX, Fun, Function, Ident, Mode, PatternX,
-    SpannedTyped, Stmt, StmtX, Typ, TypX, Typs, UnaryOp, UnaryOpr, VarAt, VirErr,
+    ArithOp, BinaryOp, CallTarget, Constant, Expr, ExprX, Fun, Function, Ident, IntRange, Mode,
+    PatternX, SpannedTyped, Stmt, StmtX, Typ, TypX, Typs, UnaryOp, UnaryOpr, VarAt, VirErr,
 };
 use crate::ast_util::{err_str, err_string};
 use crate::context::Ctx;
@@ -21,6 +21,9 @@ type Arg = (Exp, Typ);
 type Args = Arc<Vec<Arg>>;
 
 pub(crate) struct State {
+    // View exec/proof code as spec
+    // (used for is_const functions, where we let Rust --compile check for overflow)
+    view_as_spec: bool,
     // Counter to generate temporary variables
     next_var: u64,
     // Collect all local variable declarations
@@ -43,6 +46,7 @@ impl State {
         let mut rename_map = ScopeMap::new();
         rename_map.push_scope(true);
         State {
+            view_as_spec: false,
             next_var: 0,
             local_decls: Vec::new(),
             rename_map,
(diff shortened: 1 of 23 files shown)


tjhance:
could you explain & document the new ErasureMode and InferMode concepts? What problem are they solving, and how do they solve it?

source/vir/src/ast_to_sst.rs
```diff
@@ -21,6 +21,9 @@ type Arg = (Exp, Typ);
 type Args = Arc<Vec<Arg>>;
 
 pub(crate) struct State {
+    // View exec/proof code as spec
+    // (used for is_const functions, where we let Rust --compile check for overflow)
```
tjhance, line 25: Not a huge fan of the idea of relying on `--compile` for stuff like this. (If I'm a developer, I'd want to get an error for this early, before trying to verify everything else.) 

Are there other things we rely on `--compile` for? Could we make at least make a TODO item to handle this in VIR?

source/vir/src/modes.rs
```diff
@@ -32,7 +32,25 @@ pub fn mode_join(m1: Mode, m2: Mode) -> Mode {
...
+// Here, e + 1 < y is erased because the "<" comparison with s is spec, not exec.
+// Therefore, the e + 1 is a spec-mode arithmetic operation, not an exec-mode arithmetic operation.
+// However, the mode checker doesn't know this when it first descends into e + 1:
+//   - the "if" expression can take a spec or proof or exec condition, so this doesn't constrain
+//     the mode of e + 1 < s a priori
+//   - the "<" expression can take spec or proof or exec subexpressions, so this doesn't constrain
+//     the mode of e + 1 a priori
+//   - we check the left side e + 1 before we check the right side s,
+//     so when we first reach e + 1, we don't yet know that "<"'s right side is spec
+// Therefore, we use Cell to delay the decision about the erasure expected mode of e + 1.
```
tjhance, line 53: so is it right to say that the "erasure expected mode" is really the actual "final decision" about the mode? 

And the mode returned by `check_expr` is more like a "pre-mode" or an approximation of the mode in some circumstances?

source/vir/src/modes.rs
```diff
@@ -32,7 +32,25 @@ pub fn mode_join(m1: Mode, m2: Mode) -> Mode {
...
-// placeholder for the erasure expected mode, which is computed after first mode checking pass
+// Placeholder for the erasure expected mode, which is computed after first mode checking pass.
+// (See in erase_expr(..., expect: Mode, ...) in erase.rs.)
+// We compute the erasure expected mode here so that the AIR/SMT generation can use it.
+// For example, if there are arithmetic operations that are exec (and therefore not erased),
+// then the AIR/SMT code will insert overflow checks for those arithmetic operations.
+// Example:
+//   fn test(#[exec] e: u64, #[spec] s: u64) {
+//     if e + 1 < s { ... }
+//   }
```
tjhance, line 43: I don't wanna hold up the PR with too much design discussion - but I wonder if we should require the user to be explicit in this case? e.g., if they want to create a `spec` if statement in the middle of `exec` code, they should have to write it like this:

```rust
#[proof] if e + 1 < s { ... }
```

Not just to make the mode-checking code easier, but also to keep user code clearer and more explicit

source/vir/src/modes.rs
Chris-Hawblitzel: The mode returned by `check_expr` is the tightest possible mode that the expression can have (if both exec and spec are possible, return exec).  Outer expressions might relax this mode (relax exec to proof, relax exec to spec, or relax proof to spec) via the "subsumption rule" described in https://github.com/Chris-Hawblitzel/rust/wiki/Three-kinds-of-code-...-specification,-proof,-and-executable .  This is like subtyping, where a type checker computes the most precise subtype for an expression, but might later relax this type to a supertype like `Object`.

Because operations like + and < can operate in different modes, there's more than one possible "final decision" that we could choose.  The erasure chooses a decision biased towards spec whenever possible in order to erase as much code as possible.  So you could call the erasure decision the "final decision" about the mode.  The checker doesn't care, but the erasure does.

source/vir/src/modes.rs
Chris-Hawblitzel: Vale does it this way (you have to write "ghost if").  Dafny prefers to infer the ghostness.  I think we should have this discussion, but it's a broader discussion.  For example, it could be argued that _all_ proof code inside an exec function has to be marked as #[proof], even just calls to lemmas, to make the code easier for people to read and understand.  But then again, maybe not.  We should probably start a discussion page.

source/vir/src/ast_to_sst.rs
Chris-Hawblitzel: Ok, this is now fixed.

tjhance (approved):

utaal (approved):

PR #126, opened 2022-04-20
verifier(non_linear) attribute

channy412:
This PR simply adds `#[verifier(non_linear)]`, which is to set `smt.arith.nl=true` only for the annotated function.
This is expected to be useful for functions with much non-linear arithmetic operations.

(In another branch, I am also working on single line non-linear assertion:  `assert_by_nonlinear`, which is `assert_by` with `smt.arith.nl=true`)

### DIFF
--- a/source/rust_verify/tests/non_linear.rs
+++ b/source/rust_verify/tests/non_linear.rs
@@ -1,0 +1,94 @@
+// Some testcases are ported from https://github.com/secure-foundations/libraries/tree/master/src/NonlinearArithmetic
+
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+// TODO: add testcases for assert_by_nonlinear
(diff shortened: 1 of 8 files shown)


source/rust_verify/tests/non_linear.rs
```diff
@@ -0,0 +1,94 @@
...
+        #[verifier(non_linear)]
+        #[proof]
+        fn wrong_lemma_1(x: int, y: int, z: int) {
+            requires([
+                x <= y,
+                0 <= z,
+            ]);
+            ensures (x*z < y*z);
+        }
+    } => Err(_)
```
utaal: It'd be better if this was more precise. You can use `=> Err(e) => assert_one_fails(e)` to check that a specific line caused a proof failure. I pushed a commit that makes these two tests more precise this way. Consider doing the same in the future.

utaal:
Thank you!

PR #131, opened 2022-04-22
`assert_nonlinear_by`

utaal:
This feature was developed by @channy412.

Also renames `#[verifier(non_linear)]` to `#[verifier(nonlinear)]` for consistency with `assert_nonlinear_by`.

---

* [x] Rlimit exceeded reporting doesn't quite work correctly for `assert_by_forall`, because we don't have a span for it.

### DIFF
--- a/source/vir/src/well_formed.rs
+++ b/source/vir/src/well_formed.rs
@@ -111,35 +111,27 @@ fn check_one_expr(
                 referenced.extend(referenced_vars_expr(r).into_iter());
             }
 
-            fn check(expr: &Expr, referenced: &HashSet<crate::ast::Ident>) -> Result<(), VirErr> {
-                use crate::visitor::VisitorControlFlow;
+            use crate::visitor::VisitorControlFlow;
 
-                match crate::ast_visitor::expr_visitor_dfs(
-                    expr,
-                    &mut crate::ast_visitor::VisitorScopeMap::new(),
-                    &mut |scope_map, e| match &e.x {
-                        ExprX::Var(x) | ExprX::VarLoc(x) => {
-                            match scope_map.scope_and_index_of_key(&x) {
-                                None if !referenced.contains(x) => VisitorControlFlow::Stop(error(
-                                    format!("variable {} not mentioned in requires", x).as_str(),
-                                    &e.span,
-                                )),
-                                _ => VisitorControlFlow::Recurse,
-                            }
-                        }
-                        _ => VisitorControlFlow::Recurse,
-                    },
-                ) {
-                    VisitorControlFlow::Recurse => Ok(()),
-                    VisitorControlFlow::Return => unreachable!(),
-                    VisitorControlFlow::Stop(e) => Err(e),
-                }
-            }
-
-            check(proof, &referenced)?;
-            for ens in ensures.iter() {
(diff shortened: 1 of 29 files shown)


utaal:
* [ ] Rlimit exceeded reporting doesn't quite work correctly for `assert_by_forall`, because we don't have a span for it.

channy412:
`assert_by_nonlinear` imports the surrounding context using `requires` inside the second argument. In a separate query, it assumes type invariants and `requires`. Then it tries to prove the first argument using `proof` inside the second argument (with the option `smt.arith.nl=true`.)

For example, 
```
assert_by_nonlinear( (x as int) * (z as int) == ((x*z) as int), {
                requires([
                    x <= 0xffff,
                    z <= 0xffff,
                ]);
                assert(0 <= (x as int) * (z as int));
                assert((x as int) * (z as int) <= 0xffff*0xffff);
            });
```

In the main query, it assumes `requires` imply `first argument`.

channy412:
I found an example that gives different Z3 results when two non-linear queries have different type invariant ordering. It is fixed in 2ec028b994ac4462d06e9514ada5d39b3e35b66c ,but variable renaming can still change the type invariant ordering. (Since I sorted them with ident strings)
Can I change the `declared` (in `vir::sst_vars::stm_assign`) from `HashMap` to `BTreeMap` to solve this?

source/vir/src/well_formed.rs
```diff
@@ -94,6 +95,52 @@ fn check_one_expr(
...
+                use crate::visitor::VisitorControlFlow;
+
+                match crate::ast_visitor::expr_visitor_dfs(
+                    expr,
+                    &mut crate::ast_visitor::VisitorScopeMap::new(),
+                    &mut |scope_map, e| match &e.x {
+                        ExprX::Var(x) | ExprX::VarLoc(x) => {
+                            match scope_map.scope_and_index_of_key(&x) {
+                                None if !referenced.contains(x) => VisitorControlFlow::Stop(error(
+                                    format!("variable {} not mentioned in requires", x).as_str(),
```
Chris-Hawblitzel, line 124: This says "not mentioned in requires", but it looks like `referenced` contains both requires and ensures, not just requires, because of `for r in ensures.iter()` above.  So does the check `check(ens, &referenced)?` below do anything?  If not, you're just checking that `proof` doesn't mention extra variables?  Why does this matter?

Chris-Hawblitzel (approved):

source/vir/src/well_formed.rs
utaal: The ensure check was indeed spurious (leftover from a more restrictive incorrect check I wrote before). The point of this is to reject the following:
```
        #[proof]
        fn test6(x: int, y: int, z:int) {
            requires(y == 0);
            requires(z == 0);
            assert_nonlinear_by({
                requires(y == 0);
                ensures(x + y == x);
                assert(z == 0);
            });
        }
```
as I believe this is unlikely to be what the user intends, as the nonlinear query has no facts about `z`. Maybe this is superfluous?

utaal: (With the current implementation we wouldn't import typing invariants for variables only mentioned in the body.)

PR #136, opened 2022-04-25
Singular support for Verus 

channy412:
When a `proof` function is annoted with `verifier(integer_ring)`, Singular uses the Groebner basis for the ideal membership problem. This can be used only with `int` parameters. (Bounded integers are not supported now due to possible unsoundness).

For example, consider
```
    #[verifier(integer_ring)]
    #[proof]
    fn test2(a:int, s:int, R:int, M:int, RR:int, R_INV:int) {
        requires([
            (a * R - RR * s) % M == 0,
            (R_INV * R - 1) % M == 0,
            (RR - R * R % M) == 0,
        ]);
        ensures( (a - s*R) % M == 0);
    }
```
First, it registers all polynomials from requires clauses as a basis. These basis forms an ideal, and the Groebner basis is computed from it. If `ensures` polynomial is reduced to 0, it returns `Valid`, otherwise, it returns `invalid`.

This also includes some changes in CI. Test cases that use Singular are marked as `ignore`, but CI will run this every time.

(This PR is expected to be merged after https://github.com/secure-foundations/verus/pull/131)


source/air/src/printer.rs
```diff
@@ -495,3 +495,166 @@ impl NodeWriter {
...
+pub(crate) fn expr_to_singular(
+    expr: &Expr,
+    tmp_idx: &mut u32,
+    fmap: &mut HashMap<Node, String>,
+) -> String {
+    match &**expr {
+        ExprX::Const(Constant::Nat(n)) => n.to_string(),
+        ExprX::Var(x) => x.to_string(),
+        ExprX::Binary(BinaryOp::EuclideanMod, lhs, rhs) => {
+            // x % y ->  x - y*tmp
```
yizhou7, line 525: might want to use consistent temp variables for the same x and y

source/air/src/printer.rs
```diff
@@ -495,3 +495,166 @@ impl NodeWriter {
...
+                MultiOp::Mul => ") * (",
+                _ => panic!("unsupported singular multi op"),
+            };
+            for e in &**exprs {
+                ss.push(expr_to_singular(&e, tmp_idx, fmap));
+            }
+            format!("(({}))", ss.join(sop))
+        }
+        ExprX::Apply(fname, exprs) => {
+            if "Mul" == (**fname).as_str() {
```
yizhou7, line 551: maybe just do the translation in place, same applies to `EucMod`.

source/vir/src/well_formed.rs
```diff
@@ -306,6 +318,70 @@ fn check_function(ctxt: &Ctxt, function: &Function) -> Result<(), VirErr> {
         }
     }
 
+    if function.x.attrs.integer_ring {
+        if function.x.mode != Mode::Proof {
+            return err_str(&function.span, "integer_ring mode must be declared as proof");
+        }
+        if let Some(body) = &function.x.body {
+            crate::ast_visitor::expr_visitor_check(body, &mut |expr| {
+                match &expr.x {
```
yizhou7, line 327: we might want to check the function calls in the pre/post conditions don't return bounded integers.

source/air/src/printer.rs
channy412: You're right, thank you!

source/air/src/printer.rs
channy412: That's a good point, but after changing codes in that way, I just learned that constant multiplication still ends in that old path. So I think I will keep it this way.

source/vir/src/well_formed.rs
channy412: Thanks! I added more checks

source/air/src/ast.rs
```diff
@@ -181,4 +181,5 @@ pub enum CommandX {
     SetOption(Ident, Ident), // set-option option value (no colon on the option)
     Global(Decl),            // global declarations
     CheckValid(Query),       // SMT check-sat (reporting validity rather than satisfiability)
+    SingularCheckValid(Vec<String>, Vec<Expr>, Vec<Expr>, Span), // check integer-ring proof using Singular: SingularCheckValid{params, ensures, requires, span}
```
utaal, line 184: I'd rather we didn't special-case on singular here.
* Why cannot this carry a `Query`? Can we define a `Query` type for singular, if there are differences?
* Depending on that we may want a separate variant (for solvers with a different kind of Query) or a `Solver` field on `CheckValid`.

source/air/src/context.rs
```diff
@@ -33,9 +35,11 @@ pub(crate) struct AxiomInfo {
 pub enum ValidityResult {
     Valid,
     Invalid(Model, Error),
+    SingularInvalid(Error),
```
utaal, line 38: Again, I don't like special-casing, as we'll need to handle them one by one.

Is there a reason why this cannot reuse `Invalid`? Maybe with an optional `Model` on `Invalid`?

source/air/src/context.rs
```diff
@@ -33,9 +35,11 @@ pub(crate) struct AxiomInfo {
...
     Canceled,
     TypeError(TypeError),
     UnexpectedSmtOutput(String),
+    UnexpectedSingularOutput(String),
```
utaal, line 42: Can we turn these both into `UnexpectedOutput`?

source/air/src/context.rs
```diff
@@ -314,6 +318,47 @@ impl Context {
...
+    ) -> ValidityResult {
+        let nodes = singular_query_to_node(vars.to_vec(), enss.to_vec(), reqs.to_vec());
+        self.air_initial_log.log_node(&nodes);
+        self.air_middle_log.log_node(&nodes);
+        self.air_final_log.log_node(&nodes);
+
+        let query = singular_printer(vars, enss, reqs);
+        self.air_initial_log.comment(&query);
+        self.air_middle_log.comment(&query);
+        self.air_final_log.comment(&query);
```
utaal, line 336: I'm confused as what's happening here? Is this appending the singular query as a comment to the air file?
If so, why have the vars, requires, and ensures been printed as s-expr nodes? Is z3 supposed to see them?

source/air/src/main.rs
```diff
@@ -126,6 +126,10 @@ pub fn main() {
             ValidityResult::TypeError(err) => {
                 panic!("Type error: {}", err);
             }
+            ValidityResult::SingularInvalid(..) => {
+                count_errors += 1;
+                println!("SingularInvalid in air main - Singular error at: TODO");
```
utaal, line 131: There's a leftover TODO here, which I imagine is turning this into proper error reporting instead of a println.

source/air/src/main.rs
```diff
@@ -140,10 +144,16 @@ pub fn main() {
...
+            ValidityResult::UnexpectedSingularOutput(err) => {
+                panic!("Unexpected Singular output: {}", err);
+            }
         }
         if matches!(**command, CommandX::CheckValid(..)) {
             air_context.finish_query();
         }
+        if matches!(**command, CommandX::SingularCheckValid(..)) {
+            air_context.finish_query();
+        }
```
utaal, line 156: As discussed above, this is redundant with 151-153 due to having a separate CommandX variant.

source/air/src/printer.rs
```diff
@@ -495,3 +495,182 @@ impl NodeWriter {
 pub(crate) fn node_to_string(node: &Node) -> String {
     NodeWriter::new().node_to_string_indent(&"".to_string(), node)
 }
+
+pub fn singular_query_to_node(vars: Vec<String>, enss: Vec<Expr>, reqs: Vec<Expr>) -> Node {
```
utaal, line 499: Is the output of this actually ever used by any of the solvers?

source/air/src/printer.rs
```diff
@@ -495,3 +495,182 @@ impl NodeWriter {
...
+        }
+        ExprX::Binary(BinaryOp::Eq, lhs, rhs) => {
+            let s1 = expr_to_singular(lhs, tmp_idx, node_map);
+            let s2 = expr_to_singular(rhs, tmp_idx, node_map);
+            format!("({}) - ({})", s1, s2)
+        }
+        ExprX::Multi(op, exprs) => {
+            let mut ss = vec![];
+            let sop = match op {
+                MultiOp::Add => ") + (",
```
utaal, line 551: Are these parenthesis starting/ending sub-expressions? Who's opening the first and closing the second?

source/air/src/printer.rs
```diff
@@ -495,3 +495,182 @@ impl NodeWriter {
...
+                    None => {
+                        let tmp_new = format!("tmp_{}", tmp_idx.to_string());
+                        *tmp_idx += 1;
+                        node_map.insert(key, tmp_new.to_string());
+                        tmp_new
+                    }
+                }
+            }
+        }
+        _ => panic!("unsupported singular expression: {:?}", expr),
```
utaal, line 597: Can we fail more gracefully here? Or at least print a span? That may be easier if this happens in `rust_verify`. (Unless this is a redundant check of one done in an upper layer.)

source/docs/non_linear_arithmetic/non_linear_basic.rs
```diff
@@ -0,0 +1,136 @@
+// testcases and examples are ported from below two files
```
utaal, line 1: What's the purpose of this file? Is there a specific reason it isn't part of the tests?

source/rust_verify/src/verifier.rs
```diff
@@ -253,6 +253,7 @@ impl Verifier {
             (std::time::Duration::from_secs(2), report_fn)
         };
         let is_check_valid = matches!(**command, CommandX::CheckValid(_));
+        let is_singular = matches!(**command, CommandX::SingularCheckValid(..));
```
utaal, line 256: How are recommends checks handled in singular? Are they ignored?

source/vir/src/sst_to_air.rs
```diff
@@ -1621,26 +1622,60 @@ pub fn body_stm_to_air(
...
-        ]
-    } else if is_bit_vector_mode {
-        vec![
-            mk_option_command("smt.case_split", "0"),
-            Arc::new(CommandX::CheckValid(query)),
-            mk_option_command("smt.case_split", "3"),
-        ]
+    if is_integer_ring {
+        // parameters, requires, ensures to Singular Query
+        let mut vars: Vec<String> = vec![];
```
utaal, line 1627: Could these be AIR `Ident`s?

source/vir/src/well_formed.rs
```diff
@@ -306,6 +318,102 @@ fn check_function(ctxt: &Ctxt, function: &Function) -> Result<(), VirErr> {
...
+                TypX::Boxed(_) => {}
+                _ => {
+                    return err_str(&p.span, "integer_ring mode's param should be int type");
+                }
+            }
+        }
+        if function.x.ensure.len() != 1 {
+            return err_str(
+                &function.span,
+                "integer_ring mode: single ensure expression is assumed now",
```
utaal, line 348: "only a single ensures is allowed in integer_ring mode"

source/vir/src/well_formed.rs
```diff
@@ -306,6 +318,102 @@ fn check_function(ctxt: &Ctxt, function: &Function) -> Result<(), VirErr> {
...
+                    _,
+                ) = &lhs.x
+                {
+                    match &rhs.x {
+                        crate::ast::ExprX::Const(crate::ast::Constant::Nat(zero))
+                            if "0" == (**zero).as_str() => {}
+                        _ => {
+                            return err_str(
+                                &function.span,
+                                "integer_ring mode ensure expression: `X % m == 0` or `X == Y` ",
```
utaal, line 365: Can you expand on this error message a bit?

source/vir/src/well_formed.rs
```diff
@@ -306,6 +318,102 @@ fn check_function(ctxt: &Ctxt, function: &Function) -> Result<(), VirErr> {
...
+                            );
+                        }
+                    }
+                }
+            } else {
+                return err_str(
+                    &function.span,
+                    "integer_ring mode ensures expression: operator `==` is assumed",
```
utaal, line 373: What do you mean that's assumed? That it is required?

utaal (changes requested):
I think you should squash all the changes now and force-push to this branch, as it will be hard to keep track of review items otherwise.

I may have more comments on the Singular emitter as I take a more in-depth look.

utaal:
Considering the code is mostly separate from the main lowering logic, the easiest way to make singular an optional dependency may be to use a rust "feature" to gate the code and the tests: https://doc.rust-lang.org/cargo/reference/features.html

channy412:
@utaal Thank you very much for the detailed review! I will try to resolve them one by one.

source/air/src/ast.rs
channy412: I can make it fit into `Query` and make an additional boolean field(`is_singular`) in `QueryX`.

source/air/src/context.rs
channy412: Sure, I will try with `Option<Model>`

source/air/src/context.rs
channy412: I thought air log as an incorporated log file for all solvers, and imagined invoking Singular when `singular-check-valid` is encountered.  Would it be better to log the Singular part into a separate file?

source/air/src/printer.rs
channy412: I guess I didn't use vir::def::MUL, because air does not depend on vir. (If I move these to `rust_verify` this would be solved as well)

source/docs/non_linear_arithmetic/non_linear_basic.rs
channy412: I moved to docs because this lemma `LemmaMulUpperBound` failed on my machine, but it seems to work in the CI machine. (I will move to `examples`.)

Since `integer_ring` is available with `int`, users need to check bounds when they want to prove properties about bounded integers. That file is to show an example of that usage.

source/rust_verify/example/integer_ring.rs
```diff
@@ -0,0 +1,125 @@
+// most of the codes are copied from: https://github.com/utaal/verified-nrkernel/blob/main/page-table/lib.rs
```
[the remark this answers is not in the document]
channy412: Thanks, will do!

source/vir/src/well_formed.rs
channy412: I wanted to say that the most outer operator for `ensures` expression should be the equality operator  (not any functions that return boolean, especially shouldn't be inequalities)

source/rust_verify/src/verifier.rs
channy412: I think the recommends check will simply run the same thing on Singular.

source/air/src/printer.rs
```diff
@@ -495,3 +495,182 @@ impl NodeWriter {
...
+    }
+
+    // Using @ is safe. For example, `poly g1 = x2+y3` is translated as poly `g1 = x^2 + y^3`
+    for mut v in vars {
+        v.push('@');
+        vars2.push(v);
+    }
+
+    // gather polynomials that will be the basis of ideal
+    // for requires,  equality -> register ideal (rhs - lhs)
```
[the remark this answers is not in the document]
channy412: I agree that those parts will be better at `rust_verify`.   Thanks again for the reviews!

channy412:
I refactored many parts of this singular support. 

First, I expanded `CommandsWithContext`'s `is_spinoff` to be an enum, to indicate whether it is for default z3, spinoff z3, or Singular. I also merged singular commands with SMT commands.

I moved out the main logic part into `rust_verify`, and subsequently made it into a `singular` feature for conditional compilation. 

Instead of using the AIR log file, I moved out the logging part into a `.singular` log file, which can be directly run with Singular. (Similar to running `root.smt2` with z3 directly)

Then I removed most of the `panic` during the translation logic into proper error reporting.

utaal (approved):
This is great. Thanks a lot for the changes!

PR #139, opened 2022-04-27
Run some of the missing rustc checks that are currently skipped for spec or all code

utaal:
because they are in stages that are after where we stop rustc.

This breaks a lot of tests/examples. It may also be that it overrides some of our typechecking changes, so it may not be mergeable right now. However critical because it may uncover unsoundness.

* [x] try to fix tests


utaal:
May fix https://github.com/secure-foundations/verus/issues/127

utaal:
Thanks @Chris-Hawblitzel for spotting that the checks need to be after `typeck`.

PR #140, opened 2022-04-28
add some macros to the pervasive lib

tjhance:
some macros to construct boilerplate for common idioms

 * extensionality proofs for maps, sets, seqs
 * proof-by-contradiction

### DIFF
--- a/source/pervasive/set_lib.rs
+++ b/source/pervasive/set_lib.rs
@@ -143,3 +143,39 @@ pub fn lemma_int_range(lo: int, hi: int) {
         assert(set_int_range(lo, hi - 1).insert(hi - 1).ext_equal(set_int_range(lo, hi)));
     }
 }
+
+/// Prove two sets equal by extensionality. Usage is:
+///
+/// ```rust,ignore
+/// set_ext!(set1, set2);
+/// ```
+/// 
+/// or,
+/// 
+/// ```rust,ignore
+/// seq_ext!(set1, set2, elem => {
+///     // prove that set1.contains(elem) iff set2.contains(elem)
+/// });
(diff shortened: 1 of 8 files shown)


source/pervasive/set_lib.rs
```diff
@@ -143,3 +143,39 @@ pub fn lemma_int_range(lo: int, hi: int) {
...
+    };
+    ($s1:expr, $s2:expr, $elem:ident $( : $t:ty )? => $bblock:block) => {
+        let s1 = $s1;
+        let s2 = $s2;
+        ::builtin::assert_by(::builtin::equal(s1, s2), {
+            ::builtin::assert_forall_by(|$elem $( : $t )?| {
+                ::builtin::ensures(
+                    (s1.contains($elem) >>= s2.contains($elem))
+                    &&
+                    (s2.contains($elem) >>= s1.contains($elem))
```
Chris-Hawblitzel, line 174: Minor comment: isn't this just `s1.contains($elem) == s2.contains($elem)`?

Chris-Hawblitzel (approved):
Should the names `map_ext`, `set_ext`, `seq_ext` follow the conventions of `assert_forall_by`, `assert_by_contradiction`, `assert_by_nonlinear`, etc.?  For example, `assert_map_eq_by` or `assert_maps_equal_by`?

source/pervasive/set_lib.rs
tjhance: It is. I thought more fine grained ensures would help the user, though, since in practice one often needs to show both directions. (However it's a moot point now, since the forall block only allowed one 'ensures' clause.)

PR #144, opened 2022-05-09
Add `spinoff_z3` attribute

channy412:
When a function is annotated with `verifier(spinoff_z3)`, a dedicated Z3 process is spawned for that function's query only. 
Z3 is restarted at every module boundary by default, and this `spinoff_z3` attribute can be used for further separation.
The spun-off Z3 process imports prelude, as well as the current module's whole context. 

  
Not included in this pr:
We might consider Non-linear/bit-vector queries to be spinoff automatically. 
We might consider pruning out the current module's context for the dedicated new Z3.

- [x] Add more testcases

### DIFF
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -146,6 +146,8 @@ pub(crate) enum Attr {
     CheckRecommends,
     // set smt.arith.nl=true
     NonLinear,
+    // Use a new dedicated Z3 process just for this query
+    SpinoffZ3,
 }
 
 fn get_trigger_arg(span: Span, attr_tree: &AttrTree) -> Result<u64, VirErr> {
@@ -282,6 +284,9 @@ pub(crate) fn parse_attrs(attrs: &[Attribute]) -> Result<Vec<Attr>, VirErr> {
                 Some(box [AttrTree::Fun(_, arg, None)]) if arg == "nonlinear" => {
                     v.push(Attr::NonLinear)
                 }
+                Some(box [AttrTree::Fun(_, arg, None)]) if arg == "spinoff_z3" => {
+                    v.push(Attr::SpinoffZ3)
+                }
                 _ => return err_span_str(*span, "unrecognized verifier attribute"),
             },
             _ => {}
@@ -372,6 +377,7 @@ pub(crate) struct VerifierAttrs {
     pub(crate) decreases_by: bool,
     pub(crate) check_recommends: bool,
     pub(crate) nonlinear: bool,
+    pub(crate) spinoff_z3: bool,
 }
 
 pub(crate) fn get_verifier_attrs(attrs: &[Attribute]) -> Result<VerifierAttrs, VirErr> {
@@ -395,6 +401,7 @@ pub(crate) fn get_verifier_attrs(attrs: &[Attribute]) -> Result<VerifierAttrs, V
         decreases_by: false,
         check_recommends: false,
         nonlinear: false,
+        spinoff_z3: false,
     };
     for attr in parse_attrs(attrs)? {
         match attr {
@@ -419,6 +426,7 @@ pub(crate) fn get_verifier_attrs(attrs: &[Attribute]) -> Result<VerifierAttrs, V
             Attr::DecreasesBy => vs.decreases_by = true,
             Attr::CheckRecommends => vs.check_recommends = true,
             Attr::NonLinear => vs.nonlinear = true,
+            Attr::SpinoffZ3 => vs.spinoff_z3 = true,
             _ => {}
         }
     }
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -422,10 +422,10 @@ impl Verifier {
         ctx: &vir::context::Ctx,
         module_path: &vir::ast::Path,
         function_path: &vir::ast::Path,
-        datatype_commands: Vec<Arc<CommandX>>,
-        function_decl_commands: Vec<(Commands, String)>,
-        function_spec_commands: Vec<(Commands, String)>,
-        function_axiom_commands: Vec<(Commands, String)>,
+        datatype_commands: Arc<Vec<Arc<CommandX>>>,
+        function_decl_commands: Arc<Vec<(Commands, String)>>,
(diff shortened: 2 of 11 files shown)


utaal:
One thing I noticed while trying this out: the air log for the spunoff z3 doesn't have a header for the `check-valid`, which makes it trickier to figure out what was checked. Can you restore that?

channy412:
> One thing I noticed while trying this out: the air log for the spunoff z3 doesn't have a header for the `check-valid`, which makes it trickier to figure out what was checked / find the point of interest. Can you restore that?

Oh, thanks for pointing that out! I am on it now. During that, you can also find the spunoff query's span at the first line of the logs.

utaal:
Hi @channy412, is this something that's close to be merge-able?

One thing we may want to improve first are the filenames of the extra `.air` files, as they are uninformative right now, I believe.

channy412:
> @channy412: is this something that's close to be merge-able?
> 
> One thing we may want to improve first are the filenames of the extra `.air` files, as they are uninformative right now, I believe.


Thanks for the comment!  Maybe I will try to include the function's name in the file name. Also, I will add some more test cases, and then mark this as ready for review. Thanks!

utaal (changes requested):

source/rust_verify/src/verifier.rs
```diff
@@ -389,13 +390,33 @@ impl Verifier {
             air_context.blank_line();
             air_context.comment(comment);
         }
-        for CommandsWithContextX { span, desc, commands } in commands.iter().map(|x| &**x) {
+        let mut z3_idx = 0;
```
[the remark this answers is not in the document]
channy412: I agree with you, I will try to reformat the codes. Thanks!!

source/rust_verify/src/verifier.rs
```diff
@@ -591,16 +683,48 @@ impl Verifier {
...
+                    if *spinoff_z3 {
+                        let mut spinoff_air_context = self.new_air_context_with_module_context(
+                            ctx,
+                            module,
+                            &(function.x.name).path,
+                            datatype_commands.to_vec(),
+                            function_decl_commands.to_vec(),
+                            function_spec_commands.to_vec(),
+                            function_axiom_commands.to_vec(),
+                            false,
```
utaal, line 701: Should this be set if this is a rerun? Currently the recommends query air file overwrites the initial spunoff query air.

source/rust_verify/src/verifier.rs
```diff
@@ -591,16 +683,48 @@ impl Verifier {
...
+                    let command_invalidity: bool;
+                    if *spinoff_z3 {
+                        let mut spinoff_air_context = self.new_air_context_with_module_context(
+                            ctx,
+                            module,
+                            &(function.x.name).path,
+                            datatype_commands.to_vec(),
+                            function_decl_commands.to_vec(),
+                            function_spec_commands.to_vec(),
+                            function_axiom_commands.to_vec(),
```
utaal, line 700: This copies all the vectors for every `self.new_air_context_with_module_context` call, which doesn't seem ideal. Can we pass a reference / clone a wrapper `Arc` instead?

source/rust_verify/src/verifier.rs
```diff
@@ -591,16 +683,48 @@ impl Verifier {
...
+                        command_invalidity = self.run_commands_queries(
+                            compiler,
+                            error_as,
+                            air_context,
+                            command.clone(),
+                            &HashMap::new(),
+                            &snap_map,
+                            &(s.to_string() + &fun_as_rust_dbg(&function.x.name)),
+                        );
+                    }
```
utaal, line 723: What about something like:
```rust
let mut spinoff_z3_context;
let query_air_context = if *spinoff_z3 {
    spinoff_z3_context = self.new_air_context_with_module_context(
        ctx,
        module,
        &(function.x.name).path,
        &*datatype_commands,
        &function_decl_commands,
        &function_spec_commands,
        &function_axiom_commands,
        recommends_rerun,
        &span,
    )?;
    &mut spinoff_z3_context
} else {
    &mut *air_context
};
command_invalidity = self.run_commands_queries(
    compiler,
    error_as,
    query_air_context,
    command.clone(),
    &HashMap::new(),
    &snap_map,
    &(s.to_string() + &fun_as_rust_dbg(&function.x.name)),
);
```
to avoid the redundant call?

utaal (changes requested):

source/rust_verify/src/verifier.rs
channy412: Oh, that is my mistake. Thank you!

source/rust_verify/src/verifier.rs
channy412: Yes, that sounds good. Thanks!

utaal (approved):
Thank you!

source/rust_verify/src/verifier.rs
```diff
@@ -394,30 +402,113 @@ impl Verifier {
...
+            );
+            invalidity = invalidity || result_invalidity;
+            let time1 = Instant::now();
+            self.time_air += time1 - time0;
         }
+
         invalidity
     }
 
+    fn new_air_context_with_module_context(
```
Chris-Hawblitzel, line 429: This function copied-and-pasted a lot of code from `verify_module_outer`.  Could you consolidate these so that there is only one copy?

source/rust_verify/src/attributes.rs
```diff
@@ -282,6 +284,9 @@ pub(crate) fn parse_attrs(attrs: &[Attribute]) -> Result<Vec<Attr>, VirErr> {
                 Some(box [AttrTree::Fun(_, arg, None)]) if arg == "nonlinear" => {
                     v.push(Attr::NonLinear)
                 }
+                Some(box [AttrTree::Fun(_, arg, None)]) if arg == "spinoff_z3" => {
```
Chris-Hawblitzel, line 287: This isn't necessarily specific to Z3.  Should the option be named something like `spinoff_prover`?

source/rust_verify/src/verifier.rs
channy412: I agree, I will try to restructure the codes. Thank you!

PR #146, opened 2022-05-13
Small expansion to the Vec API

parno:
Add a few more items to the API.  @tjhance independently added `empty` while I added `new`.  I left them both in for now, but we could consider consolidating them.

### DIFF
--- a/source/pervasive/vec.rs
+++ b/source/pervasive/vec.rs
@@ -16,10 +16,29 @@ impl<A> Vec<A> {
     fndecl!(pub fn view(&self) -> Seq<A>);
 
     #[verifier(external_body)]
+    pub fn new() -> Self {
+        ensures(|v:Vec<A>| v.len() == 0);
+
+        Vec { vec: std::vec::Vec::new() }
+    }
+    
     pub fn empty() -> Self {
         ensures(|v: Self| equal(v.view(), Seq::empty()));
 
-        Vec { vec: std::vec::Vec::new() }
+        new()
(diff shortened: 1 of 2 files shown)


tjhance:
>  @tjhance independently added empty while I added new. I left them both in for now, but we could consider consolidating them.

`empty()` matches `Seq::empty()`, while `new()` matches the standard library `Vec::new()`. no strong feelings either way

source/pervasive/vec.rs
```diff
@@ -16,11 +16,30 @@ impl<A> Vec<A> {
...
+    pub fn push(&mut self, value: A) {
+        ensures(equal(self.view(), old(self).view().push(value)));
+        self.vec.push(value);
+    }
+
+    #[verifier(external_body)]
+    pub fn pop(&mut self) -> A {
+        requires(old(self).len() > 0);
+        ensures(|value| equal(self.view().push(value), old(self).view()));
+        self.vec.pop().unwrap()  // Safe to unwrap given the precondition above
```
tjhance, line 41: should we use [unwrap_unchecked](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_unchecked)?

Chris-Hawblitzel (approved):

source/pervasive/vec.rs
parno: Seems reasonable to me.  I didn't know that was a thing.  Thanks!

PR #149, opened 2022-05-13
Add --verify-function option

Chris-Hawblitzel:
Verify just a single function.

Example: rust_verify ... vectors.rs --verify-root --verify-function binary_search

### DIFF
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -29,6 +29,7 @@ pub struct Args {
     pub pervasive_path: Option<String>,
     pub verify_root: bool,
     pub verify_module: Option<String>,
+    pub verify_function: Option<String>,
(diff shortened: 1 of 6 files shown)


parno:
Does `--verify-function` require `--verify-root`, or are they orthogonal?

Chris-Hawblitzel:
> Does `--verify-function` require `--verify-root`, or are they orthogonal?

It requires `--verify-root` or `--verify-module`.  It will print an error if one of these is not specified.  The name of the function is relative to the module being verified.

parno:
Cool, got it -- thanks!

utaal (approved):
With these changes https://github.com/secure-foundations/verus/pull/150 if they look good to you (merge https://github.com/secure-foundations/verus/pull/150 into this first).

PR #152, opened 2022-05-17
`assert_bitvector_by`

channy412:
One pain for `assert_bit_vector` is that it is not helpful for proving properties about local variables. Although it can be used as `assert_bit_vector( (local_var == definition) >>= (assert_expr))`, this give bad readability, and sometimes made me error-prone around parenthesis(especially when `assert_expr` has `&&` or `||`). Although this pain can be relieved by using bit-vector function mode, making the lemma function can be tedious work.

`assert_bitvector_by` uses a similar format to `assert_nonlinear_by`, but without the proof body.

For example: 
```
fn concat_bits(b1: u32, b2:u32) -> u64 {
    ensures(|ret:u64| ret == (b1 as u64) * 0x100000000 + (b2 as u64));
    let b:u64 = (b1 as u64) << (32 as u64) | (b2 as u64);
    assert_bitvector_by({
        requires(b == (b1 as u64) << (32 as u64) | (b2 as u64));
        ensures(b == (b1 as u64) * 0x100000000 + (b2 as u64));
    });
    b
}
```


channy412:
I plan to adapt to the new syntax for this pr: 

for  `assert(...) by (bit_vector)`,  it cannot take the `requires` block at this point, but it would be pretty helpful for quick inlined proofs. This PR will make it possible to take the `requires` block for bit-vector assertions.

For example:
```
assert(a == b) by (bit_vector)
    requires
        a & sub(1u32 << 32,1) == b & sub(1u32 << 32,1),
{}
```

This would make the `assert(...) by(...)` syntax more consistent with `assert(...) by (nonlinear_arith)` 

I would like to deprecate the previous `assert_bit_vector`, but the users with new syntax will not notice any change.

channy412:
For the new syntax, `assert(...) by(bit_vector)`, it is now translated into this new `assert_bitvector_by`. 
I didn't remove `assert_bit_vector` on the user side, but it is internally translated into `assert_bitvector_by`.

There are example/test cases that use `assert_bit_vector` inside locally defined macros; updating it to the new syntax was not very trivial (for example, at `example/extensionality.rs`).

source/vir/src/ast.rs
```diff
@@ -355,6 +355,7 @@ pub enum InvAtomicity {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum AssertQueryMode {
     NonLinear,
+    BitVector,
```
Chris-Hawblitzel, line 358: The comment below that says `/// assert_by with smt.arith.nl=true` probably needs to be updated.

source/rust_verify/example/power_of_2.rs
```diff
@@ -0,0 +1,183 @@
...
+#[allow(unused_imports)]
+mod pervasive;
+#[allow(unused_imports)]
+use pervasive::*;
+#[allow(unused_imports)]
+use builtin_macros::*;
+
+
+#[spec]
+fn pow2(e: nat) -> nat {
```
Chris-Hawblitzel, line 16: Should this file use the `verus!` syntax?

source/rust_verify/tests/assert_bitvector_by.rs
```diff
@@ -0,0 +1,66 @@
...
+                ensures((b << 2) == b*4);
+            });
+            assert((b << 2) == b*4);
+
+            assert_bitvector_by({
+                ensures((b << 1) == b*4);  // FAILS
+            });
+        }
+    } => Err(err) => assert_one_fails(err)
+}
```
Chris-Hawblitzel, line 66: Can we also add a negative test for when the `requires` of `assert_bitvector_by` fails?

Chris-Hawblitzel (approved):

Chris-Hawblitzel:
Rather than having to write:

```
assert(a == b) by(bit_vector)
    requires
        a & sub(1u32 << 32, 1) == b & sub(1u32 << 32, 1),
{}
```

I think we should be able to write:

```
assert(a == b) by(bit_vector)
    requires a & sub(1u32 << 32, 1) == b & sub(1u32 << 32, 1);
```

This requires a small change to verus_syn, which I can make after the pull request is merged.

source/vir/src/ast.rs
channy412: Thanks, I will update the comment

source/rust_verify/example/power_of_2.rs
channy412: I think I encountered small problems around automatic widening(to `int`), but I think I should try again.

source/rust_verify/tests/assert_bitvector_by.rs
channy412: Sure, I will add some more tests. Thank you!

Chris-Hawblitzel:
> I think we should be able to write:
> 
> ```
> assert(a == b) by(bit_vector)
>     requires a & sub(1u32 << 32, 1) == b & sub(1u32 << 32, 1);
> ```
> 
> This requires a small change to verus_syn, which I can make after the pull request is merged.

Ok, I just pushed this change to `main`.

PR #153, opened 2022-05-18
Disallow arithmetic triggers

utaal:
Also group binary bitwise operators in an enum to simplify matches.

### DIFF
--- a/source/vir/src/triggers.rs
+++ b/source/vir/src/triggers.rs
@@ -34,12 +34,7 @@ fn check_trigger_expr(exp: &Exp, free_vars: &mut HashSet<Ident>) -> Result<(), V
         | ExpX::UnaryOpr(UnaryOpr::Field { .. }, _)
         | ExpX::Unary(UnaryOp::Trigger(_), _) => {}
         // allow triggers for bitvector operators
-        ExpX::Binary(BinaryOp::BitAnd, _, _)
-        | ExpX::Binary(BinaryOp::BitOr, _, _)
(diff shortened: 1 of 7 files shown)


source/vir/src/triggers.rs
```diff
@@ -34,13 +34,7 @@ fn check_trigger_expr(exp: &Exp, free_vars: &mut HashSet<Ident>) -> Result<(), V
...
-        ExpX::Binary(BinaryOp::BitAnd, _, _)
-        | ExpX::Binary(BinaryOp::BitOr, _, _)
-        | ExpX::Binary(BinaryOp::BitXor, _, _)
-        | ExpX::Binary(BinaryOp::Shr, _, _)
-        | ExpX::Binary(BinaryOp::Shl, _, _)
-        | ExpX::Unary(UnaryOp::BitNot, _) => {}
-        // REVIEW: Z3 allows some arithmetic, but it's not clear we want to allow it
+        ExpX::Binary(BinaryOp::Bitwise(_), _, _) | ExpX::Unary(UnaryOp::BitNot, _) => {}
         _ => {
             return err_str(&exp.span, "trigger must be a function call or a field access");
```
Chris-Hawblitzel, line 39: Is this initial check redundant now that the visitor is checking everything?

Chris-Hawblitzel (approved):

source/vir/src/triggers.rs
utaal: I don’t think so. The outer expression cannot be just a variable, for example.

PR #160, opened 2022-05-26
adjust trigger for `axiom_seq_push_index_same`

tjhance:
`s.push(a).index(s.len())` as a trigger seems a little too specific, specifically, this fails:

```rust
fn pust_test(t: Vec<u64>, y: u64) {
    requires([
        forall(|i: int| 0 <= i && i < t.view().len() >>= uninterp_fn(t.view().index(i))),
        uninterp_fn(y),
    ]);

    let mut t = t;
    t.push(y);

    assert(
        forall(|i: int| 0 <= i && i < t.view().len() >>= uninterp_fn(t.view().index(i))),
    );
}
```


tjhance:
yes it seems to work the same

Chris-Hawblitzel (approved):

PR #166, opened 2022-06-07
Basic version of an axiom profiler for Verus

parno:
When `--profile` or `--profile-all` is specified, this will instrument Z3 to emit a trace of its proof process, which we then analyze to determine the number of times user-level quantifiers are instantiated, as well as an estimated cost for each quantifier.


source/air/Cargo.toml
```diff
@@ -10,6 +10,8 @@ edition = "2018"
 [dependencies]
 sise = "0.6.0"
 getopts = { git = "https://github.com/utaal/getopts.git", branch = "parse-partial" }
+z3tracer = { git = "https://github.com/secure-foundations/smt2utils.git", branch = "main" }
+#z3tracer = { path = "../../../z3tracer/z3tracer" }
```
utaal, line 14: Can you remove this?

source/air/src/printer.rs
```diff
@@ -236,6 +236,12 @@ impl Printer {
                             nodes.push(str_to_node(":pattern"));
                             nodes.push(self.exprs_to_node(trigger));
                         }
+                        if let Some(s) = qid {
+                            nodes.push(str_to_node(":qid"));
+                            nodes.push(str_to_node(s));
+                            nodes.push(str_to_node(":skolemid"));
+                            nodes.push(str_to_node(format!("skolem_{}", s).as_str()));
```
utaal, line 243: Can you please add `"skolem"` to `def.rs`? And ideally check that `skolem_{}` appears correctly in the input in `parser.rs`?

source/rust_verify/example/trigger_loops.rs
```diff
@@ -0,0 +1,81 @@
+// rust_verify/tests/example.rs ignore
```
utaal, line 1: Can you add a comment to this file saying that users should expect it to timeout?
`ignore` is the right thing here, but if folks arrive here while browsing for examples it'd be nice if they were warned that this explodes.

source/vir/src/ast_util.rs
```diff
@@ -61,6 +61,16 @@ pub fn n_types_equal(typs1: &Typs, typs2: &Typs) -> bool {
 
 pub const QUANT_FORALL: Quant = Quant { quant: air::ast::Quant::Forall, boxed_params: true };
 
+// Generate a unique internal quantifier ID
+pub fn new_internal_qid(name: String) -> Option<String> {
```
utaal, line 65: I think this should go in `vir/src/def.rs`.

source/vir/src/datatype_to_air.rs
```diff
@@ -139,14 +139,16 @@ fn datatype_or_fun_to_air_commands(
         // box axiom:
         //   forall x. x == unbox(box(x))
         // trigger on box(x)
-        let bind = func_bind(ctx, &Arc::new(vec![]), &x_params(&datatyp), &box_x, false);
+        let name = format!("{}_box_axiom", path_as_rust_name(dpath));
```
utaal, line 142: Please define this and similar ones in `def.rs`

source/rust_verify/src/verifier.rs
```diff
@@ -231,6 +236,67 @@ impl Verifier {
...
+        let max = 10;
+        let delimiter = "-".repeat(100);
+        let msg = format!(
+            "Observed {} total instantiations of user-level quantifiers",
+            total.to_formatted_string(&Locale::en)
+        );
+        compiler.diagnostic().note_without_error(&msg);
+
+        for (index, cost) in profiler.iter().take(max).enumerate() {
+            println!("{}", delimiter);
```
utaal, line 256: Considering you're using diagnostics, can you remove this?

source/rust_verify/src/verifier.rs
```diff
@@ -231,6 +236,67 @@ impl Verifier {
...
+                count * cost.cost,
+                count.to_formatted_string(&Locale::en),
+                100 * count / total,
+                cost.cost,
+                index + 1,
+                num_quants
+            );
+            compiler.diagnostic().span_note_without_error(span, &msg);
+
+            let triggers = match &qexp.x {
```
utaal, line 274: I was a little uncomfortable that you're carrying a reference to the `Exp` in the `qid_map`, but I can see why.
What would you think about only stashing the `Span` and the trigs in the map, i.e. moving this logic to the point where you add quants to the map? (The span is tiny and the triggers are already `Arc`ed)

source/vir/src/sst_to_air.rs
```diff
@@ -364,6 +366,21 @@ fn assert_unsigned(exp: &Exp) {
...
+// Generate a unique quantifier ID and map it to the quantifier's span
+fn new_user_qid(ctx: &Ctx, exp: &Exp) -> String {
+    let fun_name = fun_as_rust_dbg(
+        &ctx.fun.as_ref().expect("Expressions are expected to be within a function").current_fun,
+    );
+    // In SMTLIB, unquoted attribute values cannot contain colons,
+    // and sise cannot handle quoting with vertical bars
+    let fun_name = str::replace(&fun_name, ":", "_");
+    let qcount = ctx.quantifier_count.get();
+    let qid = format!("{}{}_{}", air::profiler::USER_QUANT_PREFIX, fun_name, qcount);
```
utaal, line 378: I think I'd prefer the qid structure to be defined in `def.rs`, but obviously the rest of this function needs to be here: it may be worth splitting.

source/air/src/ast.rs
```diff
@@ -98,14 +98,16 @@ pub enum Quant {
 pub type Trigger = Exprs;
 pub type Triggers = Arc<Vec<Trigger>>;
 
+pub type Qid = Option<String>;
```
utaal, line 101: What about `Qid = Option<Ident>`? (Where `Ident` is `Arc<String>`)
I think we do the same for debugger tags, and considering you're cloning `Qid`s from time to time, I think this would be a bit more consistent (and possibly marginally faster). (If at some point we decide we need a string interner/allocator, it's easier to find all the AST Idents this way, too.)

source/air/Cargo.toml
parno: Done

source/air/src/printer.rs
parno: I added `"skolem"` to `def.rs`, but I'm not sure what you have in mind for `parser.rs`.  My understanding was that `parser.rs` should be able to handle arbitrary AIR files, which could be handwritten or generated by something other than Verus, in which case we can't necessarily expect someone to use the same convention for skloem identifiers.  Apologies if I've misunderstood the suggestion though!

source/vir/src/ast_util.rs
parno: Done.

source/vir/src/datatype_to_air.rs
parno: Done.

source/rust_verify/src/verifier.rs
parno: Done.

source/vir/src/sst_to_air.rs
parno: Just to clarify, are you suggesting that we move line 378 to its own function in `def.rs`?

source/rust_verify/src/verifier.rs
parno: That's an interesting idea -- it would certainly be more concise.  

I tried some experiments with the compiler's `Multispan`.  If I create one using `from_spans` using a vector that includes the quantifier's span and the triggers' spans, then we just get a single underline under the whole quantifier (i.e., the quantifier span subsumes the trigger spans).  If I omit the quantifier's span from the vector, it works as long as the quantifier is all on one line, but if it's on multiple lines in the source file, then only the lines that include chosen triggers show up, which seems confusing.

Things work better when I use the quantifier span as the primary span and the triggers as span_labels, or vice versa.  Let me know which of the screenshots below you two like better (if any).  Both are slightly clunky with multi-line quantifiers, but at least they both include all of the relevant info.

multi1

multi2

tjhance: i like that the first one has shorter gaps. I don't have a strong preference though

source/vir/src/sst_to_air.rs
utaal: Yes, maybe both 376 and 378? With `fun_name` and `qcount` as arguments.
This is minor obviously.

source/air/src/printer.rs
utaal: The AIR file format is pretty much currently defined by the printer and parser. Without a change to `parser.rs` an AIR file with a skolem id that doesn't match `skolem_{:qid}` would not be parsed correctly (`printer` accepts the tokens but then throws it away): I'd rather we rejected AIR that didn't match the only format AIR can currently emit and represent in its AST; otherwise if someone hand-wrote some other skolemid, it would be silently ignored, which would be confusing to debug. In fact: I'd say we should only accept AIR where either both :qid and :skolemid are present or both are absent.
Note that printer (in this case) also determines what the SMT output is.

In other words: something that would be accepted by the AIR parser right now wouldn't correctly round-trip when parsed and then printed (either as AIR or SMT).

Let me know if I've misunderstood, of this is unclear.

parno: I see; thanks for the explanation.  I've pushed a fix.

source/vir/src/sst_to_air.rs
parno: Okay, I moved the name generation into `def.rs`.

source/rust_verify/src/verifier.rs
parno: @utaal Did you have a preference between the two versions above?

utaal: (Sorry for the delay!)

I don't have a strong preference; there may be some value in having a more compact output, like @tjhance, but I personally think both would work well.

Thanks a lot for looking into this!

utaal (approved):

source/rust_verify/src/verifier.rs
parno: Thanks for all of the input!  I pushed the compact version for now.  If we dislike it down the line, we can always tweak it as needed.

parno:
Done!  Thanks again for all of the feedback.

Issue #169, opened 2022-06-08
Question about implementing a Map-like struct for exec mode


marshtompsxd:
According to https://github.com/secure-foundations/verus/issues/167, I was trying to implement a Map-like struct for exec mode. I checked how vec is implemented and was trying to implement a dict as a wrapper around std::HashMap. However, when I define the dict struct with generics, it seems that I cannot specify bounds. More details below:
```
xudongs@xudongs-a01:~/verus/source:$ cat rust_verify/example/myexample.rs                   
#[verifier(external_body)]
pub struct Dict<#[verifier(maybe_negative)] K: std::cmp::Eq + std::hash::Hash, #[verifier(strictly_positive)] V> 
{
    pub dict: std::collections::HashMap<K, V>,
}

fn main() { }

xudongs@xudongs-a01:~/verus/source:$ ./tools/rust-verify.sh rust_verify/example/myexample.rs
error: not yet supported: bounds on datatype parameters
  --> rust_verify/example/myexample.rs:21:1
   |
21 | / pub struct Dict<#[verifier(maybe_negative)] K: std::cmp::Eq + std::hash::Hash, #[verifier(strictly_pos...
22 | | {
23 | |     pub dict: std::collections::HashMap<K, V>,
24 | | }
   | |_^

error: aborting due to previous error
```
I have also tried to use `where` clause to specify bounds but `where` is not supported as well. 

Could you please let me know what is the best way for me to specify bounds since it is necessary to implement a wrapper around HashMap? If this can work I am happy to implement a dict for exec mode and send a PR if it is needed.

tjhance:
Verus support for traits is pretty limited, but it _is_ possible to add trait bounds on impls and functions:

```
#[verifier(external_body)]
pub struct Dict<#[verifier(maybe_negative)] K, #[verifier(strictly_positive)] V>
{
    pub dict: std::collections::HashMap<K, V>,
}

impl<K: std::cmp::Eq + std::hash::Hash, V> Dict<K, V> {
    #[verifier(external_body)]
    fn new() -> Self {
        Dict { dict: HashMap::new() }
    }
}
```

Unfortunately, there are going to be some more limitations.

  * For one, Verus currently doesn't handle externally-defined traits, so you won't be able to implement `Hash` or `Eq` on any datatype you define
  * Verus doesn't yet support trait bounds on types that aren't structs.

[...]

PR #173, opened 2022-06-09
have verus ignore an impl block entirely if marked with external_body

tjhance:
to help with #169


Chris-Hawblitzel (approved):
Do you want `external` or `external_body` for this?  With `external`, an entire item is invisible to Verus.  With `external_body`, Verus sees the item declaration, but not the item body.

tjhance:
ah right, `external` makes more sense here.

PR #174, opened 2022-06-10
Implement exec mode map: dict

marshtompsxd:
I was trying to implement an exec mode map, Dict. This PR is not done as some of the methods like `remove` are not implemented yet. Please let me know whether a Dict is expected to be included in pervasive and I am happy to continue working on this PR and contribute to verus.

### DIFF
--- a/source/pervasive/dict.rs
+++ b/source/pervasive/dict.rs
@@ -16,39 +16,50 @@ impl<K: std::cmp::Eq + std::hash::Hash, V> Dict<K, V> {
     fndecl!(pub fn view(&self) -> Map<K, V>);
 
     #[verifier(external_body)]
-    fn new() -> Self {
+    pub fn new() -> Self {
         ensures(|d: Self| equal(d.view(), Map::empty()));
 
         Dict { dict: std::collections::HashMap::new() }
     }
 
-    pub fn empty() -> Self {
-        ensures(|d: Self| equal(d.view(), Map::empty()));
+    #[verifier(external_body)]
+    pub fn insert(&mut self, key: K, value: V) -> crate::pervasive::option::Option<V> {
+        ensures(|r: crate::pervasive::option::Option<V>| [
+            equal(r, if old(self).view().dom().contains(key) {
+                crate::pervasive::option::Option::Some(old(self).view().index(key))
+            } else {
+                crate::pervasive::option::Option::None
+            }),
+            equal(self.view(), old(self).view().insert(key, value)),
+        ]);
 
-        Dict::new()
+        match self.dict.insert(key, value) {
+            Some(v) => crate::pervasive::option::Option::Some(v),
+            None => crate::pervasive::option::Option::None,
+        }
     }
 
     #[verifier(external_body)]
-    pub fn insert(&mut self, key: K, value: V) {
-        ensures(equal(self.view(), old(self).view().insert(key, value)));
+    pub fn remove(&mut self, key: &K) -> crate::pervasive::option::Option<V> {
+        ensures(|r: crate::pervasive::option::Option<V>| [
+            equal(r, if old(self).view().dom().contains(*key) {
+                crate::pervasive::option::Option::Some(old(self).view().index(*key))
+            } else {
+                crate::pervasive::option::Option::None
+            }),
+            equal(self.view(), old(self).view().remove(*key)),
+        ]);
 
-        self.dict.insert(key, value);
-    }
-
-    #[verifier(external_body)]
-    pub fn remove(&mut self, key: &K) {
-        ensures(equal(self.view(), old(self).view().remove(key)));
-
-        self.dict.remove(key);
+        match self.dict.remove(key) {
+            Some(v) => crate::pervasive::option::Option::Some(v),
+            None => crate::pervasive::option::Option::None,
+        }
     }
 
     #[verifier(external_body)]
     #[verifier(autoview)]    
-    pub fn contain(&self, key: &K) -> bool {
-        ensures(|r: bool|[
-            r >>= self.view().dom().contains(key),
-            self.view().dom().contains(key) >>= r,
-            ]);
+    pub fn contains(&self, key: &K) -> bool {
+        ensures(|r: bool| r == self.view().dom().contains(*key));
 
         match self.dict.get(key) {
             Some(v) => true,
@@ -59,8 +70,8 @@ impl<K: std::cmp::Eq + std::hash::Hash, V> Dict<K, V> {
     #[verifier(external_body)]
     #[verifier(autoview)]
     pub fn index(&self, key: &K) -> &V {
-        requires(self.view().dom().contains(key));
-        ensures(|r: V| equal(r, self.view().index(key)));
+        requires(self.view().dom().contains(*key));
+        ensures(|r: V| equal(r, self.view().index(*key)));
 
         match self.dict.get(key) {
             Some(v) => &v,
--- a/source/pervasive/dict.rs
+++ b/source/pervasive/dict.rs
@@ -8,8 +8,7 @@ use crate::pervasive::*;
 use crate::pervasive::map::*;
 
 #[verifier(external_body)]
-pub struct Dict<#[verifier(maybe_negative)] K, #[verifier(strictly_positive)] V> 
-{
+pub struct Dict<#[verifier(maybe_negative)] K, #[verifier(strictly_positive)] V> {
     pub dict: std::collections::HashMap<K, V>,
 }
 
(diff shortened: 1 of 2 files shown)


tjhance:
Yes, we definitely want something like `Dict` in the verus standard lib, analogously to Vec/Seq 👍

source/pervasive/dict.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[verifier(external_body)]
+    pub fn insert(&mut self, key: K, value: V) {
+        ensures(equal(self.view(), old(self).view().insert(key, value)));
+
+        self.dict.insert(key, value);
+    }
+
+    #[verifier(external_body)]
+    #[verifier(autoview)]    
+    pub fn contain(&self, key: K) -> bool {
```
tjhance, line 41: we can use `key: &K` here, similar to `std::collections::HashMap`

source/pervasive/dict.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        self.dict.remove(key);
+    }
+
+    #[verifier(external_body)]
+    #[verifier(autoview)]    
+    pub fn contain(&self, key: &K) -> bool {
+        ensures(|r: bool|[
+            r >>= self.view().dom().contains(key),
+            self.view().dom().contains(key) >>= r,
```
utaal, line 50: These can just be: `self.view().dom().contains(key) == r`

utaal (changes requested):
Thanks for working on this!

marshtompsxd:
Updated. Do we want an example to test dict?

tjhance:
A few examples in `examples/` would be nice, yeah.

I just noticed that this uses `std::cmp::Eq + std::hash::Hash`, and as discussed over in https://github.com/secure-foundations/verus/issues/169 the properties of those traits aren't verified by Verus - so this API is unsound if the user provides an implementation that doesn't match the usual equivalence-relation guarantee of Eq, or which doesn't meet the stronger condition that equality agree with SMT equality.

Can you change the API to require a stronger trait bound, like the custom trait discussed in the other discussion?

marshtompsxd:
Hi @tjhance, I just want to ensure I understand your point: Verus cannot verify the properties of Hash or Eq, so we want to add another bound `HashEq` (mentioned in https://github.com/secure-foundations/verus/issues/169#issuecomment-1150384663) to K. This bound will check whether Hash and Eq are implemented correctly. 

I have a few more questions regarding the code you posted before
```
pub trait HashEq {
  #[spec]
  fn hash(&self) -> u64 {
      no_method_body()
  }

  #[exec]
  fn compute_hash(&self) -> u64 {
      // Result matches a deterministic function
      ensures(|res: u64| res == self.hash());
      no_method_body()
  }

  #[exec]
  fn compute_eq(&self, other: &Self) -> bool {
      // equality matches the SMT equality of the objects
      ensures(|res: bool| res == equal(self, other));
      no_method_body()
  }
}
```
1. Will the postconditions be inherited implicitly when the user implement their own `compute_hash` and `compute_eq`? Or do they need to specify the same postcondition again in their impl?
2. What should be a common implementation of `compute_hash` and `compute_eq`? Should the user first implement `HashEq::compute_eq` and then implement `PartialEq::eq` like the following?
```
  fn eq(&self, other: &Self) -> bool {
      self.compute_eq(other)
  }
```
3. The signature of `Hash::hash` is `fn hash<H: Hasher>(&self, state: &mut H)`. For hash, what property should we check and how do we check the property?

tjhance:
Yes, conditions are automatically inherited. So to implement a trait, you have to meet the spec of a trait; any method that is generic over the trait can assume that specification holds.

(For example, a hypothetical verified hash table would specify internal invariants on its layout in terms of the `hash` specification fn, then it would rely on the fact that `compute_hash` computes `hash` in order to verify that these invariants are maintained.)

For the _unverified_ implementation with HashMap like we're trying to do here, we can make a blanket impl to implement `std::cmp::PartialEq`, `std::cmp::Eq`, and `std::hash::Hash` in terms of `HashEq`.

Then we'd reason (informally) as follows:

 * To use `Dict`, the type `K` has to satisfy the `HashEq` trait bound.
 * If `HashEq` is satisfied, then via the blanket impl, the `std::cmp::PartialEq`, `std::cmp::Eq`, and `std::hash::Hash` traits all have the properties they need for `std::HashMap` to behave correctly.
 * Therefore, `Dict` meets its specification.

So:

1. Yes, the preconditions are implicitly inherited.
2. The `Dict` library should provide a blanket implementation for `PartialEq` like the one you showed
3. There's no property to check for `std::hash::Hash`, since it's part of the unverified implementation. Or rather, the properties it has are specified outside Verus, but those properties follow from the formally checked properties that `HashEq` has.
4. See 2.

Hope this makes sense!

marshtompsxd:
Thanks for the explanation! 

_"For the unverified implementation with HashMap like we're trying to do here, we can make a blanket impl to implement `std::cmp::PartialEq`, `std::cmp::Eq`, and `std::hash::Hash` in terms of `HashEq`."_

This is the only part I do not fully understand.
1. what does it mean by a blanket impl? Do you mean a wrapper around HashEq::compute_eq or HashEq::compute_hash?
2. Do you mean we should implement traits PartialEq, Eq, and Hash for K without knowing what type K is in Dict? I am new to rust so I do not know whether one can implement a trait for a generic K without knowing what concrete type K is.
3. If 2 is correct, does K only need one bound HashEq instead of Eq + Hash?

tjhance:
By blanket impl I mean,

```rust
impl<K: HashEq> PartialEq for K {
    fn eq(&self, other: &Self) -> bool {
        self.compute_eq(other)
    }
}
```

i.e., for every type that implements `HashEq`, it also implements `PartialEq` with the latter defined in terms of the former.

marshtompsxd:
Hi @tjhance I am trying to refactor this PR but it seems that I can't do something like 

```
#[verifier(external)]
impl<K: HashEq> std::cmp::PartialEq for K {
    fn eq(&self, other: &Self) -> bool {
        self.compute_eq(other)
    }
}
```

rust reports error
```
error[E0210]: type parameter `K` must be used as the type parameter for some local type (e.g., `MyStruct<K>`)
  --> rust_verify/example/pervasive/dict.rs:30:6
   |
30 | impl<K: HashEq> std::cmp::PartialEq for K {
   |      ^ type parameter `K` must be used as the type parameter for some local type
   |
   = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local
   = note: only traits defined in the current crate can be implemented for a type parameter
```

It seems that we cannot use this way to implement traits that are not defined in our crate (i.e., PartialEq).

tjhance:
Shoot, I forgot about that restriction.

I think the usual way to get around this in Rust is to do define a wrapper type like this:

```
struct KeyWrapper<K> {
  key: K,
};
```

You then have to do,

```
#[verifier(external)]
impl<K: HashEq> std::cmp::PartialEq for KeyWrapper<K> {
    fn eq(&self, other: &Self) -> bool {
        self.key.compute_eq(other.key)
    }
}
```

It's a little annoying because you have to wrap & unwrap the key in your implementation, but in theory, it's a "zero cost abstraction".

marshtompsxd:
Hi @tjhance Thanks for your help again! Using the wrapper solves the problem and I do not have any other compilation errors now. But I got some error from verus this time.

```
pub trait HashEq {
    ...

    fn compute_hash<H: std::hash::Hasher>(&self, state: &mut H) {
        no_method_body()
    }
}
```
and verus reports
```
[rust_verify/src/rust_to_vir.rs:371]
error: The verifier does not yet support the following Rust feature: trait generics
  --> rust_verify/example/pervasive/dict.rs:21:5
   |
21 | /     fn compute_hash<H: std::hash::Hasher>(&self, state: &mut H) {
22 | |         no_method_body()
23 | |     }
   | |_____^

```

I tried to put `#[verifier(external)]` before `compute_hash` but the error still happens. Is there a way to workaround this error?

tjhance:
Whoops, sorry, I didn't notice your comment earlier. My bad.

As the error says, Verus doesn't yet support generics on trait methods. But then, it also doesn't support `std::hash::Hasher` anyway.

So it's kind of annoying, but for now we'll just have to do it without it and implement hash as a function from `self` to an integer.

marshtompsxd:
Hi @tjhance 
Sorry for the late response. I come back to this pending PR.

> So it's kind of annoying, but for now we'll just have to do it without it and implement hash as a function from self to an integer.

I am a bit confused here. The signature of `Hash::hash` required by `HashMap` is `fn hash<H: Hasher>(&self, state: &mut H)`, which does not return any integer. So how does a function that from `self` to an integer make `HashMap` happy here?

utaal:
Closing in favor of #308.

PR #186, opened 2022-07-01
[WIP] extensions to custom_err_req

tjhance:
I'm trying to improve the error messages on the auto-generated tracked-mode functions from tokenized_state_machine, which have elaborate preconditions. Unfortunately since these are generated as verus source, the macro doesn't have access to the error data structures in AIR/VIR, and the `custom_req_err` is missing features I'd need, like:

 * different message for each precondition
 * have a custom message while also highlighting span of the precondition
 * add annotations to arguments at the call-site

(the first point might be useful for end-users as well)

I experimented with refactoring the custom_req_err code with the aim of making it more extensible (with the side effect of consolidating the default error logic into one place as well). But I wanted to get feedback on the general direction / code changes in VIR before going any further. This WIP draft consolidates logic into a  `fn_call_error_messages` module and adds the ability to add per-precondition error messages.


Chris-Hawblitzel (approved):

tjhance:
going to close this until I return to it, to remove clutter from the PR list

PR #195, opened 2022-07-13
Error localization via expression splitting

channy412:
This pull request implements expression splitting to get better error messages for proof failures. 

When proof fails, it saves error spans and reruns `verify_module`. During VIR -> SST translation, it splits previously failed assertions. It runs this splitting by default on proof failure. It does not emit an additional error message if this additional run does not produce new information.

Not included in this pull request (TODO) :
- setting up testing for this feature
- splitting `&&&` and `|||`
- supporting some variety of `assert_by_*`. Now it only splits on `assert`, `requires`, and `ensures`.

### DIFF
--- a/source/rust_verify/example/debug_expand.rs
+++ b/source/rust_verify/example/debug_expand.rs
@@ -356,5 +356,20 @@ proof fn test_binder_inlining(b: bool)
 }
 
 
+// example15: recursive function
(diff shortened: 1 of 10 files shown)


channy412:
Plan to merge this changes(with updates) at https://github.com/secure-foundations/verus/pull/255

PR #196, opened 2022-07-15
WIP: encoding of string literals

isubasinghe:


source/rust_verify/src/context.rs
```diff
@@ -15,6 +17,7 @@ pub struct ErasureInfo {
 }
 
 type ErasureInfoRef = std::rc::Rc<std::cell::RefCell<ErasureInfo>>;
+type GlobalStrings = std::rc::Rc<std::cell::RefCell<HashMap<DefId, Arc<String>>>>;
```
isubasinghe, line 20: What is this? 
StrSlices by default are not lowered into air, at least not in any useful sense. 
When there is a "reveal" call, we then write the relevant air to represent that specific string. 
This holds all the String values along with their associated variable. So when VAR.reveal() is called, we look up the DefId for VAR and then retrieve the String, from here we can emit the air for this.

isubasinghe: migrated to vir/src/context.rs

source/vir/src/context.rs
```diff
@@ -44,6 +45,17 @@ pub struct FunctionCtx {
...
+
+/// This stores information about whether a string should be emitted or not. 
+/// To make a string verifiable, one must simply call "reveal" on that string.
+#[derive(Clone, Debug)]
+pub struct VerifiableString {
+    pub inner_str: Arc<String>,
+    pub emitted: bool
+}
+
+pub type GlobalStrings = std::sync::Arc<std::sync::Mutex<HashMap<Path, Arc<VerifiableString>>>>;
```
isubasinghe, line 57: What is this?
StrSlices by default are not lowered into air, at least not in any useful sense.
When there is a "reveal" call, we then write the relevant air to represent that specific string.
This holds all the String values along with their associated variable. So when VAR.reveal() is called, we look up the DefId for VAR and then retrieve the String, from here we can emit the air for this.

PR #200, opened 2022-07-20
support type aliases

tjhance:
Basic support for type aliases.

The difficult part of supporting this is `ty_to_vir`. I couldn't find an easy function to do all the work for me from rustc_hir - the best I found the `type_of` function from rustc that returns the type synonym as a mid_ty, but it doesn't do type param substitution.

This PR provides full support for type aliases that don't take any type parameters.

For those that do, I think it's probably easy enough to just manually perform the type substitution, if you all don't see any better ideas.


tjhance:
could we go ahead and merge this? the 'transition labels' relies partially on type synonyms. I think this PR is fine as-is and we can handle anything else as follow-up.

Chris-Hawblitzel (approved):

PR #203, opened 2022-07-21
VIR Printer options for concise printing (`--vir-log-option`)

channy412:
Span often makes it much harder to read VIR logs. This pull request adds --vir-log-config command line argument to make Verus optionally include span in VIR logs.

When vir-log-config is 0(Default), it logs all but span.
when it is 1, it logs all items (as the current version does)


parno:
In general, I think we might want to err on the side of named arguments, rather than numeric, since something like `no_span` is easier to remember than `0` (similarly `all` vs `1`).  I also think we'll probably want a few more knobs than just span/no span.  Perhaps we can look at the output without spans together to see what else might be (optionally) cleaned up.

parno:
Similarly, we'll probably want an `struct` internally to hold the configuration, rather than an integer.  That way someone can choose, for example, to show types but not spans or the other way around.

channy412:
Thanks, I agree that making named flags would be much easier than numeric values. 
I think we can consider more knobs around 
- type information
- routinary calls like box/unbox/clip
- function attributes
- formatting

channy412:
The above commit adds three flags:
`--vir-log-no-span` makes the VIR printer omit all spans.
`--vir-log-no-type` makes the VIR printer omit all types.
`--vir-log-no-routinary` makes the VIR printer omit all routinary calls. These calls are `box` , `unbox`, and `clip`.

I will add some more on formatting, and also about function's various surrounding information

utaal:
> The above commit adds three flags: `--vir-log-no-span` makes the VIR printer omit all spans. `--vir-log-no-type` makes the VIR printer omit all types. `--vir-log-no-routinary` makes the VIR printer omit all routinary calls. These calls are `box` , `unbox`, and `clip`.
> 
> I will add some more on formatting, and also about function's various surrounding information

One thing I noticed at the meeting: there's quite a few flags, and one needs all of them for the most compact output. Would it make sense to have a single flag with different levels/options?

---

Also: if I saw this correctly at the meeting, it looked like there wasn't a distinction between function calls (that come from the rust encoding) and internal s-expressions/calls (such as box); it may be good to have something that marks "spec rust" function calls.

channy412:
> One thing I noticed at the meeting: there's quite a few flags, and one needs all of them for the most compact output. Would it make sense to have a single flag with different levels/options?

Thanks, I agree that I should merge into one flag with several options. 
Something looks like `vir-log-option=all` and `vir-log-option=no_span+no_type` would work? 


> Also: if I saw this correctly at the meeting, it looked like there wasn't a distinction between function calls (that come from the rust encoding) and internal s-expressions/calls (such as box); it may be good to have something that marks "spec rust" function calls.


That's a good point. I should distinguish those two forms to avoid being ambiguous. Would it work to introduce some prefix on "rust style" function calls? -- Something like `foo(_ a  b  c)`?

utaal:
> Thanks, I agree that I should merge into one flag with several options. Something looks like `vir-log-option=all` and `vir-log-option=no_span+no_type` would work?

Something like that, yes. Can we also have an alias for the compact version, maybe "compact"?

> That's a good point. I should distinguish those two forms to avoid being ambiguous. Would it work to introduce some prefix on "rust style" function calls? -- Something like `foo(_ a b c)`?

I'd rather we didn't break the fact that the format is s-expressions, even if we don't currently plan to read this mechanically, I think the consistency is preferable. What about `(call foo ...)`?

channy412:
> Something like that, yes. Can we also have an alias for the compact version, maybe "compact"?

That sounds good, thanks! 

> I'd rather we didn't break the fact that the format is s-expressions, even if we don't currently plan to read this mechanically, I think the consistency is preferable. What about `(call foo ...)`?

Sure, that works. (I agree that having consistency is a definite plus.)

utaal:
Is this something we may want to merge soon? I'm making some changes to the AST, and it may be better to land this sooner rather than later to avoid more merge conflicts.

channy412:
> Is this something we may want to merge soon? I'm making some changes to the ASTs, and it may be better to land this sooner rather than later to avoid more merge conflicts.

Oh, I am sorry that I missed your comment. I will try to resolve conflicts and merge this soon. Thanks for the notice!

channy412:
To summarize:

This PR is to add `--vir-log-option` command line option, which helps Verus developers to debug Verus itself. 
It has five options: `no_span` `no_type` `no_encoding` `no_fn_details` `pretty_format`.
When a user gives  `--vir-log-option compact`, it will opt-in all five options.
To opt-in a few options, it uses `+`. (e.g. `--vir-log-option no_span+no_type`)

Other than `pretty_format` option, it will respect the `S-expression` format. However, `pretty_format` option is just a set of tricks that tries to make VIR logs look prettier in human eyes. (with a possibility of losing information)

utaal:
I think I'd prefer #325 over this. @channy412 can you take a look and see if it achieves similar goals? (We may still want to add `pretty_format` on top of my PR)

channy412:
> I think I'd prefer #325 over this. @channy412 can you take a look and see if it achieves similar goals? (We may still want to add `pretty_format` on top of my PR)

Thanks for working on the VIR printer!  Yes, and the new one looks great to me. (I think `pretty_format` is not that necessary)

PR #204, opened 2022-07-23
Add "let Ghost" and "let Tracked" syntax

Chris-Hawblitzel:
Support extracting tuple components from Ghost and Tracked tuples in exec mode.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -272,6 +272,48 @@ impl<A> std::ops::Deref for Tracked<A> {
     }
 }
 
+macro_rules! emit_phantom {
+    ($x:ident) => {
+        PhantomData
+    };
+}
+macro_rules! split_tuple {
+    /*
+        #[inline(always)]
+        pub fn tracked_split_tuple2<A1, A2>(_t: Tracked<(A1, A2)>) -> (Tracked<A1>, Tracked<A2>) {
+            (Tracked { phantom: PhantomData }, Tracked { phantom: PhantomData })
+        }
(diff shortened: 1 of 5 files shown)


source/builtin/src/lib.rs
```diff
@@ -272,6 +272,48 @@ impl<A> std::ops::Deref for Tracked<A> {
     }
 }
 
+macro_rules! emit_phantom {
+    ($x:ident) => {
+        PhantomData
+    };
+}
```
tjhance, line 279: what's this for?

tjhance (approved):

source/builtin/src/lib.rs
Chris-Hawblitzel: It's a hacky way to have the macro generate `$typ { phantom: PhantomData },` n times.  I'm not sure if there's a more natural way to do it.

jaybosamiya: If I understand why you've introduced this bit of code correctly, you are introducing it as a way to "eat up" the `$x`, which comes from the `$t` in `split_tuple`, because otherwise Rust wouldn't know what your `$(...)*` repetition is set up for.

If that understanding is correct, then you should be able to eliminate this `emit_phantom` macro by replacing `emit_phantom!($t)` in `split_tuple` below with `PhantomData::<$t>`. That way, you still have the `$t` anchoring the repetition.

Since the solution is to use a turbofish, please enjoy this link: [`::<$t>`](https://turbo.fish/::%3C$t%3E) 😄

PR #207, opened 2022-07-26
Z3 4.9.1 bitvector options, ci script update

channy412:
(wants to merge to `z3-4.9.1` branch) 

1. Updated bit-vector option in Z3 

I encountered this comment 
`Note: at this point tactic.default_tactic=sat sat.euf=true smt.ematching=false works best.`(https://github.com/Z3Prover/z3/issues/5660) that describes the option for bit-vector reasoning.

The garbage-collecting example that used to be working in Z3 4.8.5 fails in the latest(4.9.1) version. When I introduce the above options, it works again. 

One downside is that this option `sat.euf=true` seems not backward compatible. (When I ran the `root.smt2` with z3 4.8.5, I got the unrecognized option error for that option -- which will make Verus panic when it is with Z3 4.8.5)

I am worried that the newly introduced options might affect some internal states of Z3, affecting other parts. Therefore, I updated all bit-vector queries to be `spinoff_prover` by default.   

  

2. CI script update for Z3 4.9.1

### DIFF
--- a/source/vir/src/sst_to_air.rs
+++ b/source/vir/src/sst_to_air.rs
@@ -1183,11 +1183,13 @@ fn stm_to_stmts(ctx: &Ctx, state: &mut State, stm: &Stm) -> Vec<Stmt> {
                 stm.span.clone(),
                 "assert_bit_vector".to_string(),
                 Arc::new(vec![
+                    mk_option_command("sat.euf", "true"),
+                    mk_option_command("tactic.default_tactic", "sat"),
+                    mk_option_command("smt.ematching", "false"),
                     mk_option_command("smt.case_split", "0"),
                     Arc::new(CommandX::CheckValid(query)),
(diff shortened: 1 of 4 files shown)


utaal:
This looks good to me. The z3 argument parsing for the specific "sat" string value looked brittle, so I replaced it: does that look right to you?

channy412:
Thanks, the change looks great!

source/vir/src/sst_to_air.rs
```diff
@@ -1710,9 +1712,11 @@ pub fn body_stm_to_air(
             ]
         } else if is_bit_vector_mode {
             vec![
+                mk_option_command("sat.euf", "true"),
```
utaal, line 1715: I had forgotten one thing: can you collect this options somewhere? (They currently are duplicated for assert_by_bitvector and #[verifier(bit_vector)]

PR #210, opened 2022-07-28
remove ty_to_vir in favor of mid_ty_to_vir

tjhance:
As discussed with @utaal earlier. This makes for less duplication, plus mid_ty_to_vir supports more features, so for example, this completes the support for type aliases.

Something I forgot to account for in our discussion earlier was generic params. This ended up being the gnarliest part, and I had to rewrite `check_generic_bounds` (commit c0b99f7da9883710bb6e030969852d92d438756a). The rustc_middle `Predicate` type is pretty reasonable - it just comes in a pretty different shape than HIR's AST-based structure.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -441,8 +441,9 @@ fn check_generic_bound<'tcx>(
         // Rust language marker traits are ignored in VIR
         Ok(Arc::new(GenericBoundX::Traits(vec![])))
     } else {
-        // REVIEW: this seems wrong
-        // Why did the previous code here throw away the args?
+        // TODO we will actually need to handle these arguments at some point.
+        // Right now, this is safe only because Verus does not support having
+        // a type implement two instances of the same trait with different type args.
         let _args = args;
 
         let trait_name = def_id_to_vir_path(tcx, trait_def_id);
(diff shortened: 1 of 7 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -553,73 +427,204 @@ pub(crate) fn is_smt_arith<'tcx>(bctx: &BodyCtxt<'tcx>, id1: &HirId, id2: &HirId
...
+        || Some(trait_def_id) == tcx.lang_items().unpin_trait()
+        || Some(trait_def_id) == tcx.lang_items().sync_trait()
+        || Some(trait_def_id) == tcx.get_diagnostic_item(rustc_span::sym::Send)
+    {
+        // Rust language marker traits are ignored in VIR
+        Ok(Arc::new(GenericBoundX::Traits(vec![])))
+    } else {
+        // REVIEW: this seems wrong
+        // Why did the previous code here throw away the args?
+        let _args = args;
```
Chris-Hawblitzel, line 446: We don't yet include trait arguments in the VIR AST because we only support very simple uses of traits for now.  For example, if you say:
```
        trait T<A> { fn f(&self, x: &A); }
        struct S {}
        impl T<bool> for S { fn f(&self, x: &bool) {} }
        impl T<u8>   for S { fn f(&self, x: &u8)   {} }
```
Verus complains:
```
error: not supported: multiple definitions of same function
   --> .\rust_verify\example\features.rs:115:13
    |
112 |             fn f(&self, x: &bool) {}
    |             ---------------------
...
115 |             fn f(&self, x: &u8) {}
    |             ^^^^^^^^^^^^^^^^^^^
```

Chris-Hawblitzel (approved):

utaal (approved):
Thank you!

PR #214, opened 2022-07-28
Switch from * to @ for Ghost/Tracked and add borrow/borrow_mut for Ghost/Tracked

Chris-Hawblitzel:
This PR revamps the operations on the `Ghost` and `Tracked` wrapper types.  For example, where we used to write:

```
    let u: Ghost<int> = ghost(my_spec_fun(x as int, y as int));
    let mut v: Ghost<int> = ghost(*u + 1);
    assert(*v == x + y + 1);
    proof {
        v = Ghost::new(*v + 1); // proof code may assign to exec variables of type Ghost/Tracked
    }
```

we now write:

```
    let u: Ghost<int> = ghost(my_spec_fun(x as int, y as int));
    let mut v: Ghost<int> = ghost(u@ + 1);
    assert(v@ == x + y + 1);
    proof {
        v@ = v@ + 1; // proof code may assign to the view of exec variables of type Ghost/Tracked
    }
```

Previously, we relied on `Deref` for `Ghost` and `Tracked`, which could become problematic as we improve support for traits, since `Ghost` and `Tracked` didn't provide an exec implementation of `deref`, but rather just `unimplemented!()`.  Now we use `view` instead of `deref`, since `view` is a `spec` method rather than an `exec` method.  (Introducing a `spec_deref` spec method did not seem helpful, since our syntax macro doesn't have a good way to decide which `*` would be `deref`, as with `&` types, or `spec_deref`, as with `Ghost` and `Tracked`.)

This PR also adds `borrow` and `borrow_mut` methods for `Ghost` and `Tracked`.  As syntactic sugar, when you write `v@` on the  left side of an assignment, it gets desugared to `*(v.borrow_mut())` rather than `v.view()`.  This allows programmers to write things like `v@ = v@ + 1`, which seems more natural than the previous `v = Ghost::new(*v + 1)`.  You can also access fields this way (e.g. `v@.field = v@.field + 1`).

In the translation to VIR,  `Ghost<T>` and `Tracked<T>` are erased to `T`, and operations on them are erased to no-ops, so that `*(v.borrow_mut()) = v.view() + 1` simply becomes `v = v + 1`, leading to efficient SMT encodings.

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -106,6 +106,14 @@ pub enum TriggerAnnotation {
     Trigger(Option<u64>),
 }
 
+/// Operations on Ghost and Tracked
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ModeCoercion {
+    DerefMut,
+    BorrowMut,
+    Other,
+}
+
 /// Primitive unary operations
 /// (not arbitrary user-defined functions -- these are represented by ExprX::Call)
 #[derive(Copy, Clone, Debug)]
@@ -120,7 +128,7 @@ pub enum UnaryOp {
     /// Force integer value into range given by IntRange (e.g. by using mod)
     Clip(IntRange),
     /// Operations that coerce from/to builtin::Ghost or builtin::Tracked
-    CoerceMode { op_mode: Mode, from_mode: Mode, to_mode: Mode },
+    CoerceMode { op_mode: Mode, from_mode: Mode, to_mode: Mode, kind: ModeCoercion },
     /// Internal consistency check to make sure finalize_exp gets called
     /// (appears only briefly in SST before finalize_exp is called)
     MustBeFinalized,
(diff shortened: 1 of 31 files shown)


source/vir/src/ast.rs
```diff
@@ -106,6 +106,14 @@ pub enum TriggerAnnotation {
     Trigger(Option<u64>),
 }
 
+/// Operations on Ghost and Tracked
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ModeCoercion {
+    DerefMut,
+    BorrowMut,
+    Other,
```
tjhance, line 114: can you elaborate on what these coercions are (especially 'Other')?

Chris-Hawblitzel: I added comments to this code.  It turned out `DerefMut` was dead code so I removed it.

tjhance (approved):

utaal (approved):

PR #219, opened 2022-07-29
only macro-expand the pervasive lib once per test run

tjhance:


tjhance:
It looks like this speed up a single invocation of `cargo test` from 7 minutes to 5 minutes, judging by the CI runs. For specific tests though (like state_machines, which I run locally a lot) the speed up is significant.

source/rust_verify/tests/common/mod.rs
```diff
@@ -137,10 +139,16 @@ pub fn verify_files_and_pervasive(
     let files = files.into_iter().map(|(p, f)| (p.into(), f)).collect();
     let captured_output = std::sync::Arc::new(std::sync::Mutex::new(Vec::new()));
     let captured_output_1 = captured_output.clone();
+
+    let pervasive_path = match std::env::var("PERVASIVE_PATH") {
```
utaal, line 143: Can you rename this to TEST_PERVASIVE_PATH to distinguish it from `--pervasive-path`?

utaal (approved):
I have a bit of an uneasy feeling about this, but I suppose there isn't really a concrete way I can predict this going wrong, so let's take the test perf improvement.

Thank you!

PR #220, opened 2022-07-29
Allow extern static and GlobalAsm without verification support

ziqiaozhou:
Those features are important for writing bare-metal code in rust

```
global_asm!(
    "..."
);

extern "C" {
    #[verifier(external)]
    static extern_variable: u64;
}
```
NOTE: Rust compiler will throw away attribute for the builtin macro global_asm

### DIFF
--- a/source/rust_verify/src/rust_to_vir.rs
+++ b/source/rust_verify/src/rust_to_vir.rs
@@ -425,6 +425,11 @@ fn check_item<'tcx>(
             // type alias (like lines of the form `type X = ...;`
             // Nothing to do here - we can rely on Rust's type resolution to handle these
         }
+        ItemKind::GlobalAsm(..) =>
+        //if get_verifier_attrs(ctxt.tcx.hir().attrs(item.hir_id()))?.external =>
+        {
+            return Ok(());
+        }
         _ => {
             unsupported_err!(item.span, "unsupported item", item);
         }
--- a/source/rust_verify/src/rust_to_vir.rs
+++ b/source/rust_verify/src/rust_to_vir.rs
@@ -426,7 +426,7 @@ fn check_item<'tcx>(
             // Nothing to do here - we can rely on Rust's type resolution to handle these
         }
         ItemKind::GlobalAsm(..) =>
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (approved):

utaal:
Hi! Thanks for the PR.

Does this work for `extern "C"` functions too? (as long as they're marked `#[verifier(external)]`)?

Can you briefly explain how you'd be planning to use `global_asm` and why you'd prefer that over inline `asm!`?

ziqiaozhou:
> 



> Hi! Thanks for the PR.
> 
> Does this work for `extern "C"` functions too? (as long as they're marked `#[verifier(external)]`)?
> 
> Can you briefly explain how you'd be planning to use `global_asm` and why you'd prefer that over inline `asm!`?

I was writing OS-level code for a AMD-SNP VM. In an early stage, I want to use global_asm to define the entry point so that I can initialize stack and then call verus-verified functions.

utaal:
Thanks for the explanation; that's cool!

I'm not super comfortable with the idea of allowing global_asm anywhere and not require the user to explicitly allow (with, e.g., an attribute) it if it can influence the behavior of other code (which it probably can, if it can be the entry point).

If an attribute on `global_asm` itself doesn't work due to the builtin macro, what about a module level attribute? E.g.
`#![verifier(allow_global_asm)]`? I can look into how to make that work if you think that's reasonable.

source/rust_verify/src/rust_to_vir.rs
```diff
@@ -425,6 +425,11 @@ fn check_item<'tcx>(
             // type alias (like lines of the form `type X = ...;`
             // Nothing to do here - we can rely on Rust's type resolution to handle these
         }
+        ItemKind::GlobalAsm(..) =>
+        //if get_verifier_attrs(ctxt.tcx.hir().attrs(item.hir_id()))?.external =>
```
utaal, line 429: Please remove this comment and add a TODO to add a crate-level attribute to enable `global_asm`.

utaal (approved):

source/rust_verify/src/rust_to_vir.rs
ziqiaozhou: Fixed this comment.

PR #222, opened 2022-08-01
Warn the user when we may not have reported all latent verification errors

utaal:
This also helps the user learn about `--multiple-errors`.

@matthias-brun reported this would be useful, as the default behavior is sometimes surprising.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -450,6 +450,14 @@ impl Verifier {
                 }
             }
         }
+        if error_as == ErrorAs::Error && checks_remaining == 0 {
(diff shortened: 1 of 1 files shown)


source/rust_verify/src/verifier.rs
```diff
@@ -450,6 +450,14 @@ impl Verifier {
                 }
             }
         }
+        if error_as == ErrorAs::Error && checks_remaining == 0 {
+            let multispan = MultiSpan::from_spans(vec![from_raw_span(&context.0.raw_span)]);
+            let msg = format!(
+                "{}: not all errors may have been reported; rerun with an higher value for --multiple-errors to find other potential errors in this function",
```
Chris-Hawblitzel, line 456: should say "a higher"

Chris-Hawblitzel (approved):

PR #224, opened 2022-08-02
Add recommends checks for casts

utaal:
We _could_ only emit these for narrowing casts (by inspecting the types): do you think it's worth it?

### DIFF
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -167,6 +167,8 @@ pub(crate) enum Attr {
     IntegerRing,
     // Use a new dedicated Z3 process just for this query
     SpinoffProver,
+    // Suppress the recommends check for narrowing casts that may truncate
+    Truncate,
 }
 
 fn get_trigger_arg(span: Span, attr_tree: &AttrTree) -> Result<u64, VirErr> {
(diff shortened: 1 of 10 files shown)


Chris-Hawblitzel:
I have mixed feelings about this.  When a user says `x as u8` and it's possible that `x >= 256`, maybe the user made a mistake in writing `x as u8` and would be grateful for a warning about it.  On the other hand, maybe the user intended to truncate `x` to a `u8`, and would be annoyed by the warning (they might be doubly annoyed -- first, that they had to write `as ...` to satisfy Rust's type checking, and then second that Verus immediately warns them about the cast that Rust told them they had to put in).  In the latter case, it's not clear how the user suppresses the warning.

tjhance:
The situation seems similar to overflow checks for add - _most_ of the time, the user doesn't expect overflow / clipping, but they may occasionally intend it - so I think we should pick a similar approach for consistency of philosophy. IIRC, the consensus for overflow was:

 * emit bounds checks
 * have some specialized `overflow_add` function to be used for the case where overflow is desired

(Someone should correct me if I'm misremembering.) I think it'd be reasonable to do something similar here.

utaal:
Considering that the current encoding defines a truncating narrowing cast as an `arbitrary()` underspecified value that only fulfills the typing invariant, we (@tjhance, and others at the meeting) are leaning towards:
* adding a recommends check on the narrowing cast,
* adding an attribute to suppress the recommends check if the cast was intentional,
* and, if it ever comes up, adding a `truncating_cast` function that specifies that the result is modulo the target size, if the user wants the truncating behavior and wants to reason about it.

Chris-Hawblitzel (approved):

PR #225, opened 2022-08-02
Fix #215: do not panic on `return ()`

### LINKED ISSUE #215: `return ();` causes panic
```rust
#[allow(unused_imports)]
use builtin::*;
mod pervasive;

fn f() {
    return ();
}

fn main() {
}
```
results in
```
fhs m: source 1 ν ./tools/rust-verify.sh /st/verus/test.rs
thread 'rustc' panicked at 'internal error: missing return type', vir/src/modes.rs:800:30
```

utaal:


tjhance:
this seems fine, but I'm not much of an expert on the mode-checking at this point 😅. Chris should probably review

Chris-Hawblitzel (approved):

PR #232, opened 2022-08-03
Add support to spinoff the prover for assert by nonlinear z3 queries

utaal:
This enables
```rust
#[verifier(spinoff_prover)] assert(expr) by (nonlinear_arith) { ... }
```

cc/ @Chris-Hawblitzel


utaal:
Note: this isn't quite right because the AIR logs for the inline nonlinear queries are overwritten by the latest one in a function.

PR #234, opened 2022-08-04
Add a temporary flag to set an assumption on the size of a machine word to either 32 or 64 bits

utaal:


utaal:
@Chris-Hawblitzel I'm going to land this, but please have a look when you get the chance.

PR #238, opened 2022-08-04
Make termination checks for exec functions, add support for mutable references in parameters, and emit a warning that the check doesn't account for loops

utaal:
I do not believe this opens to any potential unsoundness, especially because it's already possible to write a non-terminating recursive exec function with a loop.

Reported by @matthias-brun.

### DIFF
--- a/source/rust_verify/example/recursion.rs
+++ b/source/rust_verify/example/recursion.rs
@@ -91,6 +91,16 @@ fn run_arith_sum(n: u64) -> u64 {
     result
 }
 
+fn exec_with_decreases(n: u64) -> u64
+    decreases 100 - n
+{
+    if n < 100 {
+        exec_with_decreases(n + 1)
+    } else {
+        n
+    }
+}
+
 #[verifier(external)]
 fn main() {
     let args = std::env::args();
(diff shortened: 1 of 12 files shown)


utaal:
@Chris-Hawblitzel I'm going to land this but please have a look when you get the chance.

utaal:
I am _not_ confident this is okay.

Chris-Hawblitzel:
Can you say what the motivation for this change is?  I could imagine having an option that either disables termination checks for both exec recursive functions and exec loops, or enables termination checks for both exec recursive functions and exec loops.

tjhance:
I was already under the impression we didn't support termination checking for `exec` functions, seeing as, `exec` functions support while loops, which don't support termination checking. As far as I can tell, Verus never complains about the existence of loops and there isn't support for opting in to exec-mode termination checking (at the moment) so permitting `decreases` on an exec function would be misleading.

tjhance:
well, i don't object to supporting it in the future. I'm just saying Andrea's change makes sense now and until such support does exist.

Chris-Hawblitzel:
I think in the design I'm proposing, though, we wouldn't be disallowing and removing the decreases clauses that already exist.

Chris-Hawblitzel:
We would, however, remove by default the requirement that exec functions have `decreases` clauses.  I support that part of the change.

utaal:
The initial impetus was due to the fact that the support was incomplete: decreases did not work for functions with mutable references as arguments. That turned out to be a simple fix, so those are now supported.

The need to check termination and the guarantees provided by the check were unclear to me. From the discussion:
* termination checks for exec functions aren't necessary for soundness,
* the current check does not take into account loops.

So, this PR, now:
* makes the termination check for exec functions optional (only enabled when there's a `decreases` clause),
* warns the user that a termination check on an exec passing doesn't guarantee that the function terminates if there are loops.

tjhance:
can you add a warning or error if the user tries to use `decreases` on a loop? It looks like that's still missing, I think.

PR #240, opened 2022-08-04
Skip recommends check for nonlinear arithmetic query, emit note when skipping recommends check

channy412:
This pull request is about making
1) non-linear queries to skip recommends check
2) Verus emit notes when skipping recommends check.

As reported by @utaal and  @matthias-brun, `assert_nonlinear_by`'s recommends check often fails to point out the cause of verification failure. It is partially due to the query separation, which makes the separate query miss many facts from the original context. Also, recommends check's purpose might differ from nonlinear queries' purpose, which is often just producing simple facts about nonlinear arithmetic.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -880,12 +880,15 @@ impl Verifier {
                 for command in commands.iter().map(|x| &*x) {
                     let CommandsWithContextX {
                         span,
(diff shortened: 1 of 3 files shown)


utaal (approved):
Looks good, thanks!

(I've tried to simplify the control flow of verify_module a bit.)

PR #244, opened 2022-08-09
Use the new Z3 arithmetic solver for nonlinear queries

utaal:
In my testing it is always either as good or significantly better at discharging nonlinear proof goals.

We need to always spinoff queries using this solver as once a query is run on a z3 instance, requests to change the arithmetic solver are ignored.

### DIFF
--- a/source/air/src/context.rs
+++ b/source/air/src/context.rs
@@ -252,10 +252,10 @@ impl Context {
             self.set_z3_param_bool(option, true, true);
         } else if value == "false" {
             self.set_z3_param_bool(option, false, true);
-        } else if let Ok(v) = value.parse::<f64>() {
-            self.set_z3_param_f64(option, v, true);
         } else if let Ok(v) = value.parse::<u32>() {
             self.set_z3_param_u32(option, v, true);
+        } else if let Ok(v) = value.parse::<f64>() {
+            self.set_z3_param_f64(option, v, true);
         } else if value.is_ascii() {
             self.set_z3_param_str(option, value, true);
         } else {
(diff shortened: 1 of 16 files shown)


channy412:
Thanks for the changes! Looks great to me. This change also makes the following test case go through (which is failing in `main` at this point).

https://github.com/secure-foundations/verus/blob/main/source/rust_verify/tests/nonlinear.rs#L15

channy412 (approved):

jaybosamiya:
@yizhou7 and I were looking at some queries, and noticed that `arith.nl` is set to false and never set to true for nonlinear queries (only the `arith.solver` is updated to 6). It appears that this PR removed the bits that would set it to true and set back to false ([here](https://github.com/verus-lang/verus/pull/244/files#diff-7d296b788454a9aaa5c4359d8852b102dc491d291b4bbb824e2b0968a98bd5f1L1315-L1317) and [here](https://github.com/verus-lang/verus/pull/244/files#diff-7d296b788454a9aaa5c4359d8852b102dc491d291b4bbb824e2b0968a98bd5f1L1873-R1868))

@utaal do you recall why you removed those `arith.nl<-true`+`arith.nl<-false` bits?

jaybosamiya:
For anyone looking at this later, @utaal pointed out that the [Z3 docs](https://microsoft.github.io//z3guide/programming/Parameters/) for `arith.nl` state:
> (incomplete) nonlinear arithmetic support based on Groebner basis and interval propagation, relevant only if smt.arith.solver=2

PR #246, opened 2022-08-09
add verifier(inline) for contains_pair; contains

tjhance:


utaal (approved):
I think @Chris-Hawblitzel is on vacation, but I think this can be merged. (@Chris-Hawblitzel obviously feel free to complain when you see this.)

utaal:
I had to revert this due to https://github.com/secure-foundations/verus/issues/248 for which there wasn't an obvious workaround.

PR #247, opened 2022-08-09
Make rust-verify.sh work from any directory

tchajed:
As a bonus, the script now passes [shellcheck](https://www.shellcheck.net/).

I did switch to using exported environment variables; was there a reason for the use of `eval`?


utaal:
> I did switch to using exported environment variables; was there a reason for the use of eval?

My confidence with bash is low, so I kept things as simple as I could. I don't think there's any issue with exporting them unless someone somehow inadvertently sources the script.

PR #255, opened 2022-08-10
`--expand-errors` for error localization 

channy412:
When a user runs Verus with `--expand-errors` flag, Verus will try to minimize the failing predicate by expression splitting. 

It saves error spans and reruns `verify_module`. During VIR -> SST translation, it splits previously failed assertions. 
During the "re-run" for `expand-errors`, it will not report errors unless it is generated from a splitted expression.

Now it only splits on `assert`, `requires`, and `ensures`.


source/vir/src/split_expression.rs
```diff
@@ -0,0 +1,546 @@
...
+        | (TypX::Tuple(_), TypX::Tuple(_))
+        | (TypX::Lambda(_, _), TypX::Lambda(_, _))
+        | (TypX::Datatype(_, _), TypX::Datatype(_, _))
+        | (TypX::TypParam(_), TypX::TypParam(_))
+        | (TypX::TypeId, TypX::TypeId)
+        | (TypX::Air(_), TypX::Air(_)) => true,
+        (TypX::Boxed(b1), TypX::Boxed(b2)) => is_same_type(b1, b2),
+        _ => false,
+    }
+}
```
tjhance, line 121: There's already a similar function ast_util, `types_equal`. This function looks like it computes something different, only comparing the root nodes of the type trees, and it's not clear to me if that's a mistake or if it's intentional. If it _is_ intentional, it isn't clear to me at all why you'd want to compute this, so please document it.

source/vir/src/split_expression.rs
```diff
@@ -0,0 +1,546 @@
...
+    body_exp: &Exp,
+    params: &Params,
+    exps: &Exps,
+) -> Result<Exp, (Span, String)> {
+    match *body_exp.typ {
+        TypX::Bool => (),
+        _ => {
+            return Err((
+                body_exp.span.clone(),
+                "Skip inlining for non-boolean expreesion".to_string(),
```
tjhance, line 133: typo, `expreesion` -> `expression`

tjhance: Also, I'm not sure what the context of this error is or why it would trigger, but it doesn't really sound like an error message. If it's "skipping" something, that makes it sound like a warning of some sort.

source/vir/src/split_expression.rs
```diff
@@ -0,0 +1,546 @@
...
+    };
+    let mut arg_map: HashMap<Arc<String>, Exp> = HashMap::new();
+    let mut count = 0;
+    for param in &**params {
+        let exp_to_insert = &exps[count];
+        if !is_same_type(&param.x.typ, &exp_to_insert.typ) {
+            return Err((
+                exp_to_insert.span.clone(),
+                "Error: arg type mismatch during expression inlining".to_string(),
+            ));
```
tjhance, line 145: Same question, in what context does this error happen? Is it a response to some user error, or does this indicate some internal VIR error?

source/vir/src/split_expression.rs
channy412: Thanks for the review!  You are right, I should use `types_equal`.

source/vir/src/split_expression.rs
channy412: I added this check just to be cautious, and I am not sure if this can be triggered.  If this is triggered, it would be some internal VIR error.

tjhance: okay that's what I expected. Can you change it to a panic or at least make it more clear it's an internal error?

As written, it looks like a user-level error which might be confusing if a user thinks it means they did something wrong.

source/vir/src/split_expression.rs
channy412: I added this after getting an error from an attempt to inline a function with "empty body". I think it consisted of `unimplemented!`, and `verifier(external_body)` attribute.  

All `Err` in `subsitute_argument` and `tr_inline_expression` are now downgraded to "Note",  specifying the reason why the function inlining failed.  I now realize I should add more documentation here. Thanks!

source/vir/src/split_expression.rs
channy412: You are right, I think it is more like a `panic`.

source/vir/src/split_expression.rs
tjhance: Can you make a specific message for `external_body`, then? (Similar to the ones about opaque functions)

channy412: That sounds great -- thanks!

kd054238 (approved):

source/vir/src/ast_to_sst.rs
```diff
@@ -657,8 +690,44 @@ fn stm_call(
...
+        // we are spliting the `requires` expression on the call site.
+        // If we split the `requires` expression on the function itself,
+        // this splitted encoding will take effect on every call site, which is not desirable.
+        //
+        // Also, note that prevasive::assert is consisted of `requires` and `ensures`.
+        // therefore, we are also splitting pervaisve::assert here
+        let params = &fun.x.params;
+        for e in &**fun.x.require {
+            let exp = crate::split_expression::pure_ast_expression_to_sst(ctx, e, params);
+            let arg_exps = Arc::new(vec_map(&args, |x| x.0.clone()));
```
Chris-Hawblitzel, line 704: This should use `small_args` rather than `args`.

source/rust_verify/src/verifier.rs
```diff
@@ -1012,8 +1080,37 @@ impl Verifier {
...
+            } else {
+                compiler.diagnostic().note_without_error(&format!(
+                    "re-verifying module {} for error localization",
+                    &module_name
+                ));
+            }
+            ctx.debug_expand_targets = self.expand_targets.to_vec();
+            ctx.expand_flag = true;
+            self.expand_flag = true;
+            self.verify_module(compiler, &poly_krate, module, &mut ctx)?;
```
Chris-Hawblitzel, line 1111: It's not clear to me why we reverify the whole module.  I would have thought that we'd reverify just the individual functions that fail, as they fail.  In other words, I would have expected this to be inside the `Create queries to check the validity of proof/exec function bodies` loop.

source/vir/src/split_expression.rs
```diff
@@ -0,0 +1,530 @@
...
+        if !crate::ast_util::types_equal(&param.x.typ, &exp_to_insert.typ) {
+            panic!("Internal error: arg type mismatch during expression inlining");
+        }
+        arg_map.insert(param.x.name.clone(), exp_to_insert.clone());
+        count = count + 1;
+    }
+    return subsitute_argument(body_exp, &arg_map);
+}
+
+pub(crate) fn pure_ast_expression_to_sst(ctx: &Ctx, body: &Expr, params: &Params) -> Exp {
```
Chris-Hawblitzel, line 135: Rather than running expr_to_exp again, you can get the already computed SST body from `fun_ssts`.

Chris-Hawblitzel (changes requested):

source/vir/src/split_expression.rs
```diff
@@ -0,0 +1,530 @@
...
+use crate::ast_to_sst::{get_function, State};
+use crate::context::Ctx;
+use crate::def::Spanned;
+use crate::sst::{Bnd, BndX, Exp, ExpX, Exps, Stm, StmX, UniqueIdent};
+use air::ast::{Binder, BinderX, Span};
+use air::errors::Error;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+fn subsitute_argument(
```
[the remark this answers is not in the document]
channy412: Thank you! I will use the one in `sst_util.rs`.

source/rust_verify/src/verifier.rs
channy412: I agree, that sounds better to me too. Thanks!

source/vir/src/split_expression.rs
channy412: I will try to use the information in `fun_ssts`. Thanks!

source/vir/src/ast_to_sst.rs
channy412: I thought I need the whole `args` to be able to split the arguments too. (If the argument of `assert` can be split, then converting `assert(A && B)` to `assert(A)` and `assert(A => B)`).
but I might be missing obvious facts here; could you explain?

Chris-Hawblitzel: Oh, I see.  Never mind, then.  Using `args` sounds like the right thing.

channy412: thank you!

source/vir/src/ast_to_sst.rs
```diff
@@ -657,8 +690,53 @@ fn stm_call(
...
+        // this splitted encoding will take effect on every call site, which is not desirable.
+        //
+        // Also, note that prevasive::assert is consisted of `requires` and `ensures`.
+        // therefore, we are also splitting pervaisve::assert here
+        let params = &fun.x.params;
+        let typ_bounds = &fun.x.typ_bounds;
+        for e in &**fun.x.require {
+            let exp = crate::ast_to_sst::expr_to_exp_as_spec(
+                &ctx,
+                &HashMap::new(),
```
Chris-Hawblitzel, line 706: This should probably be `&state.fun_ssts,`

Chris-Hawblitzel (approved):

PR #259, opened 2022-08-17
make emitting skip-recommends note optional

channy412:
"skip-recommends" notes are not very informative for proof debugging. Also, users can easily get a lot of warnings about this, especially when they use inlined bit-vector/non-linear assertions frequently. 

If a user wants to be reported for skipping recommends check explicitly, these notes can be optionally turned on by giving `--emit-skip-recommends` in the command line.

### DIFF
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -59,6 +59,7 @@ pub struct Args {
     pub profile_all: bool,
     pub compile: bool,
     pub solver_version_check: bool,
+    pub emit_skip_recommends: bool,
 }
 
 pub fn enable_default_features(rustc_args: &mut Vec<String>) {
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (approved):

utaal:
I kinda disagree with this. Hiding a note behind a default-false flag makes it hardly discoverable (I think it’s unlikely anyone would ever set it). If these were too noisy, I believe we can either:
* just remove the info message altogether, or
* only print it once by default (“recommends checks for some assert by queries were skipped” or similar).

@Chris-Hawblitzel @channy412

Chris-Hawblitzel:
> I kinda disagree with this. Hiding a warning behind a default-false flag makes it hardly discoverable (I think it’s unlikely anyone would ever set it). If these were too noisy, I believe we can either:
> 
> * just remove the warning message and the flag altogether, or
> * only print it once by default (“recommends checks for some assert by queries were skipped” or similar).
> 
> @Chris-Hawblitzel @channy412

Of these, I'd prefer removing the message and the flag altogether.

channy412:
Thanks for the feedback!  I will simply remove the warning notes than having them as an option.

channy412:
> Thanks for the feedback! I will simply remove the warning notes than having them as an option.

I just pushed this simple change to main -- https://github.com/secure-foundations/verus/commit/c2c2c1a5549f9191c6a4f64aa15861901e9f1a98 
Thanks!

utaal:
Thank you!

PR #264, opened 2022-08-22
Use subtree for syn/pretty-printing dependencies

tjhance:
Since jay's experience seemed favorable, I decided to go ahead and pull this together. This embeds the `syn` and `prettyplease` crates into the verus repo, so we can now make changes to them as normal verus commits, but we can still sync them with the external repositories at our leisure.

The PR is pretty big, because the commits that creates the subtrees look like ordinary commits that just add a bunch of files.


utaal:
Thanks for looking into this.

I'd like to take a look too before we merge, though!

jaybosamiya (approved):
Looks good to me, although I do have a couple of non-blocking minor suggestions:

1. We may want to add a `.gitattributes` file that marks everything in the `dependencies` directory as containing vendored code. Just the line `dependencies/** linguist-vendored` in a `.gitattributes` file at the root of the repository should be sufficient (or we could pick and choose which dependencies are marked vendored).

2. We might want to add `dependencies/update-syn.sh`, `dependencies/update-prettyprinting.sh` scripts that do the relevant `git subtree pull --prefix .... https://..... main --squash` for future use

jaybosamiya:
Also, copying my comments on `git-subtree` from Slack for future reference

<details><summary>Click to expand</summary>

I have started using it semi-recently in some personal projects (maybe over the past year or so), and it is a fairly different style from git-submodules. Generally, I like it.

Submodules are like a symlink, while subtree is like a full deep copy. For users of the repository, it becomes easier to use a subtree, because they don't need to do --recursive when cloning (or remember to init and update submodules).

On the other hand, it adds non-trivial bulk to the repository, so if a particular submodule is optional, it is best to leave as submodule rather than as subtree, in order to keep the repository smaller

Also, keep track of whether the subtree changed can be a tad bit more difficult than a submodule (where there needs to be an explicit change in the hash, rather than some random change somewhere in the whole dir), but this is a tiny tiny issue, that might not even be worth mentioning, since `git log` and such are sufficient tbqh.

Oh also, when doing git-subtree operations, you do need to remember to do their squashed versions, just to ensure you don't end up drastically complicating the git history (which would make some of the stuff in the "changes" thing I mentioned above waay more annoying) but other than that, the operations are fairly reasonable.

While git subtrees are an advanced feature (the command line interface for it could use some UX work), most people on the repository don't need to ever interact with it, and can use the subdir as if it were a normal subdir in the repo. the only ones who need to worry about it being a subtree are the ones who initialize it, and the ones who need to update it from upstream (or push back to upstream). This is strongly unlike submodules, where everyone needs to learn the pain.

So yeah, overall, I have started quite liking git-subtree; especially in scenarios where I've needed to fork some project to make modifications, but I don't want to deal with maintaining yet another repository that I sync up regularly. If the fork is temporary, I might just do a submodule temporarily though, since that keeps the overall bulk smaller.

---

Fwiw, the staggered updates problem is (one of the reasons) why some companies use mono repos. git-subtree gives many of the benefits of mono repos, without forcing strict mono-repo (assuming folks bother pulling from upstream once in a while).

---

Oh and finally, if we're switching to subtrees in the verus repo, I would strongly recommend keeping scripts that do the pull-and-squash operation to update. It will make everyone's lives easier in the long run, due to the aforementioned UX problems, and the fact that it is easy to forget to squash (fix to resquash is straightforward if you know what you are doing, but having to reflog while trying to get other work done is still annoying; best have a script instead).

</details>

utaal:
I'm not sure I'm sold on this yet, but one thing I didn't like was throwing away author/blame info for commits in our forks, which I suppose this would supersede; so I tried preserving them by turning them into commits into this repo. The result is this https://github.com/secure-foundations/verus/commits/main-subtrees-utaal which preserves authorship and commit dates for commits by Travis and Chris on syn and prettyplease.

jaybosamiya:
Oh yes indeed, authorship/blame info would be thrown away at the first point of the squash, and indeed making the squash happen earlier (before any changes that were done on syn/prettyplease) and moving the commits over would be better. Thanks @utaal for doing that (I had forgotten that that issue would occur; glad you caught it!).

Weirdly, running `git log dependencies/` on the new branch `main-subtrees-utaal` gives the same output as https://github.com/secure-foundations/verus/commits/main-subtrees-utaal/dependencies (i.e., does not show the changes that have happened on top of the original forks).

_However_ the blame information shows up as expected (https://github.com/secure-foundations/verus/blame/main-subtrees-utaal/dependencies/syn/src/item.rs#L127). I suspect this is due to how the merge occured, and that going forward, the log/history view in the directory would start to work better.

Overall, I do think that the new `main-subtrees-utaal` is a pure win, and this PR should probably be `git reset --hard main-subtrees-utaal && git push --force-with-lease`d before merging.

jaybosamiya (commented):
Removing the merge approval due to above

tjhance:
thanks, this is done as requested. (I also appreciate that you made the squash commits authored by 'nobody')

tjhance:
@Chris-Hawblitzel could we get your input on this? Since this is probably going to impact your workflow more than anybody else ...

Chris-Hawblitzel (approved):
I don't know enough about git to say much about the long-term impacts, but I've tried this pull request out and it certainly makes editing our version of syn much easier.  And our fork of syn is conceptually part of Verus anyway -- it's not intended to be used by itself without Verus.

utaal (approved):
I have a feeling I'm signing myself up for some git surgery in the future with this, but let's give it a shot.

PR #265, opened 2022-08-23
Support for unicode strings

isubasinghe:
This PR adds support for char's and consequently unicode. Unfortunately, this means we cannot talk about chars as bytes anymore even when it makes sense to do so, this is to be addressed in a following PR.

### DIFF
--- a/source/pervasive/string.rs
+++ b/source/pervasive/string.rs
@@ -36,16 +36,16 @@ impl<'a> StrSlice<'a> {
     #[verifier(external_body)]
     pub fn unicode_len(&self) -> (l: usize)
         ensures 
-            l as nat === self.view().len()
+            l as nat === self@.len()
     {
         self.inner.chars().count()
     }
     /// Warning: O(n) not O(1) due to unicode decoding needed
     #[verifier(external_body)]
     pub fn get_char(&self, i: usize) -> (c: char)
-        requires i < self.view().len()
+        requires i < self@.len()
         ensures
-            self.view().index(i as int) === c
+            self@.index(i as int) === c
     {
         self.inner.chars().nth(i).unwrap()
     }
@@ -54,10 +54,10 @@ impl<'a> StrSlice<'a> {
     pub fn substring_ascii(&self, from: usize, to: usize) -> (ret: StrSlice<'a>)
         requires
             self.is_ascii(),
-            from < self.view().len(),
-            to <= self.view().len(),
+            from < self@.len(),
+            to <= self@.len(),
         ensures
-            ret.view() === self.view().subrange(from as int, to as int)
+            ret@ === self@.subrange(from as int, to as int)
     {
         StrSlice {
             inner: &self.inner[from..to],
@@ -67,10 +67,10 @@ impl<'a> StrSlice<'a> {
     #[verifier(external_body)]
     pub fn substring_char(&self, from: usize, to: usize) -> (ret: StrSlice<'a>)
         requires
-            from < self.view().len(),
-            to <= self.view().len()
+            from < self@.len(),
+            to <= self@.len()
         ensures
-            ret.view() === self.view().subrange(from as int, to as int)
+            ret@ === self@.subrange(from as int, to as int)
     {
         let mut char_pos = 0;
         let mut byte_start = None;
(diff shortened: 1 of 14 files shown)


source/pervasive/string.rs
```diff
@@ -25,47 +25,82 @@ pub const fn new_strlit<'a>(s: &'a str) -> StrSlice<'a> {
...
 
+    pub spec fn is_ascii(&self) -> bool;
+    
+    // len returns the byte length, this returns
+    // the unicode length 
+    // Warning: O(n)
     #[verifier(external_body)]
-    pub fn len(&self) -> (l: usize)
-        ensures self.is_ascii() ==> l as nat === self.bytes().len()
+    pub fn unicode_len(&self) -> (l: usize)
```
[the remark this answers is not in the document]
utaal: I think this PR comment should be a docstring for the function instead.

source/pervasive/string.rs
```diff
@@ -25,47 +25,83 @@ pub const fn new_strlit<'a>(s: &'a str) -> StrSlice<'a> {
...
+            else { break; }
+        }
+        let byte_start = byte_start.unwrap();
+        let byte_end = byte_end.unwrap();
+        StrSlice {
+            inner: &self.inner[byte_start..byte_end]
+        }
+    }
+
     #[rustc_diagnostic_item = "pervasive::string::StrSlice::to_string"]
```
utaal, line 100: This isn't in this pull request, but why do you need this diagnostic_item?

source/pervasive/string.rs
```diff
@@ -25,47 +25,83 @@ pub const fn new_strlit<'a>(s: &'a str) -> StrSlice<'a> {
...
+    /// The len() function in rust returns the byte length. 
+    /// It is more useful to talk about the length of characters and therefore this function was added. 
+    /// Please note that this function counts the unicode variation selectors as characters. 
+    /// Warning: O(n)
     #[verifier(external_body)]
-    pub fn len(&self) -> (l: usize)
-        ensures self.is_ascii() ==> l as nat === self.bytes().len()
+    pub fn unicode_len(&self) -> (l: usize)
+        ensures 
+            l as nat === self.view().len()
```
utaal, line 39: You can use `@` for .view() everywhere except for the declaration.

source/pervasive/string.rs
```diff
@@ -25,47 +25,83 @@ pub const fn new_strlit<'a>(s: &'a str) -> StrSlice<'a> {
...
+        }
+        let byte_start = byte_start.unwrap();
+        let byte_end = byte_end.unwrap();
+        StrSlice {
+            inner: &self.inner[byte_start..byte_end]
+        }
+    }
+
     #[rustc_diagnostic_item = "pervasive::string::StrSlice::to_string"]
     #[verifier(external_body)]
```
utaal, line 101: I didn't notice this before, but I don't think this needs to be `external_body` as it calls a function with a requires/ensures.

source/rust_verify/tests/strings.rs
```diff
@@ -267,3 +266,156 @@ test_verify_one_file! {
...
+        use pervasive::string::*;
+        fn test() {
+            let a = new_strlit("ABC");
+            let b = a.to_string();
+            let c = b.as_str();
+            reveal_strlit("ABC");
+            assert(a@.ext_equal(c@));
+            assert(a.is_ascii());
+            // TODO assert(b.is_ascii());
+            // TODO assert(c.is_ascii());
```
utaal, line 418: I believe these should verify. I think some of the pervasive::string functions need to maintain/forward information about `is_ascii`.

utaal (changes requested):

source/pervasive/string.rs
isubasinghe: removed

source/rust_verify/example/playground.rs
```diff
@@ -1,37 +1,25 @@
-use builtin_macros::*;
```
utaal, line 1: Please remove this change.

utaal (changes requested):

source/rust_verify/example/playground.rs
isubasinghe: Sorry I should really avoid using `git add .`

utaal (approved):

PR #268, opened 2022-08-24
force using `add`, `sub`, `mul` in bit-vector inside verus!

channy412:
Inside `verus!`, integers can get widened to `int`, while bit-vector assertions favor fixed-bit integers. To use fixed bit integers, users can use `add`, `sub`, `mul` instead of `+`, `-`, `*`.  To enforce this usage, inside `verus!` macro expansion, it explicitly checks this.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -16,9 +16,9 @@ use syn_verus::{
     braced, bracketed, parenthesized, parse_macro_input, AttrStyle, Attribute, BinOp, Block,
     DataMode, Decreases, Ensures, Expr, ExprBinary, ExprCall, ExprLit, ExprLoop, ExprTuple,
     ExprUnary, ExprWhile, Field, FnArgKind, FnMode, Ident, ImplItemMethod, Invariant, Item,
-    ItemConst, ItemEnum, ItemFn, ItemStruct, Lit, Local, ModeSpec, ModeSpecChecked, Pat, Path,
-    PathArguments, PathSegment, Publish, Recommends, Requires, ReturnType, Signature, Stmt, Token,
-    TraitItemMethod, UnOp, Visibility,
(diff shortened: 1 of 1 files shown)


source/builtin_macros/src/syntax.rs
```diff
@@ -237,6 +250,7 @@ impl Visitor {
         }
 
         self.inside_ghost -= 1;
+        self.inside_bitvector = false;
```
tjhance, line 253: `self.inside_bitvector = bitvector`

source/builtin_macros/src/syntax.rs
```diff
@@ -897,18 +926,23 @@ impl VisitMut for Visitor {
...
                                 Expr::Verbatim(quote_spanned!(span => (#left).spec_sub(#right)));
                         }
-                        BinOp::Mul(..) => {
+                        BinOp::Mul(..) if !self.inside_bitvector => {
                             *expr =
                                 Expr::Verbatim(quote_spanned!(span => (#left).spec_mul(#right)));
                         }
+                        BinOp::Add(..) | BinOp::Sub(..) | BinOp::Mul(..) => {
+                            *expr = Expr::Verbatim(
+                                quote_spanned!(span => compile_error!("Inside bit-vector assertion, use `add` `sub` `mul` for fixed-bit operators, instead of `+` `-` `*`, ")),
```
Chris-Hawblitzel, line 943: Here, `add`, `sub`, and `mul` could be confused with the `.add`, `.sub`, and `.mul` *methods* on integer types, so we might want to add some additional information: `(see the functions builtin::add(left, right), builtin::sub(left, right), and builtin::mul(left, right))`.

Chris-Hawblitzel (approved):

source/builtin_macros/src/syntax.rs
channy412: Ah, I agree. Sound great, thanks!

PR #271, opened 2022-08-25
Support splitting quantifier during `expand-errors` 

channy412:
When a predicate with `forall` fails, `expand-errors` now tries to split that quantifier, and it also shrinks the binder if the split expression does not reference one of the bound variables anymore. 

The part that I want a closer review is this:
Inlining a spec function can generate a `Could not automatically infer triggers` error even if the original expression succeeds in selecting triggers. However, for just proving `forall` clause, triggers do not matter. Therefore, during the `finalize_exp` in `ast_to_sst`, it catches the trigger selection errors and makes the trigger part as an empty vector. I wonder if this behavior can make troubles later on the translation process.

### DIFF
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -250,9 +250,9 @@ impl State {
                     let trigs = match trigs {
                         Ok(trigs) => trigs,
                         Err(e) if !ctx.expand_flag => return Err(e),
-                        // ininling spec functions can make "Could not automatically infer triggers" error.
+                        // Ininling a spec function can generate "Could not automatically infer triggers" error even if the original expression succeeds to select triggers.
                         // However, for proving `forall` clause, triggers not matter. (triggers are needed when using `forall` as a hypothesis)
-                        // For split `forall` clauses, these will not be used as a fact to prove other facts
+                        // Split `forall` clauses will not be used as a fact to prove other VCs.
                         Err(_) if ctx.expand_flag => Arc::new(vec![]),
                         _ => unreachable!(),
                     };
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -246,7 +246,16 @@ impl State {
                         &vars,
                         &body,
                         quant.boxed_params,
-                    )?;
+                    );
+                    let trigs = match trigs {
+                        Ok(trigs) => trigs,
+                        Err(e) if !ctx.expand_flag => return Err(e),
+                        // ininling spec functions can make "Could not automatically infer triggers" error.
+                        // However, for proving `forall` clause, triggers not matter. (triggers are needed when using `forall` as a hypothesis)
+                        // For split `forall` clauses, these will not be used as a fact to prove other facts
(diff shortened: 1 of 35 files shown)


source/vir/src/ast_to_sst.rs
```diff
@@ -246,7 +246,16 @@ impl State {
                         &vars,
                         &body,
                         quant.boxed_params,
-                    )?;
+                    );
+                    let trigs = match trigs {
```
Chris-Hawblitzel, line 250: This should be removed.  The bug was due to expand-errors being processed during ast-to-sst translation rather than after ast-to-sst translation.  This has been fixed in main, so this workaround is no longer necessary.

source/vir/src/split_expression.rs
```diff
@@ -387,16 +388,59 @@ pub(crate) fn split_expr(ctx: &Ctx, state: &State, exp: &TracedExp, negated: boo
...
+                    // Note: trigger selection happens after the AST->SST translation(see crate::ast_to_sst::finalized_exp)
+                    BndX::Quant(
+                        Quant { quant: air::ast::Quant::Forall, boxed_params },
+                        bndrs,
+                        _trigs,
+                    ) if !negated => {
+                        let bndx = BndX::Quant(
+                            Quant { quant: air::ast::Quant::Forall, boxed_params: *boxed_params },
+                            shrink_binders(e1, bndrs),
+                            Arc::new(vec![]), // remove triggers that are already selected
```
Chris-Hawblitzel, line 418: This would be better just keeping the original binders and the original triggers.  (Same for Exists.)

source/vir/src/split_expression.rs
```diff
@@ -458,3 +502,16 @@ pub fn is_split_error(error: &Error) -> bool {
         false
     }
 }
+
+// Shrink binders if the split body does not reference one of the binded variables anymore
+// REVIEW: does this optimization can change SMT query result?
+fn shrink_binders(e1: &Exp, bndrs: &air::ast::Binders<Typ>) -> air::ast::Binders<Typ> {
```
Chris-Hawblitzel, line 508: This can be removed.

source/vir/src/sst_util.rs
```diff
@@ -198,3 +198,22 @@ pub(crate) fn subst_exp(
     assert_eq!(free_vars.num_scopes(), 0);
     e
 }
+
+// referenced `crate::ast_util::referenced_vars_expr`
+pub(crate) fn referenced_vars_exp(exp: &Exp) -> HashSet<Ident> {
```
Chris-Hawblitzel, line 203: This can be removed.

Chris-Hawblitzel (changes requested):
The issues encountered in this pull request were due to expand-errors being processed during ast-to-sst conversion rather than after ast-to-sst conversion.  I fixed the main branch so that expand-errors is handled after ast-to-sst conversion, so this pull request can now be simplified.

channy412:
Thank you, Chris! I updated this branch following the review, as well as fixing little issues with visibility tracking.

Chris-Hawblitzel (approved):

PR #275, opened 2022-09-01
Add support for proof by computation

parno:
For more details see  `source/docs/manual/proof_by_computation.md`


source/vir/src/ast.rs
```diff
@@ -458,6 +476,8 @@ pub enum ExprX {
     Block(Stmts, Option<Expr>),
     /// `assert_by` with a dedicated prover option (nonlinear_arith, bit_vector)
     AssertQuery { requires: Exprs, ensures: Exprs, proof: Expr, mode: AssertQueryMode },
+    /// Assertion discharged via computation
+    AssertCompute(Expr, ComputeMode),
```
utaal, line 480: Would it make sense to use `AssertQuery` for this, and add `Compute(ComputeMode)` as a variant of `AssertQueryMode`?

utaal:
I'd rather we didn't introduce synchronization when it's not strictly necessary. https://github.com/secure-foundations/verus/pull/276 replaces the `Mutex` and `RwLock` with `Send` alternatives.

utaal:
I'd appreciate if you squashed the commits; it makes keeping track of the review changes easier.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+                // Restore the free variables we hid during interpretation
+                let res = crate::sst_visitor::map_exp_visitor(&res, &mut |e| match &e.x {
+                    ExpX::Interp(InterpExp::FreeVar(v)) => {
+                        SpannedTyped::new(&e.span, &e.typ, ExpX::Var(v.clone()))
+                    }
+                    _ => e.clone(),
+                });
+                if debug {
+                    if exp.definitely_eq(&res) {
+                        println!();
```
utaal, line 1339: Can you use rustc's diagnostic output for this? (like we do for other notes/warnings)
This is enabled always when `--log-all` is enabled. If this is intended for internal debugging only, can you either print it to the .interp file or enable it separately (with a flag/env variable)

utaal: If I correctly recall what we discussed, it may be helpful to emit this as a note (even when debugging is disabled) when you send a partially simplified expression to z3 and z3 fails to prove it.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+    fn syntactic_eq(&self, other: &Self) -> Option<bool> {
+        // Easy case where the pointers match
+        if Arc::ptr_eq(self, other) {
+            return Some(true);
+        }
+        // If we can't definitively establish equality, we conservatively return None
+        let def_eq = |b| if b { Some(true) } else { None };
+        use ExpX::*;
+        match (&self.x, &other.x) {
+            (Const(l), Const(r)) => Some(l == r),
```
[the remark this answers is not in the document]
parno: One option would be to explicitly name the cases of `Const` here, so someone who adds another constructor would have to manually update the equality check, and would hopefully think about the consequences.  Would that suffice?

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+                },
+                Len => match &args[0].x {
+                    Interp(Seq(s)) => int_new(BigInt::from_usize(s.len()).unwrap()),
+                    _ => ok,
+                },
+                Index => match &args[0].x {
+                    Interp(Seq(s)) => match &args[1].x {
+                        UnaryOpr(crate::ast::UnaryOpr::Box(_), e) => match &e.x {
+                            Const(Constant::Int(index)) => {
+                                let index = BigInt::to_usize(index).unwrap();
```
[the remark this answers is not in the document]
parno: > this can't fail because the value is negative (type checker requires it to be nat) 

The signature for `seq.index` is: 
```
pub spec fn index(self, i: int) -> A
```
which allows a negative value, right?  So we would need to handle that too, no?

> Can you handle the case instead of panicking?

What would be the right way to handle it?  Since we're representing the sequences using a Vector, we need a `usize` index.  Or are you suggesting I just pick an arbitrary integer (e.g., 42) in the case where the provided index is too large to convert (and possibly issue a warning)?

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+
+// Automatically get syntactic equality over typs, exps, ... once we know syntactic equality over typ, exp, ...
+impl<T: SyntacticEquality> SyntacticEquality for Arc<Vec<T>> {
+    fn syntactic_eq(&self, other: &Self) -> Option<bool> {
+        if self.len() != other.len() {
+            Some(false)
+        } else {
+            self.iter()
+                .zip(other.iter())
+                .try_fold(true, |acc, (l, r)| Some(acc && l.syntactic_eq(r)?))
```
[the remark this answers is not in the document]
parno: > It looks like this always returns None, but if for any index it returns Some(false), then you can definitively return Some(false), even if some other index is indeterminate.

Ah, you're saying that if any index returns `None`, then the `l.syntactic_eq(r)?` will turn the entire result to `None`?   Looking at the [docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold), do I need to use the `ControlFlow` construct instead, if I want to stick with `try_fold`?

As a side note, I suspect it will be rare that we can establish `definitely_ne`, given that we're operating at the syntactic level.

source/vir/src/interpreter.rs
parno: > Can you use rustc's diagnostic output for this?

source/vir/src/ast.rs
parno: Good question.  I think it was added after I forked my branch, so I haven't looked at it closely.  I'll talk to Chanhee to learn more about whether it would work here.

parno: @channy412 says that he currently thinks of  `AssertQuery` as solving a specific assertion in a separate query, which is why the current signature is
```
AssertQuery { requires: Exprs, ensures: Exprs, proof: Expr, mode: AssertQueryMode }
```
In contrast, `assert_by_compute` doesn’t currently use requires or ensures, and it doesn’t spawn a separate Z3 query; it just rewrites the original assert based on the interpreter’s simplifications and then lets the normal pipleline handle the process of turning it into one or more Z3 queries.

However, we could potentially expand our notion of `AssertQuery` to signify all "special modes" of assertion handling, which would then favor a merger here.

Thoughts?

source/vir/src/interpreter.rs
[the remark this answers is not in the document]
tjhance: Whoops, you're right. Still the point stands, we should handle the case rather than panic.

I guess it would technically be sound to pick an arbitrary integer (in the sense that we've left enough things unconstrained until now) but I think that behavior would be odd and unexpected.

 You could just leave the expression unsimplified and issue a warning, like you'd otherwise do when you can't simplify, right?

source/vir/src/interpreter.rs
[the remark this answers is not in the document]
tjhance: Yeah that'd help me sleep easier 👍

source/vir/src/interpreter.rs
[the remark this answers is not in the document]
parno: > You could just leave the expression unsimplified and issue a warning, like you'd otherwise do when you can't simplify, right?

Yes, that's a good point.  Done.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+                Len => match &args[0].x {
+                    Interp(Seq(s)) => int_new(BigInt::from_usize(s.len()).unwrap()),
+                    _ => ok,
+                },
+                Index => match &args[0].x {
+                    Interp(Seq(s)) => match &args[1].x {
+                        UnaryOpr(crate::ast::UnaryOpr::Box(_), e) => match &e.x {
+                            Const(Constant::Int(index)) => {
+                                let index = BigInt::to_usize(index).unwrap();
+                                if index < s.len() { Ok(s[index].clone()) } else { ok }
```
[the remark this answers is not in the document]
parno: Good idea.  Done.

source/vir/src/interpreter.rs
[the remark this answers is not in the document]
parno: I'm always in favor of more/better sleep.  Done.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+
+                                _ => panic!(
+                                    "Type checker should not allow bitwise ops on non-fixed-width types"
+                                ),
+                            };
+                            int_new(r)
+                        }
+                        Clip { range, truncate: _ } => {
+                            let apply_range = |lower: BigInt, upper: BigInt| {
+                                if i < &lower || i > &upper { ok.clone() } else { Ok(e.clone()) }
```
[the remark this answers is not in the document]
parno: I added a placeholder warning for now.  Will fill in with something more substantial.

source/vir/src/ast.rs
utaal: That makes sense. Also, I had forgotten that `assert ... by` and `assert forall ... by` are encoded as `ExprX::Forall`, which I think this more closely resembles.

> it doesn’t spawn a separate Z3 query; it just rewrites the original assert based on the interpreter’s simplifications and then lets the normal pipleline handle the process of turning it into one or more Z3 queries.

Also, yes, I agree this is the important distinction.

`AssertCompute` sounds good, then. If other features with a similar structure come up, we can move to something like `AssertBy(AssertByMode)`, but it's not worth doing now.

source/vir/src/interpreter.rs
utaal: My take is: if it's output that the developer should see, it should go through `Diagnostics`, it should be associated with a span, and it shouldn't be conditional on whether `--log-*` is enabled (there is some machinery for VIR-generated info/warnings in `well_formed` but that needs to be extended if we need it elsewhere), if it's for debugging it should go to the log file.

If I understand this correctly (but it's very possible I don't), the "simplified to" message on failure is pretty similar to this debug info. Printing it only on failure seems like it would prevent noise when the proof goes through but still provide information to the user for debugging.

I believe the difference is that this debug info also mentions whether any simplification happened before sending the expression to z3. Ideally, I think we should print a warning (with a span) only when the computation does no simplification (if I understand correctly, this ends up being equivalent to a regular `assert`), so the user knows to either just use a regular `assert` or make changes so that computation is actually helpful.

utaal:
> I'd appreciate if you squashed the commits; it makes keeping track of the review changes easier.

Ah, I was talking about your commits (before the review). But it's probably fine. Please squash when merging, though.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1438 @@
...
+                    _ => ok,
+                },
+                Index => match &args[0].x {
+                    Interp(Seq(s)) => match &args[1].x {
+                        UnaryOpr(crate::ast::UnaryOpr::Box(_), e) => match &e.x {
+                            Const(Constant::Int(index)) => {
+                                match BigInt::to_usize(index) {
+                                    None => {
+                                        // TODO: Use Diagnostics instead of println
+                                        println!(
```
utaal, line 730: It looks like diagnostics are important for these warnings (so that we can emit a span). `well_formed` collects VIR-generated warnings in a Vec https://github.com/secure-foundations/verus/blob/main/source/vir/src/well_formed.rs#L780 and lets `rust_verify` handle the printing. We may want to do something similar, but it may be a bit trickier for this, because you'd have to thread this through the contexts, I think.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1438 @@
...
+    Len,
+    Index,
+    ExtEqual,
+    Last,
+}
+
+/// Identify sequence functions for which we provide custom interpretation
+fn is_sequence_fn(fun: &Fun) -> Option<SeqFn> {
+    use SeqFn::*;
+    match path_as_rust_name(&fun.path).as_str() {
```
utaal, line 597: Can you add a TODO to use a more robust mechanism for this matching? We may want to use either diagnostic_item or our own attribute to connect these (so that if they ever move they don't get detached from the semantics that we associate here). We should do this for other internal functions, too.

source/vir/src/interpreter.rs
parno: Indeed, I suspect it might be awkward to pass the warnings all the way back up.  At some point, you, Chris, and I had talked about defining a Diagnostics-like trait in AIR, which rust_verify could implement using its own Diagnostics.  That would let us clean up the `println`s in the profiler (e.g., https://github.com/secure-foundations/verus/blob/e72629774e0643370ff9188bb4e8920ef23d9ab6/source/air/src/profiler.rs#L39), and if we're already passing it through VIR to AIR, then it would be natural to use here too.  What do you think?

source/vir/src/interpreter.rs
parno: Just to summarize, it sounds like "Could not take advantage of compute to simplify expression before sending to z3" should always (when it applies) be printed using Diagnostics.  Then that leaves the "Was able to simplify before sending to Z3" message.  Right now, that's behind a `debug` flag, which I thought was primarily associated with the `--debug` flag.  It sounds like `--log-*` also enables it?  From the interpreter's perspective, if you enable `--log-interpreter`, then the simplified expression (if any) will be logged to the interpreter's log file, so printing it to the screen may be redundant.

source/vir/src/interpreter.rs
parno: Yep, done.

source/vir/src/interpreter.rs
utaal: That sounds good!

source/vir/src/interpreter.rs
utaal: > Right now, that's behind a debug flag, which I thought was primarily associated with the --debug flag. It sounds like --log-* also enables it?

Unless I'm misunderstanding something, I don't think this output should be related to `--debug` which "Enable[s] debugging of proof failures".

`--log-all` does not enable `--debug`; your logic enables these `println`s when `--log-interpreter` (or `--log-all`) is enabled here https://github.com/secure-foundations/verus/blob/4db2241c4ed0c5022cd018dd69eee4ae4fcdc718/source/vir/src/interpreter.rs#L1329-L1349

I thought it looked like a bit of a roundabout way to do this but I guessed you were leaving this open for more configurability in the future. It looks like the `debug` flag in the interpreter's `State` is currently equivalent to `log.is_some()`. But it sounds like that wasn't the intent.

source/vir/src/interpreter.rs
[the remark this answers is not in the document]
parno: I added some additional logic to handle this case.

source/vir/src/interpreter.rs
```diff
@@ -0,0 +1,1391 @@
...
+}
+
+impl<T: Clone + SyntacticEquality> SyntacticEquality for Vector<T> {
+    fn syntactic_eq(&self, other: &Self) -> Option<bool> {
+        if self.len() != other.len() {
+            Some(false)
+        } else {
+            self.iter()
+                .zip(other.iter())
+                .try_fold(true, |acc, (l, r)| Some(acc && l.syntactic_eq(r)?))
```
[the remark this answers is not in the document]
parno: Done.

source/vir/src/interpreter.rs
parno: Ah, I see what you mean!  Sorry, my cache on all of this was more completely flushed than I realized.  You're quite right that I apparently tied `debug` to logging.  I think `debug` initially toggled printing information to the terminal, but once we added the logging option, all of that goes to the log file.  Since `debug` doesn't seem to be doing much any more, I removed it, and now we only print a warning when the interpreter fails to do any simplification at all.

source/vir/src/interpreter.rs
parno: Cool, as discussed in yesterday's meeting, I'll plan on adding that after this lands.

source/vir/src/interpreter.rs
parno: Relevant commit: https://github.com/verus-lang/verus/pull/275/commits/6be031244a91d8d4c9f8d769fc30e1364feedca8

utaal (approved):

PR #276, opened 2022-09-02
Patch for #275: Add support for proof by computation

utaal:
Code review in the form of a convenient patch.


parno:
Very interesting!  Rust's compiler errors seemed to want me to push everything into `Arc`s and locks; I didn't realize you could make it work without all of that machinery.  Thanks for the patch.

PR #282, opened 2022-09-05
Replace ExprX::Admit with ExprX::AssertAssume

Chris-Hawblitzel:
This replaces `ExprX::Admit`, which meant `assume(false)`, with `ExprX::AssertAssume`, which can assert or assume an expression.  This allows us to put `assert` and `assume` in `builtin` rather than `pervasive`, which removes a dependency that the `verus!` syntax macro had on `pervasive` and is more consistent with the other forms of `assert` (`assert_nonlinear_by`, `assert_bitvector_by`, `assert_forall_by`, `assert_bit_vector`), which are all in `builtin`.  It allows us to write small test cases and examples that use `assert` but don't require all of `pervasive` (particularly in the tutorial; see https://github.com/secure-foundations/verus/blob/main/source/docs/guide/src/getting_started.md and https://github.com/secure-foundations/verus/blob/main/source/rust_verify/example/guide/getting_started.rs , for example).

@channy412 You might want to check that --expand-errors still works as expected; this incorporates a change to the span printing that we discussed earlier so that assertions are still split correctly in --expand-errors.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -171,6 +171,16 @@ pub fn old<A>(_: A) -> A {
 }
 
 #[proof]
+pub fn assume_(_: bool) {
+    unimplemented!();
+}
+
+#[proof]
+pub fn assert_(_: bool) {
+    unimplemented!();
+}
+
+#[proof]
 pub fn assert_by(_: bool, _: ()) {
     unimplemented!();
 }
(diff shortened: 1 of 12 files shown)


utaal (approved):

channy412:
Looks great to me, thank you!  This also looks more natural for `expand-errors` since it doesn't need to inline `perviasive::assert` into its `requires(b)`

There are a few places that `expand-errors` use `crate::def::pervasive_assert_path()`. 
Now that `verus!` does not depend on pervasive::assert, should I go ahead and remove those code paths?

Chris-Hawblitzel:
Yes, I think you can remove those code paths.

Issue #279, opened 2022-09-02
Rust to VIR doesn't respect hygienic identifiers


utaal:
Identifiers with the same name should be considered distinct if they originate from a macro expansion.

For example, in:

```rust
#[macro_export]
#[doc(hidden)]
macro_rules! assert_sets_equal_internal {
    ($s1:expr, $s2:expr $(,)?) => {
        assert_sets_equal_internal!($s1, $s2, elem => { })
    };
    ($s1:expr, $s2:expr, $elem:ident $( : $t:ty )? => $bblock:block) => {
        let s1 = $crate::pervasive::set_lib::check_argument_is_set($s1);
        let s2 = $crate::pervasive::set_lib::check_argument_is_set($s2);
        ::builtin::assert_by(::builtin::equal(s1, s2), {
            ::builtin::assert_forall_by(|$elem $( : $t )?| {
                ::builtin::ensures(
                    ::builtin::imply(s1.contains($elem), s2.contains($elem))
                    &&
                    ::builtin::imply(s2.contains($elem), s1.contains($elem))
                );
                { $bblock }
            });
            $crate::pervasive::assert(s1.ext_equal(s2));
        });
    }
}
```

`s1` and `s2` maintain their user-visible name after expansion, but they are distinct from variables with the same name in the calling scope (due to macro hygiene). We don't handle this correctly, and treat them as synonyms of variables with the same name in the same scope, i.e. in case there is:

```rust
let s1: usize = 0;
let s2: usize = 0;
...
assert_sets_equal(...);
```

This may result in invalid AIR code (and maybe, potentially unsoundness?).

Originally found by @matthias-brun.

---

* [ ] un-ignore https://github.com/secure-foundations/verus/commit/b68f8e4bcd2ec3b7a5e1de2517fc7d69a9d07375

utaal:
Minimized:

```rust
        macro_rules! assert_with_binding {
            ($s1:expr) => {
                let s1 = $s1;
                assert(s1);
            }
        }

        proof fn test() {
            let s1: nat = 0;
            assert_with_binding!(true);
            assert(s1 === 0);
        }
```

tjhance:
I think we can use this HirID to disambiguate: https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/enum.PatKind.html#variant.Binding

[...]

PR #286, opened 2022-09-09
disambiguate variables with local id to handle macro hygeine (#279)

tjhance:
With this PR, the Rust -> VIR pipeline now disambiguates *all* local variables in the form `name$$id`. The `id` comes from the `ItemLocalId` - according to the rust docs, these are supposed to be stable as long as the Item itself doesn't change.

One snafu that I ran into is that `$` isn't allowed in Singular identifiers. I decided the easiest thing to do would probably be to just write a general-purpose identifier sanitizer for singular. This does mean that all the variable names sanitize the `$` into something ugly, but the alternative would be to somehow thread the needle of unique-ified variable names that look nice in both AIR _and_ Singular ... that sounded brittle to me, whereas a general sanitizer should be pretty foolproof.

Here's one thing I wasn't sure about. I used `$$` so that it wouldn't conflict with the `$` used by the AST -> SST translation for identifier de-duplication. I suspect that—with it now the case that locals are de-duped before reaching VIR—some of that AST -> SST deduplication can be simplified. But that code is pretty complicated, and I wasn't sure I understood what every part was for, so I was wary of modifying it. So if anybody has any suggestions there, it'd be appreciated.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -138,8 +138,15 @@ pub(crate) fn hack_get_def_name<'tcx>(tcx: TyCtxt<'tcx>, def_id: DefId) -> Strin
     debug_name[last_colon + 1..].to_string()
 }
 
-pub(crate) fn ident_to_var<'tcx>(ident: &Ident) -> String {
+pub(crate) fn foreign_param_to_var<'tcx>(ident: &Ident) -> String {
     ident.to_string()
+}
+
+pub(crate) fn local_to_var<'tcx>(
+    ident: &Ident,
+    local_id: rustc_hir::hir_id::ItemLocalId,
+) -> String {
+    ident.to_string() + "$$" + &local_id.index().to_string()
 }
 
 pub(crate) fn is_visibility_private(vis_kind: &VisibilityKind, inherited_is_private: bool) -> bool {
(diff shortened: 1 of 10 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -138,10 +138,17 @@ pub(crate) fn hack_get_def_name<'tcx>(tcx: TyCtxt<'tcx>, def_id: DefId) -> Strin
...
-pub(crate) fn ident_to_var<'tcx>(ident: &Ident) -> String {
+pub(crate) fn foreign_param_to_var<'tcx>(ident: &Ident) -> String {
     ident.to_string()
 }
 
+pub(crate) fn local_to_var<'tcx>(
+    ident: &Ident,
+    local_id: rustc_hir::hir_id::ItemLocalId,
+) -> String {
+    ident.to_string() + "$$" + &local_id.index().to_string()
```
Chris-Hawblitzel, line 149: Can we do this operation (string + "$$" + string) in `def.rs` so that we keep track of the "$$" with the other prefixes, suffixes, and separators?  Also, "$$" is used for SUBST_RENAME_SEPARATOR.  Maybe local_to_var could use "~" instead?

Chris-Hawblitzel (approved):

channy412 (approved):

utaal (approved):
🎉

PR #287, opened 2022-09-13
Update the pointer for smt2utils to the verus-lang version

parno:


parno:
I moved our fork of `smt2utils`, which we use for the profiler, over the `verus-lang` org.  This PR updates the pointer in AIR's `toml` file.

utaal (approved):

PR #289, opened 2022-09-14
Support non_std code verification by replacing std with core and alloc

ziqiaozhou:
The basic idea is to remove std used in pervasive module and all rewrite ops, so that verus can be used to verify and compile non_std rust code.
* The change supports both std and non_std code verification.
* The non_std will need to set features in its project Cargo.toml with non_std and optionally no_global_allocator features
```
[features]
default = ["non_std", "no_global_allocator"]
non_std = []
no_global_allocator = []
```
* non_std feature indicates that the code cannot use modules that are not possible in non_std (e.g., thread modules).
* no_global_allocator feature indicates that the code cannot use modules depending on alloc (e.g., no_global_allocator).
* If you do not set no_global_allocator, your code should be std code or a non_std with a defined global allocator.

### DIFF
--- a/source/pervasive/ptr.rs
+++ b/source/pervasive/ptr.rs
@@ -1,5 +1,4 @@
-use core::mem::MaybeUninit;
-use core::marker;
+use core::{marker, mem, mem::MaybeUninit, alloc};
 #[allow(unused_imports)] use builtin::*;
 #[allow(unused_imports)] use builtin_macros::*;
 #[allow(unused_imports)] use crate::pervasive::*;
@@ -10,7 +9,7 @@ verus!{
 /// `PPtr<V>` (which stands for "permissioned pointer")
 /// is a wrapper around a raw pointer to `V` on the heap.
 ///
-/// Technically, it is a wrapper around `*mut core::mem::MaybeUninit<V>`, that is, the object
+/// Technically, it is a wrapper around `*mut mem::MaybeUninit<V>`, that is, the object
 /// it points to may be uninitialized.
 ///
 /// In order to access (read or write) the value behind the pointer, the user needs
@@ -140,7 +139,7 @@ unsafe impl<T> Send for PPtr<T> {}
 
 #[verifier(external_body)]
 pub tracked struct PermissionOpt<#[verifier(strictly_positive)] V> {
-    phantom: std::marker::PhantomData<V>,
+    phantom: marker::PhantomData<V>,
 }
 
 /// Represents the meaning of a [`PermissionOpt`] object.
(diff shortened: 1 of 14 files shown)


ziqiaozhou:
@tjhance  @Chris-Hawblitzel 
I do not have permission to add you as reviewers. But I guess you two should review the change.

source/pervasive/ptr.rs
```diff
@@ -362,7 +365,8 @@ impl<V> PPtr<V> {
         opens_invariants_none();
 
         unsafe {
-            dealloc(self.uptr as *mut u8, Layout::for_value(&*self.uptr));
+            #[cfg(not(feature = "no_global_allocator"))]
```
tjhance, line 368: This function won't make any sense if this line is removed. Can you just disable the entire `ptr` module for `no_global_allocator`?

source/state_machines_macros/src/to_token_stream.rs
```diff
@@ -276,14 +276,14 @@ pub fn output_primary_stuff(
             if trans.kind == TransitionKind::Init {
                 let args = post_params(&trans.params);
                 rel_fn = quote! {
-                    pub open spec fn #name (#args) -> ::std::primitive::bool {
+                    pub open spec fn #name (#args) -> bool {
```
tjhance, line 279: This needs to be a fully qualified path because the user might declare a different `bool` type. I think you can use `core::primitive::bool`.

source/pervasive/seq.rs
```diff
@@ -19,15 +21,15 @@ verus! {
 ///  * [`Seq::new`] construct a sequence of a given length, initialized according
 ///     to a given function mapping indices `i` to values `A`.
 ///  * The [`seq!`] macro, to construct small sequences of a fixed size (analagous to the
-///     [`std::vec!`] macro).
+///     [`alloc::vec!`] macro).
```
tjhance, line 24: Can you keep this as `std` here?

tjhance (changes requested):

source/state_machines_macros/src/to_token_stream.rs
ziqiaozhou: I change to core's bool. But I do not encounter problems with a non-qualified path. Would that cause bad side effect for verification result?

source/pervasive/seq.rs
ziqiaozhou: Sure.

source/pervasive/ptr.rs
ziqiaozhou: That makes sense.

source/state_machines_macros/src/to_token_stream.rs
tjhance: Generally speaking macros need to use fully qualified paths in order to be ambiguous. If it's just `bool` then it might conflict with a user-defined type.

I don't think there's a unit test for it though, so I'm not surprised you didn't run into any issues with it.

tjhance (approved):

Chris-Hawblitzel (approved):

PR #295, opened 2022-09-19
Turn on no_enhanced_typecheck by default

Chris-Hawblitzel:
Next step towards using unmodified rustc: by default, disable Verus type checking extensions to rustc, unless --deprecated-enhanced-typecheck is set.


source/rust_verify/example/invariants.rs
```diff
@@ -1,3 +1,5 @@
+// rust_verify/tests/example.rs deprecated-enhanced-typecheck
```
utaal, line 1: This implies `expect-success` right?

Chris-Hawblitzel: Yes, because `expect-success` is the default.

source/rust_verify/tests/examples.rs
```diff
@@ -56,12 +56,19 @@ fn run_example_for_file(file_path: &str) {
...
+    }
 
     if let ["//", "rust_verify/tests/example.rs", command] = &first_line_elements[..] {
         match command.strip_suffix("\n").unwrap_or("unexpected") {
             "expect-success" => mode = Mode::ExpectSuccess,
             "expect-errors" => mode = Mode::ExpectErrors,
             "expect-failures" => mode = Mode::ExpectFailures,
+            "deprecated-enhanced-typecheck" => {
+                deprecated_enhanced_typecheck = "--deprecated-enhanced-typecheck";
+            }
```
tjhance, line 71: could you clean this up? the other 3 cases each set the `mode` variable, while this one leaves it as the default value, set several lines above.

tjhance (approved):

source/rust_verify/tests/examples.rs
utaal: can you also please add a TODO/REVIEW to remove this option once we've removed `--deprecated-enhanced-typecheck`?

utaal (approved):

PR #300, opened 2022-09-26
Implement PartialEq, Eq and Hash for String and StrSlice

marshtompsxd:
It would be great if we can implement Eq and Hash for String and StrSlice since we often need to use them as keys in hash map in exec code.

### DIFF
--- a/source/pervasive/string.rs
+++ b/source/pervasive/string.rs
@@ -12,15 +12,31 @@ use builtin_macros::verus;
 
 verus! {
 
+#[derive(PartialEq, Eq)]
 #[verifier(external_body)]
 pub struct String {
     inner: string::String,
 }
 
+impl std::hash::Hash for String {
+    #[verifier(external)]
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.inner.hash(state)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


marshtompsxd:
I tried to write some postconditions for Clone but it says
```
error: trait method implementation cannot declare requires/ensures; these can only be inherited from the trait declaration
```
I am not sure how to express the condition that "the return value of Clone() === self"

utaal:
I've been looking into this, and due to the lack of support for the `core::borrow::Borrow` trait and/or a `View` trait, I think using `String`s in a `HashMap` would be difficult right now: there would be no way for the user to talk about the keys, as they wouldn't be able to make a spec `String` from a literal, for example. Ideally, we'd represent a `Dict<String, V>` with a view `Map<StrSlice<'_>, V>`, but we currently don't have sufficient trait support for this.

This use case may currently be better served by a dedicated `StringDict` that specializes for `String` keys.

lalithsuresh:
Our keys are a struct type that has Strings in them. So a dedicated `StringDict` might not work for us right now.

utaal:
Triaged in the Verus meeting. We will reconsider this in relation to the treatment of `Clone`. I'm closing for now, and will open a discussion.

PR #308, opened 2022-09-29
Implement exec mode map: dict, based on #174

utaal:
This is an attempt to support #174 with minimal extensions to the features we support.

This PR is based on @marshtompsxd's work in #174.

### DIFF
--- a/source/rust_verify/example/dict.rs
+++ b/source/rust_verify/example/dict.rs
@@ -1,0 +1,24 @@
+#[allow(unused_imports)]
+use builtin::*;
+#[allow(unused_imports)]
+use builtin_macros::*;
+mod pervasive;
+#[allow(unused_imports)]
+use crate::pervasive::{*, dict::*, map::*, string::*, option::*};
+
+verus! {
+#[derive(Eq, PartialEq, Structural, Hash)]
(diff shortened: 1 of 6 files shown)


source/rust_verify/example/dict.rs
```diff
@@ -0,0 +1,24 @@
...
+use builtin::*;
+#[allow(unused_imports)]
+use builtin_macros::*;
+mod pervasive;
+#[allow(unused_imports)]
+use crate::pervasive::{*, dict::*, map::*, string::*, option::*};
+
+verus! {
+#[derive(Eq, PartialEq, Structural, Hash)]
+pub struct ObjectKey { a: u64 }
```
Chris-Hawblitzel, line 11: It appears possible to subvert the intended `Structural` behavior:

```
#[derive(Eq, Structural)]
pub struct ObjectKey { a: u64 }

impl std::hash::Hash for ObjectKey {
    fn hash<H>(&self, h: &mut H) where H: std::hash::Hasher {
        // potentially do something nondeterministic here
    }
}

impl std::marker::StructuralPartialEq for ObjectKey {
}

impl PartialEq for ObjectKey {
    fn eq(&self, _: &Self) -> bool { true }
}
```

Issue #216, opened 2022-07-29
Panic when using forall in `ensures` of function containing `return`


matthias-brun:
This might be related to #215. The following code causes a panic. If `return true;` is replaced with `true` or if we use an ensures without a quantifier, it verifies without issues.

```rust
#[allow(unused_imports)]
use builtin::*;
mod pervasive;

#[spec]
fn f(a: nat) -> bool {
    true
}

fn g() -> bool {
    ensures(|res: bool| forall(|i: nat| f(i)));
    return true;
}

fn main() {
}
```
results in
```
fhs m: source 1 ν ./tools/rust-verify.sh /st/verus/test.rs
note: verifying root module

thread 'rustc' panicked at 'assertion failed: trigs.len() == 0', vir/src/ast_to_sst.rs:227:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
verification results:: verified: 0 errors: 0
```

tjhance:
Currently, expression-return and return statements are handled in different places, the first in func_to_air and the second in ast_to_sst. The result is there are a few differences in the way they're handled.

At one point I attempted to consolidate them into place to resolve the inconsistencies,  but I never followed through as it turned out to be kind of tricky and there were a few things i didn't know how to handle. It's probably time to revisit it, though.

tjhance:
fixed by #312

PR #312, opened 2022-10-07
fix a couple of return-related bugs

tjhance:
Fixes a couple of issues on the backlog:

 * 'return' statements cause panics in trait functions
 * Issue #216

Post-conditions are properly handled in `func_to_air` and `sst_to_air`. The post-condition handling of 'return' statements (in `ast_to_sst`) missed a couple of key details.

To make sure return statements get the proper handling, I added an `AssertPostConditions` statement to the SST. Now `ast_to_sst` just has to emit these statements, which are then handled in `sst_to_air` through mostly the same logic as before. Now return-statements and return-expressions will be handled in the same way.

### DIFF
--- a/source/rust_verify/tests/recursion.rs
+++ b/source/rust_verify/tests/recursion.rs
@@ -974,11 +974,12 @@ test_verify_one_file! {
             decreases_when(i >= 0);
             decreases_by(check_arith_sum);
 
-            if i == 0 { 0 } else { i + arith_sum(i - 1) } // FAILS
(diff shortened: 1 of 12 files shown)


Chris-Hawblitzel (approved):
I wonder if `split_expression.rs` should be revised based on `AssertPostConditions`.

utaal:
@tjhance is this still relevant? Should we merge it?
Thanks!

tjhance:
I need to address Chris's comment, which I will do after this upcoming deadline.

Issue #315, opened 2022-10-23
Cannot use datatype height for termination proofs when using non-inductive generics


utaal:
```rust
#[spec] fn max(a: nat, b: nat) -> nat {
    if a >= b { a } else { b }
}

struct Node<V> {
    left: Box<Option<Node<V>>>,
    right: Box<Option<Node<V>>>,
    value: V,
}

impl<V: SpecOrd> Node<V> {
    #[spec] pub fn height(&self) -> nat {
        decreases(self);
        max(
            if let Some(l) = *self.left { l.height() } else { 0 },
            if let Some(r) = *self.right { r.height() } else { 0 },
        )
    }
}
```

This results in "cannot prove termination" for `height`. Manually adding this axiom makes the proof go through:
```rust
(axiom (forall ((x pervasive.option.Option.)) (!
   (= (height.? (pervasive.option.Option./Some/_0 x)) (height.? (Poly%pervasive.option.Option. x)))
   :pattern ((height.? (pervasive.option.Option./Some/_0 x)))
   :qid
   prelude_datatype_height
   :skolemid
   skolem_prelude_datatype_height
)))
```

PR #317, opened 2022-10-25
generating height axiom for fields of generic type

channy412:
This PR is to fix this issue: https://github.com/verus-lang/verus/issues/315

`pervasive::option::Option` also triggers `datatype_height_axiom`, but height axioms are only generated when a field is a concrete type. I just made a small change to generate height axioms for fields of generic type

For the example of the issue, the below axiom is generated.
```
(assert
 (forall ((x pervasive.option.Option.)) (!
   (< (height.? (pervasive.option.Option./Some/_0 x)) (height.? (Poly%pervasive.option.Option.
      x
   )))
   :pattern ((height.? (pervasive.option.Option./Some/_0 x)))
   :qid
   prelude_datatype_height
   :skolemid
   skolem_prelude_datatype_height
)))
```


source/vir/src/prelude.rs
```diff
@@ -523,3 +523,21 @@ pub(crate) fn datatype_height_axiom(typ_name1: &Path, typ_name2: &Path, field: &
         )))
     )
 }
+
+pub(crate) fn datatype_height_axiom_generic(typ_name1: &Path, field: &Ident) -> Node {
```
Chris-Hawblitzel, line 527: Rather than duplicating `datatype_height_axiom`, let's make `typ_name2` an `Option`.  Then we can splice in the node for either `([box_t2] ([field] x))` or `([field] x)` based on the `Option`.

Chris-Hawblitzel (approved):

source/rust_verify/tests/recursion.rs
```diff
@@ -1215,3 +1215,29 @@ test_verify_one_file! {
         }
     } => Err(e) => assert_one_fails(e)
 }
+
+test_verify_one_file! {
+    #[test] decreases_on_non_recursive_generic_datatype verus_code! {
```
utaal, line 1220: can you please add `_regression_315` to the test name so we can track that this is a regression test?
Thanks!

utaal (approved):
Lgtm, thank you! Please squash when merging.

PR #318, opened 2022-10-31
Add axiom of function extensionality to pervasives

tchajed:
This adds the following axiom (slightly simplified; the actual implementation has a workaround for an issue with triggers):

```
  #[verifier(external_body)]
  pub proof fn fun_ext<A, B>(f1: FnSpec(A) -> B, f2: FnSpec(A) -> B)
    requires forall |x| f1(x) === f2(x)
    ensures f1 === f2
  {}
```

### DIFF
--- a/source/rust_verify/tests/functions.rs
+++ b/source/rust_verify/tests/functions.rs
@@ -1,0 +1,30 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] test_use_fun_ext verus_code! {
+        use crate::pervasive::function::*;
+
+        proof fn test_use_fun_ext(f: FnSpec(int) -> int) {
+          fun_ext::<int, int>(|i: int| i + 1, |i: int| 1 + i);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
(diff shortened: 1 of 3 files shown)


source/rust_verify/tests/functions.rs
```diff
@@ -0,0 +1,30 @@
...
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] test_use_fun_ext verus_code! {
+        use crate::pervasive::function::*;
+
+        proof fn test_use_fun_ext(f: FnSpec(int) -> int) {
+          fun_ext::<int, int>(|i: int| i + 1, |i: int| 1 + i);
```
jonhnet, line 11: All this tests is the requires line, right? Is there any useful thing to test about the ensures?

jonhnet (approved):
LGTM.

Chris-Hawblitzel (approved):

source/rust_verify/tests/functions.rs
tchajed: I could add `assert ((|i: int| i + 1) === (|i: int| 1 + i));` to this test to use the postcondition, although the postcondition is also tested by the other test in this PR currently.

tchajed:
Thanks for the reviews!

PR #320, opened 2022-11-03
Generalize Error to Message and use an AIR-level diagnostic trait consistently for reporting

parno:
As discussed at a Verus meeting several weeks ago, this PR converts the AIR `Error` type into a more generic `Message` type that also carries a reporting level with it.  It expands the helper functions for constructing such Messages, and it introduces (in AIR) a `Diagnostics` trait for reporting such messages.  One source of awkwardness is that in rust_verify, we cannot have rustc's `compiler` implement `Diagnostics`, since `compiler` and `Diagnostics` come from separate crates.  Also, the portion of the compiler we need for printing nice Rust-like errors has a compiler-related lifetime associated with it, which makes it a bit awkward to pass down into VIR and AIR.  It also can't easily be passed to the separate thread that the interpreter implements (for the sake of a larger stack), so the interpreter instead buffers messages and reports them after each run.

Renaming Error to Message touched many files superficially.  The semantically interesting changes are primarily in:
1. `source/air/src/messages.rs`
2. `source/air/src/profiler.rs`
3. `source/rust_verify/src/verifier.rs`

plus a few files that now use `Diagnostics` instead of `println`.

I've confirmed that the new output matches the old output for the following test cases, but if there are comparisons to make, please let me know.
 1. Quantifier profiler
 2. Normal assertion failures
 3. Chanhee's tracing
 5. Long running SMT query


source/vir/src/split_expression.rs
```diff
@@ -498,11 +498,12 @@ pub fn is_split_error(error: &Error) -> bool {
 fn split_call(
     ctx: &Ctx,
     state: &State,
+    diagnostics: &impl Diagnostics,
```
tjhance, line 501: is this used for anything in this file? I see it's being passed around mutually recursively, but I can't find a place where it is actually used here.

tjhance (approved):

source/vir/src/split_expression.rs
parno: It gets passed to `crate::ast_to_sst::expr_to_exp_as_spec` and `crate::ast_to_sst::expr_to_exp`

utaal (approved):
Sorry for being slow at picking this up.

Looks good to me! It really cleans up the reporting in `verifier.rs` too. Thanks!

parno:
Great, thanks very much for checking it out!

PR #325, opened 2022-11-17
Macro VIR printer (with compact mode)

utaal:
This removes the hand-written VIR printer and replaces it with a (mainly) macro-generated one.

At the same time, it supports `compact` printing mode from PR #203, but doesn't yet include `pretty_format`, which we can add back in later.

I prefer this solution for better maintainability, so changes to VIR don't always require a manual change to the printer.

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -12,6 +12,7 @@ use air::errors::Error;
 use num_bigint::BigInt;
 use std::fmt::Display;
 use std::sync::Arc;
+use vir_macros::ToNode;
 
 /// Result<T, VirErr> is used when an error might need to be reported to the user
 pub type VirErr = Error;
@@ -27,14 +28,14 @@ pub type Idents = Arc<Vec<Ident>>;
 
 /// A fully-qualified name, such as a module name, function name, or datatype name
 pub type Path = Arc<PathX>;
-#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ToNode)]
 pub struct PathX {
     pub krate: Option<Ident>, // None for local crate
     pub segments: Idents,
 }
 
 /// Describes what access other modules have to a function, datatype, etc.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, ToNode)]
 pub struct Visibility {
     /// Module that owns this item, or None for a foreign module
     pub owning_module: Option<Path>,
@@ -43,7 +44,7 @@ pub struct Visibility {
 }
 
 /// Describes whether a variable, function, etc. is compiled or just used for verification
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, Debug, ToNode, PartialEq, Eq, Hash)]
 pub enum Mode {
     /// Ghost (not compiled), used to represent specifications (requires, ensures, invariant)
     Spec,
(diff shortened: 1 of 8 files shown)


source/vir/src/printer.rs
```diff
@@ -1,24 +1,13 @@
 use crate::ast::*;
 use air::ast::Span;
```
channy412, line 2:
```suggestion

```
I think it gives warning (unused import)

source/vir/src/printer.rs
```diff
@@ -1,24 +1,13 @@
...
 use air::printer::macro_push_node;
-use air::printer::{clean_up_lines, str_to_node};
-use air::{node, nodes, nodes_vec};
+use air::printer::str_to_node;
```
channy412, line 4:
```suggestion

```
I think this also gives the same warning

source/vir/src/ast.rs
```diff
@@ -499,7 +501,7 @@ pub enum StmtX {
 /// Function parameter
 pub type Param = Arc<Spanned<ParamX>>;
 pub type Params = Arc<Vec<Param>>;
-#[derive(Debug, Clone)]
+#[derive(Debug, ToNode, Clone)]
```
channy412, line 504: Do we also want to apply compact printing to function arguments and the return type?

source/vir/src/ast.rs
```diff
@@ -406,6 +407,7 @@ pub enum ComputeMode {
 pub type Expr = Arc<SpannedTyped<ExprX>>;
 pub type Exprs = Arc<Vec<Expr>>;
 #[derive(Debug)]
+#[to_node_impl(name = ">")]
```
channy412, line 410: Just to confrim, is ">" replacing "ExprX" in the VIR log?

channy412 (approved):
Thank you for working on the VIR printer!   This looks great to me, and looks much easier to maintain. I simply added a few comments. Thanks!

source/vir/src/ast.rs
utaal: Can you elaborate? The `ToNodeOpts` are propagated in all recursive calls, but maybe that's not what you mean.

source/vir/src/ast.rs
utaal: Yes!

source/vir/src/ast.rs
channy412: Using `compact` mode, the following code snippet generates the VIR log below. It seems like `ParamX`'s type information is logged, and I wonder if we want to remove these type information of function arguments and the return type.  If we choose to do that, I am guessing that we might add a handwritten `impl ToNode for ParamX`?

```
spec fn mul(a: u64, b: u64)  -> u64 {
    builtin::mul(a, b)
}
```

```
;; rust_verify/example/assorted_demo.rs:24:6: 24:36 (#23375)
(Function
 (Fun :path "mul." :trait_path None) :mode Spec :typ_bounds () :params (
  (Param :name "a~2" :typ (Typ Boxed (Typ Int (IntRange U 64))) :mode Spec :is_mut false)
  (Param :name "b~4" :typ (Typ Boxed (Typ Int (IntRange U 64))) :mode Spec :is_mut false)
 ) :ret (Param :name "%return" :typ (Typ Int (IntRange U 64)) :mode Spec :is_mut false) :require () :ensure () :body (> Block
  () (> Binary (BinaryOp Arith (ArithOp Mul) 24) (> Var "a~2") (> Var "b~4"))))
```

Chris-Hawblitzel (approved):
Sounds like a good idea.  My only comment is that the name `ToNode` isn't very self-explanatory when used in the `derive`s in `ast.rs`.  Can we name it something that makes clear that it's for debugging or printing, like `DebugNode` or `DebugPrint` or something?

source/vir/src/ast.rs
utaal: Thanks! I'd argue that type signatures for functions are still fairly critical info, so I'd print it in compact mode too (where I'd interpret `no_type` as "no types of expressions").

PR #327, opened 2022-11-17
Generalize trait implementations to arbitrary types

tchajed:
Formerly, `impl T for U` only worked for `U` being a datatype or `StrSlice`. This PR generalizes this support to work for arbitrary types `U`, which in particular enables implementing traits for `FnSpec` types. This enables adding and overloading methods on spec functions, which can result in much clearer code in cases where functions are the natural objects in a specification.

There are a few TODOs in the PR around how to handle the generality which should be checked by someone more familiar with the code to identify what needs to be done in those cases, and ideally add tests for them as well.


source/vir/src/prune.rs
```diff
@@ -266,7 +266,10 @@ pub fn prune_krate_for_module(krate: &Krate, module: &Path) -> (Krate, Vec<MonoT
     for f in &functions {
         function_map.insert(f.x.name.clone(), f.clone());
         if let FunctionKind::TraitMethodImpl { method, datatype, .. } = &f.x.kind {
-            method_map.insert((datatype.clone(), method.clone()), f.x.name.clone());
+            // TODO(tchajed): construct a path for a method impl
```
Chris-Hawblitzel, line 269: I don't think you'll need this `method_map` to have non-`Datatype` types.  If it's not a `Datatype`, then `reached_methods` can just assume that the type is reached, so all the type's methods should be reached.  In other words, for `Datatype`s `D` we have:
```
    // If:
    // - we reach both D and T.f
    // - and D implements T.f with D.f
    // add D.f
```
and for non-`Datatype`s `D`, we have:
```
    // If:
    // - we reach T.f
    // - and D implements T.f with D.f
    // add D.f
```

source/vir/src/recursion.rs
```diff
@@ -522,6 +524,8 @@ pub(crate) fn expand_call_graph(
                                 // f1 --> D.f2
                                 Some(&method_map[&(x.clone(), datatype.clone())])
                             }
```
Chris-Hawblitzel, line 526: When `method_map` is changed to account for all types, not just `Datatype`s, then the `TypX::Datatype` case can be generalized so that it includes `Lambda`.

source/vir/src/recursion.rs
```diff
@@ -459,9 +459,11 @@ pub(crate) fn expand_call_graph(
     // Add D: T --> f and D: T --> T where f is one of D's methods that implements T
     if let FunctionKind::TraitMethodImpl { trait_path, datatype, .. } = function.x.kind.clone() {
         let t_node = Node::Trait(trait_path.clone());
-        let dt_node = Node::DatatypeTraitBound { datatype, trait_path };
-        call_graph.add_edge(t_node, dt_node.clone());
-        call_graph.add_edge(dt_node, f_node.clone());
```
Chris-Hawblitzel, line 464: I think `DatatypeTraitBound` should be generalized to allow non-`Datatype` types.  So `TypX::Datatype` would not be a special case here.

utaal:
I did some more research on this (with help from @Chris-Hawblitzel). The most challenging aspect is the `method_map` in `GlobalCtxt` (and `recursion.rs`) where we're looking for a trait implementation based on the datatype name (on the assumption that there's only one such implementation per trait and per type, which isn't true in general in Rust).

Using `Typ` directly as the target type doesn't work, because we may be dispatching based on a type with non-concrete type arguments, and the implementations may be for types with non-concrete arguments (but potentially different ones). 

Because we only allow one trait implementation per type, we could make up a unique name for each "top level" type, but I'm not sure how to do that in general (e.g. for `FnSpec`s); it's nontrivial, so it may be best to wait on this until we update trait support to understand multiple trait impls per type.

PR #328, opened 2022-11-22
implement exec closures

tjhance:
High level points:

 * Fn/FnMut/FnOnce can now be called in exec mode.
 * Closures that capture mutable references are not supported yet.
 * Every Fn/FnMut/FnOnce object now implements a trait `FnWithSpecification` which gives it `requires` and `ensures` spec functions.
 * Fn/FnMut/FnOnce can no longer be called in spec mode; only FnSpec can be called in spec mode. **This breaks some existing code.**

More details:

 * This makes the following changes to VIR AST:
   * Adds `Typ::AnonymousClosure`. The old `Lambda` type now only applies to `FnSpec` (and should probably be renamed). Similar to tuple types, these are translated into datatypes by ast_simplify.
   * Adds `CallTarget::FnExec` for any non-static function calls that are exec. The determination of whether something is FnSpec or FnExec is type-based (depending on whether the type is FnSpec).
   * Adds `CallTarget::BuiltinSpecFun`. This is for the builtin `requires` and `ensures` functions. Currently `requires` has type (Params) -> bool, while `ensures` has type (Params, Ret) -> bool.
   * Adds `ExprX::ExecClosure`.
   * **Removes** the old `GenericBoundX::FnSpec`. This was previously used to turn the Fn/FnMut/FnOnce traits into Lambda types.
 
Lowering to SST and then AIR has several components:

 * The builtin requires/ensures functions get translated to special AIR functions. These have a hardcoded name, similar to `height`.
 * When calling into a closure, we have to generate preconditions and postconditions. Instead of calling statically determined requires and ensures, we call the requires and ensures builtins using the callee as an argument.
 * Everything inside the closure gets translated into an SST `ClosureInner`, which behaves like a `DeadEnd`.
 * When creating a closure, we have to create an object that has the appropriate specifications via the `requires` and `ensures` builtins.
   * ~~This part turned out to be somewhat annoying because (i) it creates a bunch of tuples so we have to make sure those types exist and (ii) we have to do some manual poly/native coercions. I'm thinking of moving this part into ast_simplify.~~ This is now done in ast_simplify.
 
TODO:

- [ ] add special syntax to the macro
- [ ] add a check that the user doesn't reference any non-captured mutable variable in ghost code. (This would not lead to unsoundness, but in one of our discussions, we decided it might be confusing if it were allowed.) Right now I'm not really sure what the easiest way to do this check is.
- [x] potentially move some of the lowering code into ast_simplify

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -969,3 +969,21 @@ impl<Args, Output> Fn<Args> for FnSpec<Args, Output> {
 pub fn closure_to_fn_spec<Args, F: FnOnce<Args>>(_f: F) -> FnSpec<Args, F::Output> {
     unimplemented!();
 }
+
+pub trait FnWithSpecification<Args> {
+    type Output;
+    fn requires(&self, args: Args) -> bool;
+    fn ensures(&self, args: Args, output: Self::Output) -> bool;
+}
+
+impl<Args, F: FnOnce<Args>> FnWithSpecification<Args> for F {
+    type Output = F::Output;
+
+    fn requires(&self, _args: Args) -> bool {
+        unimplemented!();
+    }
+
+    fn ensures(&self, _args: Args, _output: Self::Output) -> bool {
+        unimplemented!();
+    }
+}
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -1116,10 +1116,14 @@ impl VisitMut for Visitor {
                     *expr = quote_verbatim!(span, attrs => {::builtin::assert_forall_by(|#inputs| #block);});
                 }
                 Expr::Closure(clos) => {
-                    let span = clos.span();
-                    *expr = Expr::Verbatim(quote_spanned!(span =>
-                        ::builtin::closure_to_fn_spec(#clos)
-                    ));
+                    if is_inside_ghost {
+                        let span = clos.span();
+                        *expr = Expr::Verbatim(quote_spanned!(span =>
+                            ::builtin::closure_to_fn_spec(#clos)
+                        ));
+                    } else {
+                        *expr = Expr::Closure(clos);
+                    }
                 }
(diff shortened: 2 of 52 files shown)


tjhance:
Since the main thing left to do is add an additional check, which shouldn't interfere with any of the existing code, I'm going to go ahead and mark this as ready-for-review.

I'm still not sure what's the best way to implement that check, so I'm looking for pointers here. What I want to check for is any variables which is (i) marked `mut` (ii) NOT captured in the closure by exec/proof code but (iii) IS captured by spec code. But I'm not actually sure how to get the variable modes since the mode-checker currently has this delayed choice stuff. So I'm not sure what's the easiest way to get the mode of a variable usage?

Chris-Hawblitzel:
If you want to just get variable modes, you can use `ErasureModes::var_modes`.  You can ignore `inferred_modes`; that's for `BinaryOp::Arith`.  Is this check intended to generate warnings or errors?

Chris-Hawblitzel:
I think this makes sense.  My main comment (discussed a little offline) is that, in the implementation, we could look for ways to consolidate  exec closure call code with trait call code (e.g. `StmX::DynCall` with `StmX::Call`).

tjhance:
I'll look into the possibility of consolidating some of the new code with traits code.

tjhance:
I removed the SST `DynCall` and the AST `FnExec`. I replaced them both with a fake static call to this, now defined in pervasive:

```
#[verifier(custom_req_err("Call to non-static function fails to satisfy `callee.requires(args)`"))]
#[doc(hidden)]
#[verifier(external_body)]
fn exec_nonstatic_call<Args, Output, F>(f: F, args: Args) -> Output
    where F: FnOnce<Args, Output=Output>
{
    requires(f.requires(args));
    ensures(|output: Output| f.ensures(args, output));
    unimplemented!();
}
```

Chris-Hawblitzel (approved):

PR #329, opened 2022-11-27
Support break and continue in loops

Chris-Hawblitzel:
Below is the text from the guide describing the support for `break`.  (The support for `continue` is relatively trivial).  The design for `break` continues with the philosophy that loops should be checkable in a separate SMT query for better modularity in verification, so that the loop must fully specify its invariants and, in the presence of `break`, its ensures clauses independently of the surrounding context.

-----

Loops can exit early using `return` or `break`.  Suppose, for example, we want to remove the requirement `triangle(n as nat) < 0x1_0000_0000` from the `loop_triangle` function, and instead check for overflow at run-time.  The following version of the function uses `return` to return the special value `0xffff_ffff` in case overflow is detected at run-time:

```rust
fn loop_triangle_return(n: u32) -> (sum: u32)
    ensures
        sum == triangle(n as nat) || (sum == 0xffff_ffff && triangle(n as nat) >= 0x1_0000_0000),
{
    let mut sum: u32 = 0;
    let mut idx: u32 = 0;
    while idx < n
        invariant
            idx <= n,
            sum == triangle(idx as nat),
    {
        idx = idx + 1;
        if sum as u64 + idx as u64 >= 0x1_0000_0000 {
            proof {
                triangle_is_monotonic(idx as nat, n as nat);
            }
            return 0xffff_ffff;
        }
        sum = sum + idx;
    }
    sum
}
```

Another way to exit early from a loop is with a `break` inside the loop body.  However, `break` changes the meaning of loop invariants slightly.  For simple `while` loops without a `break`, each invariant must be true before and after each loop iteration, and Verus also knows that the loop condition (e.g. `idx < n`) must be false after exiting the loop.  If there is a `break`, though, the loop condition is not necessarily false after the loop, because the `break` might cause the loop to exit even when the loop condition is true.  To deal with this, `while` loops with a `break`, as well as Rust `loop` expressions (loops with no condition), work slightly differently, as shown in the following code:

```rust
fn loop_triangle_break(n: u32) -> (sum: u32)
    ensures
        sum == triangle(n as nat) || (sum == 0xffff_ffff && triangle(n as nat) >= 0x1_0000_0000),
{
    let mut sum: u32 = 0;
    let mut idx: u32 = 0;
    while idx < n
        invariant
            idx <= n,
            sum == triangle(idx as nat),
        ensures
            sum == triangle(n as nat) || (sum == 0xffff_ffff && triangle(n as nat) >= 0x1_0000_0000),
    {
        idx = idx + 1;
        if sum as u64 + idx as u64 >= 0x1_0000_0000 {
            proof {
                triangle_is_monotonic(idx as nat, n as nat);
            }
            sum = 0xffff_ffff;
            break;
        }
        sum = sum + idx;
    }
    sum
}
```

For `while` with `break` (or for `loop`), each `invariant` must be true before each loop iteration, but need not be true on exit.  Instead, `ensures` clauses are used to indicate what must be true on exit.  (For convenience, you can also supply `invariant_ensures` clauses, in between the `invariant` clauses and `ensures` clauses, that act as both `invariant` and `ensures`.)


tjhance:
It seems a little awkward that the meaning of 'invariant' changes when a 'break' statement is added.

Could we rename 'invariant_ensures' to 'invariant' and 'invariant (in the case of a loop with a break)' to something else?

Chris-Hawblitzel:
> Could we rename 'invariant_ensures' to 'invariant' and 'invariant (in the case of a loop with a break)' to something else?

I originally implemented syntax for `requires`/`invariant`/`ensures` rather than `invariant`/`invariant_ensures`/`ensures`:

https://github.com/verus-lang/verus/blob/93c0dfa8c2f7148847c4effdd461d7217dace4b6/dependencies/syn/src/expr.rs#L542

```
    pub struct ExprLoop #full {
        pub attrs: Vec<Attribute>,
        pub label: Option<Label>,
        pub loop_token: Token![loop],
        pub requires: Option<Requires>,
        pub invariant: Option<Invariant>,
        pub ensures: Option<Ensures>,
        pub decreases: Option<Decreases>,
        pub body: Block,
    }
```

However, `requires` was confusing in practice, because it looked like it was only required upon the first entry to the loop, when in fact, it was required upon every entry to the loop.  In other words, it was an invariant, so I changed it to:

```
    pub struct ExprLoop #full {
        pub attrs: Vec<Attribute>,
        pub label: Option<Label>,
        pub loop_token: Token![loop],
        pub invariant: Option<Invariant>,
        pub invariant_ensures: Option<InvariantEnsures>,
        pub ensures: Option<Ensures>,
        pub decreases: Option<Decreases>,
        pub body: Block,
    }
```

`invariant`/`invariant_ensures`/`ensures` has some nice properties.  The word `invariant` refers to the loop head (an invariant, true n times for n iterations), the word `ensures` refers to loop exit (true once at exit, so not an invariant), and `invariant_ensures` refers to both.  `invariant_ensures` is a more cumbersome name, but it is also optional.  People can just use `invariant` and `ensures` if they want, as in the `loop_triangle_break` example above.  If we used some name other than `invariant` for the loop head invariant, then people would be forced to use that alternate name.

Chris-Hawblitzel:
Actually, I don't think I should have written "However, `break` changes the meaning of loop invariants slightly" in the documentation in the first place.  I think it makes more sense to say that `invariant` *always* means the thing that is true at the beginning of each loop iteration, and that this meaning does not change when we introduce `break`.  What changes is that without break, Verus can automatically compute the exit condition by combining the invariant with the loop condition (since the loop condition is the only thing that causes the loop to exit, and it happens to sit right at the beginning of the loop), and with break, the programmer has to specify the exit condition using `ensures`.  I've updated the documentation to say this more clearly:

Another way to exit early from a loop is with a `break` inside the loop body.  However, `break` complicates the specification of a loop slightly.  For simple `while` loops without a `break`, Verus knows that the loop condition (e.g. `idx < n`) must be false after exiting the loop.  If there is a `break`, though, the loop condition is not necessarily false after the loop, because the `break` might cause the loop to exit even when the loop condition is true.  To deal with this, `while` loops with a `break`, as well as Rust `loop` expressions (loops with no condition), must explicitly specify what is true after the loop exit using `ensures` clauses, as shown in the following code:

```rust
fn loop_triangle_break(n: u32) -> (sum: u32)
    ensures
        sum == triangle(n as nat) || (sum == 0xffff_ffff && triangle(n as nat) >= 0x1_0000_0000),
{
    let mut sum: u32 = 0;
    let mut idx: u32 = 0;
    while idx < n
        invariant
            idx <= n,
            sum == triangle(idx as nat),
        ensures
            sum == triangle(n as nat) || (sum == 0xffff_ffff && triangle(n as nat) >= 0x1_0000_0000),
    {
        idx = idx + 1;
        if sum as u64 + idx as u64 >= 0x1_0000_0000 {
            proof {
                triangle_is_monotonic(idx as nat, n as nat);
            }
            sum = 0xffff_ffff;
            break;
        }
        sum = sum + idx;
    }
    sum
}
```

(For convenience, you can also supply `invariant_ensures` clauses, in between the `invariant` clauses and `ensures` clauses, that act as both `invariant` and `ensures`.)

tjhance:
That does make a little more sense, though there's still something I'm wondering.

If you have a loop with a 'break' in it, and use `invariant` statements, what VC is generated at the end of the loop body? Is it

`assert !cond ==> invariant` (which would be okay because the invariant doesn't need to be true on exit)

or

`assert invariant` (the same as it is for a loop without a break)?

Chris-Hawblitzel:
That's interesting.  I think you could have a VC like `assert invariant || !cond` at the end of the loop body, at least if `cond` is a pure expression.  It gets more complicated in general, because Verus supports loop conditions with side effects.  For example, `cond` might be `x < y - 1`, and `y - 1` might have the side effect of underflowing.  In this case, Verus relies on the invariant (e.g. `y > 0`) to prove that it's safe to execute `cond`, which means the VC might have to be the whole invariant, without being able to use `cond` as an assumption.

tjhance (approved):

PR #334, opened 2022-12-06
support slice types `[T]`

tjhance:
There's not very much to this one.

Slices [T] turn into a datatype `slice% T` in VIR, which doesn't need any special handling, other than that we have to create it.

To add a `view()` function to `[T]`, we need to do it via a trait. To support traits for `[T]`, I had to fix a few things in rust_verify which were expecting datatypes in the Rust AST, but the changes aren't very substantial.

I didn't add any support for builtin slice operations, just external_body wrappers in pervasive.

### DIFF
--- a/source/pervasive/slice.rs
+++ b/source/pervasive/slice.rs
@@ -21,7 +21,7 @@ impl<T> SliceAdditionalSpecFns<T> for [T] {
 
 #[verifier(external_body)]
 pub exec fn slice_index_get<T>(slice: &[T], i: usize) -> (out: &T)
-    requires 0 <= (i as int) < slice.view().len(),
+    requires 0 <= i < slice.view().len(),
     ensures *out === slice@.index(i as int),
 {
     &slice[i]
--- a/source/pervasive/slice.rs
+++ b/source/pervasive/slice.rs
@@ -1,0 +1,30 @@
+#![allow(unused_imports)]
+use builtin::*;
+use builtin_macros::*;
+use crate::pervasive::seq::*;
+
+verus!{
+
+pub trait SliceAdditionalSpecFns<T> {
+   spec fn view(&self) -> Seq<T>;
+   spec fn spec_index(&self, i: int) -> T;
+}
+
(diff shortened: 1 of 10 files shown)


source/vir/src/builtins.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    // Create a fake variant with a single field of the given type,
+    // though it shouldn't matter, since transparency is Never.
+
+    let typ = Arc::new(TypX::TypParam(crate::def::slice_param()));
+    let field = ident_binder(
+        &Arc::new("DummySliceField".to_string()),
+        &(typ, Mode::Exec, visibility.clone()),
+    );
+    let fields = vec![field];
```
Chris-Hawblitzel, line 22: It's ok to leave `fields` empty.  This is what we do for external_body datatypes: ```if vattrs.external_body {
        (ident_binder(&variant_name, &Arc::new(vec![])), false)
    } else {```

source/rust_verify/tests/slices.rs
```diff
@@ -0,0 +1,29 @@
...
+        fn foo2(x: Vec<u64>)
+            requires x@.len() == 2, x[0] == 19,
+        {
+            foo(x.as_slice());
+        }
+
+        fn foo3(x: &[u64])
+        {
+            let t = *slice_index_get(x, 0); // FAILS
+        }
```
Chris-Hawblitzel, line 27: It would be good to add a test with generics (`&[T]`).

source/pervasive/slice.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    #[verifier(inline)]
+    open spec fn spec_index(&self, i: int) -> T {
+        self.view().index(i)
+    }
+}
+
+#[verifier(external_body)]
+pub exec fn slice_index_get<T>(slice: &[T], i: usize) -> (out: &T)
+    requires 0 <= (i as int) < slice.view().len(),
```
Chris-Hawblitzel, line 24: You don't need the `as int` in an inequality.

Chris-Hawblitzel (approved):

PR #349, opened 2023-01-17
support expressions with no side-effects

tjhance:
useful for macros and fixes a minor annoyance


tjhance:
@Chris-Hawblitzel this seems fine to me but i'm not sure. The implementation for `check_unit_or_never` has this comment:

```rust
            // REVIEW: we could lift this restriction by putting exp into a dummy VIR node:
            // (we can't just drop it; the erasure depends on information from VIR)
```

I don't really know what this means -- ast_to_sst doesn't propagate any information to erasure, does it? -- but it's an old comment, so I think it might just be out-of-date.

Chris-Hawblitzel:
I think `check_unit_or_never` can be removed.  I don't think the comment I wrote in it makes sense, since erasure uses VIR-AST, not VIR-SST.

Chris-Hawblitzel (approved):

PR #353, opened 2023-01-26
support word-size-related intrinsics and constants

tjhance:
* Add `unsigned_max`, `signed_min`, `signed_max`, `arch_word_bits` as builtin spec fn intrinsics that map to `uHi`, `iLo`, `iHi`, and `arch_bits` defined in our AIR prelude. (Technically `signed_max` maps to `iHi - 1` and same for `unsigned_max`.)
* Add support for MAX, MIN, and BITS constants on all the primitive integer types. For `usize` and `isize`, this means translating them into the above intrinsics as appropriate.

### DIFF
--- a/source/rust_verify/src/rust_intrinsics_to_vir.rs
+++ b/source/rust_verify/src/rust_intrinsics_to_vir.rs
@@ -1,0 +1,80 @@
+use crate::util::spanned_typed_new;
+use rustc_span::Span;
+use std::sync::Arc;
+use vir::ast::{Expr, ExprX, IntRange, IntegerTypeBoundKind, Mode, Typ, TypX, UnaryOpr};
+
+pub(crate) fn int_intrinsic_constant_to_vir(span: Span, typ: &Typ, f_name: &str) -> Option<Expr> {
+    let mk_expr = |x: ExprX| spanned_typed_new(span, &typ, x);
+
+    let lit_expr = |u: u128| Some(mk_expr(ExprX::Const(vir::ast_util::const_int_from_u128(u))));
+
+    let lit_expr_i = |i: i128| Some(mk_expr(ExprX::Const(vir::ast_util::const_int_from_i128(i))));
+
+    let arch_bits = || {
+        let arg = spanned_typed_new(
+            span,
+            &Arc::new(TypX::Int(IntRange::Int)),
+            ExprX::Const(vir::ast_util::const_int_from_u128(0)),
+        );
+
+        let kind = IntegerTypeBoundKind::ArchWordBits;
+
+        return mk_expr(ExprX::UnaryOpr(UnaryOpr::IntegerTypeBound(kind, Mode::Exec), arg));
+    };
+
+    let arch_bound = |kind| {
+        return Some(mk_expr(ExprX::UnaryOpr(
+            UnaryOpr::IntegerTypeBound(kind, Mode::Exec),
+            arch_bits(),
+        )));
+    };
+
+    match f_name {
+        // MIN
+        "core::num::<impl u8>::MIN" => lit_expr(0),
+        "core::num::<impl u16>::MIN" => lit_expr(0),
+        "core::num::<impl u32>::MIN" => lit_expr(0),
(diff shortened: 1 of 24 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2371,6 +2408,25 @@ pub(crate) fn expr_to_vir_inner<'tcx>(
...
+                        int_intrinsic_constant_to_vir(expr.span, &expr_typ(), &f_name)
+                    {
+                        let mut erasure_info = bctx.ctxt.erasure_info.borrow_mut();
+                        erasure_info
+                            .resolved_calls
+                            .push((expr.span.data(), ResolvedCall::CompilableOperator));
+                        return Ok(vir_expr);
+                    }
+                }
+                _ => {}
```
utaal, line 2426: Should this be a panic? Or an unsupported feature?

source/rust_verify/tests/arch_word_bits.rs
```diff
@@ -0,0 +1,203 @@
...
+        fn fail1() {
+            assert(arch_word_bits() == 64); // FAILS
+        }
+    } => Err(err) => assert_fails(err, 1)
+}
+
+test_verify_one_file! {
+    #[test] test_set_to_64 verus_code! {
+
+        fn test1() {  // ARCH-WORD-BITS-64
```
utaal, line 150: Minor: can you move the comment one line up to make it clear that it applies to the whole test? And can you add a comment stating it's a load-bearing comment (i.e. it affects the test environment)?

source/rust_verify/src/rust_intrinsics_to_vir.rs
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    let arch_bound = |kind| {
+        return Some(mk_expr(ExprX::UnaryOpr(
+            UnaryOpr::IntegerTypeBound(kind, Mode::Exec),
+            arch_bits(),
+        )));
+    };
+
+    match f_name {
```
utaal, line 32: Is there a way to get the `DefId`s for these, instead of matching the string names?
(I know we do this elsewhere, but I'd rather we didn't depend on how these are printed.)

utaal: So, it turns out they're defined not as intrinsics, but with this macro:

https://github.com/rust-lang/rust/blob/master/library/core/src/num/int_macros.rs#L18

```rust
pub const MIN: Self = !0 ^ ((!0 as $UnsignedT) >> 1) as Self;
```

So: nevermind, I guess? 😬

utaal: I was wondering whether it would make sense to have these in pervasive instead, with a specification that ties them to the intrinsic (we've done something similar for strings). However, I don't think that would work directly, because you'd want the actual constant from `core`.

What about something like:
```rust
#[verifier(external_body)]
pub const fn i64_max() -> (v: i64)
    ensures v == signed_max(64) {
    i64::MAX
}
```
(maybe properly grouped for each type)

My reasoning here is that these should not need a special code path once we support specification for `core`/`std`: their specification can just refer to the intrinsic.
I'm not necessarily asking for this change, just bringing it up as an option.

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: there's a panic on the line below

source/rust_verify/tests/arch_word_bits.rs
tjhance: i don't know why but it doesn't work if it's on its own line.

It's definitely jank though, I was thinking of just making a better way to pass in arguments in the test_verify_one_file! macro instead.

source/rust_verify/src/rust_to_vir_expr.rs
utaal: Sorry, I misread.

Chris-Hawblitzel (approved):

PR #357, opened 2023-01-30
partial support for usize bitvector operations

tjhance:
* support bitwise operators (&, |, !) for usizein general expressions
 * support `assert ... by(bit_vector)` for usize whenever the architecture word-size is known

### DIFF
--- a/source/vir/src/sst_to_air.rs
+++ b/source/vir/src/sst_to_air.rs
@@ -3,6 +3,8 @@ use crate::ast::{
     IntRange, IntegerTypeBoundKind, InvAtomicity, MaskSpec, Mode, Params, Path, PathX,
     SpannedTyped, Typ, TypX, Typs, UnaryOp, UnaryOpr, VarAt, VirErr, Visibility,
 };
+use crate::ast_util::bitwidth_from_int_range;
+use crate::ast_util::IntegerTypeBitwidth;
 use crate::ast_util::{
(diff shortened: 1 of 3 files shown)


source/vir/src/sst_to_air.rs
```diff
@@ -868,24 +884,8 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<
...
-                        Some(w) => w,
-                        None => {
-                            return err_string(
-                                &rhs.span,
-                                format!("error: unable to get bit-width from type {:?}", rhs.typ),
-                            );
-                        }
-                    };
+                    let width_left = bitwidth_from_type(&lhs.typ).expect("bounded integer type");
+                    let width_right = bitwidth_from_type(&rhs.typ).expect("bounded integer type");
```
channy412, line 888: I guess this will not actually panic since we have the check at beginning of `ExpX::Binary` case (line 768). Am I right? 
```
line 768: if !allowed_bitvector_type(&exp.typ) {
```

channy412 (approved):
These changes look great to me! Thanks!

source/vir/src/sst_to_air.rs
tjhance: no, these 2 lines aren't part of that case. However these should pass because we don't define bitwise ops for ints or nats in the first place.

PR #358, opened 2023-02-01
Use #[register_tool] to introduce a verus:: attribute namespace

utaal:
In order to remove our dependency on a rust fork, we need to switch away from custom builtin attributes.

This PR uses the `#[register_tool]` attribute to introduce a `#[verus::*]` attribute namespace, and a (temporary) `#[verifier::*]` attribute namespace for backwards compatibility.

It only breaks compatibility with custom attributes, which fail with:
```
error: attributes spec, proof, exec are not supported anymore; use the verus! macro instead
  --> rust_verify/example/playground.rs:60:1
   |
60 | #[spec] fn test() -> bool {
   | ^^^^^^^
```

It introduces new syntax to replace some of the attributes. Examples can be found in `examples/syntax.rs`.
The only use of `#[verus::*]` for now is for `#[verus::internal(<attr>)]` for attributes generated by the syntax macro(s).

There are concerns not yet addressed by this PR:
* porting remaining `verifier` attributes to either custom syntax or new `verus::` attributes;
* updating documentation to reflect the new recommended attributes (the current documentation remains valid for now because all the `verifier(...)` attributes remain supported for the time being.

The comment `/* vattr */` temporarily marks all locations in tests and examples that have a now-partly-deprecated verifier attribute that we need to update before we drop support for `#[verifier::*]`.

### DIFF
--- a/source/rust_verify/example/state_machines/top_sort_dfs.rs
+++ b/source/rust_verify/example/state_machines/top_sort_dfs.rs
@@ -43,7 +43,7 @@ impl<V> DirectedGraph<V> {
 }
 
 tokenized_state_machine!{
-    TopSort<#[verifier(maybe_negative)] V> {
+    TopSort<#[verus::verifier(maybe_negative)] V> {
         fields {
             #[sharding(constant)]
             pub graph: DirectedGraph<V>,
--- a/source/rust_verify/example/state_machines/top_sort_dfs.rs
+++ b/source/rust_verify/example/state_machines/top_sort_dfs.rs
@@ -171,7 +171,7 @@ struct NodeState {
     visited: bool,
     in_stack: bool,
 
-    #[proof] token: NodeToken,
+    #[verus::proof] token: NodeToken,
 }
 
 impl NodeState {
@@ -204,8 +204,8 @@ struct DfsState {
 
     node_states: Vec<NodeState>,
 
-    #[proof] top_sort_token: TopSort::top_sort<usize>,
-    #[proof] instance: TopSort::Instance<usize>,
+    #[verus::proof] top_sort_token: TopSort::top_sort<usize>,
+    #[verus::proof] instance: TopSort::Instance<usize>,
 }
 
 spec fn valid_stack_i(cur_stack: Seq<usize>, graph: DirectedGraph<usize>, i: int) -> bool {
@@ -318,7 +318,7 @@ fn visit(
     }
 
     if dfs_state.node_states.index(v as usize).visited {
-        #[proof] let tok = match &dfs_state.node_states.index(v as usize).token {
+        #[verus::proof] let tok = match &dfs_state.node_states.index(v as usize).token {
             NodeToken::Visited(tok) => tok.clone(),
             _ => proof_from_false(),
         };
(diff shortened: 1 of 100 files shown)


source/rust_verify/example/state_machines/top_sort_dfs.rs
```diff
@@ -43,7 +43,7 @@ impl<V> DirectedGraph<V> {
 }
 
 tokenized_state_machine!{
-    TopSort<#[verifier(maybe_negative)] V> {
+    TopSort<#[verus::verifier(maybe_negative)] V> {
```
utaal, line 46: @tjhance it looks like this isn't being replaced by the `verus!` macro, which now will replace `verifier` with `verus::verifier`; my understanding is that the state machine macros run (parts of) the token stream through the `verus!` macro, is that correct?

Chris-Hawblitzel:
Can we get rid of `verifier(...)`?  For example, in things like `#[verus::verifier(external_body)]`, can we just say `#[verus::external_body]` instead?  The use of `verifier(...)` was for namespacing, and it feels like we now redunantly have two levels of namespacing.

source/rust_verify/example/state_machines/top_sort_dfs.rs
tjhance: that's right, but only part of it. i think there are a few places like that cause awkwardness and need to be fixed

utaal:
The current implementation auto-prefixes '#[verifier(external_body)]' with 'verus::' in the 'verus!' macro. The intent is that the 'verus::' prefix won't ever appear in surface level code that uses the macro (which we expect to be the majority of the code).

If that's the case, I don't think keeping 'verifier' is much of a burden on the user, and would not break any existing code that uses the 'verus!' macro (though we can add temporary support for the transition).
More importantly, the only non-namespaced attributes we have left are 'trigger' and 'auto', which should be compact because they are inline (the mode attributes are replaced by the syntax). I think keeping some namespacing for the less-frequent attributes may help with clarifying which ones are verus vs regular rust attributes: these often appear intermixed on item (struct, fn) definitions.

(Sorry for the formatting, typed on mobile.)

tjhance:
It would still be convenient if there was only one namespace both for internals and externals. We could just ask the user to write `#[verus::inline]` instead of `#[verifier(inline)]`.

I guess ultimately it comes down to which one we prefer as users. I really wouldn't mind either one, though `verus::` also has the advantage it's more in keeping with what the rust ecosystem expects.

utaal:
Having looked through the macro output to debug issues, I think we should keep the internal and external namespaces separate as it helps debugging, so I'd keep the register_tool namespace (ie 'verus::') as exclusively internal.

My thinking is then: (i) if you're using the macro, you've prefixed everything with 'verus!', so we can keep 'verifier' as the namespace ("here's a verifier directive"); if we end up supporting some lighter weight macro-less syntax, the user can then use 'verus::' instead directly, but consistently with the macro.

utaal:
I now think we should replace 'verifier(external_body)' with 'verifier::external_body' though. As the call-like syntax usually carries a parameter or modifier (eg 'inline(always)').

utaal:
There's also a bit of "scary things should look scary" in my preference for the prefix: (almost?) everything in 'verifier::' are ways to mark items as trusted in some way.

Chris-Hawblitzel:
So is the proposal now that code outside the syntax macro would use `#[verifier::external_body]` or `#[verus::verifier::external_body]`?  The latter seems unnecessarily long.

utaal:
I was thinking the latter, as we (currently) do not expect people to write code outside the macro (as it won't erase properly). But if that's a constraint, I think we should go for `#[verus::external_body]` as you suggested, to avoid having multiple `register_tool`s.

utaal:
I’m reconsidering the naming based on what we discussed. I think the solution may be to be a bit more systematic in how various attributes (for different purposes) are namespaced (e.g. whether they're only ever macro-generated). Let me think about a bit more and come back with a proposal.

utaal:
Ok. This PR is getting quite large (although a lot of it is find/replace in tests, which however have a higher chance to result in merge conflicts), so I'd like to land it before we add more attribute-related syntax, or new replacement attributes for `#[verifier(...)]`.

tjhance (approved):

Chris-Hawblitzel (approved):

PR #359, opened 2023-02-03
Height intrinsic & soundness issue with variants

tjhance:
Two changes rolled in here:

 * Expose `height` intrinsic to the user. This is similar to the integer-bit-width intrinsics from last week.
 * Fix an unrelated soundness issue I noticed in passing. The datatype height axiom needs a check that a value is of the right variant in order to give height-bounds on that variant's fields.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -653,6 +653,13 @@ fn check_expr_handle_mut_arg(
             let mode = check_expr(typing, joined_mode, erasure_mode, e1)?;
             Ok(mode_join(*min_mode, mode))
         }
+        ExprX::UnaryOpr(UnaryOpr::Height, e1) => {
+            if typing.check_ghost_blocks && typing.block_ghostness == Ghost::Exec {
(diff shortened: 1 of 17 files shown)


source/vir/src/modes.rs
```diff
@@ -653,6 +653,12 @@ fn check_expr_handle_mut_arg(
             let mode = check_expr(typing, joined_mode, erasure_mode, e1)?;
             Ok(mode_join(*min_mode, mode))
         }
+        ExprX::UnaryOpr(UnaryOpr::Height, e1) => {
+            if typing.check_ghost_blocks && typing.block_ghostness == Ghost::Exec {
+                return err_str(&expr.span, "cannot test 'height' in exec mode");
+            }
+            check_expr(typing, Mode::Spec, erasure_mode, e1)
```
Chris-Hawblitzel, line 660: I think it would be clearer if this returns `Mode::Spec`.

Chris-Hawblitzel (approved):

PR #363, opened 2023-02-03
remove parens from decreases warning as they're not needed in verus! …

jonhnet:
…macro.

This came up when working with jialin & tony: the suggestion told us to type `decreases(...)
`and then I said "no don't type that because verus! lets you use nicer syntax, just as with requires".

On the other hand, what happens if users aren't using the verus! syntax wrapper? Hrm.

### DIFF
--- a/source/rust_verify/tests/consts.rs
+++ b/source/rust_verify/tests/consts.rs
@@ -35,14 +35,14 @@ test_verify_one_file! {
     #[test] test1_fails2 verus_code! {
         const C: u64 = S;
         const S: u64 = C;
-    } => Err(err) => assert_vir_error_msg(err, "recursive function must call decreases")
+    } => Err(err) => assert_vir_error_msg(err, "recursive function must have a decreases clause")
 }
 
 test_verify_one_file! {
     #[test] test1_fails3 verus_code! {
         spec const C: u64 = S;
         spec const S: u64 = C;
-    } => Err(err) => assert_vir_error_msg(err, "recursive function must call decreases")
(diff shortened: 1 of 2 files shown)


tjhance:
can you make it something like "has a `decreases` clause" instead of suggesting a "call"?

jonhnet:
I can -- I will -- but this does sort of open up a general can of worms.

One of the features that makes Rust error messages so pleasant is that, surprisingly often, they suggest the exact text of the solution to your problem. We have three choices:

- suggest correct text for the raw language that's confusing to users of the pretty verus! language
- suggest correct text for the pretty language that's confusing to users of the raw language
- abandon this niceness in our error messages (what I'll do for this PR)
- do something deeper that propagates/transforms error messages "up the stack" the same way line numbers get moved back up to the source

I submit that the pretty language is non-negotiably important. If keeping the raw language in play remains an important goal, then we probably want to take the last option, which will require some design thought.

utaal:
CI is unhappy. I think you need to update the expected error messages in `tests/consts.rs`. (Run `./tools/cargo.sh test --test consts` to see which tests are failing).

utaal:
FWIW, the raw language will soon be deprecated and will become more awkward to use (e.g. you'll need to write `#[verus::proof]`, and you won't be able to use the `spec`/`proof`/`exec` attributes on variables -- at least in exec code). We should make sure to clearly communicate that, and we should make a pass to update the error messages when we completely transition to the new syntax.

There are chunks of code that will break as a result of these changes, but unfortunately continuing to still support the raw syntax in its entirety would prevent us from moving off the rustc fork, so I think we should plan for a (near) future where the raw syntax is considered unsupported.

PR #364, opened 2023-02-04
Support ConstKind::Unevaluated

ziqiaozhou:
An example is Array<u8, MAX_LEN>.

BTW, the current const generic support does not allow exe #[external_body] fn to use N in body.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -384,15 +384,16 @@ pub(crate) fn mid_ty_to_vir<'tcx>(
 }
 
 pub(crate) fn mid_ty_const_to_vir<'tcx>(
-    _tcx: TyCtxt<'tcx>,
+    tcx: TyCtxt<'tcx>,
     cnst: &rustc_middle::ty::Const<'tcx>,
 ) -> Typ {
     use rustc_middle::mir::interpret::{ConstValue, Scalar};
     use rustc_middle::ty::ConstKind;
     match cnst.val {
         ConstKind::Param(param) => Arc::new(TypX::TypParam(Arc::new(param.name.to_string()))),
-        ConstKind::Value(ConstValue::Scalar(Scalar::Int(i))) => {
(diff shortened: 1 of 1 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -384,15 +384,16 @@ pub(crate) fn mid_ty_to_vir<'tcx>(
...
     cnst: &rustc_middle::ty::Const<'tcx>,
 ) -> Typ {
     use rustc_middle::mir::interpret::{ConstValue, Scalar};
     use rustc_middle::ty::ConstKind;
     match cnst.val {
         ConstKind::Param(param) => Arc::new(TypX::TypParam(Arc::new(param.name.to_string()))),
-        ConstKind::Value(ConstValue::Scalar(Scalar::Int(i))) => {
-            let c = num_bigint::BigInt::from(i.assert_bits(i.size()));
+        ConstKind::Unevaluated(_) | ConstKind::Value(_) => {
+            let val = cnst.try_eval_usize(tcx, rustc_middle::ty::ParamEnv::empty()).unwrap();
```
Chris-Hawblitzel, line 395: `usize` is not the only possible type, and I'm worried about `ParamEnv::empty`.  Can you try this instead?:

```
    let cnst = match cnst.val {
        ConstKind::Unevaluated(unevaluated) => {
            cnst.eval(tcx, tcx.param_env(unevaluated.def.did))
        }
        _ => cnst,
    };
    match cnst.val {
        ConstKind::Param(param) => Arc::new(TypX::TypParam(Arc::new(param.name.to_string()))),
        ConstKind::Value(ConstValue::Scalar(Scalar::Int(i))) => {
            let c = num_bigint::BigInt::from(i.assert_bits(i.size()));
            Arc::new(TypX::ConstInt(c))
        }
        _ => {
            unsupported!(format!("const type argument {:?}", cnst))
        }
    }
```

Chris-Hawblitzel (changes requested):

Chris-Hawblitzel (approved):

PR #374, opened 2023-02-13
Allow Verus's standard library to be in a separate crate.

Chris-Hawblitzel:
Currently, this is disabled by default but it can be enabled with the "--export vstd.vir" and "--import vstd=vstd.vir" command-line options. In addition, exporting the standard library requires "--compile" and the standard Rust option "--crate-type=lib". Importing requires (for now) "--cfg vstd_todo" and the standard Rust option "--extern vstd=libvstd.rlib".

For example:

Generating the standard library (vstd):

rust_verify --pervasive-path pervasive --extern builtin=../rust/install/bin/libbuiltin.rlib --extern builtin_macros=../rust/install/bin/builtin_macros.dll --extern state_machines_macros=../rust/install/bin/state_machines_macros.dll --edition=2018 --log-all --cfg erasure_macro_todo --erasure macro --crate-type=lib --compile --export vstd.vir pervasive/vstd.rs

Using the standard library:

rust_verify --pervasive-path pervasive --extern builtin=../rust/install/bin/libbuiltin.rlib --extern builtin_macros=../rust/install/bin/builtin_macros.dll --extern state_machines_macros=../rust/install/bin/state_machines_macros.dll --edition=2018 --log-all --cfg erasure_macro_todo --cfg vstd_todo --erasure macro --extern vstd=libvstd.rlib  --import vstd=vstd.vir rust_verify/example/vectors.rs

At the moment, just a few examples in rust_verify/example have support for using vstd instead of including the whole pervasive module directly; this is currently configured by "--cfg vstd_todo", as in:

```
#[cfg(not(vstd_todo))]
mod pervasive;
#[cfg(not(vstd_todo))]
#[allow(unused_imports)]
use pervasive::{*, vec::*, seq::*, modes::*};

#[cfg(vstd_todo)]
#[allow(unused_imports)]
use vstd::pervasive::{*, vec::*, seq::*, modes::*};
```

Moving all of our examples and tests to use separate crates is still to be done (after merging this pull request).  In particular, we'll have to decide on a build process for the standard library, how to have tests link to the generated library crate, whether to have just one library crate or many, a name or names for the library crate or crates (vstd?  something else?), etc.  Maintaining compatibility between the old way of handling pervasives and the new way is annoying, but we can simplify things once everything is converted.

At a high level, the support for multiple crates is straightforward.  When exporting a crate, we serialize the VIR AST to a binary file, and when importing crates, we deserialize the VIR AST for those crates.  In more detail, the main issues that this pull request deals with are:

- To allow Rust to type check references to spec and proof declarations across crates, we have to retain those declarations through compilation (we can't entirely erase them).  We still erase the function bodies, though, by replacing them with "panic!()".
- Naively serializing and deserializing Rust SpanData values would not work; instead, for each VIR AST Span we serialize a stable crate identifier and offset within the crate, and then use SourceMap to reconstruct a span correctly after deserializing.
- DefId vs LocalDefId: some places that relied on LocalDefId have been generalized to get information from non-local crates as well.
- We no longer rely exclusively on LocalDefId to fix up paths like "pervasive::vec::{impl#0}::len"; as a result, the code is now stricter in disallowing things like "{impl#0}" from showing up in a VIR Path in the first place.

Note: we don't yet attempt to prune unnecessary data from the VIR AST before serializing; this is still future work.


Chris-Hawblitzel:
> What happens if we update the verus and the veruslib becomes out-of-date? Is there a mechanism to detect out-of-date veruslib or would it silently accept the file?

There's no mechanism to detect this yet.  I think it's worth discussing how to build both verus and veruslib, perhaps with cargo.

utaal (approved):
I pushed a change that in theory should address the performance issue with reading/writing to files; it doesn't seem to make a difference on my machine, you may want to test it on yours -- if it's a regression I'm happy to drop the commit.

PR #375, opened 2023-02-14
pervasive: adding method Map::contains_value

achreto:
This is the dual of Map::contains_key but it checks whether the map contains a specific value.

### DIFF
--- a/source/pervasive/map.rs
+++ b/source/pervasive/map.rs
@@ -109,7 +109,6 @@ impl<K, V> Map<K, V> {
 
     /// Returns true if the value `v` is in the map of `self`.
 
-    #[verifier(inline)]
     pub open spec fn contains_value(self, v: V) -> bool {
         exists|i: K| #[trigger] self.dom().contains(i) && self[i] == v
     }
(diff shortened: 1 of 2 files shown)


source/pervasive/map.rs
```diff
@@ -107,6 +107,13 @@ impl<K, V> Map<K, V> {
         self.dom().contains(k)
     }
 
+    /// Returns true if the value `v` is in the map of `self`.
+
+    #[verifier(inline)]
+    pub open spec fn contains_value(self, v: V) -> bool {
+        exists|i: K| #[trigger] self.dom().contains(i) && self.index(i) == val
```
utaal, line 114: Sorry I didn't spot this earlier, can you use `self[i]` instead?

achreto: Done!

utaal:
I added a test, which ends up generating invalid AIR. I don't know what the issue is yet.

utaal:
It was a problem with `#[verifier(inline)]` which interacts badly with native types as type params here.

PR #381, opened 2023-02-17
Fix interpreter handling for usize/isize

### LINKED ISSUE #380: interpreter wrongly uses 32-bit truncation
```rust
proof fn test_shift() {
    assert((1usize << 40usize) == 0usize) by(compute);
}
```

This wrongly passes when `--arch-word-bits` is not set, because it truncates the shift to `arch_size_min_bits` which is set to 32. However, when `--arch-word-bits` is not set, the interpreter needs to account for the possibility that the arch size might be 64.

jaybosamiya:
Closes #380 which showed that previous handling was unsound

Basically, if we are in a 32-or-64-bit state, then we need to check that doing the computation at that size would lead to the same result, otherwise, we need to conservatively give up the computation at that point to prevent unsoundness.

Interestingly, it turns out conservatively disallowing the `Clip` operation from happening (in the 32-or-64-bit case) causes some tests to start failing; fixed by simply making the Clip operation a bit more precise whenever possible (i.e., when the clipped value is within 32-bit world) is sufficient.

I _am_ a tiny bit surprised though, since this came into play for:
```rs
        spec fn shifter(x: u64, amt: usize) -> u64
            decreases amt
        {
            if amt == 0 { x } else { shifter(x << 1, (amt - 1) as usize) }
        }
```
where one would think that the result should not depend on 32-bit or 64-bit architecture. Weird. I don't think that by itself should be a blocker to this fix though, but something worth looking into when we get a chance.

### DIFF
--- a/source/rust_verify/example/assert_by_compute.rs
+++ b/source/rust_verify/example/assert_by_compute.rs
@@ -390,4 +390,22 @@ mod veribetrkv_example_list_comprehension {
     }
 }
 
+//#[cfg(any())]
+mod arch_specific {
+
+    use builtin::SpecShl;
+
+    proof fn test_shift() {
+        assert((1usize << 20usize) != 0usize) by (compute_only);
+        assert((1usize << 100usize) == 0usize) by (compute_only);
+
+        // But this next assert should not work (at least without --arch-word-bits), because usize
+        // could be either 32-bit or 64-bit.
+        //
+        // assert((1usize << 40usize) == 0usize) by (compute_only);
+    }
+
+}
+
+
 } // verus!
--- a/source/rust_verify/tests/assert_by_compute.rs
+++ b/source/rust_verify/tests/assert_by_compute.rs
@@ -445,3 +445,39 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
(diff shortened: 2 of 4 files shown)


tjhance:
Thanks for picking this up. Could you add a regression test?

jaybosamiya:
I had added a test as part of the the commit above, but I don't think we currently have support for negative tests inside the same file, right? Or maybe I'm missing something

utaal:
> I had added a test as part of the the commit above, but I don't think we currently have support for negative tests inside the same file, right? Or maybe I'm missing something

You should be able to use the testing infrastructure: can you add a test here https://github.com/verus-lang/verus/blob/main/source/rust_verify/tests/assert_by_compute.rs ? (Please add `regression_<issue number>` at the end of the test name.

jaybosamiya:
Thanks, I was looking only inside the `examples` dir previously, so had missed out on the full infra. Added some regression tests to cover both positive and negative uses.

utaal:
Tests look good to me. @parno should review the interpreter changes.

jaybosamiya:
Thanks!

PR #384, opened 2023-02-21
Extensionality axioms for n-argument functions

jaybosamiya:
Adds the extensionality axioms for `n`-argument functions. Previously (via #318), we could only do extensionality for 1-argument functions, with this PR we can do it for more.

Implemented as a macro to make it easier to extend in the future if we need to, and to maintain consistency in the axioms.

### DIFF
--- a/source/pervasive/function.rs
+++ b/source/pervasive/function.rs
@@ -25,3 +25,30 @@ verus! {
     ensures f1 == f2
   {}
 }
+
+/// A macro to conveniently generate similar functional extensionality axioms for functions that
+/// take `n` arguments.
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (approved):
This looks fine to me.  Actually, we don't need `dummy_trigger` anymore; you can just delete `dummy_trigger` completely from `function.rs` and trigger on `f1(x)` or `f2(x)` directly (either one will work; the trigger doesn't matter).

jaybosamiya:
Thanks! And removed `dummy_trigger`; can be merged once CI succeeds

PR #390, opened 2023-02-22
support array type

tjhance:
Create a VIR builtin 'array' type, similar to slices. Lower Rust's `[T; N]` to the VIR type.

This mostly just adds the type; we can add more executable operations to pervasive later.

There were a couple of issues in lifetime_generate involving const generics, which I fixed.

### DIFF
--- a/source/vir/src/builtins.rs
+++ b/source/vir/src/builtins.rs
@@ -5,6 +5,11 @@ use air::ast_util::ident_binder;
 use std::sync::Arc;
 
 pub fn krate_add_builtins(no_span: &Span, krate: &mut KrateX) {
+    krate_add_slice(no_span, krate);
+    krate_add_array(no_span, krate);
+}
+
+pub fn krate_add_slice(no_span: &Span, krate: &mut KrateX) {
     // Add a datatype for 'slice'
 
     let path = crate::def::slice_type();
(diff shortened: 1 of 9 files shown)


utaal:
Should we add `array%` (and `slice%`) as new variants on `vir::ast::TypX` instead of making them look like datatypes?

Chris-Hawblitzel:
> Should we add `array%` (and `slice%`) as new variants on `vir::ast::TypX` instead of making them look like datatypes?

The way we handle string slices and regular slices is inconsistent on this point at the moment, but I'm not sure which direction to resolve it in.  Maybe when we add array construction expressions `[elem1, ..., elemn]` this will become clearer.

source/vir/src/builtins.rs
```diff
@@ -23,3 +28,26 @@ pub fn krate_add_builtins(no_span: &Span, krate: &mut KrateX) {
...
+
+    let typ_params = Arc::new(vec![
+        (crate::def::array_param_ty(), bound.clone(), is_strictly_positive),
+        (crate::def::array_param_len(), bound, is_strictly_positive),
+    ]);
+
+    let datatypex =
+        DatatypeX { path, visibility, transparency, typ_params, variants, mode: Mode::Exec };
+    krate.datatypes.push(Spanned::new(no_span.clone(), datatypex));
+}
```
utaal, line 53: The fact that we're making up e.g. a variant here seems a bit hackish to me. What is using a `Datatype` buying us? The ability to use the `typ_params` machinery?

tjhance:
I thought it would be an efficient way to make new types that I didn't expect to require any special handling in VIR. Fewer cases to maintain, piggy-backing on all the existing type parameter handling & poly stuff, and so on.

utaal:
I tried adding an AST node for Array (and Slice), and I prefer it: it distinguishes these primitive types, and ignores code paths that are irrelevant for them (e.g. the type scc computation): https://github.com/verus-lang/verus/compare/array...array-alt

This approach also works for slices with minimal additional code: https://github.com/verus-lang/verus/compare/array...slice-alt
The additional bit of logic to handle `Primitive` introduced gets offset by removing the ad-hoc datatypes for both.

However, it turns out they're both blocked behind https://github.com/verus-lang/verus/pull/327

The `SliceAdditionalSpecFns` introduced in https://github.com/verus-lang/verus/commit/2d3173e908c321b59e74faf0fd02d42215b70236 was implemented for the wrong type, I believe (`[T]` instead of `&[T]`); which I think would make these conflicting implementation if both are imported at the same time.
And because of that, my `array-alt` branch works by piggybacking on the fact that slices are treated as datatypes.

I think we should finish and merge the generalised traits PR, and then use the `slices-alt` branch that uses the new `Primitive` type.

utaal:
As discussed offline, I'm happy to merge this as-is for the time being; we can reconsider when we have a longer-term plan for trait naming.

The only concern I have right now is this, unless this isn't a problem.

> The `SliceAdditionalSpecFns` introduced in [2d3173e](https://github.com/verus-lang/verus/commit/2d3173e908c321b59e74faf0fd02d42215b70236) was implemented for the array type, I believe (`[T]` instead of `&[T]`); which I think would make these ambiguous implementations (both for `[T]`) if both are imported at the same time.

tjhance:
superseded by #398

PR #391, opened 2023-02-22
Improved `assert_*_equal!` syntax

jaybosamiya:
With this PR, we now also allow for syntax that looks like `assert_seqs_equal!(a == b)`, saving precious few seconds that would be needed to convert a `assert(a == b)` to a `assert_seqs_equal!(a, b)`, allowing one to reach a state of proof-writing flow that has never been achieved before!

A little more seriously, despite only a tiny syntactic change needed to get from `assert(a == b)` to the previous `assert_seqs_equal!(a, b)`, it was a bit annoying, especially when `a` and `b` are complicated/large. Also, the error message provided previously was not particularly great, and would be surprising to new users of Verus (the error was: `error: unexpected end of macro invocation` with a `missing tokens in macro arguments` pointing at the end of the macro).

With the new syntax, we don't have these issues anymore, since the `a == b` syntax JustWorks^TM. 

I've decided to still keep around the old syntax, and haven't updated any usages of `assert_*_equal!` macros anywhere (other than adding tests), but at some point, we _may_ consider deprecating it and replacing it with the new syntax. That sort of replacement/deprecation is out of scope for this particular PR though.

### DIFF
--- a/source/pervasive/map.rs
+++ b/source/pervasive/map.rs
@@ -352,7 +352,7 @@ pub use map;
 ///     ensures m.insert(k, v).remove(k) == m
 /// {
 ///     let m2 = m.insert(k, v).remove(k);
-///     assert_maps_equal!(m, m2);
+///     assert_maps_equal!(m == m2);
 ///     assert(m == m2);
 /// }
 /// ```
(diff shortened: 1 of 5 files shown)


tjhance (approved):
Good idea

PR #393, opened 2023-02-23
pervasive/seq_lib: adding support for Seq to Set conversion

achreto:
This adds a new method `to_set()` to the Seq<A> implementation in the seq_lib module that converts the sequence into a set. This also includes a proof that the resulting set is finite via a recursive definition of to_seq.

### DIFF
--- a/source/pervasive/seq_lib.rs
+++ b/source/pervasive/seq_lib.rs
@@ -6,6 +6,8 @@ use builtin_macros::*;
 use crate::pervasive::*;
 #[allow(unused_imports)]
 use crate::pervasive::seq::*;
+#[allow(unused_imports)]
+use crate::pervasive::set::Set;
 
 verus! {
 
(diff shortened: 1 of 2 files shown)


source/pervasive/seq_lib.rs
```diff
@@ -31,14 +33,101 @@ impl<A> Seq<A> {
...
     {
         self.subrange(0, self.len() as int - 1)
-    } 
+    }
+
+    /// returns `true` if the sequequence has no duplicate elements
+    pub open spec fn no_duplicates(self) -> bool {
+        forall(|i, j| (0 <= i < self.len() && 0 <= j < self.len() && i != j)
+            ==> self[i] != self[j])
+    }
```
Chris-Hawblitzel, line 46: minor comment: you can remove parentheses and just write
```
        forall|i, j| 0 <= i < self.len() && 0 <= j < self.len() && i != j
            ==> self[i] != self[j]
```

source/pervasive/seq_lib.rs
```diff
@@ -31,14 +33,101 @@ impl<A> Seq<A> {
...
+
+    /// Returns `true` if two sequences are disjoint
+    pub open spec fn disjoint(self, other: Self) -> bool {
+        forall |a: A| self.contains(a) ==> !other.contains(a)
```
Chris-Hawblitzel, line 50: This has three quantifiers and it has alternating forall/exists quantifiers.  It would be better to write it with just one quantifier:

```
forall|i: int, j: int| 0 <= i < self.len() && 0 <= j < other.len() ==> self[i] != other[j]
```

source/pervasive/seq_lib.rs
```diff
@@ -31,14 +33,101 @@ impl<A> Seq<A> {
...
+    /// Converts a sequence into a set
+    pub open spec fn to_set(self) -> Set<A> {
+        Set::new(|a: A| self.contains(a))
+    }
+}
+
+
+/// recursive definition of seq to set conversion
+spec fn seq_to_set_rec<A>(seq: Seq<A>) -> Set<A>
+    decreases seq.len() when seq.len() >= 0
```
Chris-Hawblitzel, line 62: I don't think you the `when seq.len() >= 0` here because `seq.len() >= 0` is always true.

source/pervasive/seq_lib.rs
```diff
@@ -31,14 +33,101 @@ impl<A> Seq<A> {
...
+
+/// proof function showing that the set obtained from the sequence is finite
+pub proof fn seq_to_set_is_finite<A>(seq: Seq<A>)
+    ensures seq.to_set().finite()
+{
+    assert(seq.to_set().finite()) by {
+        seq_to_set_equal_rec(seq);
+        seq_to_set_rec_is_finite(seq);
+    }
+}
```
Chris-Hawblitzel, line 128: In the long run, once `broadcast_forall` is generalized, I think we would want to make `seq_to_set_is_finite` `broadcast_forall` so that you don't have to call it explicitly.  Since this library file is basically trusted anyway, I think we should go ahead and add a trusted `broadcast_forall` function:

```
#[verifier(external_body)]
#[verifier(broadcast_forall)]
pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
    ensures seq.to_set().finite()
{
    // TODO: merge this with seq_to_set_is_finite when broadcast_forall is better supported
}
```
Then we can remove the calls to `seq_to_set_is_finite` from the tests.

Chris-Hawblitzel (approved):

source/pervasive/seq_lib.rs
achreto: Replaced definition with suggested forall in second commit.

source/pervasive/seq_lib.rs
achreto: Removed the `when` clause.

achreto:
Pushed the changes that include the suggestions and the broadcast_forall.

Chris-Hawblitzel:
Feel free to merge.

PR #394, opened 2023-02-24
Update VS Code docs to include info on WSL

jaylorch:


chanheec (approved):
Thanks a lot for sharing how to set this up with WSL!

PR #396, opened 2023-02-24
Add is_variant annotation to state_machine! Step enum

tenzinhl:
Allows checking of Step enum variant using is_variant methods in the auto-generated Step enum produced by the state_machine! macro.

### DIFF
--- a/source/builtin_macros/src/is_variant.rs
+++ b/source/builtin_macros/src/is_variant.rs
@@ -67,7 +67,7 @@ pub fn attribute_is_variant(
             };
 
             quote! {
-                verus! {
+                ::builtin_macros::verus! {
                     #[verus::internal(is_variant(#variant_ident_str))]
                     #[allow(non_snake_case)]
                     #[verus::internal(spec)]
--- a/source/rust_verify_test/tests/state_machines.rs
+++ b/source/rust_verify_test/tests/state_machines.rs
@@ -7037,3 +7037,43 @@ test_verify_one_file! {
 
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] step_is_variant verus_code! {
+        use state_machines_macros::state_machine;
+
+        state_machine! { X {
+            fields {
+                pub number: int,
+            }
+
+            init!{
+                initialize() {
+                    init number = 0;
+                }
+            }
+
+            transition!{
+                add(n: int) {
+                    update number = pre.number + 2*n;
+                }
+            }
+
+            transition!{
+                sub(n: int) {
+                    update number = pre.number - 2*n;
+                }
+            }
+        }}
+
+        proof fn test(s: X::Step)
+            requires s.is_add(),
+        {
+            match s {
+                X::Step::add(_) => assert(true),
+                X::Step::sub(_) => assert(false),
+                _ => {},
+            }
+        }
+    } => Ok(())
+}
--- a/source/state_machines_macros/src/to_token_stream.rs
+++ b/source/state_machines_macros/src/to_token_stream.rs
@@ -482,6 +482,7 @@ fn output_step_datatype(
 
     root_stream.extend(quote! {
         #[allow(non_camel_case_types)]
+        #[is_variant]
         pub enum #type_ident#generics {
             #(#variants,)*
             // We add this extra variant with the self_ty in order to avoid
--- a/source/state_machines_macros/src/to_token_stream.rs
+++ b/source/state_machines_macros/src/to_token_stream.rs
@@ -482,7 +482,7 @@ fn output_step_datatype(
 
     root_stream.extend(quote! {
         #[allow(non_camel_case_types)]
-        #[is_variant]
+        #[::builtin_macros::is_variant]
         pub enum #type_ident#generics {
             #(#variants,)*
             // We add this extra variant with the self_ty in order to avoid


tjhance:
i think you can fix the test by changing it to `::builtin_macros::is_variant`

tenzinhl:
Heyo! I tried running tests locally with `./tools/cargo.sh test` from the `source` directory, is that the right way to do it?

Anyways it was a bit ago, but when I ran tests using said command I don't remember running into any errors (re-running tests rn, takes a bit to build and run everything though, is there a way to run a specific test?).

Any tips on how to resolve the test that's failing on the CI here? @tjhance

utaal:
That's right. You can use `nextest` by installing the binary from here https://nexte.st, which is a bit faster: the command then becomes `./tools/cargo.sh nextest run`. You can pick a test file with `--test <test_file_name>` or you can pick a specific test with `-- test_name` (I recommend nextest if you do this, as it'll filter the test more effectively).

tenzinhl:
Okay, so it seems like in the macro-expanded output (checked using `$verus -Zunpretty=expanded top_sort_dfs.rs`, is this the right way to check the expansion?) neither the `Step` nor `Config` enums are generated.

As far as I know this happens when... something (I'm unsure what) in the parsing pipeline detects an error (then that entire chunk of the parse tree is discarded). What part of the parsing does this discarding when it detects malformed input?

Is there a way to catch where this happens?

How do I figure out why it's being considered malformed? I'm confused since this change was working for our state machine in `verified-betrfs` for Splinter port. One difference I notice here is that the test that fails uses the `tokenized_state_machine` macro, is that relevant here?

tenzinhl:
After some more digging where we messed around with the `top_sort_dfs.rs` file, it seems like the issue has to do with combining the `is_variant` macro with generic state machines. Not sure what the exact issue is though, does `is_variant` handle generics at all? If so, how do we modify the syntax for it to work with generics.

tjhance:
it looks like is_variant can't handle the `#[verifier(maybe_negative)]` attribute, which is itself a bug. @utaal do you know if that's an easy fix? I'm not familiar with the is_variant macro or the synstructure library.

tenzinhl:
If you don't mind: how did you determine that it's the interaction between `is_variant` and `#[verifier(maybe_negative)]`?

Curious about how to investigate Verus stuff.

tjhance:
well, you already told me it had something to do with generics, but i already knew is_variant ought to work with generics, so I knew it had to be something more complicated than that

the only thing special about the generics in top_sort_dfs (the failing test) was the verifier(maybe_negative), so I guessed it was probably that. So to test it out, I tried:

```
#[is_variant]
enum X<#[verifier(maybe_negative)] T> {
  ZZ(T),
}
```

and got this error message:

```
error: expected one of `>`, a const expression, lifetime, or type, found `#`
  --> test_gen.rs:16:8
   |
16 | enum X<#[verifier(maybe_negative)] T> {
   |        ^ expected one of `>`, a const expression, lifetime, or type
```

so i deduced that  the `is_variant` macro evidently doesn't know how to parse attributes on generic type parameters.

tenzinhl:
Have there been any updates on this? I'd be happy to work on this with guidance.

tjhance:
`is_variant` is a verus proc-macro, part of `builtin_macros`, which is implemented here: https://github.com/verus-lang/verus/blob/main/source/builtin_macros/src/is_variant.rs

The implementation parses the item declaration using a library called `synstructure` which apparently can't handle the type parameter with an attribute. For that matter, it probably can't handle any verus-specific stuff either. (Type attributes aren't verus-specific though, I think? So it's a little surprising synstructure doesn't support it.)

I think the easiest fix would probably be to replace the use of `synstructure` by just parsing an `ItemEnum` using `verus_syn`.

tenzinhl:
Coolio, thanks Travis :)

utaal:
I'm sorry, I completely missed this. @tenzinhl I'm happy to take a look at the `#[is_variant]` issue.

tjhance:
@utaal there's some more discussion at this issue: https://github.com/verus-lang/verus/issues/480

tenzinhl:
Now issue is this:
```
error: cannot find macro `verus` in this scope
  --> adder_generic.rs:7:1
   |
7  | / state_machine!(
8  | |     X<T> {
9  | |         fields {
10 | |             pub number: int,
...  |
39 | |     }
40 | | );
   | |_^
   |
   = help: consider importing one of these items:
           builtin_macros::verus
           crate::prelude::verus
           vstd::prelude::verus
   = note: this error originates in the attribute macro `::builtin_macros::is_variant` (in Nightly builds, run with -Z macro-backtrace for more info)

error: cannot find macro `verus` in this scope
  --> adder_generic.rs:7:1
   |
7  | / state_machine!(
8  | |     X<T> {
9  | |         fields {
10 | |             pub number: int,
...  |
39 | |     }
40 | | );
   | |_^
   |
   = help: consider importing one of these items:
           builtin_macros::verus
           crate::prelude::verus
           vstd::prelude::verus
   = note: this error originates in the attribute macro `::builtin_macros::is_variant` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The `is_variant` macro puts its contents inside of a `verus!` macro, but apparently the contexts where the `state_machine!` macro is expanded don't necessarily have `verus!` imported. What's the right way to fix this?

utaal:
This is, I believe, an issue I introduced by not using the `verus` macro's full path. And considering `#[is-variant]` may appear in code that's already inside a `verus!` macro invocation, using `verus!` in `#[is-variant]` may have been a poor choice too. Let me check.

tenzinhl:
@utaal I see you made commit https://github.com/verus-lang/verus/commit/7503f6f1fd0515140113e0b426dac048293a135e and it seems to pass checks. Should I close this MR for now?

utaal:
I merged that commit into `main_new` now, but that doesn't add the annotation to the state_machine enum, so I think you may still want that. We may want to rebase this PR on `main_new` and check that your additions now work. (Let me know if I can help.)

tjhance:
@utaal is this resolved through your other PR? Can we close this now?

PR #398, opened 2023-02-24
Use a separate ADT node, and generalised traits to support array, slice

utaal:
Candidate replacement for https://github.com/verus-lang/verus/pull/390 

Blocked behind some version of https://github.com/verus-lang/verus/pull/327


jaybosamiya:
Just adding a note here that once this is merged, we should be able to switch the [`bytes` module](https://github.com/verus-lang/verus/blob/main/source/pervasive/bytes.rs) (introduced in #453) to be more precise, and not require the otherwise-unnecessary `Vec` allocation

utaal:
I'm going to revive this, now that #571 has landed.

utaal:
I'm going to revive this, now that #571 has landed. cc/ @tjhance

utaal:
Replaced by #611.
(Please don't delete the branch for the time being, as this contains the diff for the first attempt.)

PR #402, opened 2023-02-27
Don't require main in files verified via VS Code

jaylorch:


jaylorch:
Chanhee says there's a better way to handle this, on the server side, so I'll withdraw this pull request.

utaal:
Fwiw, `--crate-type=lib` doesn't currently work (due to an interaction with erasure that we never could figure out), but it should be fixed soon when we transition to the new macro-based erasure strategy. /cc @Chris-Hawblitzel

PR #403, opened 2023-02-28
add axioms that nonlinear arithmetic operations are in-bounds where applicable

tjhance:


tjhance:
As I explain in the comment, I wrote the minimal possible axioms I could think of, but I expect that the inequality-based axioms are cheap (in contrast to other nonlinear identities), so maybe we should consider adding more common inequalities.

Chris-Hawblitzel:
For EucDiv and EucMod, should the axioms use (< 0 y)?  I'd like to avoid overspecifying the uninterpreted case of y = 0 unless necessary.

Chris-Hawblitzel:
Yes, the question is whether `(a as int / b as int) == (a / b) as int` should be unconditionally true, or only true when `b != 0`.  If `a` and `b` have type `u8`, and `a == 100` and `b == 0`, then `(a as int / b as int) == (a / b) as int` implies that `0 <= 100int / 0int < 256`.  In other words, the `u8` type tells you a fact about `int` that could not have been derived by `int` operations alone.  This isn't wrong, but it's strange.

tjhance:
I made the requested change so I'm gonna go head and merge this. Let me know if you have any other concerns

Chris-Hawblitzel:
Thanks!

Issue #405, opened 2023-03-01
Weird perf behavior: load-bearing `verus!` reinvocation, load-bearing struct definition, ...


jaybosamiya:
This originally came up in a larger example where I was chasing down weird instability.

The issue: either of the two lines marked as "load bearing" below, if commented out, makes Verus do everything quite a bit faster:

```console
$ time ./rust-verify.sh ./foo.rs 2>/dev/null
verification results:: verified: 1 errors: 4
./rust-verify.sh ./foo.rs 2> /dev/null  3.99s user 0.09s system 99% cpu 4.082 total

$ time ./rust-verify.sh ./foo_comment_out_line_1.rs 2>/dev/null
verification results:: verified: 1 errors: 4
./rust-verify.sh ./foo_comment_out_line_1.rs 2> /dev/null  1.12s user 0.05s system 99% cpu 1.174 total

$ time ./rust-verify.sh ./foo_comment_out_line_2.rs 2>/dev/null
verification results:: verified: 1 errors: 4
./rust-verify.sh ./foo_comment_out_line_2.rs 2> /dev/null  1.15s user 0.05s system 99% cpu 1.207 total
```

However, you'll notice that the two lines themselves can/should have very little to do with the rest of the code.

Also, just before finally showing the code below: I know that the code below triggers a _lot_ of verification failures, but that was the only route I've found (at least so far) to demonstrate the load bearing nature of some things I believe should not be load bearing by themselves. More context below.

### The Code
```rs
mod pervasive;

use pervasive::prelude::*;
use pervasive::vec::*;

fn main() { }

verus!{
    spec fn just_some_spec_fn<A,C>(e: Seq<A>, b: C, f: FnSpec(C, A) -> C) ->(d: C)
        decreases e.len()
    {
        just_some_spec_fn(e, f(b, e[0]), f)
    }

    fn a_fn_whose_only_job_is_to_fail<A,C>(v: Vec<A>, f: impl Fn(A) -> C) ->(d: Vec<C>)
        ensures d.len() == v.len(),
        forall | h, i | 0 <= h && v@[h] == i == f.ensures((i,), d[h])
    {
        Vec::new()
    }

    spec fn nonsense(x0: u8, x1: u8, x2: u8, x3: u8, x4: u8, x5: u8, x6: u8, x7: u8) -> Seq <u8> {
        Seq::empty().push(x0).push(x1).push(x2).push(x7)
    }

    fn p()
[...]

PR #406, opened 2023-03-01
Automating the mundane: test case minimization

jaybosamiya:
This PR adds some scripts that are able to take large examples that demonstrate issues and reduce them down to much smaller test cases that demonstrate the same issue.

More specifically, these are "interestingness test" scripts (and documentation to go with them) for use with [C-Reduce](https://embed.cs.utah.edu/creduce/).

As an example use case, this was used to minimize the originally _much_ larger (and hard to manually minimize, because it is an instability related issue) test case for https://github.com/verus-lang/verus/issues/405

### DIFF
--- a/source/tools/minimizers/.gitignore
+++ b/source/tools/minimizers/.gitignore
@@ -1,0 +1,3 @@
+foo.rs
+foo.rs.orig
+stderr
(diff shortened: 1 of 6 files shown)


utaal:
Very cool, thanks!

utaal (approved):

Issue #400, opened 2023-02-27
--expand-errors blow-up behavior for conjunct chains


tjhance:
I'm seeing behavior where running --expand error on a conjunct-list gives AIR which is ... I think cubic in the # of conjuncts, though I'm not entirely sure. I'm pretty sure it's more than quadratic, though, and it blows up to 100,000s of lines for reasonable programs.

For example, for a simple conjunct like this:

```rust
spec fn boolfn(a: bool, b: bool, c: bool, d: bool, e: bool) -> bool {
    &&& a
    &&& b
    &&& c
    &&& d
    &&& e
}
```

We could generate AIR that looks like:

```
(assert a)
(assume a)
(assert b)
(assume b)
(assert c)
(assume c)
(assert d)
(assume d)
(assert e)
```

Instead we currently make:

```rust
(assert
   ("split postcondition failure" "from this function call")
   (%B (B a~2@))
  )
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (%B (B a~2@))
    (=>
     (%B (B a~2@))
     (=>
      (%B (B a~2@))
      (=>
       (%B (B a~2@))
       (%B (B b~4@))
  )))))
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (=>
     (%B (B a~2@))
     (=>
      (%B (B a~2@))
      (=>
       (%B (B a~2@))
       (=>
        (%B (B a~2@))
        (%B (B b~4@))
    ))))
    (=>
     (=>
      (%B (B a~2@))
      (=>
       (%B (B a~2@))
       (=>
        (%B (B a~2@))
        (%B (B b~4@))
     )))
     (=>
      (and
       (%B (B a~2@))
       (%B (B b~4@))
      )
      (%B (B c~6@))
  ))))
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (=>
     (=>
      (%B (B a~2@))
      (=>
       (%B (B a~2@))
       (=>
        (%B (B a~2@))
        (=>
         (%B (B a~2@))
         (%B (B b~4@))
     ))))
     (=>
      (=>
       (%B (B a~2@))
       (=>
        (%B (B a~2@))
        (=>
         (%B (B a~2@))
         (%B (B b~4@))
      )))
      (=>
       (and
        (%B (B a~2@))
        (%B (B b~4@))
       )
       (%B (B c~6@))
    )))
    (=>
     (and
      (and
       (%B (B a~2@))
[...]

PR #409, opened 2023-03-01
fix AIR blow-up in expand-errors

tjhance:
This addresses part of issue #400.

Right now split_expressions has a lot of redundant implication chaining. In particular, in this snippet,

```
let es1 = mk_chained_implies(es1);
let es2 = mk_chained_implies(es2);
let es2 = Arc::new(es2.iter().map(|e| mk_imply_traced(e1, e)).collect());
```

Only the last line is needed to get the desired chaining effect. We don't need to `mk_chained_implies` because `es1` and `es2` will already be chained.

After this PR, trying to verify this code with `--expand-errors`,

```rust

spec fn boolfn(a: bool, b: bool, c: bool, d: bool, e: bool) -> bool {
    &&& a
    &&& b
    &&& c
    &&& d
    &&& e
}

fn test(a: bool, b: bool, c: bool, d: bool, e: bool)
    ensures boolfn(a, b, c, d, e)
{
}
```

will produce AIR that looks like:
```
(check-valid
 (declare-const a~2@ Bool)
 (declare-const b~4@ Bool)
 (declare-const c~6@ Bool)
 (declare-const d~8@ Bool)
 (declare-const e~10@ Bool)
 (axiom fuel_defaults)
 (block
  (assert
   ("split postcondition failure" "from this function call")
   (%B (B a~2@))
  )
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (%B (B a~2@))
    (%B (B b~4@))
  ))
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (and
     (%B (B a~2@))
     (%B (B b~4@))
    )
    (%B (B c~6@))
  ))
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (and
     (and
      (%B (B a~2@))
      (%B (B b~4@))
     )
     (%B (B c~6@))
    )
    (%B (B d~8@))
  ))
  (assert
   ("split postcondition failure" "from this function call")
   (=>
    (and
     (and
      (and
       (%B (B a~2@))
       (%B (B b~4@))
      )
      (%B (B c~6@))
     )
     (%B (B d~8@))
    )
    (%B (B e~10@))
))))
```

### DIFF
--- a/source/vir/src/split_expression.rs
+++ b/source/vir/src/split_expression.rs
@@ -244,21 +244,6 @@ fn mk_imply_traced(e1: &Exp, e2: &TracedExp) -> TracedExp {
     TracedExpX::new(imply_exp, e2.trace.clone())
 }
 
-fn mk_chained_implies(es: TracedExps) -> TracedExps {
-    let mut chained_vec = vec![];
-    let mut chained_e = es.first().unwrap().clone();
-    // REVIEW: change encoding order ---- (A => B) => C to A => (B => C)
-    for (idx, e) in es.iter().enumerate() {
-        if idx == 0 {
-            chained_vec.push(chained_e.clone());
-        } else {
-            chained_e = mk_imply_traced(&chained_e.e, e);
-            chained_vec.push(chained_e.clone());
-        }
-    }
-    Arc::new(chained_vec)
-}
-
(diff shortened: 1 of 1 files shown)


chanheec (approved):
The change looks great! Thanks for looking into this issue!

PR #410, opened 2023-03-01
Print a warning if using a debug build

tchajed:
We now get a static message `warning: verus was compiled in debug mode` to warn people that the numbers are suspicious.

If someone has an opinion on the error message I'm happy to change it.

I couldn't find an easy way to get a Reporter and create a nice colored warning (at least at this point in the code).

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -31,9 +31,6 @@ fn run_compiler<'a, 'b>(
 
 fn print_verification_results(verifier: &Verifier) {
     if !verifier.encountered_vir_error {
-        if cfg!(debug_assertions) {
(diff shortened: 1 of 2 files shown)


utaal:
I'd rather we used the error reporting infrastructure for this.

tjhance:
Is the intent to only print this when using `--time`? Because from the diff, it looks like it's always going to print in debug mode.

utaal:
> Is the intent to only print this when using `--time`? Because from the diff, it looks like it's always going to print in debug mode.

I was about to ask the same, but then I thought it makes sense to have it enabled always, to warn the user that they're getting worse performance due to the debug build.

utaal (approved):
Thank you!

PR #416, opened 2023-03-02
README re-organization

parno:
Since we have been seeing an uptick in people interested in using Verus (rather than hacking on Verus itself), I'd like to suggest refactoring the landing page (README.md) to be a bit more user facing, similar to what the Amazon folks have done with [Dafny](https://github.com/dafny-lang/dafny) or the main [F* page](http://www.fstar-lang.org/).  I moved the rest of the content into INSTALL.md and CONTRIBUTING.md.  I think the README should also eventually link to some examples of projects using Verus.

Thoughts and suggestions are most welcome.  Special thanks to @chanheec for the animated gif.


tjhance:
Might be a good idea to link the 'getting started' page as well: https://verus-lang.github.io/verus/guide/getting_started.html

jaybosamiya:
I actually recommend adding a few more badges:  

Also fwiw, I think the choice between the new one vs the more conservative one is based on who the target audience is. For returning users, I don't think it matters too much as long as they know where something is, so just having text is sufficient, but for new users, having the gif and so on would actually bring people in. By having the added badges linked above, we make it easier for both kinds of users (including those who have _no_ familiarity with the tool).

parno:
@utaal Thanks for the alternate proposal.  I don't feel at all strongly about the badges; many GitHub projects seem to have them, but I'm fine without them.  I think the larger question is what experience we want people to have when they arrive at the page.  For the folks who already know and love verification and Verus, I suspect that asking them to click/bookmark the installation instructions is a relatively minor hurdle.  However, I think we all hope that we'll have a much larger pool of visitors who hear about Verus (at an academic conference, a Rust meet up, Hacker News, NY Times, etc.) and want to find out what it's all about.  Most of these visitors won't even really know what verification is all about, so I'm not sure they'll immediately want to work through all of our installation instructions.  Instead, I think it would help to clearly talk through what verification is, what it can do, and generally demystify the process (which I think the gif helps with).  Once they're intrigued by the idea, then we can point them to the guide and the installation instructions (or someday, an in-browser experience), so they can try it out.  This seems to be the approach that the Amazon folks have taken in their efforts to evangelize Dafny to all of the Amazon engineers, so that also offers some support for this approach.

utaal:
I don't feel strongly about badges, and I'm on board with including badges for quick links to the tutorial and `vstd` (currently pervasive) -- I believe a failure mode for GitHub READMEs is to end up being too busy, making it hard to find the right information, so I'd like to avoid redundant info that's not directly actionable, but badges like the ones @jaybosamiya is suggesting make sense to me.

utaal:
@parno thank you for the note.
My take is that we are still at a stage where Verus would be difficult to pick up to someone who's unfamiliar with verification. Nonetheless my proposed alternate change preserves some of your text that describes Verus, to communicate the purpose of the tool to someone unfamiliar. I think that's valuable.

I think the Amazon folks are at a different stage in the project, where most beginner users will have a relatively smooth experience with Dafny even if they stray away a bit from the tutorial. They have pre-built distributions of the tooling (and their IDE extensions) which means that getting set-up for the experience they show is very quick and straightforward. We are not there yet, but we are working on bridging that gap. I believe promising potential users that experience, to then immediately disappoint them because something subtle went wrong in the installation or we haven't tested on their platform, would result in frustration. (Note that setting up the VScode extension requires a number of non-trivial steps https://verus-lang.github.io/verus/guide/ide_support.html. That's totally fine and valuable at this stage, but at the same time we should not suggest that's the experience they can immediately expect in VScode, where most users are used to just installing an extension and having it automatically do any additional set-up it needs.)

The other change that we're debating is this text:
> Furthermore, Verus allows developers to go beyond the standard Rust type system and statically check the correctness of code that, for example, manipulates [raw pointers](https://verus-lang.github.io/verus/state_machines/examples/rc.html).
Verus supports a large and growing [set of Rust features](https://verus-lang.github.io/verus/guide/features.html). It also includes support for [user-defined specifications](https://verus-lang.github.io/verus/guide/spec_functions.html), which can be written in terms of pre-eixsting [specification libraries](https://verus-lang.github.io/verus/guide/spec_lib.html).

These would become the first four clickable links on the page, and I don't think they help a new user. The first is an example that requires the user to have familiarity with Verus, and have read the first part of the concurrency verification guide; the second is a link to a mostly empty page that links to what we've used as internal tracking of supported features and it isn't easy to consume by a new user; the third (spec functions) again points to a page that requires prior understanding of modes. If a user unfamiliar with verification or with at least some of Verus clicks on these, I believe they are more likely to decide that this is all way over their head, and leave, rather than continue reading by starting with the guide. This is in part an artefact of the fact that the documentation is still incomplete, but I believe we should make some more progress on that before we link specific pages from the top of the README.

tjhance:
> These would become the first four clickable links on the page, and I don't think they help a new user.

I agree here - I don't think these links are suitable for the introductory paragraph, especially not the first two.

Like I said above, I'd prioritize linking to "getting started".

(For supported features, it's not a bad idea, but the page needs to be cleaned up first.)

parno:
Thanks everyone for all of the constructive feedback (here and elsewhere) on the new design!  I've pushed some commits trying to incorporate it all.  Overall, I suspect people just arriving at the repo for the first time will be wondering some combination of:
1. What is Verus?

jaybosamiya:
Minor update on the badges "EDIT" that I had written ~4 days ago: turns out the badge shenanigans aren't just restricted to GitHub comments as I had initially thought; GitHub does some _fun_ caching stuff that makes the badge clicking not work as intended (i.e., requiring those two clicks that I had initially mentioned).

However, turns out this is easily fixed (now fixed on this branch) which gives us the nice only-one-click-needed badges (both in the README, and in comments: [Quick Start](https://verus-lang.github.io/verus/guide/getting_started.html) [Library Documentation](https://verus-lang.github.io/verus/verusdoc/lib/))

In hindsight, using this new approach is much more sensible anyways, and I don't know why the `shields.io` folks even have a `?link=...` parameter in the first place; maybe for non-GitHub use cases. Anyways, it is now a regular markdown link whose text is the image itself.

utaal:
Thanks for looking into this again. I personally think we still may want to wait on the picture (I shared that sketch I included as part of the conversation and not as a proposal per se), but I'd be okay with including it now that it is less prominent and doesn't commit as much to providing a smooth onboarding experience (as one may expect from a VScode extension as the main interface to the tool).

> If you want to try Verus, please be prepared to ask for help by filing GitHub issues or starting GitHub discussions.

That paragraph seems good to me; however please remove GitHub discussions:

> If you want to try Verus, please be prepared to ask for help by filing GitHub issues.

We have been using Issues for questions and to track user-facing issues and bugs (we have the "question" and "bug" tags), whereas Discussions contain feature requests/discussions and other more open-ended design conversations. The exception is a Discussion to track sub-optimal error messages in a way that doesn't generate a large number of GitHub issues: https://github.com/verus-lang/verus/discussions/387

As a minor layout nit, I'd prefer we used `##` headers to save some space.

parno:
Thanks for the additional feedback!  I've incorporated all of the text edits.  I think your hybrid picture turned out nicely, so I'd like to try it out for a bit.  If we get sick of looking at it or find it leads to too much confusion, we can definitely yank it :)

parno:
@utaal I'm happy to work on a new figure with Chris' changes, unless you've already started on it.

utaal:
@parno I have the source, so I can re-do the composite quickly. Can you send me an updated VScode screenshot?

utaal:
I've updated the demo image in the README with @Chris-Hawblitzel's feedback.

parno:
Great, thanks very much!

PR #417, opened 2023-03-02
from_seq for pervasive set

jonhnet:


parno (approved):

utaal (approved):
I made a minor change to the doctoring, please squash when you merge.

utaal (changes requested):
Nevermind, I take my approval back for the moment.
We just added https://github.com/verus-lang/verus/pull/393/files `.to_set()` to `Seq`. We probably don't need both for the time being.
(In the fullness of time we may want a `From` / `Into` equivalent for these conversions.)

utaal:
@jonhnet can we close this without merging? Is `Seq::to_set` sufficient?

PR #418, opened 2023-03-02
filter for pervasive seq

jonhnet:


parno:
It might be worth ensuring that the length of the filtered sequence is `<=` to the length of the original sequence.

The Dafny standard library also throws in [a lemma](https://github.com/dafny-lang/libraries/blob/dc95fdd0c95ce6af18af708109247f179e15b030/src/Collections/Sequences/Seq.dfy#L664) about filter distributing over concatenation, which might be nice to include here too.

parno:
Another thing to consider is whether `filter_lemma` should have a `#[verifier(broadcast_forall)]`, so that you get those properties for free when you use filter, rather than having to remember to invoke the lemma.  The [Dafny version](https://github.com/dafny-lang/libraries/blob/dc95fdd0c95ce6af18af708109247f179e15b030/src/Collections/Sequences/Seq.dfy#L654) does this by providing them as postconditions on filter itself.

parno:
> The Dafny standard library also throws in [a lemma](https://github.com/dafny-lang/libraries/blob/dc95fdd0c95ce6af18af708109247f179e15b030/src/Collections/Sequences/Seq.dfy#L664) about filter distributing over concatenation, which might be nice to include here too.

Btw, I'm happy to port this over if you'd like.

utaal:
> Another thing to consider is whether `filter_lemma` should have a `#[verifier(broadcast_forall)]`, so that you get those properties for free when you use filter, rather than having to remember to invoke the lemma. The [Dafny version](https://github.com/dafny-lang/libraries/blob/dc95fdd0c95ce6af18af708109247f179e15b030/src/Collections/Sequences/Seq.dfy#L654) does this by providing them as postconditions on filter itself.

This is consistent with https://github.com/verus-lang/verus/pull/393#discussion_r1120379748 and while it does introduce two quantifiers, they seem relatively safe (as they include `.filter(_)`).
/cc @Chris-Hawblitzel, opinions?

source/pervasive/seq_lib.rs
```diff
@@ -18,6 +18,54 @@ impl<A> Seq<A> {
...
+            // we don't keep anything bad
+            // TODO(andrea): recommends didn't catch this error, where i isn't known to be in
+            // self.filter(pred).len()
+            //forall |i: int| 0 <= i < self.len() ==> pred(#[trigger] self.filter(pred)[i]),
+            forall |i: int| 0 <= i < self.filter(pred).len() ==> pred(#[trigger] self.filter(pred)[i]),
+            // we keep everything we should
+            forall |i: int| 0 <= i < self.len() && pred(self[i])
+                ==> #[trigger] self.filter(pred).contains(self[i]),
+            // we also keep them in order, but I can't think of how to write
+            // that just now.
```
utaal, line 43: The Dafny `Filter` doesn't have this property either (thanks @parno for looking it up), so I'd say it's fine as it is, until we find out it's needed. Can you remove this comment?

Chris-Hawblitzel:
> I believe we need to add an additional lemma with the same signature and an empty body (with `#[verifier(external_body)]` for the broadcast_forall (as I believe it's currently not allowed to non-trusted proof fns.

That's correct.

utaal:
Can you add a couple of minimal tests to `tests/seqs.rs` like in this PR https://github.com/verus-lang/verus/pull/393/files ?

Chris-Hawblitzel:
I'd also recommend making `filter` opaque.

Chris-Hawblitzel:
> This is consistent with [#393 (comment)](https://github.com/verus-lang/verus/pull/393#discussion_r1120379748) and while it does introduce two quantifiers, they seem relatively safe (as they include `.filter(_)`).

It's probably fine, because the triggers on the forall `i` quantifiers are very specific.

jonhnet:
@parno I added the distribution lemma, which entailed adding a couple other distribution lemmas for push and drop_last.

@utaal What did you have in mind for the tests? This is a lib file; I've proven every ensures.

@Chris-Hawblitzel I added broadcast versions of the lemmas, but as you note they have to be axioms :v( Please take a look to see if the new foralls are too broad. Also, do broadcasts work from inside impls?

utaal:
I think a `#[broadcast_forall]` lemma is missing for `filter_lemma`?

I was thinking of something like:
```
            // TODO reveal_with_fuel(Seq::<int>::filter, 10);
            // TODO assert(s1.filter(|x: int| x <= 20).len() == 3);
            
            assert(s1.filter(|x: int| x <= 20).len() <= s1.len());
```

but it turns out the `reveal_with_fuel` generates invalid AIR.

utaal:
I added a `#[verifier(broadcast_forall)]` version of the `filter_lemma`, which I think was missing. And I added a couple of usage tests.

> @utaal What did you have in mind for the tests?

Testing call-sites and that broadcast_forall lemmas are being triggered as expected, like in my last commit.

utaal (approved):

jonhnet:
@utaal thanks for the explanation about testing broadcast, that makes sense. I added a comment so future me will remember why we need tests verified libraries. :v)

Issue #411, opened 2023-03-01
Verus panic when revealing state machine transition definition: "internal error: generated ill-typed AIR code: error 'error 'error 'use of undeclared variable"


tenzinhl:
Verus panicked with strange output when I revealed the `next` function auto-generated by the `state_machine!` macro inside of a lemma.

**Verus version**: built using commit hash e2f56d1b26629ecc09c3ae5d19ce817da2ea277f (on main branch, no extra build flags, so debug build)

### How to reproduce

Repo: https://github.com/vmware-labs/verified-betrfs/tree/tenzinhl-reveal-panic
tag: `tenzinhl-reveal-panic`

(Link links to tag)

1. Clone the repo
2. cd `Splinter/src`
3. `${verus_command} bundle.rs`

Produces a panicked output, here's a snippet of the beginning:

"thread 'rustc' panicked at 'internal error: generated ill-typed AIR code: error 'error 'error 'use of undeclared variable fuel%coordination_layer.AbstractJournal_v.AbstractJournal.State.next.' in expression 'fuel%coordination_layer.AbstractJournal_v.AbstractJournal.State.next.'' in expression '(fuel_bool fuel%coordination_layer.AbstractJournal_v.AbstractJournal.State.next.)'' in statement '...
[... an expanded what looks like z3 statement]
"

Panics because of line 432 in `Splinter/src/coordination_layer/CoordinationSystemRefinement_v.rs` which reveals the `next` and `next_by` predicates auto-generated for a state machine made using the `state_machine!` macro.

### Full output

Full error message is in this file.
[reveal_panic_verus_issue.txt](https://github.com/verus-lang/verus/files/10864411/reveal_panic_verus_issue.txt)

tenzinhl:
Found a more simplified reproduction. Seems like it's the specific `next` function that I'm trying to reveal that's causing problems for some reason (specifically `AbstractJournal::State::next`). Added a dummy function which calls the reveal, panic output is then much simpler. Code is tagged as `tenzinhl-reveal-panic-simpler` in the verified-betrfs repo.

Output here:
```
[tenzinhl@LAPTOP-U355A8BD:src]$ $verus --verify-module coordination_layer::CoordinationSystemRefinement_v bundle.rs
note: verifying module coordination_layer::CoordinationSystemRefinement_v

[...]

PR #419, opened 2023-03-03
fix fuel-related malformed AIR issues

tjhance:
fixes issue #411

### DIFF
--- a/source/rust_verify/tests/regression.rs
+++ b/source/rust_verify/tests/regression.rs
@@ -190,3 +190,36 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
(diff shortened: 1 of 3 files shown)


jaybosamiya:
Minor comment (please correct me if I've misunderstood something about this change): maybe it would be good to throw a warning rather than ignore such `reveal`s? If the `reveal` won't have any effect, then one possibility can be that the programmer intended to reveal something else that looked similar?

tjhance:
the warning is a good idea ... I added a warning, although it's not perfect, since our dependency analysis is at the module level, so you could still have a single function with a useless reveal.

Chris-Hawblitzel (approved):

Issue #386, opened 2023-02-22
bitvector invalid AIR


tjhance:
Both of the following programs are invalid AIR:

```rust
proof fn test() {
    let b: u8 = 5;
    let w: u64 = 19; 

    assert(w >> (b as u64) == 0u64) by (bit_vector);
}
```

```rust
proof fn test() {
    let b: u8 = 5;
    let w: u64 = 19; 

    assert(b >= 3u64) by(bit_vector);
}
```

tjhance:
the PR #420 fixes the first issue, but the second one remains

utaal:
[Verus all-hands]

Planned resolution: allow the second pattern by widening.

chanheec:
Currently, bit-vector operators do not like operands with different widths. 
`b >= 3u64` currently gives invalid AIR as `b` has a width 8, and `3u64` has a width 64.
As Andrea mentioned above, we could simply widen the smaller operand to the width of the bigger operand.

PR #420, opened 2023-03-03
fix width-extending casts for bitvectors

tjhance:
This fixes the first half of #386. The problem was that extending a bitvector from u8 to u64 required creating a bitvector who width was a non-power-of-2.


chanheec (approved):
Thanks for fixing this issue! 

Just to elaborate on what I learned:
To cast u8 to u64, Verus internally tried concatenating a zero bit-vector of width 56 with the original u8. However, the AIR printer did not like a bit-width that is not a power of 2. (Hence this PR allows that behavior.)

PR #422, opened 2023-03-03
Move rust_verify/tests into rust_verify_test/tests (to prepare for th…

Chris-Hawblitzel:
…e coming standard library crate in a future pull request).

This is the minimal amount of movement to enable a separate vstd crate.  In the future, we will probably also move the example directory somewhere else.


source/rust_verify_test/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+air = { path = "../air" }
+vir = { path = "../vir" }
+rust_verify = { path = "../rust_verify" }
+
+[dev-dependencies]
+rust_verify_test_macros = { path = "../rust_verify_test_macros" }
+regex = "1"
```
utaal, line 16: Where is `regex` used?

utaal (approved):

source/rust_verify_test/Cargo.toml
Chris-Hawblitzel: > Where is `regex` used?

It's used in `tests/examples.rs`

```
    use regex::Regex;
    let re = Regex::new(r"verification results:: verified: (\d+) errors: (\d+)").unwrap();
```

utaal: Ah, that makes sense, thanks!

PR #425, opened 2023-03-06
Add vstd crate and use it for most tests and examples

Chris-Hawblitzel:
The first commit in this pull request adds a vstd crate that is built and verified by default during "cargo build", based on a custom build script (build.rs).  The second commit ports most of the tests and examples to use this vstd crate.

Right now, these commits are written to avoid breaking any existing code, so that the vstd-based code can co-exist with projects that are still using the non-vstd pervasive module.  Soon, I would hope to move exclusively to the vstd-based approach, eliminating the old pervasive module, but that would be a future pull request.  We can also move files from pervasive to vstd and move the example directory in a future pull request.


source/vstd/build.rs
```diff
@@ -0,0 +1,62 @@
...
+    our_args.multiple_errors = 2;
+    our_args.erasure = rust_verify::config::Erasure::Macro;
+    our_args.export = Some(INSTALL_BIN_REL_PATH.to_string() + VSTD_VIR);
+    our_args.compile = true;
+    let file_loader = PervasiveFileLoader::new(Some(PERVASIVE_PATH.to_string()));
+    let verifier = Verifier::new(our_args);
+    let (_verifier, _stats, status) = rust_verify::driver::run(verifier, rustc_args, file_loader);
+    status.expect("failed to build vstd library");
+
+    println!("cargo:rerun-if-changed={PERVASIVE_PATH}");
```
utaal, line 61: Awesome.

source/vir/src/def.rs
```diff
@@ -579,24 +582,36 @@ fn atomicity_type_name(atomicity: InvAtomicity) -> Ident {
...
-        ]),
+        segments: Arc::new(if vstd_crate_name.is_some() {
+            vec![Arc::new("invariant".to_string()), atomicity_type_name(atomicity)]
+        } else {
+            vec![
+                Arc::new("pervasive".to_string()),
+                Arc::new("invariant".to_string()),
+                atomicity_type_name(atomicity),
+            ]
+        }),
```
utaal, line 593: Is the plan to track these branches (for removal) by removing the `vstd_crate_name` bindings when we drop support for the old-style pervasive?

source/rust_verify/src/driver.rs
```diff
@@ -134,6 +134,21 @@ pub(crate) fn run_with_erase_macro_compile(
...
+        "unused_variables",
+        "unreachable_patterns",
+        "unused_parens",
+        "unused_braces",
+        "dead_code",
+        "unreachable_code",
+        "unused_mut",
+        "unused_labels",
+        "unused_attributes",
+    ];
```
utaal, line 148: Are these something we think we'd need long-term or are they temporary?

source/vir/src/def.rs
Chris-Hawblitzel: Yes, the old code here will be removed when we drop the old-style pervasive.  Anything in `rust_verify` or `vir` that refers to pervasive would be removed.  We would not have to pass `vstd_crate_name` around everywhere and test for `vstd_crate_name.is_some()`.

source/rust_verify/src/driver.rs
Chris-Hawblitzel: I think this is long term, at least as long as we're running `rustc` twice (once to verify and once to compile).  The meaningful warnings are generated only on the first run of `rustc`.  The second run, for compiling erased code, generates spurious warnings about the erased code.

At least, this is the way it appears to work for some warnings, like `unreachable_code` and `unused_imports`.  Some other warnings, unfortunately, currently don't get emitted until very late (they appear only when we use `--compile`), so I don't know how to retain these without mixing them in with the spurious warnings from erasure.  Maybe there's some way to prod rustc to generate them earlier?

If in the future we have support for erasing ghost code directly in rust, with only one run of `rustc`, then this issue goes away.

utaal: Understood, thanks.

> Maybe there's some way to prod rustc to generate them earlier?

We may be able to override rust's queries (or run them earlier); it may be easier to do once we switch to a pre-built rustc with a custom driver.

tjhance (approved):

PR #426, opened 2023-03-06
fix mode error in assert-forall-by

tjhance:


Chris-Hawblitzel:
Should `check_expr_handle_mut_arg` return `Ok((mode, None))` in the `in_forall_stmt` case?

Chris-Hawblitzel (approved):

PR #430, opened 2023-03-07
Add Option::value synonym for get_Some_0

jonhnet:
I've now gotten a bit more experience with both the rust-y and Dafny styles of using Option in spec contexts, and I've found good evidence that the Dafny approach of test-and-dereference is more readable.

Consider these two examples in match style:
``` 
       &&& match *self.prior_rec {
            Some(rec) => {
                &&& rec.wf()
                &&& rec.message_seq.can_concat(self.message_seq)
            }
            None => true
       }
```

```
            ||| match *self.prior_rec { Some(rec) => rec.valid(boundary_lsn), _ => false }
```

Compare them to the more direct translation from the test-and-reference style:
```  
        &&& self.prior_rec.is_Some() ==> {
            &&& self.prior_rec.value().wf()
            &&& self.prior_rec.value().message_seq.can_concat(self.message_seq)
            }
```

```
            ||| self.prior_rec.is_Some() && self.prior_rec.value().valid(boundary_lsn)
```

Now it's easy to see that the first example was an implication and the second was a conjunct. In the match style, that boolean relationship is encoded in the default arm of the match, making it much more subtle.

We can write this today with `get_Some_0()`, but heavens is that cumbersome to read. To claw back a little parsimony, this PR proposes adding a `value()` synonym to pervasive Option to improve readability of the test-and-dereference style.

### DIFF
--- a/source/pervasive/option.rs
+++ b/source/pervasive/option.rs
@@ -47,12 +47,14 @@ impl<A> Option<A> {
     }
 
     // A more-readable synonym for get_Some_0().
-    pub open spec fn value(self) -> A
+    #[verifier(inline)]
+    pub open spec fn spec_unwrap(self) -> A
     recommends self.is_Some()
     {
         self.get_Some_0()
     }
 
+    #[verifier(when_used_as_spec(spec_unwrap))]
     pub fn unwrap(self) -> (a: A)
         requires
             self.is_Some(),
--- a/source/pervasive/option.rs
+++ b/source/pervasive/option.rs
@@ -46,6 +46,13 @@ impl<A> Option<A> {
         }
     }
 
+    // A more-readable synonym for get_Some_0().
+    pub open spec fn value(self) -> A
+    recommends self.is_Some()
+    {
+        self.get_Some_0()
+    }
+
     pub fn unwrap(self) -> (a: A)
         requires
             self.is_Some(),


jaybosamiya (approved):
The change seems fine to me; having `.value()` around for specs seems like a good idea (although good idea to also get @utaal's quick look before merging)

Sidenote: I believe once we have the real Rust `Option`, we'll want to revisit `is_Some` as well, since Rust already has an [`Option::is_some`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some) which we'd need to keep `exec` to prevent any surprises there, but having `is_Some` be the spec version of the same (modulo capitalization) is sure to cause annoyances. I wonder if we can have spec+exec methods. If we actually do end up having them, then `.value()` is kinda equivalent to `.unwrap()` (which is the exec version of the same). I suspect this is a much longer discussion to be had though, so don't block the merge of the commit on this sidenote.

utaal:
We've had discussions about general syntax for these cases, which I think we should still pursue, but I'm not against adding something like `.value()` for the time being.

I think we can name this `spec_unwrap` (for the reasons @jaybosamiya mentioned), and make it auto-spec to `spec_unwrap`, like for `Vec::Len` https://github.com/verus-lang/verus/blob/main/source/pervasive/vec.rs#L103-L110

That way you get to `unwrap` in both spec and exec code. It should probably also be `#[verifier(inline)]`.

jaybosamiya:
Oh I somehow completely missed the `#[verifier(when_used_as_spec(...)]`; that actually already makes things a lot cleaner than I had initially thought. This isn't exposed in the docs [example](https://verus-lang.github.io/verus/verusdoc/lib/pervasive/vec/struct.Vec.html#method.len) but probably should. Opened a discussion about it: https://github.com/verus-lang/verus/discussions/434

jonhnet:
@utaal: adding something like .value() for the time being. ... I think we can name this spec_unwrap

Should I take this as "also approved"?

utaal:
Yes, but please first rename to `spec_unwrap`, add `when_used_as_spec` to `unwrap`, and add `#[verifier(inline)]`. Let me know if I can help with those, or if something's unclear.

PR #431, opened 2023-03-07
Ghost blocks locals

tjhance:
Allow a 'tracked' or 'ghost' local variable declaration in 'exec' code, as long as it not _initialized_ in exec code.

When proof blocks are erased, this will leave a single unused, uninitialized 'let' binding. With the existing macro-erasure scheme, the macro erases such 'let' bindings in order to avoid type errors; in a ghost-blocks future, an unused 'let' statement would just get optimized out.

A lot of the concurrency code is currently written in a style similar to this, so this change should make it easier to port that code to use the ghost-blocks restrictions.

I think this will probably let us simplify some of the pattern expansion code in the macro, but I haven't attempted to do that yet. I also think we can do "auto-unwrapping" of arguments so that arguments and locals can be handled uniformly (I made a janky prototype of this but it's not ready).

Specifically this PR:

 * Loosens the restriction on variable declarations in exec code
 * Adds a restriction on variable _uses_ to make sure such variables are never read or assigned in exec-mode
   * This restriction caused some problems with invariant-blocks. I didn't attempt to figure out fully how erasure works for invariant blocks, so as a temporary workaround, I had modes.rs count the invariant block argument as a 'ghost block'.
 * Removes cell_old_style and ptr_old_style. Those were explicitly written to not use the ghost-blocks style, so it's time we got rid of them.
 * Adds `let ghost x;` to go along with the existing `let tracked x;` syntax. Previously, this wouldn't have had any use, but it does now that it makes sense to declare a ghost variable in exec code.


Chris-Hawblitzel (approved):
One comment: I don't think it's strictly necessary, but it would be good if `modes.rs` can also catch assignments to non-exec variables in exec code:

```
    fn test() {
        let ghost mut x: u64;
        x = 0;
    }
```

utaal:
I'm deep in some code and would rather not context-switch right now. I trust you on this, and will try and take a look later; as far as I'm concerned, feel free to merge.

utaal:
I have one comment: can you please add an example or two on how to use this to `examples/syntax.rs`? That is effectively our most complete documentation on the syntax right now, and we should keep it up to date.

PR #438, opened 2023-03-09
Port to rustc 1.68.0 and drop dependency on compiler fork

utaal:
This mainly removes the dependency on our fork of `rustc` and uses Rust 1.68.0 instead.

That has various implications:
* removes support and code paths for ast-based erasure;
* changes the cargo dependencies and set up to use a precompiled rust;
* changes the lifetime checker and test harness to spawn a separate process because it is no longer possible to capture rust's output programmatically (this relies on stderr redirection instead);
* adapts verus to changes in the compiler API (HIR ast, and other);
* adds a cargo wrapper to handle our build process.

To run tests, from the repository root (in bash):

```
. ./tools/activate # also builds vargo
cd source; vargo test -p rust_verify_test
```

There are also scripts to source for fish `tools/activate.fish` and Windows `tools/activate.bat`.

Not in this PR:
* remove some of the unused code for deprecated syntax and erasure mechanisms;
* get all the remaining ignored tests to pass (they are marked with `TODO(main_new)`);
* determine if it's possible to avoid importing rustc_interface and rustc_driver;
* fix the remaining TODOs (`TODO(main_new)`, `REVIEW(main_new)`);
* update CI to use the self-hosted runner.


source/rust_verify_test/tests/state_machines.rs
```diff
@@ -5999,7 +5999,7 @@ test_verify_one_file! {
                 }
             }
         }}
-    } => Err(e) => assert_error_msg(e, "the first param to a 'readonly'")
```
tjhance, line 6002: what's the reason for this change?

these errors all occur during the macro phase, so they aren't technically "VIR errors". What does `assert_vir_error_msg` actually check, now? Should it be renamed?

tjhance: ok, great, that distinction makes sense 👍

utaal: we can no longer rely on getting the errors through a memory object, as we spawn `rust_verify` as a process for each test, so I had to make the distinction based on whether diagnostics carried an error code

Chris-Hawblitzel: Can we use both the error code and a message?  I find things like `assert_rust_error_code(err, "E0308")` harder to read and less specific than `assert_error_msg(err, "cannot borrow 'x' as mutable more than once at a time")`.

Chris-Hawblitzel:
Did it turn out to be necessary to include a fork of `rustc_driver` and `rustc_interface`?  This is about 5000 lines of code that we'd be forking and would have to maintain.

source/rust_verify_test/tests/cell_lib.rs
```diff
@@ -289,7 +289,7 @@ test_verify_one_file! {
         pub fn foo(r: cell::PointsTo<Foo>) {
             f(r);
         }
-    } => Err(e) => assert_error_msg(e, "the trait `std::marker::Sync` is not implemented for `Foo`")
+    } => Err(e) => assert_rust_error_code(e, "E0277")
```
Chris-Hawblitzel, line 292: I mentioned this in a different comment, but I like including the string "the trait `std::marker::Sync` is not implemented for `Foo`" in the error assertion.

Chris-Hawblitzel (approved):

tjhance (approved):
I checked out the branch, and I was able to build & run the tests successfully.

I noticed that:
 - The installation instructions need to be updated
 - Building produces a significant # of warnings

but i think we should go ahead and get this merged. i want to start working in it full time so we can start ironing out the kinks.

source/rust_verify_test/tests/state_machines.rs
utaal: I switched back to just the message. If we prefer to have both, I'm happy to re-add them after I merge.

utaal:
> Did it turn out to be necessary to include a fork of rustc_driver and rustc_interface? This is about 5000 lines of code that we'd be forking and would have to maintain.

The branch currently uses a one-line change in `rustc_interface` to access the name of the crate; if we can find another way to access that, we can drop both.
Are we okay with merging these in, and dropping them if we find an alternative?

Chris-Hawblitzel:
> Are we okay with merging these in, and dropping them if we find an alternative?

Yes, that's fine.  Go ahead.

PR #440, opened 2023-03-13
Revised ghost syntax

Chris-Hawblitzel:
This pull request implements the following syntax for `ghost`, `tracked`, `Ghost`, and `Tracked`:
- exec-mode `let ghost pat = expr;` and `let tracked pat = expr;`
- `Ghost(expr)` and `Tracked(expr)` expressions (replacing the old `ghost(expr)` and `tracked(expr)` expressions)
- `Ghost(x)` and `Tracked(x)` patterns in `let` statements
- `Ghost(x)` and `Tracked(x)` function parameters for exec functions

First, thanks to https://github.com/verus-lang/verus/pull/431 , exec local variables can be declared `ghost` or `tracked`.  This pull request adds patterns and initializers for such declarations.  The patterns and initializers are turned into proof blocks that write to the declared variables.  This means that instead of having to use `Ghost` wrappers inside an exec function:

```
fn binary_search(v: &Vec<u64>, k: u64) -> (r: usize) {
...
        let d: Ghost<int> = ghost(i2 - i1);
...
        assert(i2 - i1 < d@);
...
}
```

you can just write:

```
fn binary_search(v: &Vec<u64>, k: u64) -> (r: usize) {
...
        let ghost d = i2 - i1;
...
        assert(i2 - i1 < d);
...
}
```

The initializers for `let ghost` and `let tracked` are automatically parsed as ghost code rather than exec code, so `i2 - i1` is ghost integer subtraction here.

Second, this pull request switches to a uniform constructor-style syntax for constructing and deconstructing `Ghost<T>` and `Tracked<T>` values, so that `Ghost(expr)` and `Tracked(expr)` expressions construct `Ghost<T>` and `Tracked<T>` values and `Ghost(x)` and `Tracked(x)` patterns deconstruct them.  For example:

```
fn test() {
    let ghost x: int = 1000000000000000000000000000000000000000000000000000000000000000000int;
    let g: Ghost<int> = Ghost(x - 1);
    let Ghost(y) = g;
    assert(y < x);
}
```

Third, this pull request adds syntactic sugar for deconstructing exec function parameters of type `Ghost<T>` and `Tracked<T>`.  Specifically, you can use `Ghost(x)` and `Tracked(x)` patterns in exec function parameters, and the wrapped `Ghost<T>` or `Tracked<T>` value is unwrapped into a value `x` of type `T`, both in the function's specification and body:

```
        struct S(int);

        fn test1(Ghost(g): Ghost<int>, Tracked(t): Tracked<S>)
            requires
                g > 100,
                t.0 > 200,
        {
            assert(g >= 100);
            assert(t.0 >= 200);
        }

        fn test2(Ghost(g): Ghost<int>, Tracked(t): Tracked<S>)
            requires
                g > 100,
                t.0 > 200,
        {
            test1(Ghost(g), Tracked(t));
        }

        fn test3(g: Ghost<int>, t: Tracked<S>)
            requires
                g@ > 100,
                t@.0 > 200,
        {
            test1(g, t);
        }
```

(Note: it would also be possible to add more syntactic sugar to make the declarations more concise; for example:
```
        fn test1(Ghost(g: int), Tracked(t: S))
        fn test1(Ghost g: int, Tracked t: S)
        fn test1(g@: Ghost<int>, t@: Tracked<S>)
```
However, although `Ghost(g): Ghost<int>` is a little harder to write and takes up a little more space, I think it will be easier for newcomers to read, since it uses ordinary Rust types and patterns to document the parameter type `Ghost<int>` and the conversion from `Ghost<T>` to `T` via pattern matching syntax.)


parno:
Does this affect how we declare `ghost` members of a struct?  Currently I'm using the `x: Ghost<T>` style.

Chris-Hawblitzel:
> Does this affect how we declare `ghost` members of a struct? Currently I'm using the `x: Ghost<T>` style.

It does not affect how structs are declared.  The `x: Ghost<T>` style is still necessary for `exec` structs and we don't have any better ideas at the moment.  The only change that you might notice is that you can now pattern match inside a `Ghost` inside a struct pattern in an exec `let`.  For example: `let S { foo: Ghost(x), bar: y } = expr;`.

The `ghost x: T` style is also available (and sometimes useful) in `ghost` and `tracked` datatypes.  Specifically, distinguishing `tracked x: T` and `ghost x: T` fields is useful in `tracked` datatypes.  In `exec` datatype values, there's no way to access `ghost x: T` and `tracked x: T` fields, so in the near future we're thinking of a small change in a future pull request that would make `ghost x: T` and `tracked x: T` fields illegal except in `ghost` and `tracked` datatypes.  This would enforce that you have to use the `x: Ghost<T>` style for `exec` datatypes (since the other style won't work and would lead to confusing error messages).

parno:
Got it -- thanks for the clarification!

tjhance (approved):
great, this is gonna be a huge improvement.

I'm still not totally satisfied with `Ghost(x) : Ghost<...>` arguments, though I get the rationale.

Does the unwrapping work with `&mut` arguments? e.g., can you do: `Tracked(x) : Tracked<&mut X>`? This is the thing that gave me the most trouble when I tried to do this before.

Chris-Hawblitzel:
> Does the unwrapping work with `&mut` arguments? e.g., can you do: `Tracked(x) : Tracked<&mut X>`? This is the thing that gave me the most trouble when I tried to do this before.

I'm not sure.  The issue at the moment is that we don't support any types of the form `Foo<&mut X>` yet, including `Tracked<&mut X>`.  So we'd need to add support for `Tracked<&mut X>` first.

source/builtin_macros/src/syntax.rs
```diff
@@ -164,6 +175,40 @@ impl Visitor {
...
             }
+
+            // Check for Ghost(x) or Tracked(x) argument
+            use syn_verus::PatType;
+            if let FnArgKind::Typed(PatType { pat: box pat, .. }) = &mut arg.kind {
+                let mut tracked_wrapper = false;
+                let mut wrapped_pat_id = None;
+                if let Pat::TupleStruct(tup) = &pat {
+                    let ghost_wrapper = path_is_ident(&tup.path, "Ghost");
+                    tracked_wrapper = path_is_ident(&tup.path, "Tracked");
```
utaal, line 186: I imagine we are considering these (here and elsewhere) to be keywords, rather than constructor names in patterns, right? By that I mean: if someone hides `Ghost` or `Tracked` with a definition/import, we still treat these as Ghost/Tracked "patterns".

source/builtin_macros/src/syntax.rs
```diff
@@ -164,6 +175,40 @@ impl Visitor {
...
+                    use syn_verus::parse_quote_spanned;
+                    *pat = wrapped_pat;
+                    let x = wrapped_id.ident;
+                    let span = pat.span();
+                    view_ghost_tracked.push(stmt_with_semi!(span => let #x = #x.view()));
+                    // Note: we let subsequent processing desugar unwrap_ghost_tracked
+                    if is_trait {
+                        // do nothing
+                    } else if tracked_wrapper {
+                        let stmt = parse_quote_spanned!(span => let Tracked(#x) = #x;);
```
utaal, line 203: For the record, there's a pattern feature on arguments now: `fn foo(Constructor { a }: Type)`, however we still need custom desugaring for this, so I don't think it's applicable here.

source/builtin_macros/src/syntax.rs
Chris-Hawblitzel: Yes, they are treated as keywords (or at least "soft keywords") rather than resolved names.

PR #443, opened 2023-03-14
Introducing calc!

jaybosamiya:
I assume people reading this already know of `calc` in Dafny, F*, ..., so I am not going to motivate the "why" of us wanting calc. The guide should eventually have such a section though, but that'll be a different PR when I get some spare cycles.

The calc introduced by this PR is implemented as a macro (thus, `calc!`). 

What'll likely be the most commonly used syntax for `calc!` looks like:
```rust
calc! {
  (R)
  a_1; { /* proof that a_1 R a_2 */ }
  a_2; { /* proof that a_2 R a_3 */ }
   ...
  a_n;
}
```

In particular, this means `a_1 R a_n` via a series of proofs `a_1 R a_2`, ... `a_{n-1} R a_n`, for some transitive relation `R` (currently, `calc!` supports `==`, `<`, `<=`, `>`, `>=`).

The syntax is a bit more flexible though. In particular, rather than `a_i R a_{i+1}`, you can instead also utilize a compatible relation `Ri` to prove `a_i Ri a_{i+1}` using syntax that looks like `a_i; (Ri) { /* proof that a_i Ri a_{i+1} */ }` (i.e., skipping the `(Ri)`s is provided as a convenience in case all your `Ri`s are the same as `R`; but if you wish to, you can explicitly specify `Ri` everywhere). Specifying `Ri`s along the way can help with proof readability (for example, when proving `foo <= bar`, you may wish to do `foo == baz <= foobar < bar` or such; in theory, you _could_ do this by setting them all to be `<=`, but conveying intent, and checking that each of those steps follow that intent can lead to more readable code). See the examples and tests for more.

Note that this syntax for calc is inspired almost directly by that in F*; the relation `R` there goes _immediately_ after the `calc` keyword, rather than the first member of the body. This is a limitation with implementing `calc` as a macro, any while this _could_ be overcome by making `calc` be a keyword, that is left as out of scope for this PR (also I don't believe that making it exactly the same as there is crucial).

The macro itself is written to easily support expanding the set of `R`s that it supports. As a conscious decision though, to make sure that error reporting is nicer (things like an accidental `>` inside a `<=` chain are checked), we do not support having entirely _arbitrary_ `R`s, but instead it is strictly an allow-list.

Currently one downside to the current implementation of `calc` is that it does not provide fully localized error reporting (it currently ends up localizing onto the macro itself). I did attempt using `#[verifier(custom_err(...))]` to improve error reporting, but I was unable to get that working just yet. Imho, we should not block the introduction of `calc!` on this downside though though, since one can still localize errors via manually commenting out things and/or introduction of `assume(false)`s. Once we have better `#[verifier(custom_err(...))]` or similar, we can easily change the internals of the macro to just tag the relevant error with the right message, and I've modularized the macro enough to support the easy addition of this when we can do this better.


Chris-Hawblitzel:
I wonder why `(2 as int)` is necessary in the `calc.rs` example.  I would think that `2int` should work, but it reports `invalid suffix`.  For some of our other macros, `2int` does work, as in `assert_seqs_equal!(seq![2int], seq![2int]);`.

tjhance:
re: the `custom_err`, right now it only does anything if --expand-errors is on. So I wonder if it would work if you try that.

jaybosamiya:
@Chris-Hawblitzel

As far as I understand it, the `2int` difference is because the `seq!` or `assert_seqs_equal!` macros immediately take `tt*` of their entire input and immediately pass it into `verus_proof_macro_exprs!`, while in contrast, `calc!` (to be able to support the syntax of being able to add the intermediate relations and such) needs to parse things as an `expr` early on -- see https://github.com/verus-lang/verus/pull/443/files#diff-eccdbfc06c71abf91b466dfe10ea93b3b9cc13b1326900cdfd1c519230931406R149

We could (hypothetically) handle this by either requiring more parentheses (each of `a_1`, ... `a_n` would _require_ parens around them) which I don't think is particularly nice syntax; the alternative is to move more of the `calc` parsing into a procedural macro that can then take and work with the exact syntax we want (this would require a non-trivial rewrite of the whole macro), or alternatively, we would make our own `verus_macro_rules!` macro that gives us the right expr and so on parsing directly, rather than needing the `verus_proof_macro_exprs`/`verus_proof_macro_explicit_exprs`/etc hoops (imho this is the most principled approach, and would also clean up the `seq`/`set`/`assert_seqs_equal`/... macros too; however, I don't know how much effort this would be to actually implement/do; I suspect this would not be easy though).

I will give it some more thought to see if we can do the last of the above alternatives in some convenient easy-to-maintain way, but I suspect there isn't an easy solution there.

---

@tjhance 

Ah, I see; indeed it does support the custom error when `--expand-errors` is enabled. However, it doesn't support a dynamically generated `custom_err` (in particular, I am unable to do something like `#[verifier(custom_err(concat!("Fail: ", stringify!($a), " == ", stringify!($b))))]`, since it doesn't seem to support macro expansion at that position). Without dynamically generated custom error, having a custom error is not any more useful than what currently shows up.

Also, I am not sure we'd ever be in a scenario where the output _without_ the custom error message for a `calc` is ever more useful than the custom one, so we probably also need a way to always have the custom message added, even without `--expand-errors`, but adding something like `always_custom_err` or such is prob a separate discussion?

tjhance:
I don't know if this will help at all, but I had a similar problem for the atomic_ghost macro. What I did there was add a proc-macro to do all the parsing by calling the verus parser, then spat it all back out in the expanded form to be handled by another macro.

main macro code: https://github.com/verus-lang/verus/blob/6e70e7a940de7c7940d12cdc192d0623304dadf8/source/pervasive/atomic_ghost.rs#L256

proc-macro helper: https://github.com/verus-lang/verus/blob/main/source/builtin_macros/src/atomic_ghost.rs

> in particular, I am unable to do something like #[verifier(custom_err(concat!("Fail: ", stringify!($a), " == ", stringify!($b))))], since it doesn't seem to support macro expansion at that position

I see, that's difficult. I guess Rust doesn't expand macros in that position?

Perhaps we need to replace the attribute with a builtin function that takes a string literal.

Chris-Hawblitzel (approved):
It sounds like there may be ways in the future to integrate `calc` more closely with the rest of the `verus!` syntax.  But regardless of that, the current `calc!` implementation in this pull request already looks very useful.

utaal (approved):
Cool stuff. I'd also like to see this more closely integrated into Verus' syntax at some point, but for now this is definitely useful; and we can start getting some experience with calc-as-a-macro, which may inform the syntax design.

jaybosamiya:
Thanks for the PR reviews :)

I agree that closer integration with Verus's syntax would be nice, and Travis's pointer for `atomic_ghost` is indeed useful as a sort of template to achieve it, but even without that, indeed as both Chris and Andrea you've mentioned, the calc macro is still usable without it, so we should probably leave the closer integration for a future PR (fwiw, as I understand it, the main change that'll happen after that is that more syntax would be allowed, compared to right now, and thus that future change should likely be backwards compatible anyways).

I'll merge this branch soon, I'll just kick off a merge of `main` into this branch first to make sure nothing has been impacted by the changes on `main` and if it succeeds, I'll squash-and-merge.

PR #448, opened 2023-03-16
Add support for Ghost<&mut T> and Tracked<&mut T> parameters

Chris-Hawblitzel:
This revamps the treatment of `Ghost(x)` and `Tracked(x)` function parameters from https://github.com/verus-lang/verus/pull/440 so that `Ghost(x): Ghost<&mut T>` and `Tracked(x): Tracked<&mut T>` is supported.  For example:

```
        struct S(int);

        fn test1(Ghost(g): Ghost<&mut int>, Tracked(t): Tracked<&mut S>)
            requires
                *old(g) > 100,
                old(t).0 > 100,
            ensures
                *g == *old(g) + *old(g),
                *g > 200,
                *t == *old(t),
        {
            assert(*g >= 100);
            proof {
                *g = *g * 2;
            }
        }

        fn test2(Tracked(t1): Tracked<S>) {
            assume(t1.0 > 100);
            let tracked mut t2 = t1;
            let ghost mut i = 1000;
            assert(i == 1000);
            test1(Ghost(&mut i), Tracked(&mut t2));
            assert(i == 2000);
            assert(t2.0 > 100);
        }
```

The new treatment of `Ghost(x)` and `Tracked(x)` function parameters is not just syntactic sugar, but instead introduces a new internal header `header_unwrap_parameter` so that:

```
        fn test1(Ghost(g): Ghost<&mut int>) {
        }
```

is desugared by `syntax.rs` into:

```
        fn test1(tmp_g: Ghost<&mut int>) {
            #[verus::internal(header_unwrap_parameter)] let g;
            #[verifier(proof_block)] { g = tmp_g.view() };
        }
```

`rust_to_vir_func.rs` reads this header and changes the parameter name from `tmp_g` back to `g`, but also annotates `g` with additional mode information so that `g` has mode `exec` externally to the function but has mode `spec` inside the function, including in both the body and the requires/ensures/decreases.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -287,6 +287,12 @@ impl<A> Tracked<A> {
         unimplemented!()
     }
 
+    #[verifier::proof]
+    #[verifier(external_body)]
+    pub fn new(#[verifier::proof] _a: A) -> Tracked<A> {
+        Tracked { phantom: PhantomData }
+    }
+
     #[doc(hidden)]
     #[verifier(external)]
     #[inline(always)]
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -1109,6 +1109,8 @@ impl VisitMut for Visitor {
                     let inner = take_expr(&mut call.args[0]);
                     *expr = Expr::Verbatim(if self.erase_ghost {
                         quote_spanned!(span => Tracked::assume_new())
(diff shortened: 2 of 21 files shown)


tjhance (approved):

utaal (approved):
Looks good to me.

Can you please add a couple usage examples to `examples/syntax.rs`? I imagine we'll refer to that file when updating/adding syntax documentation in the guide.

PR #450, opened 2023-03-18
Replace "assert forall" syntax with assert-by-suppose syntax

Chris-Hawblitzel:
Replace the old `assert forall` syntax:

```
/// "assert forall by" may be used to prove foralls:
fn test_assert_forall_by() {
    assert forall|x: int, y: int| f1(x) + f1(y) == x + y + 2 by {
        reveal(f1);
    }
    assert(f1(1) + f1(2) == 5);
    assert(f1(3) + f1(4) == 9);

    // to prove forall|...| P ==> Q, write assert forall|...| P implies Q by {...}
    assert forall|x: int| x < 10 implies f1(x) < 11 by {
        assert(x < 10);
        reveal(f1);
        assert(f1(x) < 11);
    }
    assert(f1(3) < 11);
}
```

with new `assert(forall...) by(suppose)` syntax:

```
/// "assert(forall...) by(suppose)" may be used to prove foralls:
fn test_assert_forall_by() {
    assert(forall|x: int, y: int| f1(x) + f1(y) == x + y + 2) by(suppose) {
        reveal(f1);
    }
    assert(f1(1) + f1(2) == 5);
    assert(f1(3) + f1(4) == 9);

    // to prove forall|...| P ==> Q, write assert(forall|...| P ==> Q) by(suppose) {...}
    assert(forall|x: int| x < 10 ==> f1(x) < 11) by(suppose) {
        assert(x < 10);
        reveal(f1);
        assert(f1(x) < 11);
    }
    assert(f1(3) < 11);
}
```

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -995,6 +995,45 @@ impl Visitor {
     }
 }
 
+fn parse_suppose(
+    mut arg: Expr,
+) -> Result<(Punctuated<Pat, Token![,]>, Option<Expr>, Expr), String> {
+    // Parse forall|bound_vars| implies ==> ... ==> implies ==> rhs
+    while let Expr::Paren(e) = arg {
+        arg = *e.expr;
+    }
+    let (bound_vars, mut arg) =
+        if let Expr::Unary(ExprUnary { op: UnOp::Forall(..), expr, .. }) = arg {
+            if let Expr::Closure(syn_verus::ExprClosure { inputs, body, .. }) = *expr {
+                (inputs, *body)
+            } else {
+                return Err("expected closure in forall".to_string());
+            }
+        } else {
+            (Punctuated::new(), arg)
+        };
+    let mut implies: Option<Expr> = None;
+    loop {
+        while let Expr::Paren(e) = arg {
+            arg = *e.expr;
+        }
+        let span = arg.span();
+        if let Expr::Binary(ExprBinary { op: BinOp::Imply(..), left, right, .. }) = arg {
+            if let Some(imp) = implies {
+                let op = BinOp::And(Token![&&](span));
+                let bin = ExprBinary { attrs: vec![], op, left: Box::new(imp), right: left };
+                implies = Some(Expr::Binary(bin));
+            } else {
+                implies = Some(*left);
+            }
+            arg = *right;
+        } else {
+            break;
+        }
+    }
+    Ok((bound_vars, implies, arg))
+}
+
 impl VisitMut for Visitor {
     fn visit_expr_mut(&mut self, expr: &mut Expr) {
         if self.chain_operators(expr) {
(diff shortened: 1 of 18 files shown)


utaal:
In the second example, the top level `==>` in `forall|x: int| x < 10 ==> f1(x) < 11` has a special syntactic role: it separates the antecedent, which becomes an assumption, from the consequent, which becomes the proof obligation of the block. I personally think this feels a bit inconsistent with other rust / verus syntax by being more implicit, however it sounds like the Verus teams' consensus prefers this style.

Chris-Hawblitzel:
Indeed, we also discussed more explicit syntax for handling the `==>`, such as:

```
assert(forall|x: int| x < 10 ==> f1(x) < 11) by(forall, implies) { ... }
```

or

```
assert(forall|x: int| x < 10 ==> f1(x) < 11) by(forall, ==>) { ... }
```

This would be just as easy to implement; it's mainly a question of taste as to what syntax people liked when we discussed it a while back.

utaal:
Let me expand on my comment here as well (in addition to the private thread), as the conversation moved here:

When I first looked at the PR I was surprised that we are treating ==> implicitly, in particular I confused myself when looking at these two examples (which I believe to be common use cases):

```
assert(forall|x: int, y: int| f1(x) + f1(y) == x + y + 2) by(suppose) {
assert(forall|x: int| x < 10 ==> f1(x) < 11) by(suppose) {
```

This pattern matches in a way that (to me, and -- I worry -- to new users) looks like `==>` and `==` end up having a similar role, which is (in retrospect, obviously) incorrect.
suppose ended up being the consensus last time due to wanting to avoid things like intro and `==>` , the reasoning being that (among other things) we fear "intro" would be scary to users unfamiliar with theorem provers.

I think, for a similar reason, we shouldn't surprise users by moving the antecedents into the proof context for `==>` implicitly, and have a distinction between just variable introduction (for forall|vars| expr without `==>`) and variable introduction plus assumption of the antecedents (for `forall|vars| expr1 ==> expr2`).

Chris had suggested `by(intro)` for the former and `by(intro, ==>)` for the latter. I'm okay with avoiding intro, but I think we should strongly consider `by(forall)` and ` by(forall, ==>)` (or `by(forall, suppose)`) to distinguish the two "modes".

utaal:
To summarize the discussion around whether to have an explicit “mode” for just variable introduction and for assuming the antecedent, my understanding is than in the conversation so far everyone would be in favor (or not against) either `by(forall, ==>)` or `by(forall, implies)` which enables both, and `by(forall)` for just variable introduction. @parno suggests `choose` as an alternative to `forall`.
I think using `forall` and `==>` may be the simplest as it doesn’t introduce any new keyword, and matches the preceding syntax.

utaal:
Recording offline conversation with @verus-lang/devs and others: it turns out this syntax question is related to a bigger question regarding `by (_)` and support for tactics and strategies. Because of that, we are going to stick with the existing syntax for the time being, and re-consider after we have moved to `main_new`.

jaybosamiya-ms:
Just adding a quick link #1302 to aid in keeping links of the various conversations that eventually led to the PR.

PR #451, opened 2023-03-18
move bit-vector operator checking step to be consistent with syntax change

chanheec:
During verus! macro expansion with bit-vector mode, we check if the user uses `+` or `*` instead of `add` or `mul`. 

This check produces this error message to guide the user.
```
Inside bit-vector assertion, use `add` `sub` `mul` for fixed-bit operators, instead of `+` `-` `*`. (see the functions builtin::add(left, right), builtin::sub(left, right), and builtin::mul(left, right))"));
```

As we plan to deprecate `#[verifier(bit_vector)]` and instead use `fn ...... by (bit_vector)` (a38dfae), this PR moves the above checking step into the `by(bit_vector)` path.


utaal:
As I believe this is a breaking change (`#[verifier(bit_vector)]` functions that haven't been updated to syntax would be handled incorrectly after this), I've changed the base to `main_new`; let me rebase for you. We should also add a TODO to reject `#[verifier(bit_vector)]` before we switch to `main_new`.

utaal (approved):

utaal:
Merged as https://github.com/verus-lang/verus/commit/a11d804e82935388955740327a8a694989872633

PR #453, opened 2023-03-20
Add functionality to convert to/from little-endian

jaybosamiya:
Currently I've only added support for `u64` but we may wish to expand on this in the future.

Also, once we have support for Rust fixed-size arrays, we'll be able to give the same interface as Rust's standard library, like https://doc.rust-lang.org/std/primitive.u64.html#method.from_le_bytes

Finally, I am not 100% sure of whether I've picked the right files to place each of these (or whether we want to make a new file/module called `integers` or `integer_prims` such, that should hold all of these in one place.


Chris-Hawblitzel:
I'd prefer these in their own file, like `bytes.rs` or something.

jaybosamiya:
Fair point, moved it all over to a separate `bytes.rs`

Chris-Hawblitzel (approved):

PR #465, opened 2023-03-22
pervasive: adding exec methods for is_some()/is_none()

achreto:
This PR adds exec-mode methods for `is_some()` and `is_none()` for the Option type.

### DIFF
--- a/source/pervasive/option.rs
+++ b/source/pervasive/option.rs
@@ -35,6 +35,7 @@ impl<A> Option<A> {
         }
     }
 
+    #[inline(always)]
     pub const fn is_some(&self) -> (res: bool)
         ensures res <==> self.is_Some(),
     {
@@ -44,6 +45,7 @@ impl<A> Option<A> {
         }
     }
 
+    #[inline(always)]
     pub const fn is_none(&self) -> (res: bool)
         ensures res <==> self.is_None(),
     {


source/pervasive/option.rs
```diff
@@ -35,6 +35,24 @@ impl<A> Option<A> {
         }
     }
 
+    pub const fn is_some(&self) -> (res: bool)
```
jaybosamiya, line 38: We probably want to add this annotation (and similar for `is_none`)
```suggestion
    #[verifier(when_used_as_spec(is_Some))]
    pub const fn is_some(&self) -> (res: bool)
```

achreto: Agree, that would be great. Seems it doesn't quite work. Adding this, I'm getting the following error. The function is generated by `#[is_variant]` on the struct, maybe that's the reason? 

```
error: cannot find function referred to in when_used_as_spec
  --> nr/src/pervasive/option.rs:39:5
   |
39 |     pub const fn is_some(&self) -> (res: bool)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tjhance: yeah that doesn't surprise me, is_Some is treated as a special function

utaal: Please add `#[inline(always)]` to both.

utaal (approved):

source/pervasive/option.rs
achreto: I've added the `#[inline(always)]` to both.

PR #476, opened 2023-03-31
More pervasives- option and seq

jaybosamiya:
This PR adds a few more minor things to `pervasive`.

In particular, it adds:
1. A `try_option` macro, to be used while we don't yet use Rust's real `Option` type, so that we can at least give ourselves slightly nicer experience when writing executable code that is filled with a lot of optionals. It would be ideal for us to switch to real `Option`, or if not that, then use the `Try` Rust trait on our custom `Option` but that might be a more involved discussion, thus the introduction of `try_option` instead.
2. Add `fold_left` and `fold_right` to sequences. I have a few lemmas for `fold_left` implemented [here](https://github.com/verus-lang/verus-systems-code/blob/40e973f7f253d510ab0cd6a7ef757191d2f563b9/ironfleet-comparison/ironsht/src/utils_seq_v.rs) but I haven't yet gone through them all to decide which ones are the most useful generic ones, and which ones are overly specific, and which ones we really want to include in `pervasive`, thus that'll probably be a separate PR at some point (or if we think many/all of them are useful, we can move them over too, but probably as a separate PR).

### DIFF
--- a/source/pervasive/option.rs
+++ b/source/pervasive/option.rs
@@ -101,3 +101,18 @@ impl<A> Option<A> {
 }
 
 } // verus!
+
+/// A poor-person's `?` operator, until Verus switches to the "real" Rust `Option`.
+#[macro_export]
+#[allow(unused_macros)]
+macro_rules! try_option {
+    ($x:expr) => {{
+        let x = $x;
+        match x {
+            $crate::Option::None => {
+                return $crate::Option::None;
+            }
+            $crate::Option::Some(x) => x,
+        }
+    }};
+}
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (approved):
I have a slight preference for using `subrange(0, self.len() - 1)` instead of `subrange(1, self.len())` in order to keep the sequence indexing aligned and potentially make triggering easier, but I think this is fine.

jaybosamiya:
Actually I think I agree with you, having `subrange(0, self.len() - 1)` (equivalently `.drop_last()`) is probably better to keep things aligned better. Nonetheless, having the alternative form is useful too, so I've introduced both versions (with the main one being the `.drop_last()` definition, and the other considered an "alternate"); I've also introduced proofs showing that both the alternative definitions for each are equivalent.

Also added a fix for a minor trigger warning that was showing up for code that already existed; not _really_ related to this PR, but might as well fix it at the same time.

I've marked for re-review since this is a non-trivial set of changes that I just pushed and affects the publicly-visible API.

Chris-Hawblitzel (approved):

PR #479, opened 2023-04-03
improve build process (with vargo), add partial autodiscovery of builtin and vstd libraries, add Windows CI smoke test

utaal:
* in release mode, `vstd`'s build-dependencies (`builtin`, in particular) were built in a way that was incompatible with vstd as produced by verus; building them as `dependencies` fixes the issue; the current approach technically produces a potential dead-lock if cargo (vargo) is run with `-j 1` (a single concurrent job), as the vstd build script may start before the `builtin` (or macros) build and will prevent them from running (as only one job can be active at a time) -- I confirmed this happens, but it should not be an issue in any real-world scenario (including CI);
* when running with `vargo run`, or with the `tools/rust_verify.sh` script, Verus will now find and auto-import `builtin`, `builtin_macros`, `state_machines_macros`, and `vstd` (including `.vir`); `--no-vstd` can be used to disable importing `vstd` (and its `.vir`); `VERUS_ROOT` can be used to manually set the search path.


utaal:
For the record: I think there may be a way to (at least partially) do away with the artifact tracking in vargo by using a rustc wrapper for the vstd build, instead of a build script; I'll try this out later, but I don't think it's a priority right now.

Chris-Hawblitzel:
It's not clear to me why the build expects the Z3 executable to be named exactly `"../z3"`.  This didn't work under Windows, and the pull request https://github.com/verus-lang/verus/pull/477 changed this to allow configuring Z3 differently (on Windows, at least), but this new pull request undoes this partially, and thus doesn't work on Windows.  But more generally, I think we should allow people to put the Z3 executable where they want. which is the way it currently works in the `main` branch.

utaal:
> It's not clear to me why the build expects the Z3 executable to be named exactly "../z3".

That's because on *nix we ask the users to obtain `z3` with `./tools/get_z3.sh`, that stores it as `source/z3`. I thought I had addressed Windows-compatiliby by preserving the PATH-based search on Windows (after removing the `static` override), but I missed two cases. Let me fix those (and thanks for spotting it).

utaal:
Okay, I have changed things so that: there is now a script for PowerShell to get the version of z3 we expect and store it as `source/z3.exe`, like we do on *nix.

> But more generally, I think we should allow people to put the Z3 executable where they want. which is the way it currently works in the main branch.

Now, for both Windows and *nix, if the user sets `VERUS_Z3_PATH` by hand, tests don't override it anymore. They will assume `z3` is available as `source/z3(.exe)` if `VERUS_Z3_PATH` is not set.

I have also added a smoke test on Windows (build + `vargo test -p rust_verify_test --test basic`) so hopefully I won't accidentally break the Windows build again.

Chris-Hawblitzel (approved):

Issue #415, opened 2023-03-01
Incompleteness with some bitvector operations


jaybosamiya:
The following fails; changing the u64s to u32s still has it failing.

```rust
#[verifier(bit_vector)]
proof fn lemma_shift_right_u64_upper_bound(val: u64, amt: u64, upper_bound: u64)
  requires
    amt < 64u64,
    val <= upper_bound,
  ensures
    (val >> amt) <= (upper_bound >> amt)
{}
```

Changing to u16 has it succeed in 2~3 seconds, and changing to u8 has it succeed almost instantly, but that's not _too_ surprising.

Just as a quick point of comparison: a naive direct encoding into z3 is able to prove the above in ~75ms (~22ms for the 32-bit case)

tjhance:
@jaybosamiya can you post your "naive direct encoding" for comparisons' sake?

jaybosamiya:
```smt2
(declare-fun amt () (_ BitVec 64))
(declare-fun upper_bound () (_ BitVec 64))
(declare-fun val () (_ BitVec 64))
(assert (bvslt amt #x0000000000000040))
(assert (bvslt val upper_bound))
(assert (not (bvsle (bvashr val amt) (bvashr upper_bound amt))))
(check-sat)
```

Or as z3-py:
```py
from z3 import *
val, amt, upper_bound = BitVecs('val amt upper_bound', 64)
s = Solver()
s.add(amt < 64)
s.add(val < upper_bound)
s.add(Not ((val >> amt) <= (upper_bound >> amt)))
if s.check() == unsat:
    print("Much success, very proof, wow")
```

chanheec:
I think the query Verus generates for this example is very similar to the above query, although it has several differences.
1) it currently carries all the context inherited from the parent module (including the prelude)
2) It uses incremental solving(i.e. push/pop)
3) negligible encoding differences with `query` and `location_label`
4) it uses tailored z3 options for bit-vector queries

---

With Z3 4.10.1, I got these results:
1) removing incremental solving made the query go through (i.e. removing the pair of push/pop around this query)
2) removing all the context(including prelude) made the query go through. However, removing the module's context(and retaining prelude) didn't make this query success.


I think we can consider applying 1) or/and 2) to bit-vector queries

---

This is the how the query looks like:
```smt2
 (declare-const val~2@0 (_ BitVec 64))
 (declare-const amt~4@0 (_ BitVec 64))
[...]

PR #482, opened 2023-04-05
Disable incremental solving for bit-vector query

chanheec:
Summary:
This PR removes incremental solving for bit-vector queries to fix https://github.com/verus-lang/verus/issues/415

Minor details:
Some changes were made to handle two Z3 behavior changes due to disabling incremental solving.

1) "reason-unknown" message changes from
```
(:reason-unknown "(incomplete quantifiers)"
```
to
```
"(:reason-unknown "smt tactic failed to show goal to be sat/unsat (incomplete quantifiers)")"
```

2) Z3 Model can become unavailable for the same failing query

When a proof fails, even though the Z3 result is unknown, Z3 returns a model, and we have been using it for error reporting. However, for the same failing query, after removing push/pop, Z3 sometimes does not give a model .


tjhance:
What's the status on this, is it need of review? I think the best reviewer would be chris

Chris-Hawblitzel (approved):

PR #483, opened 2023-04-05
zsh + nix support `main_new`

jaybosamiya:
Minor changes:

1. Support `zsh` activation. Note that this is needed for this `tools/activate` script (but not necessary in most of our other scripts), since most other scripts can actually be invoked via `bash` (and indeed, their `#!` does just that), but this particular script is intended to be `source`d in, which means that the specific shell that is doing the `source`ing is what is in charge of running the whole script, and the interpreter-line `#!` is ignored. Thus, we need to explicitly account for zsh too.

2. Update `shell.nix` to account for `main_new`'s build requirements; much shorter now (although _technically_, depending on rustup is something hardcore nixos users might consider cheating to reduce the dependency set lol)

### DIFF
--- a/tools/activate
+++ b/tools/activate
@@ -1,5 +1,3 @@
-#! /bin/bash
-
 # Determine the directory that this script is in
 if [ "$BASH_VERSION" ]; then
   SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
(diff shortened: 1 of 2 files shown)


utaal (approved):
I should remove the #! on that script. It doesn't really make sense considering the file is intended to be sourced.

jaybosamiya:
Yep, I agree that it's prob good to remove it; have removed it.

PR #485, opened 2023-04-06
pervasive: implement vec::clear()

achreto:
Implementing `Vec::clear()`.

### DIFF
--- a/source/pervasive/vec.rs
+++ b/source/pervasive/vec.rs
@@ -161,7 +161,7 @@ impl<A> Vec<A> {
 
     #[verifier(external_body)]
     pub fn clear(&mut self)
-        ensures self.len() == 0
+        ensures self.view() == Seq::<A>::empty(),
     {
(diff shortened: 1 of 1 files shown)


source/pervasive/vec.rs
```diff
@@ -158,6 +158,13 @@ impl<A> Vec<A> {
     {
         self.vec.as_slice()
     }
+
+    #[verifier(external_body)]
+    pub fn clear(&mut self)
+        ensures self.len() == 0
```
tjhance, line 164: can you do `ensures self.view() == seq![]` instead? I know they're equivalent, but the more explicit version might prevent the user from having to call extensionality.

achreto: Sure! -- I've used `self.view() == Seq::<A>::empty()`  as `seq![]` seems to have had problems with the type parameter.

tjhance (approved):

PR #486, opened 2023-04-07
Add -L dependency line to rustc args

tjhance:
This seems to fix the issue I was talking about on slack, so that verus source code doesn't need to explicitly `use builtin::*`.


utaal (approved):
Ty!

PR #490, opened 2023-04-07
Automatically select profile in rust-verify.sh

### LINKED ISSUE #488: main_new rust_verify.sh requires --release
Currently, if Verus is compiled in release mode, rust_verify.sh requires explicitly using it (and it will fail if Verus is not compiled in debug mode). I propose we change that so that the script tries release mode and falls back to debug if it isn't available, so that a flag is generally not required. We can then also have a `--debug` flag to force the debug profile (and/or  `--release` to force release profile), if there are use cases for those flags.

Thoughts on this? If folks agree on the plan I can go ahead and implement it.

tchajed:
Default is release unless it isn't built, in which case we fall back to debug. The `--release` and `--debug` flags (as the first flag) can override this.

Fixes #488.

### DIFF
--- a/source/tools/rust-verify.sh
+++ b/source/tools/rust-verify.sh
@@ -2,15 +2,23 @@
 
 set -e
 
-if [ "$1" = "--release" ]; then
-    shift
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SOURCE="$DIR/.."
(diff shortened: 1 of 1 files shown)


utaal (approved):

jaybosamiya:
Since this is a minor change that has already been approved, I'm just gonna go ahead and hit merge here rather than waiting on @tchajed to hit merge.

Issue #303, opened 2022-09-28
Existence of functions in trait impls isn't checked


matthias-brun:
Omitting any of the `proof` or `exec` functions defined in a trait from a trait implementation should cause an error.

```rust
use builtin::*;
use builtin_macros::*;

verus! {

trait T {
    proof fn unprovable(&self)
        ensures false;
}
struct S { }
impl T for S { }

fn main() {
}
}
```

```
verification results:: verified: 1 errors: 0
```

tjhance:
Here's a more explicit program showing the soundness issue:

```rust
use builtin::*;
use builtin_macros::*;

verus! {

trait T {
    proof fn unprovable(&self)
        ensures false;
}
struct S { }
impl T for S { }

proof fn foo<J: T>(t: J)
ensures(false),
{
    t.unprovable();
    assert(false);
}

proof fn some_proof() {
    let s = S { };
    foo::<S>(s);
    assert(false);
}

fn main() {
}
}
```

PR #493, opened 2023-04-09
Encode trait specifications in separate method, so that Rust's trait …

Chris-Hawblitzel:
…checking doesn't see the specifications as a default implementation of the method (fixes https://github.com/verus-lang/verus/issues/303 ).

For this code:
```
trait T {
    fn f(b: bool) requires X;
}
```
the old encoding was:
```
trait T {
    fn f(b: bool) { requires(X); no_method_body() }
}
```
the new encoding is:
```
trait T {
    fn VERUS_SPEC__f(b: bool) { requires(X); no_method_body() }
    fn f(b: bool);
}
```

### DIFF
--- a/source/rust_verify/src/lifetime_generate.rs
+++ b/source/rust_verify/src/lifetime_generate.rs
@@ -27,6 +27,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use vir::ast::{Fun, FunX, Function, Mode, Path};
 use vir::ast_util::get_field;
+use vir::def::VERUS_SPEC;
 
 impl TypX {
     fn mk_unit() -> Typ {
@@ -1398,6 +1399,10 @@ fn erase_fn_common<'tcx>(
     } else {
         def_id_to_vir_path(ctxt.tcx, id)
     };
+    let is_verus_spec = path.segments.last().expect("segment.last").starts_with(VERUS_SPEC);
+    if is_verus_spec {
+        return;
+    }
     let fun_name = Arc::new(FunX { path: path.clone(), trait_path });
     if let Some(f_vir) = &ctxt.functions[&fun_name] {
         if f_vir.x.mode == Mode::Spec {
@@ -1416,7 +1421,7 @@ fn erase_fn_common<'tcx>(
         } else {
             None
         };
-        let body_exp = if external_body {
+        let body_exp = if body.is_none() || external_body {
             force_block(Some(Box::new((sig_span, ExpX::Panic))), sig_span)
         } else {
             let body = &body.expect("body");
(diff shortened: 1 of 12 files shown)


source/rust_verify/src/lifetime_generate.rs
```diff
@@ -1398,6 +1399,10 @@ fn erase_fn_common<'tcx>(
     } else {
         def_id_to_vir_path(ctxt.tcx, id)
     };
+    let is_verus_spec = path.segments.last().expect("segment.last").starts_with(VERUS_SPEC);
```
utaal, line 1402: Is it correct that the reason we rely on the name here is that the synthetic VERUS_SPEC methods do not appear in VIR, and are not otherwise tracked?

source/rust_verify/src/rust_to_vir.rs
```diff
@@ -65,7 +66,7 @@ fn check_item<'tcx>(
                 None,
                 None,
                 generics,
-                body_id,
+                Either::Left(body_id),
```
utaal, line 69: My understanding is that the general rust ethos is to use dedicated descriptive enums for cases like these, so that it's un-ambiguous what `Left` and `Right` represent (which is also why there's no `Either` in `std`).
It would also clarify what's the role of each branch of the `match` on `body_id` in `rust_to_vir_func`.

utaal (approved):

source/rust_verify/src/lifetime_generate.rs
Chris-Hawblitzel: Yes, the `VERUS_SPEC__f` trait methods are not considered real methods in VIR (I just added a test for this, confirming that programs can't call them directly) but are rather considered more like attributes on the `f` methods.

source/rust_verify/src/rust_to_vir.rs
Chris-Hawblitzel: I just removed the use of `Either` and replaced it with a dedicated enum for this case.

source/vir/src/headers.rs
```diff
@@ -238,3 +240,112 @@ impl Header {
...
+        );
+    }
+    for (b1, b2) in methodx.typ_bounds.iter().zip(typ_bounds.iter()) {
+        let ((x1, g1), (x2, g2)) = (b1, b2);
+        match (&**g1, &**g2) {
+            (GenericBoundX::Traits(ps1), GenericBoundX::Traits(ps2)) => {
+                if x1 != x2 || ps1 != ps2 {
+                    return err_str(
+                        &spec_method.span,
+                        "method specification different type parameters from method",
```
tjhance, line 281: i think this should say 'type bounds' not 'type parameters' right?

Also can we add test cases for both inconsistent type bounds & inconsistent type parameters?

tjhance (changes requested):

source/vir/src/headers.rs
Chris-Hawblitzel: It's hard to make a real test case yet because trait method type parameters aren't actually supported, but I can add a test case that will work in the future.

tjhance (approved):
i'll just go ahead and accept (don't wait for me after you've added the test)

PR #499, opened 2023-04-10
Fix --time when there are VIR errors, add option for machine-readable json output of verification outcome and timing

utaal:


tchajed:
This didn't work for me at first because it was using an older version of serde_json that didn't support u128. Can you constrain the version to be after [v1.0.93](https://github.com/serde-rs/json/releases/tag/v1.0.93)? Perhaps just take 1.0.95, the current version?

Chris-Hawblitzel (approved):

Issue #497, opened 2023-04-10
Ill-typed AIR for unit variable when trait execution


jaybosamiya:
Consider:
```rs
use vstd::prelude::*;

verus! {

    trait Foo {
        exec fn foo(&self);
    }

    struct X;

    impl Foo for X {
        exec fn foo(&self) {
            // do nothing
        }
    }

    exec fn bar() {
        let z: X = X;
        let x: () = z.foo();
    }

}

fn main() {}
```

Produces:
```console
$ rust-verify.sh --release a.rs
note: verifying root module

thread 'rustc' panicked at 'internal error: generated ill-typed AIR code: error 'in call to %Poly%tuple%0., argument #1 has type "tuple%0." when it should have type "Poly"' in expression '(%Poly%tuple%0. (tuple%0./tuple%0))'', rust_verify/src/verifier.rs:375:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@Chris-Hawblitzel this is the minimized reproducer for the bug we mentioned a short while ago (not particularly high priority btw, since the fix is to just not have the trivial let binding)

tjhance:
I think this is a duplicate of #392. The trait isn't actually necessary for the bug.

jaybosamiya:
Ah, I didn't try without the trait (should've run the automated minimizer and that'd have figured out that the trait wasn't necessary).

Anyways, the message here seems to somewhat different (in #392 it seems to be about number of arguments, while here it seems to be the wrong type), so I am not 100% sure it is a duplicate; but indeed seems related.

I suggest leaving this open so that we check this too when we fix up #392.

utaal:
It turns out they are slightly different code paths, but conceptually the same issue. https://github.com/verus-lang/verus/commit/f2d802ac331df27dcdafaceeefada1f61b7c8db1 should fix both.

utaal:
Fixed in `main_new`.

PR #502, opened 2023-04-11
Fix #392, #497: do not emit an AIR assignment for `let x: () = function_with_no_return()`

### LINKED ISSUE #392: invalid AIR for let-assignment with function-returning-unit
```
fn returns_nothing(a: u8, b: u8)
    requires a == b,
    ensures a == b,
{
}

fn test() {
    let x = returns_nothing(0, 0);
}
```

Gives invalid AIR in `expected 2 arguments, found 3 arguments' in expression '(ens%returns_nothing. 0 0 x~8@)'`

utaal:


utaal:
Good idea. Added!

tjhance (approved):

source/vir/src/sst_to_air.rs
```diff
@@ -1471,21 +1471,23 @@ fn stm_to_stmts(ctx: &Ctx, state: &mut State, stm: &Stm) -> Result<Vec<Stmt>, Vi
             }
             let mut ens_args: Vec<_> =
                 typ_args.into_iter().chain(ens_args_wo_typ.into_iter()).collect();
-            if let Some(Dest { dest, is_init }) = dest {
```
tjhance, line 1474: Actually, I do wanna double-check I understand this right:

It looks like in this situation, the SST/AIR creates a 'dest' variable, but if there's no return-value, then we skip this whole chunk of code, which means we end up not assigning to the variable in AIR.

But that's fine, because it's a unit variable, and the type invariant exists, so it's only constrained to be that one value anyway.

Is that right?

utaal: That's my understanding too, yes.
However, now that you mention it, skipping havoc-ing here is only sound if this case can only happen if the destination variable is unit (rather, `tuple%0` here). We do rely on the type checker for this. We could conceivably always havoc in this case, as a defensive strategy.

source/rust_verify_test/tests/functions.rs
```diff
@@ -28,3 +28,67 @@ test_verify_one_file! {
...
+test_verify_one_file! {
+    #[test] test_ensures_returns_default_regression_392_1 verus_code! {
+        fn returns_nothing(a: u8, b: u8)
+            requires a == b,
+            ensures a == b,
+        {
+        }
+
+        fn test() {
+            let x: () = returns_nothing(0, 0);
```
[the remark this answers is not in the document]
utaal: done.

Chris-Hawblitzel (approved):

PR #509, opened 2023-04-12
Make Seq::filter opaque, fix instructions for fmt check

jaylorch:
This pull request does two things:

1) It makes Seq::filter opaque instead of closed, so that it isn't useless. This follows the suggestion from @Chris-Hawblitzel in https://github.com/verus-lang/verus/pull/418.

2) It fixes the instructions for doing a format-check before checking in new Verus code.


utaal (approved):
Thanks!

jonhnet (approved):

PR #510, opened 2023-04-12
Replace some HashMap with IndexMap for stable SMT

Chris-Hawblitzel:
There were some obvious places where HashMap/HashSet iteration led to random ordering of items in the generated AIR/SMT files.  There may be more such places, but this is at least a starting point.


achreto (approved):
I've ran it a couple of times on a problematic code that previously triggered spurious assertion failures. Seems to be stable as of now.

tjhance (approved):

utaal (approved):

PR #511, opened 2023-04-12
Add Seq::map_values

jonhnet:
Most seq map calls ignore the index. Add a cleaner interface.


utaal:
This should probably be called `map`, once what's currently called `map` is renamed to `map_entries`. Can you add a TODO to that effect?

utaal (approved):

PR #512, opened 2023-04-13
Make rust_verify easier to run on Windows

Chris-Hawblitzel:


utaal:
As discussed, I'm going to closed this because we went with a separate binary instead. Feel free to reopen if I missed something.

PR #518, opened 2023-04-14
Add spans to all "unsupported" error messages

Chris-Hawblitzel:


jaylorch:
Cool! I tested it on the Ironfleet porting repo, and it works. And I reproduced an error that used to cause an information-free panic and now it points directly at the appropriate span. Thanks!

utaal (approved):

Issue #517, opened 2023-04-14
Can't `reveal` in `exec` mode


jaylorch:
Just as we can `assert` in exec mode (by implicitly switching to proof mode), it seems like we should be able to `reveal` in exec mode as well.

utaal:
This is currently working as expected, but the error message we're emitting isn't helpful. You're expected to use a proof block:
```
#[verifier(opaque)]
spec fn is_true(a: bool) -> bool {
    a
}

exec fn foo() {
    let a = true;
    let ghost b = is_true(a);
    proof {
        reveal(is_true);
    }
    assert(b);
}
```

Do you consider a better error message (https://github.com/verus-lang/verus/pull/520) a sufficient fix, or would you expect to use `reveal` directly?

jaylorch:
I should have tagged this as a feature request, but I didn't see such a tag available. I just thought that, since it's obvious that a `reveal` implies it's part of a proof, there was no need to require the developer to waste the characters putting a `proof` block around it. We already do that for `assert`, so I figured it would be analogously acceptable and natural to do it for `reveal`.

jaylorch:
To more directly answer your question, I appreciate the error message but I was really hoping that this could be made to be not an error.

PR #520, opened 2023-04-14
Rustc-like help in error messages

utaal:
This adds a facility to add "help" messages to diagnostics, and adds help text to errors of the form "cannot use <x> in exec mode" (which, at least in part, addresses https://github.com/verus-lang/verus/issues/517).

For the program:
```
#[verifier(opaque)]
spec fn is_true(a: bool) -> bool {
    a
}

exec fn foo() {
    let a = true;
    let ghost b = is_true(a);
    reveal(is_true);
    assert(b);
}
```

It prints:
```
error: cannot use reveal/hide in exec mode
  --> rust_verify/example/playground.rs:14:5
   |
14 |     reveal(is_true);
   |     ^^^^^^^^^^^^^^^
   |
   = help: wrap the reveal call in a `proof` block
```

It also cleans up error reporting code by leveraging `Into<String>` everywhere.

### DIFF
--- a/source/air/src/messages.rs
+++ b/source/air/src/messages.rs
@@ -44,6 +44,7 @@ pub struct MessageX {
     pub note: String,
     pub spans: Vec<Span>,          // "primary" spans
     pub labels: Vec<MessageLabel>, // additional spans, with string annotations
+    pub help: Option<String>,
 }
 pub type Message = Arc<MessageX>;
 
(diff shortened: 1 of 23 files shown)


utaal:
@jaylorch if this helps for now, please have a look and approve (or request changes); once we get around to allowing `reveal` without a proof block, we can remove the related diagnostic.

jaylorch:
This is a much clearer error message, thanks!

jaylorch (approved):

PR #522, opened 2023-04-14
Binary entry point

utaal:
This introduces a `verus` binary that leverages `rustup` to set-up the linker environment for `rust_verify` so it can find the rustc dynamically-linked libraries (on *nix and Windows).

The wrapper binary also tries to auto-discover `z3` in the `target-verus` folder, where it's placed by `vargo` if it finds it in `source`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -79,10 +79,18 @@ vargo run -p rust_verify --release -- rust_verify/example/recursion.rs
 
 This will make sure that the Verus and `vstd` builds are up-to-date, then run the verifier.
 
-You can also run the verifier directly (skipping the up-to-date check) on **Linux and macOS**:
+You can also run the verifier directly (skipping the up-to-date check) with:
+
+on Linux and macOS:
 
 ```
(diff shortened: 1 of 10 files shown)


Chris-Hawblitzel (approved):
I've confirmed that this works on Windows, and that the INSTALL.md changes work on Windows.  Minor comment: we might want to suggest example files that produce nice output when you verify and run them.  Right now, `recursion.rs` prints a warning about `decreases` clauses.  `doubly_linked_xor.exe` prints nothing when you run it.  We could add print statements to `doubly_linked_xor`.  Or we could use `vectors.rs` as the example.

utaal:
@Chris-Hawblitzel FYI, I added a commit after your review that attempts to follow symlinks to the executables, so users can symlink the binaries somewhere convenient to them.

Chris-Hawblitzel:
> @Chris-Hawblitzel FYI, I added a commit after your review that attempts to follow symlinks to the executables, so users can symlink the binaries somewhere convenient to them.

I checked again on Windows, and it continues to work.

tjhance (approved):

PR #541, opened 2023-04-21
remove the verus macro invocation in is_variant, switch the is/get_variant attributes to verus::internal

utaal:


utaal:
I've tried this on the NR proof and it seems to work there too, I'll go ahead and merge this to unblock https://github.com/verus-lang/verus/pull/396

Issue #538, opened 2023-04-20
recursive_types.rs isn't correctly catching the lack of a base-case variant in datatypes


Chris-Hawblitzel:
Verus currently allows the following proof of false, because the `check_well_founded` function in `recursive_types.rs` isn't working properly:

```
struct I<A>(A);
struct R(Box<I<R>>);

proof fn bad(r: R)
    ensures false
    decreases r
{
    bad(r.0.0);
}

spec fn make_r() -> R;

proof fn test()
    ensures false
{
    bad(make_r())
}
```

Here's some background on where this bug came from.  In order to support decreases clauses on datatypes,  Verus associates a "height" (aka a "rank") with datatype values.  Verus also adds an axiom that the height of a value in a field of a datatype value is strictly smaller than the height of the datatype value itself.  This axiom is only sound if the definition of height is well founded.  In particular, Verus requires that the datatype have a "base case" -- a non-recursive variant of the datatype that can be used to construct a default value of the datatype.  (For more technical details and a soundness proof, see Section 10.3 of https://arxiv.org/pdf/2303.05491.pdf , which talks about how the the typing rules enforce "that recursive structs have default values".)  Enforcing this requirement is the job of the `check_well_founded` function in `recursive_types.rs`, which is supposed to reject the definition `struct R(Box<I<R>>)` for failing to contain a base case variant.

The reason that `check_well_founded` is currently broken is that it was designed for an earlier version of the `height` definition and it was not correctly updated when we made `height` more general.  Specifically, the original version of `height` was fairly restrictive, and only introduced the height axiom for fields that had the same type as the outer datatype.  For example, in the following definition, there was an axiom that the inner `Box<List<A>>` field value of `Cons` had a smaller height than the outer `List<A>`:

```
enum List<A> { Nil, Cons(A, Box<List<A>>) }
```

However, in the original version, there was no axiom that the `A` field of `Cons` was smaller than the outer `List<A>`.  (This meant that the verifier correctly rejected the `bad` example shown above, because `I`'s field wasn't smaller than `I` itself.)  We later added this axiom but didn't correspondingly update `check_well_founded`, which needed to be made more restrictive to compensate for the additional height axiom.

We could go back to the older, more limited height axioms, but I think the better fix is to update `check_well_founded` to account for the more general height axioms.

PR #542, opened 2023-04-22
Fix check_well_founded in recursive_types.rs to correctly enable more flexible treatment of decreases on datatypes

Chris-Hawblitzel:
This fixes https://github.com/verus-lang/verus/issues/538 .  It replaces the old attributes:
- `#[verifier::maybe_negative]`
- `#[verifier::strictly_positive]`

with new attributes:
 - `#[verifier::reject_recursive_types]`
 - `#[verifier::reject_recursive_types_in_ground_variants]`
 - `#[verifier::accept_recursive_types]`

Partly this is a renaming (`maybe_negative` -> `reject_recursive_types`) because I found the old names confusing from the perspective of someone using types like `Set<A>` and `Seq<A>`.  For example, I found it hard to keep track of what `strictly_positive` meant when instantiating `Seq<A>` with `A = int` or `A = Foo`.  The new names attempt to make it clear that `Set<A>`, which declares `A` as `reject_recursive_types`, will reject an attempt to define a recursive type `my_recursive_type` with a field of type `Set<my_recursive_type>` while `Seq<A>`, which declares `A` as `accept_recursive_types`, will accept such an attempt.

Mainly, though, this pull request introduces a distinction between `reject_recursive_types_in_ground_variants` (the default attribute) and `accept_recursive_types`.  This distinction used to be absent; both of these were previously considered just `strictly_positive`.  The distinction is not important for ensuring strictly positive uses of recursive types, but it is important for ensuring that heights (ranks) work soundly.  In particular, the distinction captures the difference between types that have some "ground" variant, such as Option's None variant:

```
enum Option<#[verifier::accept_recursive_types] A> {
    None,
    Some(A),
}
```

and those that do not have such a variant, like Pair:
```
struct Pair<A, B> { first: A, second: B }
```

This distinction is also made by Dafny, and this pull request is similar to Dafny's approach ( see https://github.com/dafny-lang/dafny/blob/f116640b96d9d5d1af015d810d77e9637069f2d5/Source/DafnyCore/Resolver/Resolver.cs#L5302 ).  The main difference with Dafny is that this pull request introduces explicit attributes for `reject_recursive_types_in_ground_variants` and `accept_recursive_types` so that datatypes with private fields or external_body datatypes can express the right constraints on their type parameters.

I expect that the default `reject_recursive_types_in_ground_variants` will be used almost exclusively, except in cases where strict positivity requirements force the use of `reject_recursive_types`.  The attribute `accept_recursive_types` can be thought of as an optional (and rare) upgrade for types like `Option` that might be used to build other recursive types, as in:

```
struct List<A>(A, Box<Option<List<A>>>);
```

`Option` is the main use case for `accept_recursive_types`, although the attribute is also useful for `Seq` and `Map` in some cases:

```
struct Tree<A>(A, Box<Seq<Tree<A>>>);
```

## Detailed explanation
(At some point, this explanation will be moved into the guide.)

If treated naively, recursive types can lead to nonterminating proofs:
```
struct R { f: FnSpec(R) -> int }
proof fn bad()
    ensures false
{
    let f1 = |r: R| -> int {
        (r.f)(r) + 1
    };
    let r = R { f: f1 };
    // from r == R { f: f1 }:
    assert( r.f     == f1   );
    assert((r.f)(r) == f1(r));
    // from the definition of f1:
    assert(f1(r) == (r.f)(r) + 1);
    // taken together:
    assert(f1(r) == f1(r) + 1);
}
```
To prevent this, Verus prohibits recursion in "negative positions" in a recursive type.  Roughly, a negative position is anything on the left-hand side of a function type ->.  For example, the `R` in `FnSpec(R) -> int` is in a negative position.  Therefore, Verus rejects the definition `struct R { f: FnSpec(R) -> int }` with an error.

If generics are treated naively, they could encode recursion in negative positions.  For example, we could try to wrap the function type in a new type to hide the negative use of R:
```
struct FnWrapper<A, B> { f: FnSpec(A) -> B } // error: A not allowed in negative position
struct R { f: FnWrapper<R, int> }
```
To prevent this, Verus requires that type parameters used in negative positions (like A) be annotated with `#[verifier::reject_recursive_types]`:
```
struct FnWrapper<#[verifier::reject_recursive_types] A, B> { f: FnSpec(A) -> B } // ok
struct R { f: FnWrapper<R, int> } // error: R not allowed in negative position
```
Based on this annotation on A, Verus knows that the recursive R in `FnWrapper<R, int>` should be rejected, and it reports an error in the definition of R.

Recursive types can be used in decreases clauses of recursive specifications and recursive proofs:
```
enum List<A> {
    Nil,
    Cons(A, Box<List<A>>),
}

spec fn len<A>(list: &List<A>) -> nat
    decreases list // decreases can be used on values of type List<A>
{
    match list {
        List::Nil => 0,
        List::Cons(_, tl) => 1 + len(tl),
    }
}
```

To support this, Verus requires that struct and enum datatypes have a well-defined height (rank).  For this, Verus requires that struct and enum datatypes have a non-recursive "ground" variant that can be used as a base case for defining the height. For example, the Nil variant in List can be used to construct List values of height 0, and then the Cons variant can be repeatedly applied to construct bigger and bigger values with height > 0.  Attempting to declare a datatype with no ground variant will cause an error:
```
enum UngroundedList<A> {
    // error: no ground variant; the only variant is Cons, which recursively uses UngroundedList
    Cons(A, Box<UngroundedList<A>>),
}
```

If generics are treated naively, they could encode datatypes with no ground variant:
```
struct DataWrapper<A> { a: A }
enum UngroundedList<A> {
    // error: no ground variant; the only variant is Cons, which still recursively uses UngroundedList
    Cons(A, Box<DataWrapper<UngroundedList<A>>>),
}
```
To prevent this, Verus rejects a recursive type definition's ground variant from instantiating a type parameter A with the recursive type (UngroundedList) unless the type parameter A is marked `#[verifier::accept_recursive_types]`.  However, if DataWrapper marks A `accept_recursive_types`,  then DataWrapper must have a ground variant that is not built from A.  Because of this, Verus rejects the following:
```
struct DataWrapper<#[verifier::accept_recursive_types] A> { a: A } // error: no ground variant without A
```
However, by adding a ground variant, we can provide a correct wrapper, making both DataOption and GroundedList properly grounded:
```
enum DataOption<#[verifier::accept_recursive_types] A> { None, Some(A) } // ok
enum GroundedList<A> {
    Cons(A, Box<DataOption<GroundedList<A>>>), // ok
}
```

Overall, Verus parameters have one of three levels of acceptance of recursive types:
 - `#[verifier::reject_recursive_types]`
 - `#[verifier::reject_recursive_types_in_ground_variants]`
 - `#[verifier::accept_recursive_types]`

`reject_recursive_types` is added to types that use the type parameter negatively, and `accept_recursive_types` may (optionally) be added to types that have a ground variant that doesn't use the type parameter.

Typical example of reject_recursive_types:
```
struct Set<#[verifier::reject_recursive_types] A> {
    f: FnSpec(A) -> bool,
}
```

Typical example of reject_recursive_types_in_ground_variants (which is the default):
```
struct Pair<A, B> { first: A, second: B }
```

Typical example of accept_recursive_types:
```
enum Option<#[verifier::accept_recursive_types] A> {
    None,
    Some(A),
}
```

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -239,12 +239,12 @@ pub fn internal_arbitrary<A>(_: u64) -> A {
 //
 
 #[verifier::external_body]
-pub struct Ghost<#[verifier::strictly_positive] A> {
+pub struct Ghost<#[verifier::reject_recursive_types_in_ground_variants] A> {
     phantom: PhantomData<A>,
 }
 
 #[verifier::external_body]
-pub struct Tracked<#[verifier::strictly_positive] A> {
+pub struct Tracked<#[verifier::reject_recursive_types_in_ground_variants] A> {
     phantom: PhantomData<A>,
 }
 
--- a/source/pervasive/cell.rs
+++ b/source/pervasive/cell.rs
@@ -51,7 +51,7 @@ verus!{
 /// ### Example (TODO)
 
 #[verifier(external_body)]
-pub struct PCell<#[verifier(strictly_positive)] V> {
+pub struct PCell<#[verifier::accept_recursive_types] V> {
     ucell: UnsafeCell<MaybeUninit<V>>,
 }
 
@@ -69,7 +69,7 @@ unsafe impl<T> Send for PCell<T> {}
 // (Note: this depends on the current behavior that #[verifier::spec] fields are still counted for marker traits)
 
 #[verifier(external_body)]
-pub tracked struct PointsTo<#[verifier(strictly_positive)] V> {
+pub tracked struct PointsTo<#[verifier::reject_recursive_types_in_ground_variants] V> {
     phantom: marker::PhantomData<V>,
     no_copy: NoCopy,
 }
@@ -255,7 +255,7 @@ impl<T> InvariantPredicate<(Set<T>, PCell<T>), PointsTo<T>> for InvCellPred {
     }
 }
 
-pub struct InvCell<#[verifier(maybe_negative)] T> {
+pub struct InvCell<#[verifier::reject_recursive_types] T> {
     possible_values: Ghost<Set<T>>,
     pcell: PCell<T>,
     perm_inv: Tracked<LocalInvariant<(Set<T>, PCell<T>), PointsTo<T>, InvCellPred>>,
--- a/source/pervasive/invariant.rs
+++ b/source/pervasive/invariant.rs
@@ -7,7 +7,7 @@
 
 // An invariant storing objects of type V needs to be able to have some kind of configurable
 // predicate `V -> bool`. However, doing this naively with a fully configurable
-// predicate function would result in V being maybe_negative,
+// predicate function would result in V being reject_recursive_types,
 // which is too limiting and prevents important use cases with recursive types.
 
 //
@@ -22,7 +22,7 @@
 //  * V - Type of the stored 'tracked' object 
 //  * Pred: InvariantPredicate - provides the predicate (K, V) -> bool
 //
-// With this setup, we can now let both K and V be strictly_positive.
+// With this setup, we can now declare both K and V without reject_recursive_types.
 // To be sure, note that the following, based on our trait formalism,
 // is well-formed CIC (Coq), without any type polarity issues:
 //
@@ -119,7 +119,7 @@ pub trait InvariantPredicate<K, V> {
 
 #[verifier::proof]
 #[verifier::external_body] /* vattr */
-pub struct AtomicInvariant<#[verifier::strictly_positive] /* vattr */ K, #[verifier::strictly_positive] /* vattr */ V, #[verifier::strictly_positive] /* vattr */ Pred> {
+pub struct AtomicInvariant<#[verifier::accept_recursive_types] K, #[verifier::accept_recursive_types] V, #[verifier::accept_recursive_types] Pred> {
     dummy: builtin::SyncSendIfSend<V>,
     dummy1: core::marker::PhantomData<(K, Pred)>,
 }
@@ -168,7 +168,7 @@ impl<K, V, Pred> AtomicInvariant<K, V, Pred> {
 
 #[verifier::proof]
 #[verifier::external_body] /* vattr */
-pub struct LocalInvariant<#[verifier::strictly_positive] /* vattr */ K, #[verifier::strictly_positive] /* vattr */ V, #[verifier::strictly_positive] /* vattr */ Pred> {
+pub struct LocalInvariant<#[verifier::accept_recursive_types] /* vattr */ K, #[verifier::accept_recursive_types] /* vattr */ V, #[verifier::accept_recursive_types] /* vattr */ Pred> {
     dummy: builtin::SendIfSend<V>,
     dummy1: core::marker::PhantomData<(K, Pred)>, // TODO ignore Send/Sync here
 }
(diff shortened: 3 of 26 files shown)


tjhance:
Let me rewrite this in my own words to check that I understand:

1. Verus requires all types to be inhabited. Furthermore, Verus treats all datatypes as inductive datatypes (no co-inductive datatypes), which allows the height axioms. As a consequence of both of these things, we need to ensure that every datatype definition has a "ground" term (unlike say, Coq, which just allows uninhabited inductive types).
2. If we write something like `struct Foo { a: Bar<Foo> }` we need to be able to check if `Bar<Foo>` contains a ground term.
3. Thus we need to able to answer queries like "does `Bar<T>` have a ground term that does not depend on its type parameter `T`?"
    * In general, the answer to such a query for `Bar<T, U, V, W, X, ...>` could be really complicated, since it could have a bunch of potential ground variants, each of which depends on an arbitrary subset of its type parameters.
    * The approach you took here is more conservative. Specifically, if we consider each type parameter that doesn't have negative occurrences, each one is marked either `accept_recursive_types` or `reject_recursive_types_in_ground_variants`. Furthermore, it must be possible to construct a ground type that does not depend on any of the `accept_recursive_types` parameters.
    * Now, we can answer the query: if we instantiate `Bar<...>` where some subset of the parameters is recursive, we check if each of those parameters is `accept_recursive_types`, and if so, the answer to the query is 'yes'. Otherwise, we conservatively say 'no'.

Does that all seem right to you?

---

On an unrelated note, I've been finding the attribute style makes the declarations a bit unwieldy, and it's only going to get worse with the new long name. What do we think of a style like:

```
#[verifier::reject_recursive_types(A)] 
struct Set<A> {
    f: FnSpec(A) -> bool,
}
```

Chris-Hawblitzel:
> Does that all seem right to you?

Yes, that sounds right to me.

> On an unrelated note, I've been finding the attribute style makes the declarations a bit unwieldy, and it's only going to get worse with the new long name. What do we think of a style like:
> 
> ```
> #[verifier::reject_recursive_types(A)] 
> struct Set<A> {
>     f: FnSpec(A) -> bool,
> }
> ```

I agree with this suggestion; this makes it more readable.

tjhance (approved):

utaal:
Considering that this has been open for a while (which is my fault), I'm happy to move this along.
One question: am I correct in interpreting the code changes in that this will not immediately break existing code by continuing to support the old-style attributes so that we can properly deprecate them before they go away?
Do you think (not necessarily in this PR) we should add a deprecation warning?

Chris-Hawblitzel:
> Considering that this has been open for a while (which is my fault), I'm happy to move this along.
> One question: am I correct in interpreting the code changes in that this will not immediately break existing code by continuing to support the old-style attributes so that we can properly deprecate them before they go away?
> Do you think (not necessarily in this PR) we should add a deprecation warning?

Yes, I'd like to add a deprecation warning and eventually remove the old attributes.  I'm not sure what the best way is to mark things as deprecated in Verus.

utaal (approved):
> I'm not sure what the best way is to mark things as deprecated in Verus.

I think we may be able to use the generalized `air::Diagnostics` interface to emit a warning when we encounter one of the deprecated attributes. We may want to add a "deprecated" function to `Diagnostics` (or to a wrapper) to be able to track all of the deprecation warnings in the codebase, and periodically review them to eventually remove them.

PR #543, opened 2023-04-23
Clarify check_function error message about private fields

tenzinhl:


tenzinhl:
For learning purposes: could I get some help understanding what this block does?
```rust
    if let Some(body) = &function.x.body {
        // Check that public, non-abstract spec function bodies don't refer to private items:
        let disallow_private_access = match (&function.x.publish, function.x.mode) {
            (Some(_), Mode::Spec) => {
                let msg = "public spec function cannot refer to private fields, if it is marked #[verifier(publish)]";
                Some((&function.x.visibility.restricted_to, msg))
            }
            _ => None,
        };
        check_expr(ctxt, function, body, disallow_private_access)?;
    }
    Ok(())
```

So the first line is an if let statement, checks if `function.x.body` is `Some`, pulls out the actual body inside of the `Option<body>`, then passes it to `check_expr` which seems to do most of the work (with disallow_private_access containing some information about visibility and an error message to send if the check fails).

What's going on in `check_expr`?

```rust
fn check_expr(
    ctxt: &Ctxt,
    function: &Function,
    expr: &Expr,
    disallow_private_access: Option<(&Option<Path>, &str)>,
) -> Result<(), VirErr> {
    crate::ast_visitor::expr_visitor_check(expr, &mut |_scope_map, expr| {
        check_one_expr(ctxt, function, expr, disallow_private_access)
    })
}
```

Is it mapping over the expressions in the function it's given? (Since I see it takes a lambda). Then it seems like `check_expr`, well, checks if an expression is valid...

Which `ExprX` does the private field access correspond to?

tenzinhl:
Oh woops, guess I need to change the tests too.

tjhance:
i think the message can be even clearer. "cannot call constructor of struct with private fields"

tenzinhl:
> i think the message can be even clearer. "cannot call constructor of struct with private fields"

Would this involve more logic than just a text substitution? (It's not immediately obvious to me that where the error message currently crops up it can only be triggered by constructing a struct)

tjhance:
> What's going on in check_expr?

The `crate::ast_visitor::expr_visitor_check` will recursively traverse the given expr, and call the closure on each Expr node it finds. The `check_one_expr` then has a bunch of checks for that node.

> Would this involve more logic than just a text substitution?

Yeah, it looks like it. Right now, the `disallow_private_access` argument to `check_expr` exists because different calls will specify different contexts, e.g., `public function requires` or `public spec function ... if it is marked #[verifier(publish)]`. There are a bunch of hard-coded error messages with different context, but each one uses the vague `cannot refer to private items` phrasing. The `refer to private items` part should be more specific, e.g.,

* `cannot refer to private constant item`
* `cannot refer to private function`
* `cannot access private field`
* `cannot call constructor of struct with private fields`

The ideal error message here would be specific both in the 'context' and in the reason that the given expression refers to a private item. This would probably require some messages constructed with format!()

tjhance:
this error message has now been fixed in 59e29ec18b62d91afdccb42c98add575a319bfe7

PR #544, opened 2023-04-26
Implement methods for testing and unwrapping result

marshtompsxd:
Implement exec methods for testing and unwrapping result, similar to option.

### DIFF
--- a/source/pervasive/result.rs
+++ b/source/pervasive/result.rs
@@ -35,14 +35,14 @@ impl<T, E> Result<T, E> {
 
     // A more-readable synonym for get_Ok_0().
     #[verifier(inline)]
-    pub open spec fn spec_unwrap_ok(self) -> T
+    pub open spec fn spec_unwrap(self) -> T
     recommends self.is_Ok()
     {
         self.get_Ok_0()
     }
 
-    #[verifier(when_used_as_spec(spec_unwrap_ok))]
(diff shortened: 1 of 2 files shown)


source/pervasive/result.rs
```diff
@@ -1,9 +1,79 @@
...
+    // A more-readable synonym for get_Ok_0().
+    #[verifier(inline)]
+    pub open spec fn spec_unwrap_ok(self) -> T
+    recommends self.is_Ok()
+    {
+        self.get_Ok_0()
+    }
+
+    #[verifier(when_used_as_spec(spec_unwrap_ok))]
+    pub fn unwrap_ok(self) -> (t: T)
```
utaal, line 45: Can you please rename this to `unwrap`, to be consistent with https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap ?

marshtompsxd: Updated

utaal:
Can you please add a couple of tests of the intended usage of these, in particular of the `when_used_as_spec` functions? (Sorry I didn't catch it the first time.)

marshtompsxd:
Yes. Should I add tests in `rust_verify_test`?

> in particular of the when_used_as_spec functions?

I added it because I see similar things in option.rs. If this is undesired I can also remove it.

utaal (approved):
Thank you! Please use "Squash and merge" to merge.

PR #545, opened 2023-04-26
ignore OpaqueTy items to support external async functions

utaal:
they implicitly also define an opaque type (e.g. `type Foo = impl Bar;`)


tjhance:
what happens if the user declares an opaque type of their and then uses it in non-external code? do they get a reasonable error message?

Chris-Hawblitzel (approved):

PR #546, opened 2023-04-27
Implement `as_rust_string_ref` for String

marshtompsxd:
Implement `into_rust_string_ref` for String to extract the inner String as a reference.

### DIFF
--- a/source/pervasive/string.rs
+++ b/source/pervasive/string.rs
@@ -125,8 +125,8 @@ impl<'a> StrSlice<'a> {
     }
 
 
-    
-    // TODO:This should be the as_bytes function after 
+
+    // TODO:This should be the as_bytes function after
     // slice support is added
     // pub fn as_bytes<'a>(&'a [u8]) -> (ret: &'a [u8])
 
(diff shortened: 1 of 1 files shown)


source/pervasive/string.rs
```diff
@@ -247,6 +247,12 @@ impl String {
     {
         self.inner
     }
+
+    #[verifier(external)]
+    pub fn into_rust_string_ref(&self) -> &std::string::String
```
jaybosamiya, line 252: Just a suggestion wrt naming:
```suggestion
    pub fn as_rust_string_ref(&self) -> &std::string::String
```
For a borrowed->borrowed conversion (that is not expensive, and exposes a view into the underlying data), the API convention in Rust is to use an `as_*` name https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

I don't know if the `rust_string_ref` part is ideal or not (I _personally_ think it is fine), but I do think maintaining the `as_` portion would likely be better than `into_`

utaal (approved):

PR #549, opened 2023-04-30
In lifetime_generate, generalize treatment of Copy

Chris-Hawblitzel:


source/rust_verify/src/lifetime_ast.rs
```diff
@@ -136,7 +137,8 @@ pub(crate) struct DatatypeDecl {
     pub(crate) name: Id,
     pub(crate) span: Span,
     // Does the type implement the Copy trait? (e.g. impl<A: Copy> Copy for S<A> {})
-    pub(crate) implements_copy: bool,
+    // If so, for each GenericParam A say whether clone and copy require A: Clone and A: Copy
+    pub(crate) implements_copy: Option<Vec<bool>>,
```
tjhance, line 141: are you only checking the `Copy` bounds here? It'd be unusual, but in principle it's possible for the `Copy` impl to depend on other traits.

Chris-Hawblitzel: Yes, the comment "For now, only allowed predicates are Copy and Sized" still holds.  It would be possible for the trait bound reasoning to get almost arbitrarily complicated and the full generality is not supported (yet).

tjhance (approved):

PR #550, opened 2023-04-30
allow adding requires/ensures to external exec functions

tjhance:
Not hard in concept, although there are a bunch of sanity checks that need to be done. Making sure functions don't get spec'ed twice, treating visibility the right way when the requires/ensures and the actual definition is in a different module, and so on.

TODO:
 - [x] Add sanity checks that the type signatures match
 - [x] Fix other misc. tests

~~Have a cleaner syntax that lowers into the encoding?~~ (I'll do this later if it turns out to be desired)

### The encoding

A spec looks like this:

```
pub fn swap_requires_ensures<T>(a: &mut T, b: &mut T)
    ensures *a == *old(b), *b == *old(a),
{   
    core::mem::swap(a, b)
}   
```

I called this a "proxy" in the code. The function given inside the body is the one that the specification is applied to. This forces rustc to do some typechecking for us, although we need to check to make sure that the proxy function doesn't have extra trait bounds or anything. (Though it may be a good idea to relax this later.)

The specification can be attached to any function as long as it wouldn't otherwise appear in the VIR. So for example, anything from an external crate, or any `verifier::external` function.

### VIR changes

A function now has a `proxy` field which indicates if the spec was declared with a proxy. So in the above example, you'd have the `proxy` be `crate::swap_requires_ensures` and the real "name" would be `core::mem::swap`. The previous assumption that every function in a "krate" would have a `Path` that matches the krate name is no longer going to hold.

### Visibility

The proxy has to be at least as visible as the function it is specifying. That way, anything that can call the real function can also access the specification.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -1,7 +1,8 @@
 use crate::attributes::{
-    get_fuel, get_mode, get_publish, get_ret_mode, get_var_mode, get_verifier_attrs,
+    get_fuel, get_mode, get_publish, get_ret_mode, get_var_mode, get_verifier_attrs, VerifierAttrs,
 };
 use crate::context::{BodyCtxt, Context};
+use crate::rust_to_vir_base::mk_visibility;
 use crate::rust_to_vir_base::{
     check_generics_bounds_fun, def_id_to_vir_path, foreign_param_to_var, mid_ty_to_vir,
 };
@@ -10,9 +11,11 @@ use crate::util::{err_span, unsupported_err_span};
 use crate::{unsupported_err, unsupported_err_unless};
 use rustc_ast::Attribute;
 use rustc_hir::{
-    def::Res, Body, BodyId, Crate, FnDecl, FnHeader, FnRetTy, FnSig, Generics, HirId, MaybeOwner,
-    MutTy, Param, PrimTy, QPath, Ty, TyKind, Unsafety,
+    def::Res, Body, BodyId, Crate, ExprKind, FnDecl, FnHeader, FnRetTy, FnSig, Generics, HirId,
+    MaybeOwner, MutTy, Param, PrimTy, QPath, Ty, TyKind, Unsafety,
 };
+use rustc_middle::ty::Predicate;
+use rustc_middle::ty::SubstsRef;
 use rustc_middle::ty::TyCtxt;
 use rustc_span::def_id::DefId;
 use rustc_span::symbol::{Ident, Symbol};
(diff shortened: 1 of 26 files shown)


utaal:
I imagine we are going to need additional traits and types (e.g. `Seq`) to specify the standard library (and other external libraries, e.g. for `Vec`). Do you have a plan on what that would look like? i.e. in what crate(s) and crate metadata files will the additional types and specs will live?

Chris-Hawblitzel (approved):

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -162,23 +172,186 @@ pub(crate) fn check_item_fn<'tcx>(
...
+    let this_path = def_id_to_vir_path(ctxt.tcx, id);
+
+    let is_verus_spec = this_path.segments.last().expect("segment.last").starts_with(VERUS_SPEC);
+    let is_new_strlit =
+        ctxt.tcx.is_diagnostic_item(Symbol::intern("pervasive::string::new_strlit"), id);
 
     let vattrs = get_verifier_attrs(attrs)?;
+    let mode = get_mode(Mode::Exec, attrs);
+
+    let (path, proxy, visibility) = if vattrs.external_fn_specification {
```
utaal, line 184: Would it make sense to factor out this branch of the `if`, to clarify that it handles proxies and to avoid making this function even longer?

tjhance: sure

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -519,6 +700,91 @@ fn is_mut_ty<'tcx>(
...
+        ExprKind::Block(block_body, _) => match &block_body.expr {
+            Some(body_value) => body_value,
+            None => {
+                return err();
+            }
+        },
+        _ => &body.value,
+    };
+
+    // TODO handle MethodCall
```
utaal, line 761: Should this TODO be resolved? Or is it intentionally left for later?
I believe `MethodCall` is only relevant for associated methods (or trait impls), so I think that branch is currently guarded by the "unsupported" error in `rust_to_vir_func`. If this is intentionally left for later; can you expand on the comment, to point out that `MethodCall` would be unreachable currently?

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -519,6 +700,91 @@ fn is_mut_ty<'tcx>(
...
+    id1: rustc_span::def_id::DefId,
+    id2: rustc_span::def_id::DefId,
+    substs: SubstsRef<'tcx>,
+) -> bool {
+    let preds1 = all_predicates(tcx, id1, substs);
+    let preds2 = all_predicates(tcx, id2, substs);
+    if preds1.len() != preds2.len() {
+        return false;
+    }
+    // TODO should allow this to work if the predicates are in a different order
```
utaal, line 714: With the code as-is, what happens if they aren't?

source/rust_verify/src/rust_to_vir_func.rs
tjhance: it's left for later, just a matter of scoping down the PR.

source/rust_verify/src/rust_to_vir_func.rs
tjhance: there will be an error about trait mismatch

utaal (approved):

PR #553, opened 2023-05-08
Add decorated types to SMT encoding to represent &, Box, Rc, etc.

Chris-Hawblitzel:
For the most part, Verus's SMT encoding does not have to distinguish between types `T`, `&T`, `Box<T>`, `Rc<T>`, and so on.  However, there are some situations where it would be useful to distinguish these in the SMT encoding.  For example, the `size_of<T>` function returns different values depending on `T`.  As another example, trait implementations can depend on the difference between, say, `S` and `&S` (although this is currently disallowed by Verus):

```
trait T { spec fn f(&self) -> int; }
struct S {}
impl T for S { spec fn f(&self) -> int { 3 } }
impl T for &S { spec fn f(&self) -> int { 4 } }
```

This commit adds types to the SMT encoding that are "decorated" with `&`, `Box`, `Rc`, etc., to allow functions like `size_of<T>`, and in the future, to allow trait implementations like `impl T for &S`.

For efficiency's sake, the new encoding distinguishes between undecorated and decorated types.  Undecorated types are still used for boxing/unboxing operations and has_type, just as they were before this commit.  This ensures that decorations don't add more axiom instantiations for dealing with boxing/unboxing and has_type.  So instead of replacing undecorated types with decorated types, the new encoding uses both side by side.  Specifically, in the new SMT encoding, generic functions now take two type-id parameters for each Rust generic type parameter, one undecorated and one decorated.  So for this Verus code:

```
spec fn sizeof<A>() -> nat;

proof fn test() {
    assume(sizeof::<&u8>() == 8);
}
```

the SMT encoding is:

```
;; Function-Decl crate::sizeof
(declare-fun sizeof.? (Type Type) Int)

;; Function-Axioms crate::sizeof
(axiom (forall ((A& Type) (A&. Type)) (!
   (<= 0 (sizeof.? A& A&.))
   :pattern ((sizeof.? A& A&.))
   :qid
   internal_sizeof.?_pre_post_definition
   :skolemid
   skolem_internal_sizeof.?_pre_post_definition
)))

;; Function-Def crate::test
(check-valid
 (declare-const no%param@ Int)
 (axiom fuel_defaults)
 (assume
  (= (sizeof.? (UINT 8) (REF (UINT 8))) 8)
))
```

It may be possible to optimize away the decorated type in some situations.  However, the SMT overhead is not necessarily that bad, because the decorated type doesn't really do much.  It doesn't cause any additional quantifier instantiations or theory work.  All it does is to make the existing quantifier instantiation process (e-matching) a little more selective.

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -253,10 +253,10 @@ where
 {
     if !build_test_mode {
         if let Some(VerusRoot { path: verusroot, in_vargo }) = find_verusroot() {
-            verifier
-                .args
-                .import
-                .push((format!("vstd"), verusroot.join("vstd.vir").to_str().unwrap().to_string()));
+            if !verifier.args.no_vstd {
+                let vstd = verusroot.join("vstd.vir").to_str().unwrap().to_string();
+                verifier.args.import.push((format!("vstd"), vstd));
+            }
             rustc_args.push(format!("--edition"));
             rustc_args.push(format!("2018"));
             let externs = VerusExterns { path: &verusroot, has_vstd: !verifier.args.no_vstd };
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -89,9 +89,30 @@ pub enum IntRange {
     ISize,
 }
 
+/// Type information relevant to Rust but generally not relevant to the SMT encoding.
+/// This information is relevant for resolving traits.
+#[derive(Debug, Serialize, Deserialize, Hash, ToDebugSNode, Clone, Copy, PartialEq, Eq)]
+pub enum TypDecoration {
+    /// &T
+    Ref,
+    /// &mut T
+    MutRef,
+    /// Box<T>
+    Box,
+    /// Rc<T>
+    Rc,
+    /// Arc<T>
+    Arc,
+    /// Ghost<T>
+    Ghost,
+    /// Tracked<T>
+    Tracked,
+    /// !, represented as Never<()>
+    Never,
+}
+
 /// Rust type, but without Box, Rc, Arc, etc.
 pub type Typ = Arc<TypX>;
-
 pub type Typs = Arc<Vec<Typ>>;
 // Deliberately not marked Eq -- use explicit match instead, so we know where types are compared
 #[derive(Debug, Serialize, Deserialize, Hash, ToDebugSNode)]
@@ -108,6 +129,9 @@ pub enum TypX {
     AnonymousClosure(Typs, Typ, usize),
     /// Datatype (concrete or abstract) applied to type arguments
     Datatype(Path, Typs),
+    /// Wrap type with extra information relevant to Rust but usually irrelevant to SMT encoding
+    /// (though needed sometimes to encode trait resolution)
+    Decorate(TypDecoration, Typ),
     /// Boxed for SMT encoding (unrelated to Rust Box type), can be unboxed:
     Boxed(Typ),
     /// Type parameter (inherently SMT-boxed, and cannot be unboxed)
--- a/source/vir/src/context.rs
+++ b/source/vir/src/context.rs
@@ -21,6 +21,9 @@ use std::fs::File;
 use std::rc::Rc;
 use std::sync::Arc;
 
+// Use decorated types in addition to undecorated types (see sst_to_air::typ_to_id)
+pub(crate) const DECORATE: bool = true;
+
 pub type ChosenTrigger = Vec<(Span, String)>;
 #[derive(Debug, Clone)]
 pub struct ChosenTriggers {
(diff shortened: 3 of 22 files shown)


source/rust_verify/src/driver.rs
```diff
@@ -253,10 +253,10 @@ where
...
     if !build_test_mode {
         if let Some(VerusRoot { path: verusroot, in_vargo }) = find_verusroot() {
-            verifier
-                .args
-                .import
-                .push((format!("vstd"), verusroot.join("vstd.vir").to_str().unwrap().to_string()));
+            if !verifier.args.no_vstd {
+                let vstd = verusroot.join("vstd.vir").to_str().unwrap().to_string();
+                verifier.args.import.push((format!("vstd"), vstd));
+            }
```
Chris-Hawblitzel, line 259: @utaal This is independent of the rest of the commit but it came up during testing -- I hope this is the correct change to disable loading of vstd with --no_vstd.

source/vir/src/recursion.rs
```diff
@@ -431,8 +433,8 @@ pub(crate) fn check_termination_exp(
             let varx = ExpX::Var(unique_local(&str_ident(FUEL_PARAM)));
             let var_typ = Arc::new(TypX::Air(str_typ(FUEL_TYPE)));
             args.push(SpannedTyped::new(&exp.span, &var_typ, varx));
-            let rec_name = prefix_recursive_fun(&x);
-            let callx = ExpX::Call(CallFun::Fun(rec_name), typs.clone(), Arc::new(args));
+            let callx =
+                ExpX::Call(CallFun::CheckTermination(x.clone()), typs.clone(), Arc::new(args));
```
Chris-Hawblitzel, line 437: @utaal This turned out not to be need by the commit, but it seems neater than the way this worked previously, so I left this change in.

source/vir/src/ast.rs
```diff
@@ -89,9 +89,30 @@ pub enum IntRange {
...
 }
 
+/// Type information relevant to Rust but generally not relevant to the SMT encoding.
+/// This information is relevant for resolving traits.
+#[derive(Debug, Serialize, Deserialize, Hash, ToDebugSNode, Clone, Copy, PartialEq, Eq)]
+pub enum TypDecoration {
+    /// &T
+    Ref,
+    /// &mut T
+    MutRef,
```
tjhance, line 99: Why is MutRef here?

(I guess that this does nothing at the moment, due to all the current restrictions on `&mut`. But either why I don't think it makes sense to consider it isomorphic to `T` and `&T`)

source/vir/src/ast.rs
```diff
@@ -89,9 +89,30 @@ pub enum IntRange {
...
+    Rc,
+    /// Arc<T>
+    Arc,
+    /// Ghost<T>
+    Ghost,
+    /// Tracked<T>
+    Tracked,
+    /// !, represented as Never<()>
+    Never,
+}
```
tjhance, line 112: Should we also have type decoration to distinguish different integer types?

tjhance (changes requested):

source/vir/src/ast.rs
Chris-Hawblitzel: Rust lets trait implementations specialize on `&` and `&mut`:

```
trait T { fn f(&self) -> u64; }
struct S {}
impl T for S { fn f(&self) -> u64 { 3 } }
impl T for &S { fn f(&self) -> u64 { 4 } }
impl T for &mut S { fn f(&self) -> u64 { 5 } }
```

source/vir/src/ast.rs
Chris-Hawblitzel: I think this should already be handled by `typ_to_id` and `range_to_id`, for example generating `(UINT 32)` or `NAT`.

source/vir/src/context.rs
```diff
@@ -21,6 +21,9 @@ use std::fs::File;
 use std::rc::Rc;
 use std::sync::Arc;
 
+// Use decorated types in addition to undecorated types (see sst_to_air::typ_to_id)
+pub(crate) const DECORATE: bool = true;
```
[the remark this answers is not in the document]
Chris-Hawblitzel: Mainly, it's just for future benchmarking.  I'd like to check whether this commit needs further optimizations at some point when we have larger benchmarks, so this a way to easily turn this feature on or off to check the performance impact.

source/vir/src/ast.rs
tjhance: Is `impl T for &mut S { fn f(&self) -> u64 { 5 } }` actually supported right now? Given our limited support for `&mut` types, it's hard to see how that would work.

What I'm really asking is ... I thought type decorations were for "trivial" type constructors, like `&` or `Box<_>` which would otherwise be dropped entirely from the encoding, but that can't be the case for `&mut`.

source/vir/src/ast.rs
tjhance: oh I see, I didn't realize this distinction was already made

utaal (approved):

PR #555, opened 2023-05-09
Support external types such as std::Option

tjhance:
~~This branch is currently branched on the 'external functions' branch (#550) but implementationwise, it's mostly independent.~~ Now branched on `main`

### Overview

The user can pull in externally defined datatypes like so:

```rust
#[verifier(external_type_specification)]
struct ProxyOption<V>( core::option::Option<V> );
```

This tells Verus to support `core::option::Option<V>`. For the most part, the "proxy type" mechanism works similar to the "proxy function" mechanism in #550. The user can supply Verus attributes on the proxy type, which are applied to the external type. The internal representation, handling of conflicts, and so on is similar to in #550. That is, the entry in the `KrateX`'s `datatypes` field will have a `name` pointing to `core::option::Option` and a `proxy` field pointing to `ProxyOption` which is just used for user errors.

One key difference is that for datatypes, we support datatypes _transparently_. That is:

 * For functions, every `external_fn_specification` was implicitly `external_body` (though not marked as such), with the specification being "trusted".
 * For types, each type pulled in via `external_type_specification` may or may not be marked `external_body`.

### Implementation details

 * First, the handling of datatypes in `rust_to_vir_adts.rs` had to be changed to primarily rely on `rustc_middle` so that we can support enums defined outside the crate. As of this PR, datatype handling only relies on `rustc_hir` for the mode attributes. This isn't a problem for external types, because they can be assumed to be fully `exec`-mode.
 * As I was testing, I ran into a small quirk with the way we've been handling datatype visibility. Specifically, `DatatypeTransparency::Always` did not distinguish between "completely public" and "public to _this crate_". This was a problem since these values would get serialized as VIR when handling multi-crates, although it didn't really come up in practice since we weren't defining any datatypes in the root of `vstd`. To address this, I made these changes:
   * `DatatypeTransparency` is now either `Never` or `WhenVisible(visibility)`. The `visibility` can be fully public (`restricted_module = None`) or can be restricted to a particular crate.
   * The `owning_module` field is moved from the `Visibility` datatype to `FunctionX` and `DatatypeX`.

### DIFF
--- a/source/pervasive/std_specs/core.rs
+++ b/source/pervasive/std_specs/core.rs
@@ -9,4 +9,17 @@ pub fn ex_swap<T>(a: &mut T, b: &mut T)
     core::mem::swap(a, b)
 }
 
+#[verifier(external_type_specification)]
+pub struct ExOption<V>(core::option::Option<V>);
+
+#[verifier(external_type_specification)]
+pub struct ExResult<T, E>(core::result::Result<T, E>);
+
+#[verifier(external_type_specification)]
+pub struct ExRange<Idx>(core::ops::Range<Idx>);
+
+#[verifier(external_type_specification)]
+#[verifier(external_body)]
+pub struct ExDuration(core::time::Duration);
+
 }
--- a/source/pervasive/std_specs/alloc.rs
+++ b/source/pervasive/std_specs/alloc.rs
@@ -1,0 +1,11 @@
+use crate::prelude::*;
+extern crate alloc;
+
+verus!{
+
+// TODO need to figure out how to handle the Allocator type param
+//#[verifier(external_type_specification)]
+//#[verifier(external_body)]
+//pub struct ExVec<V>(alloc::vec::Vec<V>);
+
+}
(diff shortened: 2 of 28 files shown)


source/pervasive/std_specs/core.rs
```diff
@@ -9,4 +9,17 @@ pub fn ex_swap<T>(a: &mut T, b: &mut T)
...
+pub struct ExOption<V>(core::option::Option<V>);
+
+#[verifier(external_type_specification)]
+pub struct ExResult<T, E>(core::result::Result<T, E>);
+
+#[verifier(external_type_specification)]
+pub struct ExRange<Idx>(core::ops::Range<Idx>);
+
+#[verifier(external_type_specification)]
+#[verifier(external_body)]
```
Chris-Hawblitzel, line 22: Does `external_body` add anything here?  `Duration` already has no public fields.

Chris-Hawblitzel (approved):

source/pervasive/std_specs/core.rs
tjhance: actually, the only reason I added this at all was that I wanted to test this particular combination of features in a cross-crate context

Issue #248, opened 2022-08-10
`#[verifier(inline)]` produces trigger selection failures, and nonsensical error messages


utaal:
If `Map::contains_pair` is inlined, and you try and trigger on it, you get:

```
error: trigger must be a function call, a field access, or a bitwise operator
    --> /st/verus/verified-nrkernel/page-table/spec.rs:1048:50
     |
1048 |                 forall(|base: nat, p: MemRegion| (#[trigger] self.interp_of_entry(i).map.contains_pair(base, p)) ==> between(base, self.entry_base(i), self.entry_base(i+1))),
     |       
```

It's also not clear how you'd select a proper trigger in this case.

utaal:
Candidate fixes discussed in the meeting (@tchajed, @parno, @matthias-brun, @tjhance):

* only allow `#[verifier(inline)]` on functions where it's trivial to transfer the user requested function to the inlined expressions, e.g. restrict to only function applications;
* expand the error message and mention that part of the expression had been inlined and ask the user to provide triggers manually.

If we do not restrict the inlining feature, we need to be careful about where we introduce `#[verifier(inline)]` in the standard library, to not surprise the user.

tjhance:
One of the obstacles pointed out in the meeting today, is that if `map.contains_pair(k, v)` gets inlined as `map.dom().contains(k) && map.index(k) === v`, then it may not be possible to write a trigger that includes `v`.

One idea I had is that we don't actually _need_ to find a trigger for `v` in this situation, because `v` can be simplified out of the expression. If I were writing Verus code today, I would do this manually; I'd remove `v` as a quantified variable and substitute it for `map.index(k)`, and then specify the trigger as `map.dom().contains(k)`.

More specifically, we could have Verus follow the general simplification rule: suppose `forall a, b, ..., c :: EXPR` is a forall statement and `EXPR` uniquely determines `c` in terms of the other variables (e.g., because it includes `c === f(a, b)` as a hypothesis). Then we remove `c` as a quantified variable, substitute it for `f(a, b)` - and in particular, we don't require that any trigger pattern include `c`.

utaal:
[Verus all-hands]

* [ ] the name `inline` gives the wrong intuition (it's about the compiler? and it's not a performance optimization); consider `#[trigger_transparent]`, `#[pre_expand]`.
* [ ] tell the user that they cannot use certain `#[trigger_transparent]` functions as a trigger, at the call site.

PR #562, opened 2023-05-19
Use rustc's impl numbering in paths for more flexible handling of traits

Chris-Hawblitzel:
1. This pull request uses rustc-style paths like `vstd::set::impl&%0::empty` instead of `vstd::set::Set::empty`.  This disambiguates paths that used to be disallowed due to ambiguity.  (Note that this disambiguation relies on https://github.com/verus-lang/verus/pull/553/commits/a103e776c266adf30d296705a672b216cbc5216d for soundness.)  For example, the following is now allowed:

```
        trait T { spec fn f() -> bool; }
        struct S<A> { a: A }
        impl T for S<u8> { spec fn f() -> bool { true } }
        impl T for S<bool> { spec fn f() -> bool { false } }
        proof fn test() {
            assert(S::<u8>::f());
            assert(!S::<bool>::f());
        }
```

2. This pull request uses rustc's `Instance::resolve` rather than the old `method_map`.  This removes the reliance on `method_map` that was blocking https://github.com/verus-lang/verus/pull/327 .

3. This pull request implements `#[verifier::inline]` for trait methods (based on `Instance::resolve`).  For example, the `spec_index` implementation in this code is now actually inlined when called directly (it was never inlined before):

```
pub trait SliceAdditionalSpecFns<T> {
   spec fn spec_index(&self, i: int) -> T;
}

impl<T> SliceAdditionalSpecFns<T> for [T] {

    #[verifier(inline)]
    open spec fn spec_index(&self, i: int) -> T {
        self.view().index(i)
    }
}
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,3 +81,16 @@ VERUS_EXTRA_ARGS="--log-all" vargo test -p rust_verify_test --test refs -- --noc
 
 This will output the log files in `rust_verify_test/.verus-log`. Only run one test at
 a time when using this flag, so that the logs are not overwritten by other tests.
+
+## Other tips
+
+You can use `--vstd-no-verify` to skip verification of the `vstd` library. This is pretty useful if you're building or running tests a lot. Note that it will still _build_ `vstd`—it just skips the SMT step. For example:
+
+```
+# for building
+vargo build --vstd-no-verify
+# for tests
+vargo test --vstd-no-verify -p rust_verify_test --test <test file> <test name>
+```
+
+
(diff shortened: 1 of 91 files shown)


tjhance (approved):

utaal:
Not strictly relevant for this PR, but I think we should determine how to make `inline` less prone to cause issues, or make it "look more dangerous" so that it's mainly a tool for `vstd` developers and advanced users, due to https://github.com/verus-lang/verus/issues/248

utaal:
Is it correct that there's a `impl%<number>` per impl block? I.e. rust groups all defs in a single impl block in the same `impl%<number>` path?

PR #565, opened 2023-05-19
support function types ("FnDef")

tjhance:
Besides closures, rust also has "function types", called TypeKind::FnDef in rustc_middle. A FnDef type is basically a zero-sized singleton type that corresponds to some function.

By supporting them, we can support stuff like:

```rust
fn a<F: Fn(u8) -> bool>(f: F) { ... }

fn b(value: u8) -> bool { ... }

fn main() {
  a(b);

  // Previously you would have had to create a closure:
  a(|x: u8| b(x));
}
```

It also lets us reason about the requires/ensures of named functions e.g., `a.requires(...)` and `a.ensures(...)` using the `requires` and `ensures` trait functions introduced with exec-closures. It remains to be seen how useful this actually is, though I think it might give us a way to add extra specs to external traits like `Eq`. (The _actual_ reason I'm implementing this now is that I was trying to write guide documentation for exec-closures and thought it would be easier to explain if we support these too.)

## Implementation

A `FnDef` type basically just becomes a datatype like the unit tuple, with a singleton element.

We also need to emit axioms saying that `f.requires(...)` is equivalent to the `req%f` function we generate in AIR. Naturally, I'm only emitting these axioms for FnDef types that are actually instantiated by the user.

We also need to implement some kind of recursion checking to make sure the precondition of `f` doesn't recursively depend on `f.requires(...)` ~(this hasn't been implemented yet)~ See below

## Recursion checking

We need to do recursion checking to rule out stuff like this:

```rust
fn stuff()
  requires !stuff.requires(())
```

This means we need to augment recursion checking to check requires/ensures clauses for the sake of ensuring that spec code terminates. Right now, each function gets a single Node, given by the `Node::Fun` variant. We've got to split this node in two somehow.

~~I made the choice to add a new variant `Node::Exec` so that, when `f` is an exec-mode function, the `Node::Fun(f)` is the requires/ensures signature of the function, while `Node::Exec(f)` is the body. I made this choice under the theory that now `Node::Fun` is always "spec/proof" content while `Node::Exec` is always the executable content. However, it _does_ mean that the body of a proof function is on the `Fun` node while the body of the exec function is on the `Exec` mode, so now the recursion-checking code has to keep casing on the function mode. So, not entirely sure this was the right decision.~~

What we do is add a node that represents the trait bound `FnDef(f) : FnOnce(...) -> ...`. This trait bound represents the implementations of `call_requires(f, ...)` and `call_ensures(f, ...)`, that is, the requires and ensures clauses. You'll also see, in commands.rs, that that definitions of req% and ens% _for exec functions_ is moved to the new trait bound node.

This new trait bound node is represented by the existing `Node::TraitImpl` node. To do this, this PR generalizes the notion of "impl path". Rather than just being a path to a trait impl, it can be that _or_ a path to a function indicating the implies `FnDef(f) : FnOnce` trait bound. See the new `ImplPath` enum.

Now, if the user writes a cyclic definition `fn f() requires call_requires(f, ...)`, we'll end up with a cycle where the node
`Node::TraitImpl(ImplPath::FnDefPath(f))` points to itself, resulting in an error from existing trait recursion-checking code.

## Todo

 - [x] Fill in cases marked `todo!()` that resulted from rebasing (Mostly just need to understand what the new functions are supposed to do)
 - [x] Add additional tests for interactions with various features
 - [x] Fix known issues with recursion checking
   - [x] `TypX::FnDef` should take an ImplPaths type argument, similar to `TypX::Datatype` (this is done now, though it's actually unused because it's not possible to use a FnDef type in a struct definition)
   - [x] Update recursion checking to add the appropriate edges relating to ImplPaths
   - [x] Current implementation is inconsistent with the current expectations regarding partial-exec-termination-checking
   - [x] Don't ignore the FnOnce trait bound (which conceptually carries the requires/ensures in it)
 - [x] Support trait functions (i.e., functions defined in trait impl) - now that #562 is merged in, this is probably not much work or might already work as-is. Just need to write tests for it.
 - [x] Error messages need to properly distinguish between errors related to requires/ensures vs other trait cycles


tjhance:
Converting this back to a 'draft' because I realized there's a problem with trait functions

tjhance:
OK. This should be ready now

source/vir/src/ast.rs
```diff
@@ -159,6 +159,11 @@ pub enum TypX {
     Lambda(Typs, Typ),
     /// Executable function types (with a requires and ensures)
     AnonymousClosure(Typs, Typ, usize),
+    /// Corresponds to Rust's FnDef type
+    /// Typs are generic type args
+    /// If Fun is a trait function, then the Option<Fun> has the statically resolved
+    /// function if it exists.
+    FnDef(Fun, Typs, ImplPaths, Option<Fun>),
```
Chris-Hawblitzel, line 166: I didn't expect ImplPaths in TypX::FnDef, but I think it's actually not used.  Can it be removed?

tjhance: I put it there because TypX::Datatype has it as well (although I think that may also be unused?)

Chris-Hawblitzel: The TypX::Datatype one is used in one place for checking datatype definitions (`recursive_types::check_positive_uses`, although it's not actually about positive uses).  Since TypX::FnDef can't appear in datatype definitions, I think that won't need it.

Chris-Hawblitzel (approved):
My only remaining comment is the Node trait encoding that we discussed offline (and any associated cleanup).  I think everything else is ready to merge.

PR #566, opened 2023-05-21
Add unified extensional equality operators

Chris-Hawblitzel:
Previously, Verus's support for extensional equality was fragmented:
- Seq, Set, Map, Multiset each had their own `.ext_equal` method.
- FnSpec didn't have a `.ext_equal` method, but instead had separate `fun_ext_1`, `fun_ext_2`, etc. lemmas for each arity.
- structs and enums supported extensional equality via Z3's datatype mechanism, but this mechanism doesn't handle Seq, Set, Map, Multiset, and FnSpec values embedded inside datatype fields.  Support for this sort of embedded extensional equality (as implemented in Dafny) has been a highly requested feature for Verus.

This pull request unifies all extensional equality into operators that are shared among Seq, Set, Map, Multiset, FnSpec, concrete datatypes, and user-defined abstract datatypes.  There are now three levels of equality, in increasing order of automation and SMT solving expense:
- `==` means the SMT solver's equality (as before).
- `=~=` means extensional equality at the level of Dafny's `==` operator; it handles Seq/Set/Map/Multiset/FnSpec nested inside datatypes, but does not recursively handle, say, a Seq inside a Seq.
- `=~~=` is "deep" extensional equality that handles arbitrary nesting of Seq/Set/Map/Multiset/FnSpec and datatypes.  (This is beyond Dafny's `==` as far as I understand.)

For example, the `=~=` and `=~~=` axioms for `Seq` now say:

```
#[verifier(external_body)]
#[verifier(broadcast_forall)]
pub proof fn axiom_seq_ext_equal<A>(s1: Seq<A>, s2: Seq<A>)
    ensures
        #[trigger] (s1 =~= s2) <==> {
            &&& s1.len() == s2.len()
            &&& forall|i: int| 0 <= i < s1.len() ==> s1[i] == s2[i]
        },
{
}

#[verifier(external_body)]
#[verifier(broadcast_forall)]
pub proof fn axiom_seq_ext_equal_deep<A>(s1: Seq<A>, s2: Seq<A>)
    ensures
        #[trigger] (s1 =~~= s2) <==> {
            &&& s1.len() == s2.len()
            &&& forall|i: int| 0 <= i < s1.len() ==> s1[i] =~~= s2[i]
        },
{
}
```

These operators are used in the same way that the old `.ext_equal` methods or the Dafny `==` operator:

```
        proof fn test(s1: Seq<u8>, s2: Seq<u8>) {
            assume(s1.len() == 1);
            assume(s2.len() == 1);
            assume(s1[0] == s2[0]);
            assert(s1 =~= s2);
        }
        proof fn test_deep(s1: Seq<Seq<u8>>, s2: Seq<Seq<u8>>) {
            assume(s1.len() == 1);
            assume(s2.len() == 1);
            assume(s1[0].len() == 1);
            assume(s2[0].len() == 1);
            assume(s1[0][0] == s2[0][0]);
            assert(s1 =~~= s2);
        }
```

This pull request is designed to be "pay for what you use" in terms of SMT performance.  It doesn't change the SMT encoding of the regular `==`, and it doesn't change the SMT encoding for datatypes by default.  Datatypes still have regular SMT extensional equality via Z3 datatypes, but they can also opt in to the additional `=~=` and `=~~=` functionality using the `#[verifier::ext_equal]` attribute in case they need extensionality on Seq/Set/Map/Multiset/FnSpec fields or fields of other `#[verifier::ext_equal]` datatypes:

```
#[verifier::ext_equal]
struct S<A> {
    a: Seq<u8>,
    b: u16,
    c: A,
}

proof fn test3(x: S<Seq<bool>>, y: S<Seq<bool>>) {
    assume(x.a.len() == 1);
    assume(y.a.len() == 1);
    assume(x.a[0] == y.a[0]);
    assume(x.b == y.b);
    assume(x.c.len() == 1);
    assume(y.c.len() == 1);
    assume(x.c[0] == y.c[0]);
    assert(x =~= y);
}
```

Semantically, `=~=` and `=~~=` are simply defined to be equivalent to `==` (the only difference is in triggering), and there is an axiom in the AIR prelude stating this equivalence:

```
        (declare-fun [ext_eq] (Bool [typ] [Poly] [Poly]) Bool)
        (axiom (forall ((deep Bool) (t [typ]) (x [Poly]) (y [Poly])) (!
            (= (= x y) ([ext_eq] deep t x y))
            :pattern (([ext_eq] deep t x y))
        )))
```

Because of this equivalence, untrusted abstract collection datatypes can provide their own lemmas that establish `=~=` and `=~~=` by establishing ordinary `==` equality.  Here is an example of an abstract sequence-of-4-elements type that provides its own `=~=` and `=~~=` lemmas:

```
mod m1 {
    use vstd::prelude::*;
    #[verifier::ext_equal]
    pub struct Seq4<A>(A, A, A, A);
    impl<A> Seq4<A> {
        pub closed spec fn spec_index(&self, i: int) -> A {
            if i == 0 {
                self.0
            } else if i == 1 {
                self.1
            } else if i == 2 {
                self.2
            } else {
                self.3
            }
        }
    }
    // TODO: make this broadcast_forall
    pub proof fn lemma_ext_equal<A>(x: Seq4<A>, y: Seq4<A>)
        ensures
            x =~= y <==> (forall|i: int| 0 <= i < 4 ==> x[i] == y[i]),
    {
        if (forall|i: int| 0 <= i < 4 ==> x[i] == y[i]) {
            assert(x[0] == y[0]);
            assert(x[1] == y[1]);
            assert(x[2] == y[2]);
            assert(x[3] == y[3]);
        }
    }
    // TODO: make this broadcast_forall
    pub proof fn lemma_ext_equal_deep<A>(x: Seq4<A>, y: Seq4<A>)
        ensures
            x =~~= y <==> (forall|i: int| 0 <= i < 4 ==> x[i] =~~= y[i]),
    {
        lemma_ext_equal(x, y);
    }
}
mod m2 {
    use vstd::prelude::*;
    use crate::m1::*;
    proof fn test(x: Seq4<u8>, y: Seq4<u8>)
        requires
            x[0] == y[0],
            x[1] == y[1],
            x[2] == y[2],
            x[3] == y[3],
    {
        lemma_ext_equal(x, y);
        assert(x == y);
    }
}
```

(When broadcast_forall is generalized, the client test program will simply be able to say `assert(x =~= y)` without having to call `lemma_ext_equal(x, y)` explicitly.)


source/vir/src/ast.rs
```diff
@@ -300,6 +300,14 @@ pub enum BinaryOp {
...
 }
 
+/// More complex unary operations (requires Clone rather than Copy)
+/// (Below, "boxed" refers to boxing types in the SMT encoding, not the Rust Box type)
+#[derive(Clone, Debug, Serialize, Deserialize, Hash, ToDebugSNode)]
+pub enum BinaryOpr {
+    /// extensional equality ext_equal (true ==> deep extensionality)
+    ExtEq(bool, Typ),
+}
+
```
tjhance, line 310: I know this matches what we did for Unary, but is there any purpose to this split?

Chris-Hawblitzel: It was mainly a convenience for some code that traverses AST and SST, where the code traverses into `UnaryOpr`'s op but not `Unary`'s op:

```
        ExpX::Unary(op, e1) => ok_exp(ExpX::Unary(*op, fe(env, e1)?)),
        ExpX::UnaryOpr(op, e1) => {
            let op = match op {
                UnaryOpr::Box(t) => UnaryOpr::Box(ft(env, t)?),
                UnaryOpr::Unbox(t) => UnaryOpr::Unbox(ft(env, t)?),
                ...
```

It's not a big deal, though, and we could get the same convenience in other ways if we wanted.

tjhance (approved):

Chris-Hawblitzel:
During a discussion this morning, someone made a good point about the terminology: we're not introducing multiple semantic notions of equality -- semantically, "extensional equality" is just the same equality as `==`.  So rather than using the term "extensional equality", it may be clearer to phrase this as "using extensionality to prove equality" or "proving equality via extensionality".

utaal (approved):

PR #592, opened 2023-06-08
Specify functionality for vec, option, result in vstd

tjhance:
depends on #555 and #584

### Details

This PR deprecates `vstd::vec`, `vstd::option`, and `vstd::result` and adds support for `std::vec`, `std::option`, and `std::result`.

It also changes `vstd::prelude` to no longer export these by default. The expectation is to eventually remove `vstd::vec`, `vstd::option`, and `vstd::result` if everybody is able to switch with no problems.

**To switch to the new API:** Make sure you are _not_ importing `vstd::vec`, `vstd::result`, or `vstd::option`. Also make the following changes:

 * Replace `vec.index(i)` with `&vec[i]` (or `*vec.index(i)` with `vec[i]`)
 * Replace `vec.pop()` with `vec.pop().unwrap()`
 * Replace `vec.swap(i, value)` with `vec.set_and_swap(i, value)`

**To continue using the old API:** Make sure to explicitly import the old API with `use vstd::{vec::*, result::*, option::*};`


source/pervasive/std_specs/alloc.rs
```diff
@@ -1,11 +1,221 @@
...
 
-// TODO need to figure out how to handle the Allocator type param
-//#[verifier(external_type_specification)]
-//#[verifier(external_body)]
-//pub struct ExVec<V>(alloc::vec::Vec<V>);
+#[verifier(external_type_specification)]
+#[verifier(external_body)]
+#[verifier::accept_recursive_types(T)]
+#[verifier::reject_recursive_types(A)]
+pub struct ExVec<T, A: Allocator>(Vec<T, A>);
```
Chris-Hawblitzel, line 17: Should the `Vec` declarations go in `std_specs/vec.rs`?  They might be hard for people to find in `alloc.rs`.

source/pervasive/std_specs/alloc.rs
```diff
@@ -1,11 +1,221 @@
...
+            *value == old(self)@.index(i as int)
+    {
+        core::mem::swap(&mut self[i], value);
+    }
+}
+
+// TODO this should really be a 'external_fn_specification' function
+// but it's difficult to handle vec.index right now because
+// it uses more trait polymorphism than we can handle right now.
+// And `index` is too important that I don't want do
```
Chris-Hawblitzel, line 75: typo in comment

Chris-Hawblitzel (approved):

utaal (approved):

PR #595, opened 2023-06-08
Add vec append method

hayley-leblanc:


source/pervasive/vec.rs
```diff
@@ -160,6 +160,15 @@ impl<A> Vec<A> {
     {
         self.vec.clear();
     }
+
+    #[verifier(external_body)]
+    pub fn append(&mut self, other: &mut Vec<A>)
+        ensures
+            self@ == old(self)@ + other@,
```
tchajed, line 167: I think you mean this?
```suggestion
            self@ == old(self)@ + old(other)@,
```

hayley-leblanc: Oops, thank you!

tjhance (approved):

PR #597, opened 2023-06-09
Nonlinear arithmetic and bitvector document

chanheec:
Starting from the old documentation(https://github.com/verus-lang/verus/blob/main/source/docs/bv_experiments/bit-vector-reference.md) (https://github.com/verus-lang/verus/blob/main/source/docs/non_linear_arithmetic/NonLinear_Singular.md), this PR adds non-linear arithmetic and bit-vector documentation.

depends on https://github.com/verus-lang/verus/pull/635

### DIFF
--- a/source/docs/guide/src/nonlinear_bitvec.md
+++ b/source/docs/guide/src/nonlinear_bitvec.md
@@ -10,12 +10,37 @@ Nonlinear arithmetic involves equations that multiply, divide, or take the remai
 is undecideable in general, meaning that general-purpose SMT solvers like Z3 can only make a best-effort attempt
 to solve them.  These attempts rely on heuristics that can be unpredictable.  Hence, by default, Verus 
 disables Z3's nonlinear arithmetic heuristics.  When you need to prove such properties, Verus offers the two dedicated
-proof strategies described below.  The first can reliably prove a limited subset of nonlinear properties.  For properties
+proof strategies described below.  First, `integer_ring` feature can reliably prove a limited subset of nonlinear properties.  For properties
 outside that subset, Verus offers a way to invoke Z3's nonlinear heuristics in a way that will hopefully provide
 better reliability.
 
-## 1. Proving Ring-based Properties with Singular (optional)
-***WARNING: This feature is currently under maintenance; this feature might be broken.***
+## 1. Proving General Properties with Z3 
+To prove a nonlinear formula that cannot be solved using `integer_ring` feature,
+you can selectively turn on Z3's nonlinear reasoning heuristics.
+As described below, you can do this either inline in the midst of a larger
+function, or in a dedicated proof function.
+
+### Inline Proofs with `assert(...) by(nonlinear_arith)`
+To prove a nonlinear property in the midst of a larger function,
+you can write `assert(...) by(nonlinear_arith)`.  This creates
+a separate Z3 query just to prove the asserted property,
+and for this query, Z3 runs with its nonlinear heuristics enabled.
+The query does not include ambient facts (e.g., knowledge that stems
+from the surrounding function's `requires` clause
+or from preceding variable assignments) other than each variable's type invariants
+(e.g., the fact that a `nat` is non-negative).  To include additional
+context in the query, you can specify it in a `requires` clause for the `assert`,
+as shown below.
+```rust
+{{#include ../../../rust_verify/example/guide/nonlinear_bitvec.rs:bound_checking}}
+```
+
+### Modular Proofs with `proof fn ... by(nonlinear_arith)`
+You can also use `by(nonlinear_arith)` in a proof function's signature. By including `by(nonlinear_arith)`, the query for this function runs with nonlinear arithmetic reasoning enabled.
+
+
+
+## 2. Proving Ring-based Properties with Singular
 
 While general nonlinear formulas cannot be solved consistently, certain
 sub-classes of nonlinear formulas can be.  For example, nonlinear formulas that
@@ -25,9 +50,8 @@ divisor `n`).  As a simple example, we might like to show that `a % n == b % n
 
 Verus offers a deterministic proof strategy to discharge such obligations.
 As shown below, to use this strategy, you must state the desired property
-as a proof function annotated with `#[verifier(integer_ring)]`.
+as a proof function annotated with `by(integer_ring)`.
 
-(TODO: add example based on the equations above)
 
 Verus will then discharge the proof obligation using a dedicated algebra solver
 called [Singular](https://www.singular.uni-kl.de/).  As hinted at by the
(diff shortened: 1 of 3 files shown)


parno:
Thanks for working on incorporating this documentation into the Guide!  I checked in a few edits.  There are a few TODOs that should be cleaned up before merging.  For example, it looks like we need an example of using `#[verifier(integer_ring)]`.  On a related note, it looks like we support `proof fn ... by(bit_vector)`; should we also support `proof fn ... by(integer_ring)` instead of using an annotation?

In general, it would be nice to include inline-code, rather than pointing to files.  If the code is too large to inline, we should link to the appropriate file, rather than just giving the path.

parno (commented):
@hayley-leblanc Thanks for adding these examples!  As a small note, I think that if you start the code blocks with
```
```rust
```
we'll get slightly nicer formatting when `mdbook` runs.

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+Nonlinear arithmetic involves equations that multiply, divide, or take the remainder of integer variables 
+(e.g., `x * (y * z) == (x * y) * z`).  As discussed earlier in this guide, determining the truth of such formulas
+is undecideable in general, meaning that general-purpose SMT solvers like Z3 can only make a best-effort attempt
+to solve them.  These attempts rely on heuristics that can be unpredictable.  Hence, by default, Verus 
+disables Z3's nonlinear arithmetic heuristics.  When you need to prove such properties, Verus offers the two dedicated
+proof strategies described below.  First, `integer_ring` feature can reliably prove a limited subset of nonlinear properties.  For properties
+outside that subset, Verus offers a way to invoke Z3's nonlinear heuristics in a way that will hopefully provide
+better reliability.
+
+## 1. Proving General Properties with Z3
```
parno, line 17: It seems odd to start with the general case, rather than the specific case, especially since the last two sentences above do it in the other order.  Hence, I would swap the order of Z3 vs. Singular.

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+As described below, to tackle these properties, Verus offers several dedicated proof strategies.
+
+# Nonlinear Arithmetic in Verus
+
+Nonlinear arithmetic involves equations that multiply, divide, or take the remainder of integer variables 
+(e.g., `x * (y * z) == (x * y) * z`).  As discussed earlier in this guide, determining the truth of such formulas
+is undecideable in general, meaning that general-purpose SMT solvers like Z3 can only make a best-effort attempt
+to solve them.  These attempts rely on heuristics that can be unpredictable.  Hence, by default, Verus 
+disables Z3's nonlinear arithmetic heuristics.  When you need to prove such properties, Verus offers the two dedicated
+proof strategies described below.  First, `integer_ring` feature can reliably prove a limited subset of nonlinear properties.  For properties
```
parno, line 13: "First" -> "First, the"

source/rust_verify/example/guide/nonlinear_bitvec.rs
```diff
@@ -0,0 +1,66 @@
+#[allow(unused_imports)]
+use builtin_macros::*;
+#[allow(unused_imports)]
+use builtin::*;
+
+verus! {
+
+// ANCHOR: bound_checking
```
parno, line 8: This example is confusing.  What is the `mul` function for?  Why do we want to prove this?  What are the assertions on lines 22 and 23 doing?  Are they necessary for the proof or are they making some other point?  

Perhaps we should use an inline version of a lemma like [LemmaMulUpperBound](https://github.com/verus-lang/verus/blob/2dc6a17858e241ded90617375d9ce2784576afb7/source/rust_verify/example/integer_ring/integer_ring_bound_check.rs#L24)

Also, instead of saying
```
// note that y <= 0xffff is not available inside this proof block
```
perhaps:
```
// assert(y <= 0xffff);    <- Fails, since this property is not included in the `requires` clause
```

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+
+
+
+
+### Details/Limitations
+- This can be used only with **int** parameters.
+- Formulas that involve inequalities are not supported.   
+- Division is not supported.
+- Function calls in the formulas are treated as uninterpreted functions.  If a function definition is important for the proof, you should unfold the definition of the function in the proof function's `requires` clause.
+- When using `integer_ring` lemma, the divisor of `%` operator must not be zero. If a divisor can be zero in the ensures clause of the `integer_ring` lemma, the fact in ensures clause will not be available in the callsite.
```
parno, line 89: "When using" -> "When using an"

parno: "the divisor of `%` operator " -> "the divisor of a modulus operator (`%`)"
" the fact in ensures " -> " the facts in the ensures "
"in the " -> "at the"

parno: The bit about 0 divisors seems like an unfortunate user trap.  Could we automatically add a `requires` clause to any uses of `integer_ring` that insists that all divisors in the ensures clause (and other requires clauses?) are non-zero?

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Using `integer_ring` as a helper lemma for nonlinear proofs
+As `integer_ring` feature has several limitations, it is not possible to get an arbitary nonlinear property only with `integer_ring` feature. Instead, it is a common pattern to have `by(nonlinear_arith)` function as a main lemma for the desired property, and use `integer_ring` lemma as a helper function.
```
parno, line 92: "As" -> "As the", "only with" -> "only with the", "to have" -> "to have a"

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+
+To work around the lack of support for inequalities and division, you can often write a helper proof discharged with `integer_ring` and use it to prove properties that are not directly supported by `integer_ring`. Furthermore, you can also add additional variables to the formulas. For example, to workaround division, one can introduce `c` where `b = a * c`, instead of `b/a`. Experiences indicate that introducing many additional variables can help with the proof.
```
parno, line 94: "to workaround" -> "to work around"

parno: > Experiences indicate that introducing many additional variables can help with the proof.

I'm not sure what this is trying to communicate.  I should just add a bunch of random variables to my proof?

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Example 1: `integer_ring` as a helper lemma to provide facts on modular arithmetic
+In this below `lemma_mod_difference_equal` function, we have four inequalities inside the requires clauses, which cannot be encoded into `integer_ring`. In the ensures clause, we want to prove `y % d - x % d == y - x`. The helper lemma `lemma_mod_difference_equal_helper` simply provides that `y % d - x % d` is equal to `(y - x)` modulo `d`. The rest of the proof is done by `by(nonlinear_arith)`.
```
parno, line 97: Here and in the next paragraph:
"In this below `lemma_mod_difference_equal` function" -> "In the `lemma_mod_difference_equal` function below,"

parno: "is done by" -> "is done via"

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+{
+    let small_x = x % d;
+    let small_y = y % d;
+    let tmp1 = (small_y - small_x) % d;
+    let tmp2 = (y - x) % d;
+    lemma_mod_difference_equal_helper(x,y,d, small_x, small_y, tmp1, tmp2);
+}
+```
+
+In the below `lemma_mod_between`, we want to prove that `x % d <= z % d < y % d`. However, `integer_ring` only supports equalities, so we need cannot prove `lemma_mod_between` directly. Instead, we provide facts that can help assist the proof. The helper lemma provides 1) `x % d - y % d == x - y  (mod d)` and 2) ` y % d - z % d == y - z  (mod d)`. The rest of the proof is done by `by(nonlinear_arith)`.
```
parno, line 126: "we need" -> "we"
"is done by" -> "is done via"

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+
+    let tmp2 = (small_z - small_y) % d;
+    lemma_mod_between_helper(z,y,d, small_z, small_y, tmp2);    
+}
+```
+
+
+#### Example 2: Proving properties on bounded integers with the help of `integer_ring`
+
+Since `integer_ring` proofs only support `int`, you need to include explicit bounds when you want to prove properties about bounded integers. For example, in order to use the proof `lemma_mod_after_mul` on `u32`s, `lemma_mod_after_mul_u32` must ensure that all arguments are within the proper bounds before passing them to `lemma_mod_after_mul`.
```
parno, line 164: "For example, " -> "For example, as shown below,"

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+
+#### Example 2: Proving properties on bounded integers with the help of `integer_ring`
+
+Since `integer_ring` proofs only support `int`, you need to include explicit bounds when you want to prove properties about bounded integers. For example, in order to use the proof `lemma_mod_after_mul` on `u32`s, `lemma_mod_after_mul_u32` must ensure that all arguments are within the proper bounds before passing them to `lemma_mod_after_mul`.  
+
+If a necessary bound (e.g., `m > 0`) is not included, Verus will fail to verify the proof.
+
```
parno, line 167: I'm not sure I understand this example.  It seems like `lemma_mod_after_mul` is a generic property, and `lemma_mod_after_mul_32` is just a specialization?  I'm not sure what lesson is being taught.

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+        z <= 0xffff,
+        m <= 0xffff,
+    ensures (x*z - y*z) % (m as int) == 0
+{ 
+  lemma_mod_after_mul(x as int, y as int, z as int, m as int);
+  // rest of proof body omitted for space
+}
+```
+
+The desired property for `nat` can be proved similarly. Note that we introduce several additional variables(`ab`, `bc`, and `abc`) to help with the `integer_ring` proof.
```
parno, line 190: If I'm interpreting this correctly, this paragraph should be split into one piece that finishes the discussion of the example above, and a new piece that introduces the next example.   E.g.,:

> The desired property for `nat` can be proved similarly.
>
> The next example is similar, but note that we introduce several additional variables(`ab`, `bc`, and `abc`) to help with the `integer_ring` proof.

parno: It's unclear to me why the variable introductions are needed in the example.  The advice earlier was to introduce additional variables to deal with division.

source/docs/guide/src/nonlinear_bitvec.md
```diff
@@ -0,0 +1,287 @@
...
+- `let` binding
+- Quantifiers
+- `if-then-else` 
+
+Note that function calls are not supported. As a workaround, you may consider using a macro instead of a function. 
+
+
+### Bitwise Operators As Uninterpreted Functions
+Outside of `by(bit_vector)`, bitwise operators are translated into uninterpreted functions in Z3, meaning Z3 knows nothing about them when used in other contexts. 
+As a consequence, basic properties such as the commutativity and associativity of bitwise-AND will not be applied automatically. To make use of these properties, please refer to [this example file](https://github.com/verus-lang/verus/blob/main/source/rust_verify/example/bitvector_basic.rs), which contains basic properties for bitwise operators.
```
parno, line 281: Someday, it would be nice to put a version of that file in `vstd`, ideally with some macro magic so that we can easily create versions for all of the different integer sizes.

parno (commented):
Thanks for all of the work you've put into the documentation.  I've add a variety of mostly small comments and suggestions.

PR #599, opened 2023-06-10
Version flag

ahuoguo:


jaybosamiya (changes requested):
Thanks for the PR! This is a good step towards us getting a useful `--version` from Verus!

A few suggested changes:
+ Rather than using `VERUS_BUILD_SHA` and `VERUS_BUILD_DIRTY` separately as env vars, and then passing that along, I recommend just a single `VERUS_BUILD_VERSION` that contains both of them in it, taking care of adding the dirty info to it directly.
+ I would also prefer if the build date could be included in the version string. I don't have any strong preferences on whether the build date should also be included separately in the output or not, but I do feel that including it in the `VERUS_BUILD_VERSION` string would be good, since it would allow for sorting if necessary.
+ Having the full SHA hash isn't crucial, so we can shorten the SHA hash a bit while doing still being useful

To also account for [semver](https://semver.org/) and potential future 1.x release, we might want to keep everything under a 0.x version. Also accounting for date, sha-hash, and (optional) dirty flag (and making sure everything gets sorted/ordered correctly) leads to a version format that looks like `0.2023.06.12-61abcc0` (for a clean build of [this commit](https://github.com/verus-lang/verus/commit/61abcc0c273999802f0d1a466f947ba37fcc48d8)) or like `0.2023.06.12-61abcc0.dirty` (for a dirty build of the same).

Thus, the output for `verus --version` might look something like:
```
Verus
  Platform: macos_arm64
  Version: 0.2023.06.12-61abcc0
```

Also, instead of `option_env!("VERUS_BUILD_VERSION").unwrap_or_default()` you probably want something like `option_env!("VERUS_BUILD_VERSION").unwrap_or("Unknown")`, so that if the environment variable doesn't exist, you get the string "Unknown" rather than an empty string in the output.

Finally, I don't know how it'll behave on Windows, and if the potential panics (when trying to run the git commands) could break the build on it (I expect Linux and MacOS users to have those commands work, just am concerned about Windows since I don't know enough about how people tend to set things up on it); one way to guarantee handling this with good fallback is to have failures succeed silently (instead just not set the `VERUS_BUILD..` environment vars); then that way the build won't fail, but will just lead to a less-useful `--version` info.

source/rust_verify/src/main.rs
```diff
@@ -27,6 +27,15 @@ pub fn main() {
     let internal_program = internal_args.next().unwrap();
     let build_test_mode = if let Some(first_arg) = internal_args.next() {
```
utaal, line 28: `internal_args` is intended for calls to `verus` originated by the build system, or verus itself (for alternative drivers for verifying the standard library, or our mechanism to run the borrow checker only on proof code). Can you please add `version` to the args in https://github.com/verus-lang/verus/blob/2285a9d3769c530ed0461251e82694327ce641de/source/rust_verify/src/config.rs#L32
and parse it from https://github.com/verus-lang/verus/blob/2285a9d3769c530ed0461251e82694327ce641de/source/rust_verify/src/config.rs#L110 instead?

utaal (changes requested):
Hi! Thanks for the PR. I have a couple of comments of my own to try and keep things consistent with how we otherwise handle arguments and for the communication with `vargo`.
Here's my first comment; I'll get back to you in a bit with a more precise note on the env var from `vargo` and how to hopefully do something similar to `VARGO_BUILD_PROFILE`.

source/rust_verify/src/config.rs
```diff
@@ -251,9 +254,9 @@ pub fn parse_args(program: &String, args: impl Iterator<Item = String>) -> (Args
                 print_usage();
                 std::process::exit(0);
             }
-            if m.free.len() == 0 {
-                print_usage();
-                std::process::exit(-1);
+            if m.free.len() == 0 && !m.opt_present("version") {
+            print_usage();
+            std::process::exit(-1);
```
jaybosamiya, line 259: Why did the indentation change here?

source/rust_verify/src/main.rs
```diff
@@ -99,6 +99,16 @@ pub fn main() {
     let program = if build_test_mode { internal_program } else { args.next().unwrap() };
     let (our_args, rustc_args) = rust_verify::config::parse_args(&program, args);
     let pervasive_path = our_args.pervasive_path.clone();
+    
+    if our_args.version {
+        println!("Verus");
+        println!("  Platform: {}_{}", std::env::consts::OS, std::env::consts::ARCH);
+
+        let date = option_env!("VERUS_BUILD_VERSION").unwrap_or("Unknown");
+        println!("  Version: {}", date);
```
jaybosamiya, line 108: This isn't the date, but the entire version, so probably should not be called `date`

tools/vargo/src/main.rs
```diff
@@ -140,6 +140,13 @@ const Z3_FILE_NAME: &str = if cfg!(target_os = "windows") {
 };
 
 fn run() -> Result<(), String> {
+    let version_info = match util::commit_info() {
+        Ok(info) => info,
+        Err(_) => format!("could not get commit info"),
+    };
+
+    std::env::set_var("VERUS_BUILD_VERSION", version_info);
```
jaybosamiya, line 148: The other side of this already marks things as `unknown` if the environment variable is not set, so there is no reason to set the env var in the situation where it couldn't get the info.

Instead, you could do something like this:
```suggestion
    if let Ok(version_info) = util::commit_info() {
      std::env::set_var("VERUS_BUILD_VERSION", version_info);
    }
```

Also, the function is getting you version information as a whole, not just commit info, so probably good to call it `version_info` instead of `commit_info`?

tools/vargo/src/util.rs
```diff
@@ -66,3 +66,65 @@ pub fn mtime_recursive(path: &Path) -> Result<FileTime, String> {
         .unwrap_or_else(|| FileTime::from_last_modification_time(&meta));
     Ok(max_meta)
 }
+
+pub fn commit_info() -> Result<String, std::io::Error> {
```
jaybosamiya, line 70: Since this function's err value is going to be ignored by the caller, it is better to have it return an `Option<String>` rather than a `Result<String, ...>`

You'll have to change up the current `?` places to `.ok()?` to account for this change, but `Option` should better indicate what the intention of this function is.

tools/vargo/src/util.rs
```diff
@@ -66,3 +66,65 @@ pub fn mtime_recursive(path: &Path) -> Result<FileTime, String> {
...
+        ))
+    } else {
+        Ok(format!(
+            "0.{}.{}.{}-{} (dirty)",
+            year,
+            month,
+            day,
+            String::from_utf8(sha_msg).unwrap()
+        ))
+    }
```
jaybosamiya, line 129: A somewhat cleaner way to handle this:
```suggestion
    let dirty = if status.success() { "" } else { ".dirty" };
    Ok(format!("0.{year}.{month}.{day}{dirty}"))
```

source/rust_verify/src/config.rs
```diff
@@ -150,8 +151,10 @@ pub fn parse_args(program: &String, args: impl Iterator<Item = String>) -> (Args
     const OPT_PROFILE_ALL: &str = "profile-all";
     const OPT_COMPILE: &str = "compile";
     const OPT_NO_SOLVER_VERSION_CHECK: &str = "no-solver-version-check";
+    const OPT_VERSION: &str = "version";
 
     let mut opts = Options::new();
+    opts.optflag("", OPT_VERSION, "print version information");
```
jaybosamiya, line 157: For consistency with the capitalization of other options
```suggestion
    opts.optflag("", OPT_VERSION, "Print version information");
```

jaybosamiya (changes requested):
Thanks for making the updates @ahuoguo; some more comments added, mostly just minor changes.

@utaal thanks for reviewing it; I was planning to ping you for a review once it was in a nicer shape anyways (one important question was whether this version info should be grabbed by vargo, or by the `build.rs`; I _think_ having vargo do that makes sense, but wanted to confirm), but yeah I appreciate you taking a look at it already

jaybosamiya (approved):
Looks good to me, but we should wait for @utaal's comments

Also, I would prefer if someone on Windows (@Chris-Hawblitzel?) can checkout and build to confirm that this isn't breaking the build (our CI doesn't test on Windows, and since this affects the build process, might be good to confirm that it still builds, before we commit it into `main`)

utaal:
I have a few suggestions, conveniently in PR form: https://github.com/ahuoguo/verus/pull/2
If those all look good, you can merge that PR (it's based against this PR's branch, so merging would append the commit to this PR).

One thing I did is to restore the full sha, so that we can always (with e.g. a script) uniquely identify a version, even in the (unlikely) presence of a short-sha collision. @jaybosamiya does that make sense?

> Also, I would prefer if someone on Windows (@Chris-Hawblitzel?) can checkout and build to confirm that this isn't breaking the build (our CI doesn't test on Windows, and since this affects the build process, might be good to confirm that it still builds, before we commit it into `main`)

This would indeed be good, but FWIW we do now run a build and "smoke test" on Windows (e.g. https://github.com/verus-lang/verus/actions/runs/5274309721/jobs/9538648267) to catch changes that would break the Windows build (which I did, previously, a couple times :).

utaal (approved):
Looks good to me, but please clean up the history before merging.

Edit: As discussed, please re-add `--short` to `git rev-parse HEAD`. Thank you!

jaybosamiya:
Summarizing an out-of-band discussion (w/ @utaal) on the SHA hash in the version:
+ Having a shorter version string is useful, so that people can get a chance to notice the `.dirty` if it exists
+ Longer SHA is better because uniqueness
+ Purely by numbers (and birthday collision problem), 7 hex chars is _probably_ good enough anyways, due to needing ~16k commits to accidentally hit a collision
+ But `--short` (from `git rev-parse`) gives us a unique prefix! (thanks @utaal; TIL) In particular, it will give a longer length if necessary
+ _Technically_, since HEAD is baked in at time of the build, it _could_ become ambiguous in the future, but this is almost certainly ok, unless folks are actively trying to break version numbering (and even then, they'll only break a pre-built copy of an older version)

TLDR: we should reintroduce `--short` (probably explicitly use `--short=7` to guarantee that as a minimum length, rather than relying on the system default, which can vary based on user/OS/package-manager defaults)

jaybosamiya:
Re the Windows smoke test, I hadn't noticed this was added recently, but nice addition to prevent the kinds of problems we were worried about, thanks for pointing it out!

jaybosamiya (approved):
Modulo minor changes (commented), looks good to me

Also you might find the removal-of-comma commit to be a bit funny: 12345678f30137b5a4899f56d695d1a2a3b08124

Issue #588, opened 2023-06-07
Error defining a predicate on nat and Seq


zhengyao-lin:
Got an error when trying to define a predicate
```
use vstd::{prelude::*, seq::*};

verus! {

spec fn pred(a: nat, s: Seq<int>) -> bool
{
    a < s.len()
}

#[verifier(external)]
fn main() {}

} // verus!
```
Error:
```
error: The verifier does not yet support the following Rust feature: method call to method not defined in this crate
 --> test.rs:7:5
  |
7 |     a < s.len()
  |     ^
```

If I replace one of `a` and `s.len()` with a constant, then it works.

Chris-Hawblitzel:
Here's a workaround for the moment: `#[allow(unused_imports)] use builtin::*;`

utaal:
Triaged with @Chris-Hawblitzel.
I've landed a temporary hack as a workaround. We're working on a more permanent fix.

utaal:
We overhauled the way we match items in https://github.com/verus-lang/verus/pull/605, which should address cases like this.

PR #605, opened 2023-06-13
Verus items

utaal:
This PR replaces all the string-based matching to determine whether a `def_id` is a specific verus or rust item (a "lang item", in rust terminology). (`TyCtxt::def_path_str`, which we often used previously, can return different paths for the same item that's re-exported (we think depending on what's imported) which makes it unsuitable for this. This bug https://github.com/verus-lang/verus/issues/588 was caused by one such failed match.)

This PR primarily introduces two types and related functions that identify verus and rust items (respectively): `VerusItem` and `RustItem`:

* Because we control `builtin` and `vstd`, `VerusItem`s are found using `rustc_diagnostic_item`s added to the definitions; `verus_items.rs` maps the rustc_diagnostic_items strings to enum variants, returned by `verus_items.id_to_name.get`.
* For rust items, we can sometimes use lang items (when available), otherwise we fall back to a new function that prints an item's path in a stable manner (`ty_to_stable_string_partial`): it doesn't support all items, and doesn't attempt to be friendly, it's written specifically to return stable string identifiers for rust items that we need to match.

Because `fn_call_to_vir` heavily relied on `def_path_str` it has been heavily refactored to now match on the `VerusItem`, and moved to a new file (`rust_to_vir_expr` was getting very large). There are probably more opportunities to refactor `def_path_str` away from its current pattern of a long list of conditionals and early returns, but this PR is already quite big, so we can attack that later.

We used to also rely on `rustc_diagnostic_item` to assign well-known paths to `vstd` items referenced in `vir`. To avoid a potentially confusing name spacing issue with `rustc_diagnostic_item`, these well-known paths are now assigned by `verus_items` (some variants carry and `Ident`, which is the canonical path).

I've also replaced all other uses I could find of `def_path_str`, `def_path_debug_str`, and similar. As far as I can tell, we'll have no more string-based item matching in the codebase after this PR is merged, and I think we should consider it a bug if any are found later.


source/rust_verify/src/verus_items.rs
```diff
@@ -0,0 +1,638 @@
...
+                one_impl_block_in_path = true;
+                let self_ty = tcx.type_of(tcx.parent(def_id));
+                match ty_to_stable_string_partial(tcx, &self_ty)? {
+                    TyToStableStringPartialResult::Path(path) => {
+                        segments.clear();
+                        segments.push(path);
+                    }
+                    TyToStableStringPartialResult::Segment(segment) => {
+                        segments.push(segment);
+                    }
```
tjhance, line 98: I don't totally get the logic here - if the type is something like `bool`, then we get a 'Segment', and if the type is an ADT we get a `Path` ... why do those types have different behavior? I'd have expected we'd want to do TYPE_PATH::method_name in either case.

utaal: Because the ADT printer will print its full path, so we replace the path we've seen so far with the ADT's full path. This is similar to what we do for friendly names, but more focused to produce stable identifier.

utaal: I'm not sure I understand the question, but let me be more precise, what I said before wasn't quite right: when we print the path for a impl function, we want to print the "self" type path, and not the path with the disambiguator (`impl%0`), so when I reach an `impl` in the path I call out to the printer function recursively, but having replaced the `def_id` with that of the the "self" type. When that happens, we want to throw away the path constructed so far, and use the one for the "self" type instead.

tjhance: the TyToStableStringPartialResult::Path case - which does what you just described - makes sense to me. I'm confused about the TyToStableStringPartialResult::Segment case, which doesn't clear the path.

utaal: Ah! Good catch. That case doesn't really make sense if we want to stay close to the rust paths.
I had a bit of a broken model of what was happening because the paths I saw made sense (`core::num::u64::MAX` vs `u64::MAX`), so I didn't see what you meant initially.

utaal: Fixed in the commit I just pushed. Feel free to un-resolve if something still doesn't make sense.

tjhance: Ok, the path construction makes more sense now, but now that it's changed, I'm wondering, is `u8` properly unambiguous on its own? Maybe ty_to_stable_string_partial should return `"core::num::u64"`?

utaal: Why would it be ambiguous? i.e. what could it collide with? (I think `core::num::u64` is the deprecated path, see https://doc.rust-lang.org/std/u64/constant.MAX.html)

tjhance: oh, i see. i'm not really sure, then

tjhance (approved):

Chris-Hawblitzel (approved):

source/rust_verify/src/fn_call_to_vir.rs
```diff
@@ -0,0 +1,1461 @@
...
+                    None
+                } {
+                    return mk_expr(ExprX::RevealString(Arc::new(s.to_string())));
+                } else {
+                    return err_span(args[0].span, "string literal expected".to_string());
+                }
+            }
+        },
+        Some(VerusItem::Expr(expr_item)) => match expr_item {
+            crate::verus_items::ExprItem::Choose => {
```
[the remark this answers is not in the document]
utaal: They are. Thanks!

source/rust_verify/src/verus_items.rs
tjhance: wait, actually, i do think it's possible for users to have crates called 'bool' or 'u8' or something.

is there a check that the rust item really lives in core or std?

PR #612, opened 2023-06-15
Error report

ahuoguo:
This PR implements an error-reporting tool into Verus to simplify the process of reporting unexpected Verus errors/bugs. It essentially takes a “snapshot” of a given file – including the file itself, all of the file’s dependencies, the user’s Z3 version and Verus version, the command line arguments, and stdout/stderr information – and compresses it into a zip file that preserves the original file hierarchy. All relevant information is recorded into a .toml file that is zipped in with the dependencies. The created zip file (named YYYY-MM-DD-hh-mm-ss.zip) will be saved to the current working directory.

This PR puts the flag `--record` into Verus itself. With this PR, if I am in directory `bar` and want to report a Verus bug that I am experiencing in `foo.rs`, I could call in the command line `verus --record foo.rs`. This would produce a zip file in the directory `bar` that includes `foo.rs`, any dependency of `foo.rs`, and a file called `error_report.toml` that would look something like the following:

```
report-schema-version = "1"
title = "Error report file - details and dependencies"

[command-line-arguments]
args = "verus rust_verify/example/playground.rs"

[verification-results]
encountered-vir-error = false
errors = 0
success = true
verified = 3

[versions]
z3-version = """
Z3 version 4.12.2 - 64 bit
"""

[versions.verus-version]
commit = "07f6e9959e4398c67a2307e4bc8123662580960a"
profile = "debug"
toolchain = "1.68.0-aarch64-apple-darwin"
version = "0.2023.07.14.07f6e99"

[versions.verus-version.platform]
arch = "aarch64"
os = "macos"

[verus-output]
stderr = """
...
"""

[verus-output.stdout.times-ms]
...

[verus-time]
verus-time = 0.58403575
```

Usage: `verus --record INPUT [options]`

### DIFF
--- a/source/error_report/src/main.rs
+++ b/source/error_report/src/main.rs
@@ -13,6 +13,9 @@ use toml::{map::Map, value::Value};
 use zip::write::FileOptions;
 
 fn main() {
+    if cfg!(windows) && !yansi::Paint::enable_windows_ascii() {
+        yansi::Paint::disable();
+    }
     match run() {
         Ok(()) => (),
         Err(err) => {
@@ -32,10 +35,13 @@ fn run() -> Result<(), String> {
     if args.len() > 1 {
         for argument in &args {
             if argument.ends_with(".rs") {
+                if file_path.is_some() {
+                    Err("multiple .rs files passed, unsupported by the --record flag")?;
+                }
                 file_path = Some(argument.clone());
             }
             if argument.starts_with("-o") || argument.starts_with("--out-dir") {
-                Err("error report does not support `-o` or `--out-dir` flag")?;
+                Err("--record does not support `-o` or `--out-dir` flag")?;
             }
         }
         our_args = args[2..].to_vec();
@@ -110,7 +116,7 @@ fn run() -> Result<(), String> {
         Err(err) => {
             // remove temp file if created
             fs::remove_file(&temp_dep_file).map_err(|x| {
-                format!("failed to delete toml file with this error message: {}", x)
+                format!("failed to delete `.d` file with this error message: {}", x)
             })?;
             Err(format!(
                 "failed to execute verus with error message {}, verus path is at {:?}, args are {:?}",
@@ -126,7 +132,7 @@ fn run() -> Result<(), String> {
         Err(err) => {
             // remove temp file if created
             fs::remove_file(&temp_dep_file).map_err(|x| {
-                format!("failed to delete toml file with this error message: {}", x)
+                format!("failed to delete `.d` file with this error message: {}", x)
             })?;
             Err(err)?
         }
--- a/source/docs/guide/src/reference-flag-record.md
+++ b/source/docs/guide/src/reference-flag-record.md
@@ -1,7 +1,6 @@
 # Record flag
 
-Verus is still under development, so when you get into an error
-it can be quite frustrating to manually record the versions of your z3 and verus to reproduce the error. This tool attempts to record this data automatically.
+Sometimes, you might wish to record an execution trace of Verus to share, along with all the necessary dependencies to reproduce an execution.
+This might be useful for either packaging up your verified project, or to report a Verus bug to the [issue tracker](https://github.com/verus-lang/verus/issues).
 
-The `--record` flag will rerun verus and package source files of the current crate to the current directory, alongside with output and version information. The file will be named YYYY-MM-DD-HH-MM-SS.zip. If you are reporting an error, please keep the original arguments in addition to this flag 
-
+The `--record` flag will do precisely this.  In particular, to record an execution of Verus (say, `verus foo --bar --baz`), simply add the `--record` flag (for example, `verus foo --bar --baz --record`).  This will re-run Verus, and package all the relevant source files, along with the execution output and version information into a zip file (`yyyy-mm-dd-hh-mm-ss.zip`) in your current directory.


source/error_report/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    let date = local.to_string();
+    let mut zip_file_name = date[0..19].to_string();
+    zip_file_name.push_str(".zip");
+    zip_file_name.replace_range(10..11, "-");
+    zip_file_name.replace_range(13..14, "-");
+    zip_file_name.replace_range(16..17, "-");
+
+    let path = std::path::Path::new(&zip_file_name);
+    let file = std::fs::File::create(path).unwrap();
+    //let deps = vec!["folder/a.rs", "src/b.rs", "src/c.rs", "src/main.rs"];
```
jaybosamiya, line 188: Potential old comment left behind?

source/error_report/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    let mut zip = zip::ZipWriter::new(file);
+    let options = FileOptions::default()
+        .compression_method(zip::CompressionMethod::Stored)
```
jaybosamiya, line 191: Actually allow the zip file to compress things, rather than just store things
```suggestion
        .compression_method(zip::CompressionMethod::Bzip2)
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+    zip.finish().expect("Could not finish up zip file");
+    zip_file_name
+}
+
+///Turns a file path into a string
+fn read_file_string(filepath: &str) -> Result<String, Box<dyn std::error::Error>> {
+    let data = fs::read_to_string(filepath)?;
+    Ok(data)
+}
```
jaybosamiya, line 209: Isn't this exactly equivalent to `fs::read_to_string(filepath)`? Not sure why this is a separate function

source/rust_verify/src/main.rs
```diff
@@ -110,6 +110,27 @@ pub fn main() {
...
 
+    if our_args.error_report {
+        let mut args: Vec<String> = std::env::args().collect();
+        args.remove(0);
+
+        let index = args.iter().position(|x| *x == "--error-report").unwrap();
+        args.remove(index);
+
+        let verus_path = std::env::current_exe().unwrap();
+        let exe = verus_path.parent().unwrap().join("../../error_report/target/debug/error_report");
```
jaybosamiya, line 121: This expects `error_report` to have been built with debug-only; might want to consider checking whether the debug or release builds are around, and pick the right one.

source/rust_verify/src/main.rs
```diff
@@ -110,6 +110,27 @@ pub fn main() {
...
+
+        let mut res = std::process::Command::new(exe)
+            // .current_dir(std::env::current_dir().unwrap())
```
jaybosamiya, line 124: Out of curiosity, why is this comment here? I'd assume `current_dir` to be set to the current dir anyways (which is what this comment implies)

source/rust_verify/src/util.rs
```diff
@@ -153,4 +153,4 @@ pub const fn const_str_equal(lhs: &str, rhs: &str) -> bool {
     }
     true
 }
-// ==================================================================================================
+// ==================================================================================================
```
jaybosamiya, line 156: Not sure why this diff showed up, should be cleaned up

jaybosamiya (commented):
Thanks @elaustell and @ahuoguo; this is a good first draft for the error reporting functionality. I've added a bunch of comments inline that should be addressed, before we consider merging this (also please make sure to see the "hidden comments" that GitHub collapses down too)

Also, the main PR message says "Note: this error-report relies on an old version of --version flag. This issue will be addressed after --version flag is incorporated."; is this still the case, or has the `--version` version (heh) been fixed up.

Thanks again! And looking forward to have `--error-report` merged in soon!

source/error_report/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    let child = Command::new(verus_path)
+        .stdin(Stdio::null())
+        .args(our_args)
+        .arg("--emit=dep-info")
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("failed to execute process");
+    let verus_output: std::process::Output =
```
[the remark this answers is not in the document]
elaustell: Maybe but the output itself doesn't have to do with the dep info, this output relates specifically to what verus outputs to stdout and stderr when run on the given file. So I'm not sure the name verus_dep_info_output is accurate.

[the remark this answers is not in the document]
jaybosamiya: Oh indeed; ignore this comment.

source/error_report/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[workspace]
+members = [
+]
+
```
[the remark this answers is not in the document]
jaybosamiya: Ah, I see! Well this is now replaced out with the `vargo build` changes, so I guess it is resolved either way.

source/error_report/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    let args: Vec<String> = env::args().collect();
+    if args.len() > 1 {
+        for argument in &args {
+            if argument.ends_with(".rs") {
+                file_path = argument.clone();
+            }
+        }
+        our_args = args[2..].to_vec();
+    } else {
+        println!("Usage: error_report <file_name>");
```
jaybosamiya, line 29: Now that the expected call is `verus --error-report <args..>`, it should probably be reflected here:
```suggestion
        println!("Usage: verus --error-report <args..>");
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+        return;
+    }
+
+    let program_dir = args[1].clone();
```
jaybosamiya, line 33: I assume this is guaranteed to be passed in via the implementation of the `--error-report` flag, but that is still an assumption being made, so prob should be documented as a comment

source/error_report/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    let z3_path = Path::new(&program_dir).join("z3");
+    let verus_path = Path::new(&program_dir).join("verus");
+
+    let mut verus_call = args[2..].to_vec();
```
jaybosamiya, line 38: `args[2..]` is placed into `our_args`; maybe reuse that via a `clone` or such?

source/error_report/src/main.rs
jaybosamiya: On further thought, maybe this should be passed in as an environment variable? For example `VERUS_HOME` or such?

source/rust_verify/src/config.rs
```diff
@@ -261,6 +263,11 @@ pub fn parse_args_with_imports(
     );
 
     opts.optflag("h", "help", "print this help menu");
+    opts.optflag(
+        "",
+        OPT_ERROR_REPORT,
+        "Create zip file to reproduce verus error (with version info)",
```
jaybosamiya, line 269:
```suggestion
        "Zip relevant files triggering a Verus error to aid in reporting it",
```

source/rust_verify/src/main.rs
```diff
@@ -124,6 +124,45 @@ pub fn main() {
...
+        args.remove(0);
+
+        let index = args.iter().position(|x| *x == "--error-report").unwrap();
+        args.remove(index);
+
+        // error_report binary will always add --time flag
+        if let Some(index) = args.iter().position(|x| *x == "--time")
+        {
+            args.remove(index);
+        }
```
jaybosamiya, line 138: This probably should be handled by the `error_report` binary?

jaybosamiya (commented):
Some more comments added inline

source/error_report/src/main.rs
jaybosamiya: Fair, I think it is fine to leave it as an argument here

source/error_report/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+/// Turns the .d file that lists each of the input files' dependencies
+/// and turns them into a vector of Strings for easier data manipulation
+///
+/// Parameters: file_name: The name of the previously generated .d file
+///
+/// Returns:    a vector containing each dependency of the input file as an
+///             individual string
+fn d_to_vec(file_name: String) -> Vec<String> {
```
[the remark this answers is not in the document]
jaybosamiya: that's somewhat surprising, but I think a `String` is fine for now

source/rust_verify/src/main.rs
jaybosamiya: The intention with placing the `--time` removal in the `error_report` binary is that we want to report (to the `error_report` binary) something that fully captures the command the user used for Verus, as closely as possible; the fact that an `--error-report` flag was passed is only a way to invoke the `error_report` binary; but whether the user _explicitly_ passed in `--time` or not _may_ potentially be something that we might want/need to actually track (for `--time` I guess it is not _as_ crucial, but hopefully you see my reasoning on why the two flags are somewhat different in their removal)

source/rust_verify/src/main.rs
```diff
@@ -124,6 +124,44 @@ pub fn main() {
...
+        let exe: std::path::PathBuf;
+        let release_exe = verus_path.parent().unwrap().join("../../target/release/error_report");
+        let debug_exe = verus_path.parent().unwrap().join("../../target/debug/error_report");
+        if release_exe.exists() {
+            exe = release_exe;
+        } else if debug_exe.exists() {
+            exe = debug_exe;
+        } else {
+            panic!(
+                "error_report not found, try running `cargo build --release` in source/error_report"
```
jaybosamiya, line 149:
```suggestion
                "error_report executable not found; try running `vargo build --release`?"
```

source/rust_verify/src/main.rs
```diff
@@ -124,6 +124,44 @@ pub fn main() {
...
+        let mut args: Vec<String> = std::env::args().collect();
+        args.remove(0);
+
+        let index = args.iter().position(|x| *x == "--error-report").unwrap();
+        args.remove(index);
+
+        // error_report binary will always add --time flag
+        if let Some(index) = args.iter().position(|x| *x == "--time") {
+            args.remove(index);
+        }
```
jaybosamiya, line 137: Copying over comments from https://github.com/verus-lang/verus/pull/612#discussion_r1254769157 which is now marked as outdated:

@jaybosamiya:
> This probably should be handled by the `error_report` binary?

@ahuoguo:
> I removing `--error-report` flag at the same time, so these two might want to stay in the same level?

@jaybosamiya:
> The intention with placing the `--time` removal in the `error_report` binary is that we want to report (to the `error_report` binary) something that fully captures the command the user used for Verus, as closely as possible; the fact that an `--error-report` flag was passed is only a way to invoke the `error_report` binary; but whether the user _explicitly_ passed in `--time` or not _may_ potentially be something that we might want/need to actually track (for `--time` I guess it is not _as_ crucial, but hopefully you see my reasoning on why the two flags are somewhat different in their removal)

jaybosamiya (commented):
Thanks for making the updates.

Modulo above minor comments above (for the message in a panic, and the place for `--time`), I think the PR is in reasonable-enough shape for folks to start trying out on `main`. I still think some more cleanup can be done, but imho that shouldn't block this PR from being merged and being used by folks, so will do that at a separate point in time.  I think I can approve the PR once those changes are made.

Wrt merging itself, I think we should get at least an approval from @utaal since this touches command line arguments and such.  `source/error_report` doesn't actually impact regular Verus usage, so the total scale of changes to review should be fairly minimal.

Sidenote: when squash-merging this PR (after approvals), feel free to drop me off the co-author list; my changes in the changeset were fairly minor fixes here.

utaal:
Thank you for working on this! I have a few high-level comments and some specifics.

1. I wasn't completely comfortable with having to emit a `.d` file and having to find it by name, but I've had a look at what rustc libraries offer and it seems there isn't a straightforward way to obtain this information otherwise (I believe they're computed here: https://github.com/verus-lang/verus/blob/main/dependencies/rustc_interface/src/passes.rs#L549-L662) We might want to try and ask rustc for the `.d` output path (via `OutputFilenames::path`), instead of guessing it but this isn't critical.

2. Generally speaking, we should only use `.expect` (on `Option` and `Result`) for cases we never expect to happen (and `.unwrap` only for trivial cases, e.g. when you've just otherwise checked that an `Option` is `Some`). For example, invoking `verus` with `--error-report` and an invalid command line argument results in a panic (because `error_report` cannot find the `.d` file.)

3. We cannot expect that `error_report` is at `../../target/(debug|release)/error_report` relative to the `verus` path, the regular `target` can get clobbered by successive builds (in fact, this fails if you run `vargo run -p rust_verify -- --error-report rust_verify/example/playground.rs`). The only stable directory is `verus-target/(debug|release)` (which gets wiped by vargo if it becomes stale or incomplete). `vargo` should copy `error_report` into `verus-target/(debug|release)/.`, like it does for, e.g. `rust_verify` (https://github.com/verus-lang/verus/blob/main/tools/vargo/src/main.rs#L674-L675); then `verus` can find it via the `verus_root`, computed here https://github.com/verus-lang/verus/blob/main/source/rust_verify/src/main.rs#L91

utaal:
4. The way rustc collects dependencies may be surprising (i.e. files mentioned by a macro); shall we (a.) gracefully handle a case in which we cannot find one of the files, and (b.) print a list of files included in the archive? Also, I think it would be good to mention which information is included in the archive (in brief) in the output, so that the user knows what they're about to submit (they can obviously go ahead and check the zip file themselves).

5. The outputs for `time` and `version` are currently just text, which would make it harder to parse later. If we use `--output-json` (which I've recently updated for this purpose) we can include its output directly in the toml, ideally by parsing the json and re-emitting it as toml. This way the toml file can be parsed directly by a tool. (This may sound tricky but should be relatively straightforward; let me know if I can help).

source/error_report/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+        our_args = args[2..].to_vec();
+    } else {
+        println!("Usage: verus --error-report <args..>");
+        return;
+    }
+
+    // assumption: when verus is invoking error_report, the dir of verus path should be put in the first argument
+    let program_dir = args[1].clone();
+
+    let z3_path = Path::new(&program_dir).join("z3");
```
utaal, line 36: This may not be the version of `z3` used by `verus`, as it depends on whether `VERUS_Z3_PATH` is set. We can have `verus` pass this information along to `error-report`.

source/error_report/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+/// Creates a zip file from a given list of files to compress
+///
+/// Parameters: deps: A vector of strings representing files to be compressed
+///                    (in this context, each file is a dependency of the input)
+///
+/// Returns:    The name of the created zip file
+fn write_zip_archive(deps: Vec<String>) -> String {
+    let local: DateTime<Local> = Local::now();
+    let date = local.to_string();
+    let mut zip_file_name = date[0..19].to_string();
```
utaal, line 200: Can we construct the format we want, instead of just truncating the output of `to_string` and replacing characters?

source/error_report/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+    zip_file_name.push_str(".zip");
+    zip_file_name = zip_file_name.replace(" ", "-");
+    zip_file_name = zip_file_name.replace(":", "-");
+
+    let path = std::path::Path::new(&zip_file_name);
+    let file = std::fs::File::create(path).unwrap();
+    let mut zip = zip::ZipWriter::new(file);
+    let options = FileOptions::default()
+        .compression_method(zip::CompressionMethod::Bzip2)
+        .unix_permissions(0o644);
```
utaal, line 210: Why does this need to be set explicitly? What happens on Windows?

utaal:
Also, thoughts on naming this `--package-report`, or similar? (It isn't necessarily only useful for errors, and it'd be good to communicate that it "packages" the report, that the user is in charge of sending/uploading.)

source/error_report/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+        verus_call,
+        z3_version_output,
+        verus_version_output,
+        verus_output,
+        verus_duration,
+    );
+    let (dependencies_path, zip_path) = zip_setup(file_path);
+    println!("Stored error report to {}\n", zip_path);
+
+    fs::remove_file("error_report.toml").expect("failed to delete toml file\n");
```
utaal, line 79: If we panic between the time these files are created and when this line is reached, they won't be cleaned up (this has happened to me while trying this out). Error handling with `Result` should help with this.

parno:
> > Also, thoughts on naming this `--package-report`, or similar? (It isn't necessarily only useful for errors, and it'd be good to communicate that it "packages" the report, that the user is in charge of sending/uploading.)
> 
> Shed by the lights of Jon, I suggest `--report`

`--report` could still reasonably be interpreted as doing the reporting, i.e., sending information off to a Verus server somewhere.  `--package-report` makes it clearer that the option is just assembling the report, not doing anything with it.  If we want something more concise, perhaps `--collate` or `--collect`?

utaal:
Recording offline conversation: @ahuoguo, @jaybosamiya, and @parno suggested `--record`. That sounds good to me.

source/error_report/README.md
utaal, line 1: I'd suggest replacing this with a section in the guide. It's unlikely that users will find it here.

source/error_report/src/main.rs
```diff
@@ -0,0 +1,320 @@
...
+        Ok(()) => (),
+        Err(err) => {
+            eprintln!("{}", yansi::Paint::red(format!("error: {}", err)));
+            std::process::exit(1);
+        }
+    }
+}
+
+fn run() -> Result<(), String> {
+    let mut file_path = String::new();
```
utaal, line 26: Not a big deal, but the idiomatic rust to do is probably for this to be `Option<String>`. Then your check later can be:

```rust
let Some(file_path) = file_path else { return Err(...); };
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,320 @@
...
+            .to_string_lossy()
+            .to_string(),
+    );
+
+    let z3_version_output =
+        Command::new(z3_path.clone()).arg("--version").output().map_err(|x| {
+            format!("failed to execute z3 with error message {}, path is at {:?}", x, z3_path)
+        })?;
+
+    // mandating --time and --output-json flag, we remove both flags here to prevent verus two many options
```
utaal, line 84:
```suggestion
    // mandating --time and --output-json flag, we remove both flags here to prevent passing duplicates to verus
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,320 @@
...
+    let version_info: toml::Value = serde_json::from_str(&stdout_json["verus"].to_string())
+        .map_err(|x| {
+            format!("failed to parse version info to toml file with error message {}", x)
+        })?;
+
+    // get verification result
+    let verification_result: toml::map::Map<String, Value> =
+        serde_json::from_str(&stdout_json["verification-results"].to_string()).map_err(|x| {
+            format!(
+                "failed to pretty print verification result to toml file with error message {}",
```
utaal, line 177:
```suggestion
                "failed to parse verification results with error message {}",
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,320 @@
...
+                x
+            )
+        })?;
+
+    // the serde_json stdout is pretty bad
+    let stdout_toml_text = serde_json::from_str(&stdout_json.to_string()).map_err(|x| {
+        format!("failed to pretty print stdout to toml file with error message {}", x)
```
utaal, line 184: Like above, the failure here is parsing the json, not pretty-printing.

source/rust_verify/src/config.rs
```diff
@@ -265,6 +267,11 @@ pub fn parse_args_with_imports(
     );
 
     opts.optflag("h", "help", "print this help menu");
+    opts.optflag(
+        "",
+        OPT_ERROR_REPORT,
+        "Rerun verus that zip relevant files of the crate to the current directory. The file will be named YYYY-MM-DD-HH-MM-SS.zip. If you are reporting an error, please keep the original arguments in addition to this flag",
```
utaal, line 273:
```suggestion
        "Rerun verus and package source files of the current crate to the current directory, alongside with output and version information. The file will be named YYYY-MM-DD-HH-MM-SS.zip. If you are reporting an error, please keep the original arguments in addition to this flag",
```

source/rust_verify/src/config.rs
```diff
@@ -163,6 +164,7 @@ pub fn parse_args_with_imports(
     const OPT_COMPILE: &str = "compile";
     const OPT_NO_SOLVER_VERSION_CHECK: &str = "no-solver-version-check";
     const OPT_VERSION: &str = "version";
+    const OPT_ERROR_REPORT: &str = "record";
```
utaal, line 167:
```suggestion
    const OPT_RECORD: &str = "record";
```

source/rust_verify/src/config.rs
```diff
@@ -402,6 +409,7 @@ pub fn parse_args_with_imports(
         compile: matches.opt_present(OPT_COMPILE),
         no_vstd,
         solver_version_check: !matches.opt_present(OPT_NO_SOLVER_VERSION_CHECK),
+        error_report: matches.opt_present(OPT_ERROR_REPORT),
```
utaal, line 412:
```suggestion
        error_report: matches.opt_present(OPT_RECORD),
```

source/rust_verify/src/config.rs
```diff
@@ -265,6 +267,11 @@ pub fn parse_args_with_imports(
     );
 
     opts.optflag("h", "help", "print this help menu");
+    opts.optflag(
+        "",
+        OPT_ERROR_REPORT,
```
utaal, line 272:
```suggestion
        OPT_RECORD,
```

source/rust_verify/src/config.rs
```diff
@@ -71,6 +71,7 @@ pub struct ArgsX {
     pub no_vstd: bool,
     pub compile: bool,
     pub solver_version_check: bool,
+    pub error_report: bool,
```
utaal, line 74:
```suggestion
    pub record: bool,
```

source/rust_verify/src/main.rs
```diff
@@ -127,6 +127,33 @@ pub fn main() {
 
     let pervasive_path = our_args.pervasive_path.clone();
 
+    if our_args.error_report {
```
utaal, line 130:
```suggestion
    if our_args.record {
```

source/rust_verify/src/main.rs
```diff
@@ -127,6 +127,33 @@ pub fn main() {
...
+                let mut res = std::process::Command::new(exe)
+                    .arg(&verusroot.path)
+                    .args(args)
+                    .spawn()
+                    .expect("running error_report");
+
+                res.wait().expect("error_report failed to run");
+                return;
+            }
+        }
```
utaal, line 153: Don't fail silently, either panic (which I think is okay here, as you can expect a non-broken verus build) or print a message that says what went wrong.

utaal (approved):
Thank you for addressing the comments.
I have added a couple minor comments/suggestions. Please address those, then feel free to squash and merge. (If GitHub suggests adding me as a co-author because of the suggestions, please remove me.)

source/error_report/src/main.rs
```diff
@@ -0,0 +1,302 @@
...
+        .arg("--output-json")
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .output()
+    {
+        Ok(output) => output,
+        Err(err) => {
+            // remove temp file if created
+            fs::remove_file(&temp_dep_file).map_err(|x| {
+                format!("failed to delete toml file with this error message: {}", x)
```
jaybosamiya, line 113:
```suggestion
                format!("failed to delete `.d` file with this error message: {}", x)
```

source/error_report/src/main.rs
```diff
@@ -0,0 +1,302 @@
...
+    };
+
+    let verus_duration = start_time.elapsed();
+
+    match toml_setup_and_write(verus_call, z3_version_output, verus_output, verus_duration) {
+        Ok(()) => (),
+        Err(err) => {
+            // remove temp file if created
+            fs::remove_file(&temp_dep_file).map_err(|x| {
+                format!("failed to delete toml file with this error message: {}", x)
```
jaybosamiya, line 129:
```suggestion
                format!("failed to delete `.d` file with this error message: {}", x)
```

jaybosamiya (approved):
Thanks for all the updates; some minor remaining suggestions (all of which should just be easily accepted suggestions). Feel free to squash-and-merge once you get a CI green after making the updates based on the suggestions.

PR #616, opened 2023-06-16
Support nightly built rust toolchain

ziqiaozhou:
@utaal Would it be okay to change your script to support the nightly built rust toolchain?
I have tested and confirmed that the nightly-2023-01-20 would provide exact compiler API for verus.


utaal (approved):
I made a couple of minor changes, but otherwise this should work for now.
Thanks!

Issue #619, opened 2023-06-21
&F for exec-functions doesn't work


tjhance:
This should work, but doesn't:

```rust
fn merge_sorted_with<T: Copy, F>(left: Vec<T>, right: Vec<T>, leq: &F) 
    where F: Fn(T, T) -> bool
    requires left.len() > 0, right.len() > 0,
        forall |x, y| leq.requires((x, y)) 
{
    let x = leq(left[0], right[0]);
}
```

```
error: Call to non-static function fails to satisfy `callee.requires(args)`
  --> ms.rs:19:13
   |
19 |     let x = leq(left[0], right[0]);
   |             ^^^^^^^^^^^^^^^^^^^^^^
```

If you remove the & from `&F`, it does work:

```rust
fn merge_sorted_with<T: Copy, F>(left: Vec<T>, right: Vec<T>, leq: &F) 
    where F: Fn(T, T) -> bool
    requires left.len() > 0, right.len() > 0,
        forall |x, y| leq.requires((x, y)) 
{
    let x = leq(left[0], right[0]);
}
```

PR #621, opened 2023-06-22
explicitly handle 'adjustments' in rust_to_vir_expr

tjhance:
The rustc_hir AST `Expr` nodes don't contain implicit conversions, like auto-inserted * and & and so on. Instead, those are found in the "adjustments", which we can look up for each node. We've been ignoring them until now, which was _mostly_ fine, since we basically ignore * and & anyway, though it was kind of annoying to handle the implicit conversions when I added support for [] recently. The adjustments also include auto derefs, which _are_ semantically meaningful, and I think it was probably unsound before because we were ignoring them.

Anyway, with this PR we now check the adjustments explicitly. Thus we handle `&` in a uniform way whether it's implicit or explicit, and we can error if we find any conversion that is unsupported.

TODO:

 - [x] Wait for the verus items PR #605 to be merged in first, to avoid rebasing problems
 - [x] fix tests related to exec-closures (also probably related to #619)

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -1082,9 +1082,6 @@ pub(crate) fn auto_deref_supported_for_ty<'tcx>(
     match ty.kind() {
         TyKind::Adt(AdtDef(adt_def_data), _args) => {
             let did = adt_def_data.did;
-            if Some(did) == tcx.lang_items().owned_box() {
-                return true;
-            }
             let rust_item = verus_items::get_rust_item(tcx, did);
             return matches!(rust_item, Some(RustItem::Box | RustItem::Rc | RustItem::Arc));
         }
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -1074,3 +1074,20 @@ pub(crate) fn check_generics_bounds_fun<'tcx>(
     let typ_params = typ_params.iter().map(|(x, _)| x.clone()).collect();
     Ok((Arc::new(typ_params), typ_bounds))
 }
+
+pub(crate) fn auto_deref_supported_for_ty<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    ty: &rustc_middle::ty::Ty<'tcx>,
+) -> bool {
+    match ty.kind() {
+        TyKind::Adt(AdtDef(adt_def_data), _args) => {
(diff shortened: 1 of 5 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -1069,3 +1069,20 @@ pub(crate) fn check_generics_bounds_fun<'tcx>(
...
+    ty: &rustc_middle::ty::Ty<'tcx>,
+) -> bool {
+    match ty.kind() {
+        TyKind::Adt(AdtDef(adt_def_data), _args) => {
+            let did = adt_def_data.did;
+            if Some(did) == tcx.lang_items().owned_box() {
+                return true;
+            }
+            let rust_item = verus_items::get_rust_item(tcx, did);
+            return matches!(rust_item, Some(RustItem::Box | RustItem::Rc | RustItem::Arc));
```
utaal, line 1084: I think the check with `owned_box` is redundant (the `matches!` `RustItem::Box` should have the same effect).

```suggestion
            let rust_item = verus_items::get_rust_item(tcx, did);
            return matches!(rust_item, Some(RustItem::Box | RustItem::Rc | RustItem::Arc));
```

source/rust_verify_test/tests/refs.rs
```diff
@@ -485,3 +485,33 @@ test_verify_one_file! {
...
+test_verify_one_file! {
+    #[test] mutref_arg_ref_unsupported verus_code! {
+        fn stuff(x: &mut &u8) { }
+
+        fn test() {
+            let mut y: u8 = 0;
+            stuff(&mut &y); // this does NOT modify y
+        }
+    } => Err(err) => assert_vir_error_msg(err, "complex arguments to &mut parameters are currently unsupported")
+}
```
utaal, line 517: Are there tests we can add for the non-overloaded autoderef and for the auto-borrow cases?

source/vir/src/well_formed.rs
```diff
@@ -219,6 +219,7 @@ fn check_one_expr(
                     _ => false,
                 };
                 if !is_ok {
+                    println!("arg: {:#?}", arg);
```
utaal, line 222: Remove this?

source/rust_verify_test/tests/refs.rs
tjhance: what kind of test do you have in mind?

Chris-Hawblitzel (approved):

utaal (approved):

PR #622, opened 2023-06-23
Port over Dafny std library

elaustell:
Having a vast standard library can help prevent reinventing the wheel when it comes to proving behavior about built-in types. This PR will add a number of proven axioms and lemmas about Verus' seq, set, and map types (as well as a few for multisets). This will make it easier for Verus users to write proofs that use these types.

# Ported Dafny Lemmas

## Sequences

The following table demonstrates the sequence lemmas found in the Dafny library and the corresponding Verus equivalent. The verus equivalent will be implemented by this PR unless otherwise clarified to be already implemented.

| Dafny                  |    Verus         |
| ---------------------- | -----------------|
| LemmaLast              | lemma_add_last_back |
| LemmaAppendLast     |  lemma_append_last   |
| LemmaConcatIsAssociative | lemma_concat_associative |
| IndexingImpliesMembership | lemma_indexing_implies_membership |
| MembershipImpliesIndexing | lemma_membership_implies_indexing |
| LemmaSplitAt | lemma_split_at |
| LemmaElementFromSlice | lemma_element_from_slice |
| LemmaSliceOfSlice  | lemma_slice_of_slice |
| LemmaCardinalityOfSet | lemma_cardinality_of_set |
| LemmaCardinatliyOfEmptySetIs0 | lemma_cardinality_of_empty_set_is_0 |
| LemmaNoDuplicatesInConcat | lemma_no_dup_in_concat |
| LemmaCardinalityOfSetNoDuplicates | unique_seq_to_set (already implemented) |
| LemmaNoDuplicatesCardinalityOfSet | lemma_no_dup_set_cardinality |
| LemmaMultisetHasNoDuplicates | lemma_multiset_has_no_duplicates |
| LemmaZipOfUnzip | lemma_zip_of_unzip |
| LemmaMaxOfConcat | lemma_max_of_concat |
| LemmaMinOfConcat | lemma_min_of_concat |
| LemmaSubseqMax | lemma_subseq_max |
| LemmaSubseqMin | lemma_subseq_min |
| LemmaFlattenConcat | lemma_flatten_concat |
| LemmaFlattenReverseConcat | lemma_flatten_reverse_concat |
| LemmaFlattenAndFlattenReverseAreEquivalent | lemma_flatten_and_flatten_reverse_are_equivalent |
| LemmaFlattenLengthGeSingleElementLength | lemma_flatten_length_ge_single_element_length | LemmaFlattenLengthLeMul | lemma_flatten_length_le_mul |
| LemmaMapDistributesOverConcat | ignored |
| LemmaFilterDistributesOverConcat | ignored |
| LemmaFoldLeftDistributesOverConcat | ignored |
| LemmaInvFoldLeft | ignored |
| LemmaFoldRightDistributesOverConcat | ignored |
| LemmaInvFoldRight | ignored |
| SortedUnique | lemma_sorted_unique |

The ignored lemmas deal with higher-order sequence functions, which I was advised to ignore.

## Sets

| Dafny                  |    Verus         |
| ---------------------- | -----------------|
| LemmaSubset   | definition of subset_of (already implemented) |
| LemmaSubsetSize | lemma_len_subset (already implemented) |
| LemmaSubsetEquality | lemma_subset_equality |
| LemmaSingletonSize | ignored  |
| LemmaSingletonEquality | ignored |
| LemmaIsSingleton | ignored |
| LemmaMapSize | lemma_map_size |
| LemmaFilterSize | lemma_len_filter (already implemented) |
| LemmaUnionSize | lemma_len_union_ind |
| LemmaBoundedSetSize | lemma_int_range |
| LemmaGreatestImpliesMaximal | lemma_greatest_implies_maximal |
| LemmaLeastImpliesMinimal | lemma_least_implies_minimal |
| LemmaMaximalEquivalentGreatest | lemma_maximal_equivalent_greatest |
| LemmaMinimalEquivalentLeast | lemma_minimal_equivalent_least |
|  LemmaLeastIsUnique | lemma_least_is_unique |
|  LemmaGreatestIsUnique | lemma_greatest_is_unique |
| LemmaMinimalIsUnique | lemma_minimal_is_unique |
| LemmaMaximalIsUnique | lemma_maximal_is_unique |
| LemmaFindUniqueMinimal | spec fn find_unique_minimal and proof fn _find_unique_minimal_ensures |
| LemmaFindUniqueMaximal | spec fn find_unique_maximal and proof fn _find_unique_maximal_ensures |

## Maps

| Dafny                  |    Verus         |
| ---------------------- | -----------------|
| LemmaDisjointUnionSize | lemma_disjoint_union_size |
| LemmaInvertIsInjective | lemma_invert_is_injective |

# Ported Dafny Axioms

The Dafny prelude includes a vast number of axioms that relieve some responsibility from the user attempting to get proofs to pass. This PR implements these axioms in the form of proven lemmas. Because they are no longer axiomatized, the facts they provide are not automatically included in the context. The solution to this is a "magic" lemma that, when included in a proof body, brings in the same context as the Dafny prelude. These magic lemmas are so named:
- lemma_seq_properties
- lemma_set_properties
- lemma_map_properties
- lemma_multiset_properties
  
Lemmas brought in from the Dafny prelude axioms are currently labeled `This verified lemma used to be an axiom in the Dafny prelude`.

# Ported Dafny functions

This PR also implements a few more functions for `seq`, `set`, `map`, and `multiset` so that the user has a broader range of options and functionality when dealing with these built-in datatypes.

- Seq:
  - fill (constructor)
  - take
  - drop
  - first
  - is_prefix_of
  - is_suffix_of
  - sort_by for generic types and sort for ints
  - max_via for generic types and max for ints
  - min_via for generic types and min for ints
  - index_of_option
  - first_index_of and first_index_of_option
  - last_index_of and last_index_of_option
  - drop_first
  - to_multiset
  - multiplicity
  - remove_value
  - reverse
  - zip_with
  - unzip
  - flatten and flatten_reverse
- Set
  - to_seq and to_sorted_seq
  - is_singleton
  - find_unique_minimal and find_unique_maximal
  - to_multiset
- Map
  - values
  - len
  - is_equal_on_key
  - is_injective
  - invert
  - is_monotonic and is_monotonic_from
- Multiset
  - new (constructor)
  - update
  - intersection_with
  - difference_with
  - disjoint_with
  - dom


tjhance (changes requested):
I'm a little worried about adding so many broadcast_forall axioms into code that everybody will be importing. I know Dafny had some trouble with too many sequence-related triggers, causing some really slow verification times whenever you had too many sequence subranges.

In particular, I think `lemma_element_from_slice` looks especially worrying, since it will add `new[pos-a]` into the context when `new[pos]` matches.

And then there are a handful of others, like `lemma_split_at` or `lemma_slice_of_slice` that probably aren't that useful anyway because they prove equalities that would be easy for Verus anyway as long as the user is using `=~=`, which is pretty reliable.

For now, I think we should probably remove the `broadcast_forall` attribute, and make them  normal lemmas with proof bodies that are checked by Verus.

tjhance:
@Chris-Hawblitzel do you have a second opinion on what makes for good broadcast_forall axioms? My feeling is that the minimal-ish set we have right now is close to what we want, at least as far as the basic operations go (like `add` and `push` and whatnot).

Chris-Hawblitzel:
I think we should avoid adding broadcast_forall universally to basic operations like `append`, because lemmas like `lemma_concat_associative` can easily create a lot of quantifier instantiations.  Instead, in the long run, we should be able selectively import sets of lemmas using the proposed https://github.com/verus-lang/verus/discussions/37 .  So for now the safe thing to do is to omit the broadcast_foralls and wait for https://github.com/verus-lang/verus/discussions/37 .

source/pervasive/seq_lib.rs
```diff
@@ -365,6 +504,153 @@ impl<A> Seq<A> {
...
+        }
+    }
+
+    /// Flattens a sequence of sequences into a single sequence by concatenating 
+    /// subsequences in reverse order, i.e. starting from the last element.
+    pub open spec fn flatten_reverse(self) -> Seq<A>
+        decreases self.len()
+    {
+        if self.len() == 0 {Seq::empty()}
+        else {self.drop_last().flatten_reverse().add(self.last())}
```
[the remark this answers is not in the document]
elaustell: I'm not so sure about that. Consider the sequence `[[1, 2, 3], [4, 5, 6], [7, 8, 9]]`. Flatten_reverse should result in the sequence `[7, 8, 9, 4, 5, 6, 1, 2, 3]`. Flatten and then reverse would result in `[9, 8, 7, 6, 5, 4, 3, 2, 1]`. Or I might just not be understanding your suggestion...

utaal:
Thank you for porting these over! I have a small suggestion: can we tag the functions and lemmas that you've imported from dafny with a comment? (I'm kind of tempted to maybe even add a marker attribute, for this purpose) This would help us distinguish between the ones that had been added because someone needed them and those that we're inheriting. This can, for example, be helpful in deciding what goes into the import sets when we implement https://github.com/verus-lang/verus/discussions/37 in August/September.

jaybosamiya:
Just making a quick note here that there is a lot of added context to this code that happens in IRL discussions that hasn't made its way to GitHub just yet (esp since this is still a draft, not yet ready for review). Once it is ready for review, we'll explicitly mark it as such and add more reviewers.

Fwiw, multiple things added to this branch at the current moment are not going to directly make it to the final PR (such as many(/most?) of the added `broadcast_forall`s), since some of the code is being used to test things out (a direct port of the Dafny std library wouldn't quite work, due to _reasons_, thus a bunch of testing is needed to see what might work well).

The aim overall though is that things will be prepared well enough for #37.  The way we've been proceeding for that is somewhat different from adding a marker attribute or comment (it'll probably make more sense when the PR is marked ready-to-review; we're still testing more things out).  But yeah, good reminder to track things (our current tracking strategy, at least until we get closer to final PR is just that _everything_ in this branch is tainted, so no need to add comments/attributes 😄)

source/pervasive/set.rs
```diff
@@ -1,3 +1,7 @@
+// TODO: make a table of every Dafny lemma/axiom and the verus equivalent
+// include notes of any skipped lemmas
+// include this in PR description
```
tjhance, line 3: can you remove these notes

source/pervasive/set.rs
```diff
@@ -184,6 +189,16 @@ pub proof fn axiom_set_insert_different<A>(s: Set<A>, a1: A, a2: A)
...
 }
 
+/// The result of inserting element `a2` into set `s` must contain `a1` if and only if
+/// the set contained `a1` before the insertion of `a2` or if `a1` is equal to `a2`
+#[verifier(external_body)]
+#[verifier(broadcast_forall)]
+pub proof fn axiom_set_insert_contains<A>(s: Set<A>, a1: A, a2: A)
+    ensures
+        #[trigger] s.insert(a1).contains(a2) <==> a1 == a2 || s.contains(a2),
+{}
```
jaybosamiya, line 199: Isn't this equivalent to `axiom_set_insert_same` + `axiom_set_insert_same`. I don't think this should be a separate axiom.

source/pervasive/set_lib.rs
```diff
@@ -46,8 +61,112 @@ impl<A> Set<A> {
             arbitrary()
         }
     }
+
+    /// Converts a set into a sequence with an arbitrary ordering.
+    pub open spec fn to_seq(self) -> Seq<A> 
+        // recommends
+        //     self.finite()
```
jaybosamiya, line 68: Why is this commented out?

source/pervasive/std_specs/vec.rs
```diff
@@ -220,5 +222,31 @@ pub fn ex_vec_as_slice<T, A: Allocator>(vec: &Vec<T, A>) -> (slice: &[T])
...
+}
+
+// #[verifier::external_fn_specification]
+// pub fn ex_vec_to_vec<T: std::clone::Clone, A: Allocator>(vec: &mut Vec<T, A>) -> (return_value: Vec<T, ExGlobal>)
+//     ensures
+//         vec@ == old(vec)@,
+//         return_value@ =~= old(vec)@,
+// {
+//     vec.to_vec()
+// }
```
jaybosamiya, line 248: Why is this commented out?

jaybosamiya (commented):
Thanks for all the work on this! This definitely will make using Verus a lot easier for projects by having a well rounded standard library.

I've added comments inline (use the "Files Changed" view to easily see them all), but also have some open questions that I am not sure I've actually made up my mind about:
1. Do we want lemmas to be methods? I think it might actually make usage nicer (eg: rather than  `lemma_remove_key_len(m, k)`, `m.lemma_remove_key_len(k)` would require you to import fewer things via `use` declarations, and _as a personal preference_ seems to also feel like a cleaner style). Not sure how strong others opinions on this are, so I'm opening this up as a question here.
2. Do we even need to specify the "Ported from Dafny prelude" comment?  Since there are now the `lemma_X_properties` functions that are equivalent to those, I don't think we need to mark the specific individual ones.

Also, I am yet to do a full detailed pass over every single function and comment, but hopefully the current set of comments already point towards the style in which I am looking over things.  I'll do a full pass later on after the next set of changes are pushed.

source/pervasive/map.rs
```diff
@@ -296,6 +318,116 @@ impl<K, V> Map<K, V> {
...
+    /// Returns `true` if and only if a map is monotonic, only considering keys greater than
+    /// or equal to start
+    pub open spec fn monotonic_from(self, start: int) -> bool {
+        forall |x: int, y: int| self.dom().contains(x) && self.dom().contains(y) && start <= x <= y 
+            ==> #[trigger] self[x] <= #[trigger] self[y]
+    }
+
+}
+
+// Proven lemmas
```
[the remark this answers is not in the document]
tjhance: If a spec fn can be defined with a body, it probably makes sense to put it in the `lib` file. Likewise if a lemma has a body, it makes sense to put it in `lib`. The `set.rs` and `seq.rs` files are mostly for spec fns with no bodies, and proof fns marked external_body.

source/pervasive/seq_lib.rs
```diff
@@ -157,15 +234,107 @@ impl<A> Seq<A> {
...
 
+    /// Returns some index if an element occurs at least once in a sequence
+    /// and occurs at index i. Otherwise returns an arbitrary value.
     pub open spec fn index_of(self, needle: A) -> int {
         choose|i: int| 0 <= i < self.len() && self[i] == needle
     }
 
+    /// Returns Some(i), if an element occurs at least once in a sequence and
+    /// occurs at index i. Otherwise the return is None.
+    pub open spec fn index_of_option(self, needle: A) -> (result: Option<int>)
```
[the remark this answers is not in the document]
elaustell: for both the regular one and the option one?

source/pervasive/std_specs/vec.rs
elaustell: Oh I forgot about this cause I did it very early on, I think I got stuck and confused on it. Anyways I'm just going to delete it because it's not really relevant to this project anyways

tjhance (approved):

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+            self.len() == 1 ==> self.flatten() == self.first(),
+    {
+        if self.len() == 1
+        {
+            assert(self.flatten() =~= self.first().add(self.drop_first().flatten()));
+        }
+    }
+
+    /// The length of a flattened sequence of sequences x is greater than or 
+        /// equal to any of the lengths of the elements of x.
```
parno, line 901: There's an extra tab here

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+            if self[0] >= later_max {self[0]}
+            else {later_max}
+        }
+    }
+
+    /// Proof of correctness and expected properties for max function
+    pub proof fn lemma_max_properties(self)
+        ensures
+            forall |x: int| self.contains(x) ==> x <= self.max(),
+            forall |i: int| 0<= i < self.len() ==> self[i] <= self.max(),
```
parno, line 983: "0<=" => "0 <="

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+                assert(i==0 || self[i] == self.drop_first()[i-1]);
+                assert(forall |j: int| 0<= j < self.drop_first().len() 
+                        ==> self.drop_first()[j] <= self.drop_first().max()) by {
+                    self.drop_first().lemma_max_properties()
+                }
+            }
+        }
+    }
+
+    /// Returns the maximum integer value in a non-empty sequence of integers.
```
parno, line 1004: "minimum"

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+            0 <= from < to <= self.len(),
+        ensures
+            self.subrange(from, to).min() >= self.min(),
+    {
+        self.lemma_min_properties();
+        self.subrange(from, to).lemma_min_properties();
+    }
+}
+
+/// Helper function to aid with merge sort
```
parno, line 1082: Probably remove `pub` and use `//`

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+    if left.len() == 0 {right}
+    else if right.len() == 0 {left}
+    else if leq(left.first(), right.first()){
+        Seq::<A>::empty().push(left.first()) + merge_sorted_with(left.drop_first(), right, leq)
+    } else {
+        Seq::<A>::empty().push(right.first()) + merge_sorted_with(left, right.drop_first(), leq)
+    }
+}
+
+proof fn lemma_merge_sorted_with_ensures<A>(left: Seq<A>, right: Seq<A>, leq: FnSpec(A,A) -> bool)
```
parno, line 1100: In Dafny, I suspect this function was executable; here it's a spec.  In theory, for a spec sort, we can use something less efficient than merge sort.  I think we can leave it for now though (maybe it'll make someone's job of writing an exec version easier).

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+        assert(s.to_multiset() =~= Multiset::<A>::empty());
+        assert(s.len() == 0);
+    } else {
+        to_multiset_len(s.drop_first());
+        assert(s.len() == s.drop_first().len() + 1);
+        assert(s.to_multiset().len() == s.drop_first().to_multiset().len() + 1);
+    }
+}
+
+/// to_multiset() contains only all elements of the sequence
```
parno, line 1258: "only all" => "only the"?

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+    }
+}
+
+/// The last element of two concatenated sequences, the second one being non-empty, will be the 
+/// last element of the latter sequence.
+pub proof fn lemma_append_last<A>(s1 :Seq<A>, s2 :Seq<A>)
+    requires
+        0 < s2.len(),
+    ensures
+        #[trigger] (s1 + s2).last() == s2.last(),
```
parno, line 1297: No trigger

source/pervasive/seq_lib.rs
```diff
@@ -363,10 +646,664 @@ impl<A> Seq<A> {
...
+{}
+
+/// The concatenation of sequences is associative
+pub proof fn lemma_concat_associative<A>(s1 : Seq<A>, s2 :Seq<A>, s3 :Seq<A>)
+    ensures
+        #[trigger] s1.add(s2.add(s3)) =~= #[trigger] s1.add(s2).add(s3),
```
parno, line 1303: No trigger

source/pervasive/seq_lib.rs
```diff
@@ -444,24 +1380,465 @@ pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
     // TODO: merge this with seq_to_set_is_finite when broadcast_forall is better supported
 }
 
-/// A sequence of unique items, when converted to a set, produces a set with matching length
-pub proof fn unique_seq_to_set<A>(seq:Seq<A>)
-    requires seq.no_duplicates(),
-    ensures seq.len() == seq.to_set().len()
-    decreases seq.len(),
+/// If sequences a and b don't have duplicates and there are no
```
parno, line 1383: comma between "duplicated" and "and

source/pervasive/seq_lib.rs
```diff
@@ -444,24 +1380,465 @@ pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
...
+    if a.len() == 0 {
+        assert(a+b =~= b);
+    }
+    else {
+        lemma_multiset_commutative(a.drop_first(),b);
+        assert(a.drop_first() + b =~= (a+b).drop_first());
+    }
+}
+ 
+/// Proves that any two sequences that are sorted by a total order and that have the same elements are equal.
```
parno, line 1462: Drop "Proves that"

source/pervasive/seq_lib.rs
```diff
@@ -444,24 +1380,465 @@ pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
...
+        }
+    }
+}
+
+// This verified lemma used to be an axiom in the Dafny prelude
+/// After pushing an element onto a sequence, the sequence contains that element
+pub proof fn lemma_seq_contains_after_push<A>(s: Seq<A>, v: A, x: A)
+    ensures 
+        (#[trigger] s.push(v).contains(x) <==> v==x || s.contains(x))
+            && #[trigger] s.push(v).contains(v),
```
parno, line 1546: No triggers

source/pervasive/seq_lib.rs
```diff
@@ -444,24 +1380,465 @@ pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
...
+        if exists |i: int| 0 <= start <= i < stop <= s.len() && s[i] == x
+        {
+            let index = choose |i: int| 0 <= start <= i < stop <= s.len() && s[i] == x;
+            assert(s.subrange(start,stop)[index - start] == s[index]);
+        }
+    }
+}
+
+/************************** Lemmas about Take/Drop ***************************/
+
```
parno, line 1575: Many of these still have extraneous triggers

source/pervasive/set_lib.rs
```diff
@@ -46,22 +61,367 @@ impl<A> Set<A> {
...
+            Self::decreases_proof
+    {
+        if self.len() == 0 {Seq::<A>::empty()}
+        else {
+            let x = self.choose();
+            Seq::<A>::empty().push(x) + self.remove(x).to_seq()
+        }
+    }
+
+    /// Helper function to prove termination of function to_seq
```
parno, line 83: Probably should use `//` and no `pub`

parno: Similar comment for other helper-style functions below.

source/pervasive/set_lib.rs
```diff
@@ -46,22 +61,367 @@ impl<A> Set<A> {
...
+
+    /// Converts a set into a multiset where each element from the set has
+    /// multiplicity 1 and any other element has multiplicity 0.
+    pub open spec fn to_multiset(self) -> Multiset<A> 
+        decreases 
+            self.len()
+        when
+            self.finite()
+    {
+        if self.len() == 0 {Multiset::<A>::empty()}
```
parno, line 275: if-else formatting

source/pervasive/set_lib.rs
```diff
@@ -163,6 +545,325 @@ pub proof fn lemma_int_range(lo: int, hi: int)
...
+    lemma_set_properties::<A>();
+    lemma_set_properties::<B>();
+    if x.len() != 0 {
+        let a = x.choose();
+        lemma_map_size(x.remove(a), y.remove(f(a)), f);
+    }
+}
+
+// This verified lemma used to be an axiom in the Dafny prelude
+/// Taking the set union of sets `a` and `b` and then taking the set union of the result with `b`
```
parno, line 591: "set union" -> "union"

source/pervasive/set_lib.rs
```diff
@@ -163,6 +545,325 @@ pub proof fn lemma_int_range(lo: int, hi: int)
...
+                let a = s.choose();
+                assert(s.remove(a).len() == s.len() -1) by {
+                    axiom_set_remove_len(s, a);
+                }
+            }
+        }
+    }
+    assert(s.len() == 0 <== s =~= Set::empty());
+}
+
```
parno, line 665: Some extra trigger annotations in the proofs below

source/pervasive/set_lib.rs
```diff
@@ -163,6 +545,325 @@ pub proof fn lemma_int_range(lo: int, hi: int)
...
+// This verified lemma used to be an axiom in the Dafny prelude
+/// If set `s` already contains element `a`, the inserting `a` into `s` does not change the length of `s`.
+pub proof fn lemma_set_insert_same_len<A>(s: Set<A>, a: A)
+    requires
+        s.finite(),
+    ensures
+        s.contains(a) ==> #[trigger] s.insert(a).len() =~= s.len(),
+{}
+
+// magic lemma from spec function
```
parno, line 675: Not sure what this comment means

source/rust_verify/example/multiset.rs
```diff
@@ -0,0 +1,51 @@
...
+proof fn multiset_ext_eq2() {
+
+    let a: Seq<int> = seq![3, 2, 1, 1, 2, 3];
+    let b: Seq<int> = seq![1, 2, 3, 1, 2, 3];
+    
+    lemma_seq_properties::<int>(); //Provides the necessary lemmas for seq to multiset conversion
+    assert(a.to_multiset() =~= b.to_multiset());
+}
+
+proof fn sorted_by_eg() {
```
parno, line 26: "eg" -> "leq"?

source/pervasive/multiset.rs
```diff
@@ -264,6 +520,57 @@ macro_rules! assert_multisets_equal {
...
+        lemma_left_pseudo_idempotence(a, b);
+    }
+    assert forall |a: Multiset<V>, b: Multiset<V>| #[trigger] a.intersection_with(a.intersection_with(b)) == a.intersection_with(b) by {
+        lemma_right_pseudo_idempotence(a, b);
+    }
+    assert forall |a: Multiset<V>, b: Multiset<V>, x: V| #[trigger] a.difference_with(b).count(x) == clip(a.count(x) - b.count(x)) by {
+        lemma_difference_count(a, b, x);
+    }
+}
+
```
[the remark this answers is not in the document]
elaustell: Sure, my only thought is it's so ugly that it would be best to hide it at the bottom. Like here are all these lemmas and down here is them tied in a little bow. But I'm happy to move it up if you think that fits better with convention.

source/pervasive/seq_lib.rs
```diff
@@ -157,15 +269,90 @@ impl<A> Seq<A> {
...
 
+    /// For an element that occurs at least once in a sequence, if its first occurence
+    /// is at index i, Some(i) is returned. Otherwise, None is returned
+    pub open spec fn find_first(self, needle: A) -> (result: Option<int>)
+    {
+        if (self.contains(needle)) {Some(self.first_index_helper(needle, 0))}
+        else {None}
+    }
+
+    /// Recursive helper function for first_index_of
```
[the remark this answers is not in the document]
elaustell: I get the error "in pub open spec function, cannot refer to private function"

source/pervasive/seq_lib.rs
elaustell: Yes sorry I forgot to take all the triggers off after removing broadcast formal

source/rust_verify/example/multiset.rs
elaustell: I think the point was it's an example of using the function sorted_by. But I can change it if you think that is more appropriate

source/pervasive/multiset.rs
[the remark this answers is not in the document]
elaustell: My only thought is that these make more sense being called from the math library, but when I try to call that verus complains that it's not supported. But these functions don't make a whole lot of sense in the Multiset library, which is why I hid them at the bottom. Thoughts on this?

source/rust_verify/example/multiset.rs
parno: Yes, I think the `eg` looks odd without the periods and in a non-formal writing setting.

source/pervasive/multiset.rs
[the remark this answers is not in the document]
parno: Hmm, what error do you see when you put the definitions in the math library?  I think that should be feasible to do.

utaal:
Thank you for all your work on this. Having an extensive standard (spec) library is going to significantly improve the experience of using Verus.

I've left a few comments inline, and I have two general questions:
* there are a number of lemmas that have no proof: did they also have no proof in Dafny?
* there are a number of `open spec` functions with an accompanying `*_ensures` lemma which appear to carry the properties of the function that are useful to the user: were they also not opaque in Dafny? Should we consider marking (some of them) closed? (@jonhnet is working on adding ensures to spec functions, which would make these `_ensures` lemmas more directly usable without needing to call them).

source/pervasive/seq_lib.rs
```diff
@@ -157,15 +270,156 @@ impl<A> Seq<A> {
 //        }
     }
 
+    /// Returns the maximum value in a non-empty sequence, given sorting function leq
+    pub open spec fn max_via(self, leq: FnSpec(A,A) -> bool) -> A
```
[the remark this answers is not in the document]
jaybosamiya: This one in particular, I had explicitly suggested _against_ `max_by`, and specifically had suggested `max_via`, see https://github.com/verus-lang/verus/pull/622#discussion_r1275397096 but relevant part of the comment:
> leaves room in the future for a `max_by` which can use a convention similar to that in Rust's standard library (see [`max_by`](https://doc.rust-lang.org/std/cmp/fn.max_by.html)) that would use a `FnSpec(A,A)->Ordering`)

source/pervasive/seq.rs
```diff
@@ -169,6 +189,15 @@ impl<A> Seq<A> {
     {
         self[self.len() as int - 1]
     }
+
+    /// Returns the first element of the sequence.
+    
+    #[rustc_diagnostic_item = "vstd::seq::Seq::first"]
```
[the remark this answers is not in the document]
elaustell: I have no idea why this is there, I wasn't the one who put it there. I'm happy to leave it or remove it.

Chris-Hawblitzel (approved):

utaal (approved):
Thank you again for all the work!

(At the meeting, we discussed moving some of the empty lemmas derived from Dafny axioms to tests; we can do this either now or after this PR is merged.)

source/pervasive/seq_lib.rs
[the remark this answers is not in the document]
utaal: Sgtm. Thanks for the clarification, and sorry for missing your previous comment.

source/pervasive/seq.rs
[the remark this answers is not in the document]
utaal: Please remove it -- it currently has no effect, thanks!

utaal:
Please clean up the history before merging. (Let me know if I can help.)

PR #625, opened 2023-06-26
Allow arithmetic in quantifier triggers

Chris-Hawblitzel:
Allow integer arithmetic in quantifier triggers.  Also fix bitwise arithmetic in quantifier triggers to use native `int` representations of the relevant bound quantifier variables in the triggers rather than `Poly`.  The arithmetic is allowed as long as each bound quantifier variable is used unambiguously as either arithmetic or non-arithmetic in all the quantifier's triggers.  This should subsume and eventually replace `forall_arith`.


source/vir/src/ast.rs
```diff
@@ -621,8 +620,10 @@ pub enum ExprX {
     Header(HeaderExpr),
     /// Assert or assume
     AssertAssume { is_assume: bool, expr: Expr },
-    /// Forall or assert-by statement; proves "forall vars. ensure" via proof.
-    Forall { vars: Binders<Typ>, require: Expr, ensure: Expr, proof: Expr },
+    /// Assert-forall or assert-by statement; proves "assumption" via proof,
+    /// where "assumption" is "forall vars. require ==> ensure"
+    /// ("assumption" is redundant, but is kept separately because poly.rs treats it differently)
```
utaal, line 625: Can you briefly expand on how it is treated differently? Either here, or in poly.rs.

utaal:
@Chris-Hawblitzel I've made a couple of minor changes, please check they look good to you.

utaal (approved):

tjhance (approved):

PR #631, opened 2023-06-27
Documentation for assert ( ... ) by { ... }

jaylorch:


parno (approved):
Generally looks good to me.  Perhaps we should replace "Z3" with "SMT solver" to keep open the possibility that some day in the future we'll support other solvers too?

jaylorch:
> Generally looks good to me. Perhaps we should replace "Z3" with "SMT solver" to keep open the possibility that some day in the future we'll support other solvers too?

Good idea, done.

PR #633, opened 2023-06-27
Correctly calculate the default number of threads

achreto:
use the std::cmp::max function instead of std::cmp:min to calculate the amount of parallelism.


utaal (approved):
Oops. This was my fault.

PR #634, opened 2023-06-27
Update guide to represent extensional equality

elaustell:


tchajed (approved):
This PR looks good to me to fix https://verus-lang.github.io/verus/guide/exec_lib.html fully to use the new `=~=` syntax (there's no mention of extensional equality in the text).

PR #637, opened 2023-06-27
Change verification output

ahuoguo:


jaybosamiya:
Weird; CI succeeded (on 07326f47a465604ae0354cf2aadb7a40fa6f0b43), but locally on mine and @ahuoguo's machine, we get 
```
     Summary [   1.629s] 2 tests run: 0 passed, 2 failed, 1716 skipped
        FAIL [   1.442s] rust_verify_test::state_machines relation_codegen_special
        FAIL [   1.623s] rust_verify_test::state_machines relation_codegen_special_general
error: test run failed
```

Either there is some internal state being maintained across (testing now with `git clean -fdX .` locally) but if there is internal state that isn't being cleared, it might be worth looking into.

utaal (changes requested):
I believe this was discussed at the all-hands, but I'd rather we didn't change the verifier's main output as part of a PR that looks like is only changing the docs. Can you please split the changes to the code into a separate PR? (I'm happy to help.)

PR #639, opened 2023-06-27
remove an old code path when searching for Z3

utaal:
One step towards addressing https://github.com/verus-lang/verus/discussions/387#discussioncomment-6298063

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -266,16 +266,7 @@ where
             let externs = VerusExterns { path: &verusroot, has_vstd: !verifier.args.no_vstd };
             rustc_args.extend(externs.to_args());
             if in_vargo && !std::env::var("VERUS_Z3_PATH").is_ok() {
-                std::env::set_var(
-                    "VERUS_Z3_PATH",
-                    verusroot
-                        .parent()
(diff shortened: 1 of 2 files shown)


jaybosamiya:
It appears that switching this code path to a panic actually causes it to panic:

```console
$ ./tools/get-z3.sh
...
$ vargo run -p rust_verify -- ./rust_verify/example/calc.rs
...
warning: verus was compiled in debug mode, which will result in worse performance
thread 'main' panicked at 'we are in vargo, but VERUS_Z3_PATH is not set; this is a bug', /home/jay/verus/simplify-verusroot/source/rust_verify/src/driver.rs:269:17
```

Looking within the vargo source itself, I don't see anything actually setting that environment variable; there is only a check for if it is set, but that check only happens if `./z3` doesn't already exist: https://github.com/verus-lang/verus/blob/ce715d977f801747d12b35079b4ed073f48b9e43/tools/vargo/src/main.rs#L235-L237

utaal:
Great catch!
That explains why that code path was there. I think this other change should have the same effect but still reduce the number of different ways VERUS_Z3_PATH can be set.

jaybosamiya (approved):
Thanks for updating it and adding code to set the env var.  LGTM.

PR #640, opened 2023-06-27
make closed spec tut more understandable

elaustell:


source/docs/guide/src/spec_functions.md
```diff
@@ -47,6 +48,8 @@ error: assertion failed
    |                ^^^^^^^^^^^^^^^^^ assertion failed
 ```
 
+After the call to `lemma_min`, the assertion that `min(10, 20) <= 10` succeeds because `lemma_min` is marked `pub`, and thus can be seen by `M2`.
```
utaal, line 51: I'd suggest expanding on how the lemma helps "export" that fact about `min`. Does the following make sense to you?

```suggestion
After the call to `lemma_min`, the assertion that `min(10, 20) <= 10` succeeds because `lemma_min` exposes `min(x,y) <= x` as a post-condition. `lemma_min` can prove because this postcondition because it can see the body of `min` despite `min` being `closed`, as `lemma_min` and `min` are in the same module.
```

PR #642, opened 2023-06-27
fix modes table

elaustell:


source/docs/guide/src/modes.md
```diff
@@ -28,16 +28,10 @@ Every function in Verus is either a `spec` function, a `proof` function, or an `
...
-| can call `spec` functions   | yes            | yes              | yes              |
-| can call `proof` functions  | no             | yes              | yes              |
-| can call `exec` functions   | no             | no               | yes              |
+the three modes:
+
+|                        | spec code      | proof code       | exec code        |
+|------------------------|----------------|------------------|------------------|
+| can call `spec` code   | yes            | yes              | yes              |
+| can call `proof` code  | no             | yes              | yes              |
+| can call `exec` code   | no             | no               | yes              |
```
utaal, line 37: This simplification is good, but I don't think it's quite right. We'd want to mention that `spec`, `proof`, `exec` functions can contain code in certain other modes.
What about:
```suggestion
|                        | spec code      | proof code       | exec code        |
|------------------------|----------------|------------------|------------------|
| can contain `spec` code, call `spec` functions   | yes            | yes              | yes              |
| can contain `proof` code, call `proof` functions | no             | yes              | yes              |
| can contain `exec` code, call `exec` functions   | no             | no               | yes              |
```

elaustell: Yes I agree this is more clear.

utaal (approved):
Thank you. Please squash when you merge.

PR #643, opened 2023-06-27
Add guide section on breaking proofs into pieces

jaylorch:


utaal (approved):
What do you think about adding a note suggesting to factor out `mid1`, `mid2`, ... as spec functions to avoid repetition?

Issue #471, opened 2023-03-28
broadcast_forall lemma fails to fire


parno:
In this code:

```
#[allow(unused_imports)]
use builtin_macros::*;

#[cfg(not(vstd_todo))]
mod pervasive;
#[cfg(not(vstd_todo))]
#[allow(unused_imports)]
use pervasive::{*, seq::*, seq_lib::*};

verus! {
proof fn weird_broadcast_failure(seq:Seq<usize>)
{
    //seq_to_set_is_finite_broadcast::<usize>(seq);
    assert(seq.to_set().finite());
}
}

fn main() { }
```

I expect the assertion to go through, thanks to this [proof function in `seq_lib.rs`](https://github.com/verus-lang/verus/blob/77e518bbd21e0466e226d6ccd510570751ab6be1/source/pervasive/seq_lib.rs#L275):
```
#[verifier(external_body)]
#[verifier(broadcast_forall)]
pub proof fn seq_to_set_is_finite_broadcast<A>(seq: Seq<A>)
    ensures #[trigger] seq.to_set().finite()
```
However, the assertion only goes through when I manually invoke the lemma (by uncommenting the line preceding the assertion).  One possibly unrelated observation is that if I don't add the `::<usize>` annotation to the call to `seq_to_set_is_finite_broadcast`, then Verus/Rust complains that it
```
cannot infer type for type parameter `A` declared on the function `seq_to_set_is_finite_broadcast`
```

parno:
Actually, I think the type annotation is unrelated.  The more important problem is that without the manual invocation of `seq_to_set_is_finite_broadcast`, that lemma doesn't show up at all in the generated SMT output.  All of the other `pub` functions inside the `impl<A> Seq<A> { ... } ` in `seq_lib.rs` do show up, but it looks like `seq_to_set_is_finite_broadcast` is the only `pub` function that's in the file but not inside the `impl<A> Seq<A> { ... } `.  Does this mean it should be moved inside, or that this is a bug in how Verus handles imports?

As a side note, I tried adding an explicit
```
use pervasive::seq_lib::seq_to_set_is_finite_broadcast;
```
to the file, but Rust issues a warning that it's an unused import, and it still doesn't appear in the SMT output.

utaal:
[Verus all-hands]

This will hopefully be fixed by import sets.

In the meantime, we can move the axioms to `seq.rs`.

utaal:
[Verus all-hands]

This now passes on `main`. Great!

PR #650, opened 2023-06-28
Add a regression test for issue #471

tchajed:
Add a regression test for the broadcast_forall issue in #471, ported to the new vstd imports.


utaal (approved):
Thank you! Can you please remove the main function?

PR #659, opened 2023-06-29
Add support for compound assignments

jcp19:
With this PR, the syntax for assignments like `x += 1` should now be supported. For now, only variables(`x`) and derefs of mut borrows (`*x`) are supported on the lhs of the assignment. To avoid repetition, I moved some pre-existing code into functions `binopkind_to_binaryop` and `expr_assign_to_vir_innermost`.


source/vir/src/ast_simplify.rs
```diff
@@ -436,6 +436,31 @@ fn simplify_one_expr(ctx: &GlobalCtx, state: &mut State, expr: &Expr) -> Result<
...
+                        SpannedTyped::new(
+                            &expr.span,
+                            &expr.typ,
+                            ExprX::Assign {
+                                init_not_mut: *init_not_mut,
+                                lhs: lhs.clone(),
+                                rhs: SpannedTyped::new(
+                                    &expr.span,
+                                    &lhs.typ,
+                                    ExprX::Binary(op.clone(), lhs.clone(), rhs.clone())
```
tjhance, line 452: It looks like you're cloning `lhs` as-is, but I think `lhs` might be a VarLoc, so should be changed to a `Var`

tjhance (changes requested):

tjhance:
oh sorry! I only just realized this was marked as a 'draft'.

source/vir/src/ast_simplify.rs
jcp19: done :+1:

Chris-Hawblitzel (changes requested):
Looks good overall.  It would be worth adding proof-mode tests as well (such as `test1b` below).  I think there should be clipping inserted for add/sub/mul/div/mod so that the `x += y` behavior matches the `x = x + y` behavior.  For example, `test3a` below succeeds but `test3b` fails:

```
proof fn test1a() {
    let mut x: u8 = 200;
    x = (x + 100u8) as u8;
    assert(x < 256);
}

proof fn test1b() {
    let mut x: u8 = 200;
    x += 100u8;
    assert(x < 256);
}

fn test2a() {
    let mut x: u8 = 200;
    x = x + 100u8;
    assert(x < 256);
}

fn test2b() {
    let mut x: u8 = 200;
    x += 100u8;
    assert(x < 256);
}

fn test3a() {
    let mut x: u8 = 200;
    x = x / (x + 1);
    assert(x < 256);
}

fn test3b() {
    let mut x: u8 = 200;
    x /= (x + 1);
    assert(x < 256);
}
```

Chris-Hawblitzel (approved):

PR #663, opened 2023-06-30
builtin is_variant / get_variant_field functions

tjhance:
- [x] fix issue where is_variant creates functions that are too public


source/rust_verify_test/tests/lifetime.rs
```diff
@@ -149,8 +149,8 @@ test_verify_one_file! {
             x
         }
         proof fn test<A>(tracked x: A) {
-            let s = id(Option::Some(x)).is_None();
-            let s = id(Option::Some(x)).is_None();
+            let s = builtin::is_variant(id(Option::Some(x)), "None");
+            let s = builtin::is_variant(id(Option::Some(x)), "None");
```
utaal, line 153: Why are these change necessary?

utaal:
Do we intend users to call the `is_variant` and `get_variant_field` builtins directly? Maybe only for things like extensions to external objects? If so, shall we move them into a separate builtin module? (Something like `builtin::internal`, but they aren't quite "internal", I think.)

utaal:
I've added a couple of tests, please double check they make sense to you. Thanks!

tjhance:
Yeah, it's mostly useful as an implementation detail of is_variant, or for handling external structs. I imagine it could also be useful in macros.

> If so, shall we move them into a separate builtin module? (Something like builtin::internal, but they aren't quite "internal", I think.)

Most stuff in builtin is already internal, isn't it?

utaal:
> Most stuff in builtin is already internal, isn't it?

I guess I had not updated my model of this in a while, you're right. We may want to warn on direct usages of builtin functions at some point, but not a concern for this PR.

utaal (approved):

source/rust_verify_test/tests/lifetime.rs
tjhance: It has to do with whether or not the spec function allows proof arguments ... normal spec functions don't, but apparently field access & variant checks do. The test needs something that does accept proof arguments.

`is_None` is a "normal" spec function now, so the test can't work while using it. On the other hand, `builtin::is_variant` lowers to the same VIR as is_None used to lower

The behavior is (and was) a bit inconsistent and could maybe use a revisit later.

utaal: Maybe add this as a comment to that change? (Not critical, though.)

PR #665, opened 2023-06-30
Use same RUSTFLAGS for each build step

hayley-leblanc:
Fix/workaround for a vargo issue introduced in commit 533f0a212dee3c13121f17be367d8f51b782bf8c.

Cargo automatically clears some caches when building crates with different `RUSTFLAGS` values. The aforementioned commit added a `RUSTFLAGS` argument to one crate's build command but not to the rest, which caused a lot of things to be unnecessarily rebuilt when rebuilding Verus with vargo. This PR adds the same `RUSTFLAGS` argument to all build commands issued by vargo so that the cache is not cleared out between them. ~~I still see *some* seemingly unnecessary rebuilding with this fix, but less than before.~~ After running it a few times it uses the cache as expected.

I think this is probably a workaround that should be removed when the unstable `proc_macro_span` feature is stabilized.

### DIFF
--- a/tools/vargo/src/main.rs
+++ b/tools/vargo/src/main.rs
@@ -480,6 +480,7 @@ fn run() -> Result<(), String> {
                 let cargo = cargo
                     .env("RUSTC_BOOTSTRAP", "1")
                     .env("VARGO_TARGET_DIR", target_verus_dir_absolute)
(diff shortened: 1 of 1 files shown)


utaal (approved):
Thank you for diagnosing and fixing this!

PR #666, opened 2023-06-30
chained equality for integers

ahuoguo:
Draft: only restricted to integers as this [PR](https://github.com/verus-lang/verus/commit/3e64d4860c8f018e1a6f0b5cd500ba0ab1772676).

-  separate the `MultiOp` field in `InequalityOp`, which requires some refactoring here and there.

### DIFF
--- a/source/rust_verify_test/tests/integers.rs
+++ b/source/rust_verify_test/tests/integers.rs
@@ -109,7 +109,7 @@ test_verify_one_file! {
     #[test] test_chained1 verus_code! {
         proof fn test1(n: nat) {
             assert(n as int == n as int + 0 == 0 + n as int);
-            assert(2 == 1 + 1 == 3 - 1);
+            assert(2 == 1 + 1 == 3 - 1 < 4);
             assert(n as int == (n as int) * 1 == (n as int) * 1 == 1 * (n as int));
             assert(1 + n as int == n as int + 1 == 1); // FAILS
         }


source/rust_verify_test/tests/integers.rs
```diff
@@ -101,6 +101,10 @@ test_verify_one_file! {
         proof fn test1(n: nat) {
             assert(0 <= n < n as int + 1 < n as int + 2);
             assert(0 <= n as int + 1 < n < n as int + 2); // FAILS
+            assert(n as int == n as int + 0 == 0 + n as int);
+            assert(1 + n as int == n as int + 1 == 1); // FAILS
+            assert(2 == 1 + 1 == 3 - 1);
+            assert(n as int == (n as int) * 1 == (n as int) * 1 == 1 * (n as int));
```
Chris-Hawblitzel, line 107: These should be split into a separate test.  Otherwise, the second `FAILS` actually succeeds because the first `FAILS` shadows it.

source/vir/src/ast_simplify.rs
```diff
@@ -354,7 +355,12 @@ fn simplify_one_expr(ctx: &GlobalCtx, state: &mut State, expr: &Expr) -> Result<
             }
             let mut conjunction: Expr = es[0].clone();
             for i in 0..ops.len() {
-                let op = BinaryOp::Inequality(ops[i]);
+                let op: BinaryOp;
+                if let ChainedOp::CmpOp(a) = ops[i] {
+                    op = BinaryOp::Inequality(a);
+                } else {
+                    op = BinaryOp::Eq(Mode::Spec);
+                }
```
Chris-Hawblitzel, line 363: This can be written as `let op = match ops[i] { ChainedOp::CmpOp(a) => BinaryOp::Inequality(a), ChainedOp::MultiEq => BinaryOp::Eq(Mode::Spec), };` so that it explicitly covers all the cases in `ChainedOp`.

source/vir/src/ast.rs
```diff
@@ -305,6 +305,12 @@ pub enum InequalityOp {
     Gt,
 }
 
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, ToDebugSNode)]
+pub enum ChainedOp {
+    CmpOp(InequalityOp),
```
Chris-Hawblitzel, line 310: `Inequality(InequalityOp)` might be a clearer name.

Chris-Hawblitzel (approved):
I suggested some minor edits.  Afterwards, go ahead and squash and merge.

PR #667, opened 2023-06-30
Introduce `calc` to the guide

jaybosamiya:


source/docs/guide/src/calc.md
```diff
@@ -0,0 +1,76 @@
...
+`a_1` and `a_n`, where it might be easier to do this in a series of steps, `a_1`
+to `a_2`, `a_2` to `a_3`, ... all the way to `a_n`. One might do this by just
+doing all the steps at once, but as mentioned in the section on
+[assert-by](./assert_by.md), a better approach might be to split it into a
+collection of restricted contexts. This is better, but still might not be ideal,
+since you need to repeat each of the intermediate expressions at each point.
+
+## `calc!`ulations, to Reduce Redundant Redundancy
+
+The `calc!` macro supports structured proofs through calculations.
```
tchajed, line 15: Can you add a simple example with concrete assertions here, similar to the one at the end?

source/docs/guide/src/calc.md
```diff
@@ -0,0 +1,76 @@
...
+calc! {
+  (R)
+  a_1; { /* proof that a_1 R a_2 */ }
+  a_2; { /* proof that a_2 R a_3 */ }
+   ...
+  a_n;
+}
+```
+
+Notice that you mention of `a_1`, `a_2`, ... `a_n` only once each. Additionally,
```
tchajed, line 34:
```suggestion
Notice that you mention `a_1`, `a_2`, ... `a_n` only once each. Additionally,
```

source/docs/guide/src/calc.md
```diff
@@ -0,0 +1,76 @@
...
+the proofs for each of the steps is localized, and restricted to only its
```
tchajed, line 35:
```suggestion
the proof for each of the steps is localized, and restricted to only its
```

source/docs/guide/src/calc.md
```diff
@@ -0,0 +1,76 @@
...
+particular step, ensuring that proof context is not polluted.
+
+The body of the function where this `calc` statement is written only gets to see
+`a_1 R a_n`, and not any of the intermediate steps (or their proofs), further
+limiting proof context pollution.
```
tchajed, line 40:
```suggestion
limiting proof-context pollution.
```
this is really pedantic, sorry

source/docs/guide/src/calc.md
```diff
@@ -0,0 +1,76 @@
...
+
+Currently, the `calc!` macro supports common transitive relations for `R` (such
+as `==` and `<=`). This set of relations may be extended in the future.
```
tchajed, line 43: Just out of curiosity, does it support `=~=` and `=~~=`?

source/docs/guide/src/calc.md
jaybosamiya: Thanks, added

source/docs/guide/src/calc.md
jaybosamiya: Oops, good catch; I had this sentence phrased differently before, and rewriting it this way I left in an `of`

source/docs/guide/src/calc.md
jaybosamiya: Good catch, thanks!

source/docs/guide/src/calc.md
jaybosamiya: Nw, I appreciate being pedantic about this; it does help clarify things for folks who might not be _as_ used to these things; thanks!

source/docs/guide/src/calc.md
jaybosamiya: At the moment, no; `calc` is older than `=~=` and `=~~=`.  Adding it is not difficult, but there is an interesting decision to be made as to what it compiles to: should saying R=`=~=` mean `=~=` _everywhere_ or should it mean `==` outside and all the internal ones being `=~=`? The former is less surprising wrt expansion itself (for anyone reading the `calc` code), but the latter would be more stable (and lead to less context pollution, and also the fact that we're restricting it should be practically undetectable by any user of the feature). I like the latter, but it requires a bit more work to implement than the former, so I'm holding off on that for a bit while I think through which of the designs work better.

utaal (approved):
Can you use `includes from a file tested as part of the examples for the two non-pseudocode blocks? An example is here: https://github.com/verus-lang/verus/blob/main/source/docs/guide/src/equality.md

jaybosamiya:
Oh, I wasn't aware of this feature, thanks for pointing it out; updated

utaal:
I think you can go ahead and squash and merge.

PR #668, opened 2023-06-30
Add .envrc to auto-source activation for vargo

jaybosamiya:
This PR adds auto-activation of vargo when you enter the `source` directory (along with some documentation about this).

Additionally, it fixes up the activation script to prevent `PATH` from being updated if vargo build fails, to prevent polluting it.

Also reminder @utaal to test this on fish to make sure we don't accidentally break things; I didn't see it do anything bad in fish on my setup, but since I am not _really_ a fish user, I don't know how people normally have it configured.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,6 +53,8 @@ source ../tools/activate.fish  # for fish
 
 This command builds (or re-builds) `vargo`, our cargo wrapper, and adds it to the `PATH` for the current shell.
 
+If you use [direnv](https://direnv.net/), then this activation is performed automatically (i.e., you don't need to `source ../tools/activate`; instead, `vargo` will automatically be in your `PATH` as long as you are in the `source` subdirectory or any directory under it).
+
 Now, simply run,
 
 ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,7 +42,7 @@ process does not expect.
 
 You should be in the `source` subdirectory.
 
-First, activate the development environment with one of the following:
+First, activate the development environment with one of the following: (This is automatic for [direnv](https://direnv.net/) users[^1])
 
 ```
 source ../tools/activate       # for bash and zsh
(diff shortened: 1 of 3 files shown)


tchajed:
I briefly tested this on fish (which is my main shell) and it seems to work. direnv always runs the script in bash and just exports the resulting environment variables in a shell-specific way, which helps.

It is a bit verbose because every time it activates it prints the following:
```txt
direnv: loading ~/code/sw/verus/source/.envrc
building vargo
    Finished release [optimized] target(s) in 0.06s
direnv: export ~PATH ~XPC_SERVICE_NAME
```
but I don't think this is a big problem if you're using direnv.

INSTALL.md
```diff
@@ -53,6 +53,8 @@ source ../tools/activate.fish  # for fish
 
 This command builds (or re-builds) `vargo`, our cargo wrapper, and adds it to the `PATH` for the current shell.
 
+If you use [direnv](https://direnv.net/), then this activation is performed automatically (i.e., you don't need to `source ../tools/activate`; instead, `vargo` will automatically be in your `PATH` as long as you are in the `source` subdirectory or any directory under it).
```
utaal, line 56: I'd prefer we kept these instructions as streamlined as possible for all users, and this paragraph is only relevant to direnv users.

What about we added a short note earlier (line 45):

```
First, activate the development environment with one of the following: (This is automatic for [direnv](https://direnv.net/) users[^1])
```

that points to a footnote:

```
[^1]: If you are a [direnv](https://direnv.net/) user this activation is performed automatically, i.e. you don't need to `source ../tools/activate`; instead, `vargo` will automatically be in your `PATH` as long as you are in the `source` subdirectory.
```

(GitHub will render it as an actual footnote.)

jaybosamiya: Good point, updated

jaybosamiya:
Thanks folks; updated the docs, and glad to know that this Just Works^TM with `fish` too :)

As far as it printing output: I think this is acceptable for direnv users (they expect prints to happen whenever going into and out of directories that impact env vars). Yes we are printing a bit more than just the "PATH changed", but I think it is useful for us to not suppress the fact that a (re)build is happening.

utaal (approved):
Thank you!

PR #669, opened 2023-06-30
Commit Cargo.lock

jaybosamiya:
This commit introduces `Cargo.lock`, removing it from the `.gitignore`; it does this only for the executables (not for the dependencies, since they are libraries), following regular Rust convention.

`Cargo.lock` helps make sure we're all in agreement about the _exact_ set of dependencies being used at any particular time.

Also as a reminder, since we haven't had a `Cargo.lock` for this long, we will probably have to remind users that they should drop their local copies of `Cargo.lock` when pulling `main` after this PR is merged.


utaal (approved):
Thank you!

Please run `cargo update` and amend the commit.

Please don't merge yet. Let's coordinate the merge at the meeting on Wednesday (2023/07/10).

jaybosamiya:
`cargo update`d and `git commit --ammend`ed; and yep let's coordinate on this Wednesday

PR #677, opened 2023-07-10
docs for spec closure

ahuoguo:
Splitted from PR https://github.com/verus-lang/verus/pull/637.


source/docs/guide/src/spec_closures.md
```diff
@@ -0,0 +1,25 @@
...
+Spec Closures have the [same restrictions](modes.md) as normal sepc functions.
+
+## FnSpec is a type
+
+It is worth noticing that `FnSpec` is not a trait as Rust does for `Fn`. Therefore, ghost code can return a spec closure without [a lot of trouble](https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html#:~:text=clearer%20to%20you.-,Returning%20Closures,return%20value%20of%20the%20function.).
+
+
+<!--https://github.com/verus-lang/verus/wiki/Status%3A-currently-supported-Rust-features>
+
+<!--https://github.com/verus-lang/verus/blob/main/CONTRIBUTING.md>
```
utaal, line 25: Why are these here?

source/docs/guide/src/spec_closures.md
```diff
@@ -0,0 +1,25 @@
...
+    assert(s[3] == 30);
+}
+```
+Since ghost code can only contain ghost code, the closure `|i: int| 10 * i` passed into `Seq::new` is a spec closures, which has type `FnSpec(int -> bool)`. You can declare these closures as normal rust Code.
+
+Spec Closures have the [same restrictions](modes.md) as normal sepc functions.
+
+## FnSpec is a type
+
+It is worth noticing that `FnSpec` is not a trait as Rust does for `Fn`. Therefore, ghost code can return a spec closure without [a lot of trouble](https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html#:~:text=clearer%20to%20you.-,Returning%20Closures,return%20value%20of%20the%20function.).
```
utaal, line 20: Can you say "`Fn`, `FnOnce`, `FnMut`"?
Can you be more specific? E.g. "without the need for a dynamic or anonymous return type (`dyn` or `impl`)" (while still having the link).

utaal (changes requested):

utaal (approved):
Thanks for the changes. LGTM. Please squash and merge (feel free to remove me as a co-author if suggested by GitHub).

PR #678, opened 2023-07-10
Specs for i32 and u32 methods

hayley-leblanc:
This PR adds specs for some checked methods for `i32` and `u32`, plus tests to check them. This includes checked addition and subtraction (including signed/unsigned variants) and both Euclidean and non-Euclidean division and remainder operations.

### DIFF
--- a/source/pervasive/std_specs/num.rs
+++ b/source/pervasive/std_specs/num.rs
@@ -1,0 +1,141 @@
+#![allow(unused_imports)]
+use builtin::*;
+use builtin_macros::*;
+use crate::prelude::*;
+
+verus!{
+
+// == u32 methods ==
+
+#[verifier::external_fn_specification]
+pub fn ex_u32_checked_add(lhs: u32, rhs: u32) -> (result: Option<u32>)
+    ensures 
+        lhs + rhs > u32::MAX ==> result.is_None(),
+        lhs + rhs <= u32::MAX ==> {
+            match result {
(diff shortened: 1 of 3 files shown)


source/pervasive/std_specs/num.rs
```diff
@@ -0,0 +1,281 @@
...
+verus!{
+
+// == u32 methods ==
+
+#[verifier::external_fn_specification]
+pub fn ex_u32_checked_add(lhs: u32, rhs: u32) -> (result: Option<u32>)
+    ensures 
+        lhs + rhs > u32::MAX ==> result.is_None(),
+        lhs + rhs <= u32::MAX ==> {
+            match result {
```
tjhance, line 15: you could do result == Some(lhs + rhs)

source/pervasive/std_specs/num.rs
```diff
@@ -0,0 +1,281 @@
...
+                0
+            } else if x > 0 && d > 0 {
+                x / d
+            } else if x < 0 && d < 0 {
+                ((x * -1) / (d * -1))
+            } else if x < 0 {
+                ((x * -1) / d) * -1
+            } else { // d < 0
+                (x / (d * -1)) * -1
+            };
```
tjhance, line 208: probably a good idea to factor the (lhs, rhs) -> output computation as a spec function on ints.

tjhance (approved):

PR #683, opened 2023-07-13
fn_call_to_vir rewrite

tjhance:
as discussed in the slack a while ago, rewrite fn_call_to_vir to split by cases with less entanglement


utaal (approved):
Thank you! Looks good to me.

I have a suggestion to remove some (more) of the explicit control flow and ensure that all branches return a value / error: https://github.com/verus-lang/verus/compare/fn-cleanup...fn-cleanup-suggestion

PR #686, opened 2023-07-14
hide the "reporting diagnostics for module" messages unless --trace or --verify-module are present

utaal:


jonhnet (commented):
Was the `multithreaded` argument disappearing a collateral change or somehow integral to this commit?

utaal:
It was only used to decide how to print the "reporting diagnostics" message, in that function, so it became unused after this change.

jonhnet (approved):

PR #689, opened 2023-07-18
WIP: Local storage of intermediate/ongoing verification

ahuoguo:
This PR aimed to collect user data on the local end. In particular, after the user specifies a flag during build time (eg: `vargo build --release --local-store`), ***and*** after the user invoked verus with the `--local-store` flag, then everytime verus is invoked, it will put the files necessary to generate the crate root (together with `reports.toml`) to `<data_local_dir>/verus/reports/<uuid>/<SHA256(CRATE_ROOT)>/`

The uuid is the in `<data_local_dir>/verus/reports/userid`, verus will create one if it doesn't exists

CRATE_ROOT is the canonicalized path to the crate root (assumed to be the only `.rs` file in the arguments)

There are certain cases where the user will not have `<data_local_dir>/verus/reports/<uuid>/<SHA256(CRATE_ROOT)>/` created (and changes will not be recorded):
1. the user does not have `git` in PATH
2. the user do not build verus with `--local-store`
3. the user did not run project with `--local-store` at some point (the recording will start after the user includes this flag)

TODO:
- [x] add a per-project opt-in option


source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+        _ => {
+            eprintln!("{}: cannot find input file", yansi::Paint::red("error"));
+            None // passed down to rust_verify for error message
+        }
+    };
+
+    let input_file_path = match input_file.clone() {
+        Some(input_file_name) => std::fs::canonicalize(input_file_name).ok(),
+        None => None,
+    };
```
jaybosamiya, line 271: You can likely write this more cleanly using `Option::and_then` (eg: `input_file.as_ref().and_then(|f| std::fs::canonicalize(f).ok())`), also `Path` has a [`canonicalize` method](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize), so you can just use `input_file.as_ref().and_then(|f| f.canonicalize().ok())`

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+
+    let project_name = match input_file_path.clone() {
+        Some(path) => {
+            Some(format!("{:x}", sha2::Sha256::digest(path.to_str().unwrap().as_bytes())))
+        }
+        None => None,
+    };
```
jaybosamiya, line 278: You can use `Option::map` to clean this up (eg: `input_file_path.as_ref().map(|f| ......)`)

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+
+    let proj_dir = match project_name {
```
jaybosamiya, line 280: Similarly can be cleaned up with `Option::map`

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+            Some(Reports { proj_path: proj_dir, crate_root: crate_root_path })
         }
+        _ => None,
     }
 }
+
+fn get_dependencies(dep_file_path: &std::path::Path) -> Result<Vec<PathBuf>, String> {
+    // update to better error message
+    let file = std::fs::File::open(dep_file_path)
+        .map_err(|x| format!("{}, dependency file name: {:?}", x, dep_file_path))?;
```
jaybosamiya, line 303: Looks like you missed this TODO you'd set for yourself?

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+    let mut reader = std::io::BufReader::new(file);
+    let mut dependencies = String::new();
+    reader.read_line(&mut dependencies).map_err(|x| {
+        format!("Could not read the first line of the dependency file with message: {}", x)
+    })?;
+    let dep: Vec<String> = dependencies.split_whitespace().skip(1).map(|x| x.to_string()).collect();
+    println!("\n{} {:?}", yansi::Paint::blue("dependencies:"), dep);
```
jaybosamiya, line 310: Looks like this is debug output for yourself; I don't think it should be included in the final code
```suggestion
```

jaybosamiya (changes requested):
Thanks for the working on a first draft of this!  Some comments added inline that need to be looked at and relevant updates done. Once that's done, please mark me as reviewer again so I can take a look at the next draft.

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+
+    // clone is not implemented for Args, so I grabbed the arguments here again
+    let new_args = std::env::args().into_iter();
+
+    let rs_files = new_args.filter(|arg| arg.ends_with(".rs")).collect::<Vec<_>>();
+
+    let input_file: Option<String> = match rs_files.len() {
+        1 => Some(rs_files[0].clone()),
+        _ => {
+            eprintln!("{}: cannot find input file", yansi::Paint::red("error"));
```
tjhance, line 263: Do we need this error message? I think it makes more sense to leave this kind of thing to Verus's main argument-parsing code

tjhance (commented):
Like jay says, even though this is "just" local storage, there still needs to be a way to disable this. Also, I think it should probably be disabled by default for building vstd.

tjhance:
Also, what happens if they don't have `git` installed?

utaal:
Please first describe what the set-up flow for this is, what is stored, and where. We should only act on their behalf and only take actions that they requested, unless there's a very good reason *and* we have their consent.

I am asking to please describe what the exact flow is first because it looks like this will automatically start storing copies of their code elsewhere on disk (including files that are not tracked by their source control). For some users, that, on its own, may be problematic.

utaal (changes requested):
Please first describe what the set-up flow for this is, what is stored, and where. We should only act on behalf the user and only take actions that they requested, unless there's a very good reason *and* we have their consent.

I am asking to please describe what the exact flow is first because it looks like this will automatically start storing copies of their code elsewhere on disk (including files that are not tracked by their source control). For some users, that, on its own, may be problematic.

This should be opt-in, not opt-out. We can maybe remind the user to opt-in in the output of verus, if we are not too intrusive.

utaal:
You're welcome to keep this open (maybe as draft) if it's helpful (but you're also welcome to re-open later). 

FWIW I realize it was work in progress: I only commented about some high level points as they may influence the design.

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+}
+
+pub fn exec(cmd: &mut Command, reports: ReportsPath) -> Result<(), String> {
+    use std::io::BufReader;
+    use std::io::Write;
+    use std::process::Stdio;
+    use std::thread;
+
+    #[cfg(windows)]
+    {
```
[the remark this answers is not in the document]
jaybosamiya: Huh, interesting; yeah feel free to skip this change then.

source/verus/src/main.rs
```diff
@@ -75,19 +51,263 @@ fn main() {
...
+
+    let thread_err = thread::spawn(move || {
+        let mut tmpstr = String::new();
+        err.lines().for_each(|line| {
+            let line = line.unwrap();
+            eprintln!("{}", line);
+            let to_push = anstream::adapter::strip_str(&line).to_string() + "\n";
+            tmpstr.push_str(&to_push);
+        });
+        return tmpstr;
```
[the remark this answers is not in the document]
jaybosamiya: Oh the x was just a placeholder for the newly named variable

source/rust_verify/src/config.rs
```diff
@@ -275,6 +277,11 @@ pub fn parse_args_with_imports(
         OPT_RECORD,
         "Rerun verus and package source files of the current crate to the current directory, alongside with output and version information. The file will be named YYYY-MM-DD-HH-MM-SS.zip. If you are reporting an error, please keep the original arguments in addition to this flag",
     );
+    opts.optflag(
+        "",
+        OPT_LOCAL_STORE,
+        "Record your changes to a specific crate (and its dependencies) everytime it invokes verus.",
+    );
```
jaybosamiya, line 284: Rather than a command-line flag, we probably want it to be something project specific (with the possibility of the user opting in everywhere if they wish to). My suggestion is to check if there is a `.verus-opt-in` (as the full name of the file; this is not an extension, just the whole file name, including the `.`) in the same directory, or any ancestor directory, of the `foo.rs` file that was passed. If that file exists, then it should store to local store.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    if !std::env::args().any(|arg| arg.starts_with("--color")) {
+        if is_terminal::is_terminal(&std::io::stderr())
+            || is_terminal::is_terminal(&std::io::stdout())
+        {
+            cmd.arg("--color=always");
+        } else {
+            cmd.arg("--color=never");
+        }
+    }
```
jaybosamiya, line 82: This doesn't account for if the user explicitly ran either `verus --color auto foo.rs` or `verus --color=auto foo.rs` (in which case you should remove those argument(s) and then do this manual update you are doing)

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+            let x = anstream::adapter::strip_str(&line).to_string() + "\n";
+            stdout_output.push_str(&x);
+        });
+        return stdout_output;
+    });
+
+    let err_string = thread_err.join().unwrap();
+    let out_string = thread_out.join().unwrap();
+
+    child.wait().map_err(|x| format!("verus failed to run with error: {}", x))?;
```
jaybosamiya, line 173: You want to capture the resulting `ExitCode` you get as an output into a variable, and then pass it along to the calling function so that the exit code is handled correctly. Currently it always exits with `0` irrespective of how the child exited, which is incorrect.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
-            std::process::exit(code.0);
+        Ok(_) => {
+            std::process::exit(0);
+        }
+    }
+}
+
+pub fn exec(cmd: &mut Command, reports: Option<Reports>) -> Result<(), String> {
+    #[cfg(windows)]
+    if cfg!(windows) {
```
jaybosamiya, line 97: Only one of these two lines is needed, pick one and remove the other.

If it is messing with your rust-analyzer setup (which is what I recall to be the reason why you'd done something like this), what I might try to do is factor this portion out into a separate function (named `configure_windows_for_exec`, marked with the `#[cfg(windows)]` attribute), and then have the call here be done with a `#[cfg(windows)] configure_windows_for_exec()` or similar), and hopefully rust-analyzer is happy with that?

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    let err_string = thread_err.join().unwrap();
+    let out_string = thread_out.join().unwrap();
+
+    child.wait().map_err(|x| format!("verus failed to run with error: {}", x))?;
+
+    writeln!(file, "{}", err_string)
+        .map_err(|x| format!("failed to write to reports.toml: {}", x))?;
+    writeln!(file, "{}", out_string)
+        .map_err(|x| format!("failed to write to reports.toml: {}", x))?;
```
jaybosamiya, line 178: Why are the strings being added to the reports.toml directly? I would expect this to be a valid TOML file.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    let deps = get_dependencies(&dep_file_path)?;
+
+    let deps_path = deps.iter().map(|x| Path::new(x)).collect::<Vec<_>>();
+
+    // compute common ancester
+    let common = match common_path::common_path_all(deps_path) {
+        Some(common) => common,
+        None => PathBuf::from(""),
+    };
```
jaybosamiya, line 190: It is _extremely_ unlikely that you can't find a common path to dependencies (the only situation I can think of where this can happen is if you are on Windows, and are using many separate drives for your dependencies), and if this ever happens, it is better to be loud about it, rather than silently replace it with the empty string.

In general, it is good to be loud about horrifying and rare cases, rather than silently replace them with _potentially_ valid information, making detection and debugging harder.

```suggestion
    let common = common_path::common_path_all(deps_path)
        .expect("Getting common ancestor for dependencies");
```

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    if deps.clone().into_iter().any(|dep| dep.is_absolute()) {
```
jaybosamiya, line 192: This is likely an unnecessary clone:
```suggestion
    if deps.iter().any(|dep| dep.is_absolute()) {
```

But more importantly, I think this portion makes a few assumptions that I am not sure would always hold. Currently, it assumes that if one file is absolute, then all files are absolute (is this the case?). If it _is_ what you've observed so far, then it might be good to add that as an explicit `assert!(...)` in the body of this `if` block.

jaybosamiya: More generally, one way to handle this is by also adding the current working directory to the set of paths you are taking the common path from, so that you can then remove the outer check of whether all/any are absolute, and do a check inside to check each one.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+        .map_err(|x| format!("localstorage: failed to git commit with error message: {}", x))?;
+
+    Ok(())
+}
+
+#[derive(Debug)]
+pub struct Reports {
+    proj_path: PathBuf,
+    dep_path: PathBuf,
+}
```
jaybosamiya, line 250: Use some more descriptive names (you already know they are paths due to the type, it is prob more important to track if it is a file or dir, if you really want to hold on to that info; in this case, I'd recommend holding on to it)
```suggestion
pub struct Reports {
    project_directory: PathBuf,
    dependencies_listing_file: PathBuf,
}
```

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
         }
     }
+
+    std::fs::remove_file(&dep_file_path).map_err(|err| {
+        format!(
+            "failed to remove dependency file {} in local data cache with error message: {}",
+            dep_file_path.display(),
+            err
+        )
+    })?;
```
jaybosamiya, line 225: It's not just the dependency file, but the whole temporary directory that is supposed to be removed, right?

jaybosamiya: Also, if removing it fails, that is not the end of the world, and probably should just be reported via a `eprintln!` and then continued with doing the latter steps.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    Command::new("git").current_dir(&proj_path).arg("add").arg(".").output().map_err(|x| {
+        format!("localstorage: failed to stage current project with error message: {}", x)
+    })?;
+
+    std::process::Command::new("git")
```
jaybosamiya, line 231: Why is one of these `Command::` and the other is `std::process::Command`; good to use things consistently.

Also, if you are going to do the operations `git add .` _immediately_ followed by `git commit -m ...`, then the preferred way to do it is `git commit -am ...`.

Also, it is generally good practice to use the full versions of short options when doing them via code, since they are self-documenting a lot better than short options. Thus, instead of `-am ...` you would use `--all --message ...`.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+        .current_dir(&proj_path)
+        .arg("commit")
+        .arg("-m")
+        .arg("\"verus\"")
```
jaybosamiya, line 235: Why does this have extra quotes? This doesn't need extra quotes. Also updated to be _slightly_ more descriptive (just in case we add alternative ways to add to this in the future)

```suggestion
        .arg("verus command invocation")
```

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+pub struct Reports {
+    proj_path: PathBuf,
+    dep_path: PathBuf,
+}
+
+fn repo_path(store: bool) -> Option<Reports> {
+    // check if user has git as executable
+    if Command::new("git").arg("--version").output().is_err() {
+        return None;
+    }
```
jaybosamiya, line 256: Assuming we change to the file based opt-in described in another comment, we should not silently stop if the user doesn't have `git` available; instead we should warn (with `eprintln!("NOTE: Opted into Verus analytics, but no `git` available on system. Please make sure git is installed to be able to use analytics")` or similar), before returning `None`.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+
+    // check if user has local storage enabled during vargo build
+    if option_env!("VERUS_LOCAL_STORE").is_none() {
+        return None;
+    }
```
jaybosamiya, line 261: You can use `VERUS_IN_VARGO` to check if you are in vargo, and return `None` in that case.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+            .expect("Initializing git in local data dir");
+    }
+
+    let uuid_file = reports_dir.join("userid");
+    if !uuid_file.is_file() {
+        let uuid = uuid::Uuid::new_v4();
+        write(&uuid_file, uuid.to_string()).expect("Writing user id file");
+    }
+
+    let uuid = read_to_string(&uuid_file).expect("Reading user id file");
```
jaybosamiya, line 283: Rather than calling it a `uuid` in your variables, can you call it `user_id` or such? Best to use "what is this" rather than "what format is this in".

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+            None // passed down to rust_verify for error message
+        }
+    };
+
+    let input_file_path = input_file.and_then(|f| std::fs::canonicalize(f).ok());
+
+    let project_name = input_file_path
+        .as_ref()
+        .map(|path| format!("{:x}", sha2::Sha256::digest(path.to_str().unwrap().as_bytes())));
+    // not sure how to write a ? to replace unwrap
```
jaybosamiya, line 302: You can use `to_string_lossy` instead, which is ideal for this use case, since we don't actually want a crash/panic as part of the sha computation.

source/verus/src/main.rs
```diff
@@ -75,19 +63,270 @@ fn main() {
...
+        project_dir
+    });
+
+    let dir = std::env::temp_dir();
+    let temp_file = dir.join("verus-dep-info");
+
+    match (input_file_path, proj_dir) {
+        (Some(_), Some(proj_dir)) => Some(Reports { proj_path: proj_dir, dep_path: temp_file }),
+        _ => None,
+    }
```
jaybosamiya, line 318: Why is this not handled via `?`s when you make `input_file_path` and `proj_dir`?

tools/vargo/src/main.rs
```diff
@@ -296,6 +296,16 @@ fn run() -> Result<(), String> {
...
+    let local_store = args_bucket
+        .iter()
+        .position(|x| x.as_str() == "--local-store")
+        .map(|p| args_bucket.remove(p))
+        .is_some();
+
+    if local_store {
+        std::env::set_var("VERUS_LOCAL_STORE", "1");
+    }
+
```
jaybosamiya, line 308: I recommend removing this entirely. We shouldn't be touching vargo for this feature.

jaybosamiya (changes requested):
Thanks for the updates! This is now in a nicer state; some more comments added inline.

source/verus/src/main.rs
jaybosamiya: Cool, if you can confirm that it is not relevent, perfectly fine to drop this. I mentioned this about vargo because your comment is explicitly talking about vargo here.

source/verus/src/main.rs
jaybosamiya: Ah, good point; yeah I missed that, indeed you are correct.  Nonetheless, the other parts of this comment still do hold.

utaal:
Thank you for all the work on this. After a significant refactor, I have re-implemented a similar mechanism re-using the `--record` machinery extensively. The new implementation is https://github.com/verus-lang/verus/pull/798

I am closing this, as https://github.com/verus-lang/verus/pull/798 supersedes it.

PR #693, opened 2023-07-19
Generate error report zip file when verus do not have stdout

ahuoguo:
json parse will fail when stdout is empty, but we still want the files for error report purpose.

I don't think there's a concern for `stderr` since it is not parsed and just wrapped with a `toml::Value::String` constructor

### DIFF
--- a/source/error_report/src/main.rs
+++ b/source/error_report/src/main.rs
@@ -1,4 +1,5 @@
 use chrono::{prelude::*, DateTime};
+use regex::Regex;
 use std::time::{Duration, Instant};
 use std::{
     env,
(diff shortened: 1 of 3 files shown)


source/error_report/src/main.rs
```diff
@@ -164,22 +164,37 @@ fn toml_setup_and_write(
...
 
-    let version_info: toml::Value = serde_json::from_str(&stdout_json["verus"].to_string())
-        .map_err(|x| {
+    let stdout_json = match stdout_string.len() {
+        0 => {
+            eprintln!("{}: verus did not generate stdout", yansi::Paint::yellow("warning"));
+            serde_json::Value::Null
+        }
+        _ => serde_json::from_str::<serde_json::Value>(&stdout_string)
+            .map_err(|x| format!("failed to parse stdout to json file with error message {}", x))?,
```
utaal, line 175: Instead of failing, in this case, should we store the stdout as a string in the toml? (like we do for stderr)
In that case I think you don't need to check the length of the output and then just store the stdout as a string if you cannot parse it as json (I think we should print a warning in this case).

source/error_report/src/main.rs
```diff
@@ -164,22 +164,37 @@ fn toml_setup_and_write(
...
     let verification_result: toml::map::Map<String, Value> =
-        serde_json::from_str(&stdout_json["verification-results"].to_string()).map_err(|x| {
-            format!("failed to parse verification results with error message {}", x)
-        })?;
-
-    let stdout_toml_text = serde_json::from_str(&stdout_json.to_string())
-        .map_err(|x| format!("failed to parse stdout (json) with error message {}", x))?;
+        match stdout_json {
+            serde_json::Value::Null => toml::map::Map::new(),
+            _ => serde_json::from_str(&stdout_json["verification-results"].to_string()).map_err(
```
utaal, line 188: Why do we need to string-ify and then parse stdout again?

Issue #449, opened 2023-03-17
panic for non-mut variables when --no-lifetime is specified


tjhance:
This is pretty low-priority, since it only applies when --no-lifetime is specified, but Verus panics if you try to mutate a non-mut variable:

```rust
fn test() {
    let x: u32 = 5;
    takesmut(&mut x); 
}
```

utaal:
[Verus all-hands]

@tjhance has started on a patch.

PR #696, opened 2023-07-20
Initial support for mut params

jcp19:
This PR introduces initial support for mutable parameters. Given a function named `f` with mut parameters `x_0`, ..., `x_n` and body `f_body`, we rewrite it to a function without mut params and body `let mut x_0 = x_0; ...; let mut x_n = x_n; f_body`.

There are a few features that are still not supported, which I plan to add in a future PR:
- support for `mut self`
- support for mut parameters of `&mut T` types
    - Currently, Verus does not allow for declaring a variable of type `&mut T` and assigning to it. As such, the transformation that I described above does not work for parameters of these types yet; 
    - Because of this, I could not yet implement support for special warnings when referring to mutable parameters of type `&mut T` in postconditions


Chris-Hawblitzel (approved):

source/rust_verify/src/lifetime_emit.rs
```diff
@@ -694,6 +694,11 @@ pub(crate) fn emit_fun_decl(state: &mut EmitState, f: &FunDecl) {
     state.push_indent();
     for (span, x, typ) in f.params.iter() {
         state.newline();
+        // For checking lifetimes, we make all function parameters mutable.
+        // This way, we do not have to closely track which parameters are
+        // mutable and which are not. This should not have an impact on
+        // the result of the lifetime analysis.
+        state.write("mut ");
```
tjhance, line 701: I'm not sure this is correct. `mut` checking is one of the things that is done by the "lifetime analysis" so I think we need to preserve this information from the source.

jcp19: can you elaborate on how `mut` is checked at this point? I thought that assigning to a non-mutable variable would have been caught at an earlier point.

tjhance: Well I don't know why rustc does it this way, but it does. mut-checking isn't done by this point, but it gets caught in the lifetime-checking stage instead.

Furthermore, Verus actually relies on the mut analysis being done correctly ... (there's actually an issue for it https://github.com/verus-lang/verus/issues/449) but even setting that issue aside, we still need to get the mut checks done in any case

jcp19: Oh, I see. I will address this. Thank you!

jcp19: done

source/rust_verify/src/lifetime_ast.rs
```diff
@@ -187,7 +187,7 @@ pub(crate) struct FunDecl {
     pub(crate) name: Id,
     pub(crate) generic_params: Vec<GenericParam>,
     pub(crate) generic_bounds: Vec<GenericBound>,
-    pub(crate) params: Vec<(Option<Span>, Id, Typ)>,
+    pub(crate) params: Vec<(Option<Span>, Id, Typ, bool)>,
```
utaal, line 190: Can you please change this to a struct, so that we can name the `bool`?

source/rust_verify_test/tests/basic.rs
utaal, line 1: Can you add a test where the parameter's type is not primitive? E.g. a `struct` with an assignment to a field in the body.

utaal (approved):
I have a couple of comments, but looks good to me otherwise. Thank you!

utaal:
Please squash when merging. Thank you!

Issue #698, opened 2023-07-24
Turn const values to upper cases in the guide.


HaoYang670:
https://github.com/verus-lang/verus/blob/2dc6a17858e241ded90617375d9ce2784576afb7/source/rust_verify/example/guide/modes.rs#L353-L369

To follow rust's naming style, we should rename them to `ONE` and `SPEC_ONE`

PR #700, opened 2023-07-24
Style fix in guide

ahuoguo:
Fixes https://github.com/verus-lang/verus/issues/698

To follow rust's naming style, we should rename them to ONE and SPEC_ONE


jaybosamiya (approved):
Thanks, looks good to me. Please squash-and-merge. Thanks!

Magic to make sure it closes the issue: Closes #698

PR #706, opened 2023-07-25
Remove redundant type variables in the guide.

HaoYang670:
I am reading https://verus-lang.github.io/verus/guide/spec_lib.html and find a TODO is left: `TODO: lemma_len_intersect::<A>(...) should not need the ::<A>`. I removed those type variables and verus can also work.  It seems that verus has supported type inference.


utaal:
Hi! Thank you for fix!

Issue #701, opened 2023-07-24
=~= not doing its thing on a simple Option; maybe an extensionality problem for additive types?


jonhnet:
In the before file, Verus fails to prove this extensional equality:
```
        assert( post.iptr(ptr) =~= PagedJournal_v::JournalRecord::discard_old_journal_rec(self.iptr(ptr), lsn) );
```
In the after file, I "fix" it by beginning a manual proof of extensional equality on the branches of the Option. Once I prove that the Option legs match, Verus is happy.

[2023-07-24-15-01-48.zip](https://github.com/verus-lang/verus/files/12152862/2023-07-24-15-01-48.zip)
[2023-07-24-15-01-54.zip](https://github.com/verus-lang/verus/files/12152863/2023-07-24-15-01-54.zip)

utaal:
It seems verus usually needs
```
        assert(post.iptr(ptr).is_Some() == PagedJournal_v::JournalRecord::discard_old_journal_rec(self.iptr(ptr), lsn).is_Some());
```
to prove extensional equality. Can you spell out the argument why this is true? (which definitions / lemmas this comes from)

PR #707, opened 2023-07-25
add verifier::ext_equal to external type spec for core::option::Option

utaal:


utaal:
This is an attempt to address https://github.com/verus-lang/verus/issues/701, but seems insufficient. I think #701 is in part due to instability (the outcome changes by adding/removing `#[verifier::spinoff_prover]`).

tjhance (approved):

PR #710, opened 2023-07-25
Add line number in comments for some functions (to support Mariposa's Verus integration features)

jetline0:
This feature is intended to make it easier for [Mariposa](https://github.com/secure-foundations/mariposa) to determine the original location of unstable Verus queries from their corresponding SMT2 files.

Specifically, the additional comment is for queries labeled as Function-Defs, Function-Decl-Check-Recommends, and Function-Terminations.
Ex:
```
;; Function-Def main::impl_u::l1::Directory::lemma_resolve_structure_assertions
;; /home/gelatin/verified-nrkernel/page-table/impl_u/l1.rs:744:5: 744:80 (#0)
```

A similar comment is already present in some SMT2 files generated by Verus, but the comment only appears in queries that use the spinoff prover:
Ex:
```
;; MODULE 'impl_u::l2_impl'
;; /home/gelatin/verified-nrkernel/page-table/impl_u/l2_impl.rs:358:9: 358:15 (#0)
```

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -591,7 +591,7 @@ impl Verifier {
         qid_map: &HashMap<String, vir::sst::BndInfo>,
         module: &vir::ast::Path,
         function_name: Option<&Fun>,
-        comment: (&str, &str),
+        comment: &str,
         desc_prefix: Option<&str>,
     ) -> bool {
         if let Some(verify_function) = &self.args.verify_function {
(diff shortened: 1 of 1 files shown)


source/rust_verify/src/verifier.rs
```diff
@@ -591,7 +591,7 @@ impl Verifier {
         qid_map: &HashMap<String, vir::sst::BndInfo>,
         module: &vir::ast::Path,
         function_name: Option<&Fun>,
-        comment: &str,
+        comment: (&str, &str),
```
utaal, line 594: Can you take the span as a new parameter instead, and stringify it here instead? This way we cannot forget to add it if we add new calls to `run_commands_queries`.

utaal (changes requested):
This looks fine to me, modulo one hopefully minor change. Thanks!

source/rust_verify/src/verifier.rs
jetline0: I think the commit below should fix this problem! Turns out `commands_with_context` contains the span information I needed, so no span information needs to be passed into `run_commands_queries.`

utaal:
Can you please run `vargo fmt`?

utaal:
Thank you!

Issue #702, opened 2023-07-24
extensionality doesn't extend to dom()


jonhnet:
In the before file, Verus can't prove
```
            assert( other.entries =~~= tight.entries );  // flaky
```
In the after file, the proof starts working if I ask it to work on domain equality, which I'd think is pretty much step 1 of extensionality:
```
            assert( tight.entries.dom() =~~= other.entries.dom() );
```
[2023-07-24-15-06-14.zip](https://github.com/verus-lang/verus/files/12152881/2023-07-24-15-06-14.zip)
[2023-07-24-15-06-23.zip](https://github.com/verus-lang/verus/files/12152882/2023-07-24-15-06-23.zip)

utaal:
[Mariposa](https://github.com/secure-foundations/mariposa) reports:

* before adding `assert( tight.entries.dom() =~~= other.entries.dom() );`:

```
query: gen/journal__LinkedJournal_v.smt2_/split.1.smt2
| mutation   | status     | unsat       | timeout   | unknown      | mean(second)   | std(second)   |
|------------|------------|-------------|-----------|--------------|----------------|---------------|
| overall    | unstable   | x           | x         | x            | x              | x             |
| shuffle    | unstable   | 47/61 77.0% | 0/61 0.0% | 14/61 23.0%  | 0.02           | 0.0           |
| rename     | unsolvable | 0/61 0.0%   | 0/61 0.0% | 61/61 100.0% | 0.02           | 0.0           |
| reseed     | unsolvable | 0/61 0.0%   | 0/61 0.0% | 61/61 100.0% | 0.02           | 0.0           |
| quake      | unstable   | 42/61 68.9% | 0/61 0.0% | 19/61 31.1%  | 0.01           | 0.0           |
```

* after adding `assert( tight.entries.dom() =~~= other.entries.dom() );`

```
query: gen/journal__LinkedJournal_v.smt2_/split.1.smt2
| mutation   | status   | unsat        | timeout   | unknown   | mean(second)   | std(second)   |
|------------|----------|--------------|-----------|-----------|----------------|---------------|
| overall    | stable   | x            | x         | x         | x              | x             |
| shuffle    | stable   | 61/61 100.0% | 0/61 0.0% | 0/61 0.0% | 0.02           | 0.0           |
| rename     | stable   | 61/61 100.0% | 0/61 0.0% | 0/61 0.0% | 0.02           | 0.0           |
[...]

PR #713, opened 2023-07-26
allow #![trigger f(x)] on ensures (of broadcast_forall) functions

utaal:
This started from an attempt to add additional triggers to a broadcast forall for map extensional equality, to address this https://github.com/verus-lang/verus/issues/702 but seems independently useful.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -315,23 +315,63 @@ impl Visitor {
                 ));
             }
         }
-        if let Some(Ensures { token, mut exprs }) = ensures {
+        if let Some(Ensures { attrs, token, mut exprs }) = ensures {
             if exprs.exprs.len() > 0 {
                 for expr in exprs.exprs.iter_mut() {
                     self.visit_expr_mut(expr);
                 }
-                if let Some((p, ty)) = ret_pat {
-                    stmts.push(Stmt::Semi(
-                        Expr::Verbatim(
-                            quote_spanned!(token.span => ::builtin::ensures(|#p: #ty| [#exprs])),
-                        ),
-                        Semi { spans: [token.span] },
-                    ));
-                } else {
-                    stmts.push(Stmt::Semi(
-                        Expr::Verbatim(quote_spanned!(token.span => ::builtin::ensures([#exprs]))),
-                        Semi { spans: [token.span] },
-                    ));
+                let cont = match self.extract_quant_triggers(attrs, token.span) {
+                    Ok(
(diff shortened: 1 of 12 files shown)


source/builtin_macros/src/syntax.rs
```diff
@@ -315,23 +315,62 @@ impl Visitor {
...
-                        Expr::Verbatim(quote_spanned!(token.span => ::builtin::ensures([#exprs]))),
-                        Semi { spans: [token.span] },
-                    ));
+                let cont = match self.extract_quant_triggers(attrs, token.span) {
+                    Ok(
+                        found @ (ExtractQuantTriggersFound::Auto
+                        | ExtractQuantTriggersFound::Triggers(..)),
+                    ) => {
+                        if exprs.exprs.len() != 1 {
+                            let err = "when using #![trigger f(x)], only one ensures is allowed";
```
Chris-Hawblitzel, line 329: Suggestion:

```
                        if exprs.exprs.len() == 0 {
                            let err = "when using #![trigger f(x)], at least one ensures is required";
```

This works fine:

```
            #[verifier::external_body]
            #[verifier::broadcast_forall]
            pub proof fn ab(a: A)
                ensures #![trigger a.f1()]
                    a.f3(),
                    a.f1() ==> a.f2(),
            {
            }
```

Chris-Hawblitzel (approved):

source/builtin_macros/src/syntax.rs
utaal: This will produce 
```
::builtin::ensures([
  ::builtin::with_triggers( ( (a.f1(),), ), a.f3() ),
  a.f1() ==> a.f2(),
])
```
The two expressions are conjoined (&&) when we generate the AIR. It sounds like we interpret it correctly anyway, but it felt like it may subtly break our implicit assumptions on what goes in `with_triggers`, which may result in (silent) breakage later.

Chris-Hawblitzel: We already handle this for the requires anyway.  When you write:

```
            pub proof fn ab(a: A)
                requires
                    a.f1(),
                ensures #![trigger a.f1()]
                    a.f2() && a.f3(),
```

poly.rs turns this into `a.f1() ==> with_triggers[a.f1()](a.f2() && a.f3())`.  The processing in triggers.rs is designed to handle this.  Even if at some point we want to change how we process with_triggers internally, I don't think that this should be visible as a restriction on user code written in Verus.

utaal: Ah, that makes sense, I missed that part of the code path. Thanks for the clarification.

PR #716, opened 2023-07-26
Clean up the `docs` directory

parno:
Now that we've started to consolidate our outward facing documentation in the online tutorial and reference guide, it seems useful to clean up some of the extraneous files in the `verus/source/docs` directory, so that users don't stumble on outdated/internal notes.

Thus far, I have removed documents that appear to be outdated, and I moved talk/poster information into the verus-papers repo.   Apologies in advance if I accidentally deleted a file or files that are still useful/relevant (e.g., @chanheec I deleted files related to bit vectors and non-linear arith on the theory that I believe they're all subsumed by your documentation PR #597).  Please feel free to point out things that should be saved!

After the purge above, there are three remaining doc files to deal with.

@tjhance Do we still need [trait-notes.md](https://github.com/verus-lang/verus/blob/main/source/docs/trait-notes.md)?  If so, perhaps we can create a new `internal` directory and put it there to communicate that we don't expect Verus users to read such docs.

@Chris-Hawblitzel Has [modes.md](https://github.com/verus-lang/verus/blob/main/source/docs/design/modes.md) been subsumed by the official tutorial and/or the upcoming OOPSLA paper?

@utaal Could [syntax.md](https://github.com/verus-lang/verus/blob/main/source/docs/manual/syntax.md) possibly be incorporated into your attribute documentation on #638 ?


tjhance:
> @tjhance Do we still need [trait-notes.md](https://github.com/verus-lang/verus/blob/main/source/docs/trait-notes.md)? If so, perhaps we can create a new internal directory and put it there to communicate that we don't expect Verus users to read such docs.

I've been meaning to return to that file and do a more thorough write-up. Can you make an internal directory like you suggested?

utaal:
> @utaal Could [syntax.md](https://github.com/verus-lang/verus/blob/main/source/docs/manual/syntax.md) possibly be incorporated into your attribute documentation on https://github.com/verus-lang/verus/pull/638 ?

Yes! I'll work on that.

parno:
> I've been meaning to return to that file and do a more thorough write-up. Can you make an internal directory like you suggested?

Done!

> Yes! I'll work on that.

Great, thanks!

Chris-Hawblitzel:
I think `modes.md` has some value but would be better in the `internal` directory.

parno:
Thanks all!  I implemented the actions discussed above and in the Verus meeting, so I'm going to merge this in.

PR #721, opened 2023-07-27
add a --shuffle flag that performs a simplified Mariposa-style stability analyis on a function

utaal:
This is inspired by the [Mariposa](https://github.com/secure-foundations/mariposa) methodology, and is based on @yizhou7's shuffling technique
(https://github.com/secure-foundations/mariposa/blob/eb2c020/src/main.rs#L126-L141)

It's likely not as effective as Mariposa's in its shuffling, it doesn't do any of the hypothesis testing, but it's a way to do a quick stability check with shuffling.

### DIFF
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -1256,6 +1256,7 @@ dependencies = [
  "getopts",
  "num-bigint 0.4.3",
  "num-format",
+ "rand 0.8.5",
  "regex",
  "rust_verify_test_macros",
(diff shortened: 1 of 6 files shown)


jaybosamiya:
Just a minor comment: might be good to consider `--stability shuffle` or `--stability-test shuffle` as the flag, in order to leave space in the future to support other stability tests without increasing the number of flags itself?

Issue #656, opened 2023-06-29
Can't add spec for external function with ~const trait bound


jaylorch:
Some external functions, like `Option::unwrap_or`, have trait bounds involving `~const`. If you don't take this into account when writing a specification for it, you get an error about a trait-bound mismatch. For example, if you write:

```
#[verifier(when_used_as_spec(spec_unwrap_or))]
#[verifier::external_fn_specification]
pub fn unwrap_or<T>(option: Option<T>, default: T) -> (t: T)
    ensures
        option.is_Some() ==> t == option.get_Some_0(),
        option.is_None() ==> t == default
{
    option.unwrap_or(default)
}
```
then you get the error:
```
Proxy function trait bounds: [
    Binder(TraitPredicate(<T as std::marker::Sized>, polarity:Positive), []),
]
External function trait bounds: [
    Binder(TraitPredicate(<T as std::marker::Sized>, polarity:Positive), []),
    Binder(~const TraitPredicate(<T as std::marker::Destruct>, polarity:Positive), []),
]
error: external_fn_specification requires function type signature to match exactly (trait bound mismatch)
   --> pervasive/std_specs/option.rs:144:1
    |
144 | pub fn ex_unwrap_or<T>(option: Option<T>, default: T) -> (t: T)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
(By the way, the format of this error is just a temporary hack to report the specific trait bound. It should probably be reformatted for clarity.)

But if you do add the trait bound, e.g.,:
```
use std::marker::Destruct;

#[verifier(when_used_as_spec(spec_unwrap_or))]
#[verifier::external_fn_specification]
pub fn unwrap_or<T>(option: Option<T>, default: T) -> (t: T)
    where
        T: ~const Destruct,
    ensures
        option.is_Some() ==> t == option.get_Some_0(),
        option.is_None() ==> t == default
{
    option.unwrap_or(default)
}
```
you get lots of errors suggesting Verus doesn't really support this:
```
error: `~const` is not allowed here
   --> pervasive/std_specs/option.rs:148:12
    |
148 |         T: ~const Destruct,
    |            ^^^^^^^^^^^^^^^
    |
note: this function is not `const`, so it cannot have `~const` trait bounds
   --> pervasive/std_specs/option.rs:146:8
    |
[...]

PR #726, opened 2023-07-28
Specify Option::unwrap_or by ignoring Destruct trait bound mismatches

jaylorch:


jaylorch:
Fixes #656

tjhance:
I think this might not be right. After looking into exactly what `~const` is for (https://internals.rust-lang.org/t/pre-rfc-revamped-const-trait-impl-aka-rfc-2632/15192) it seems that `A: ~const B` should probably just be treated as `A: B` for our purposes, rather than being ignored.

However, the special `Destruct` trait (which iirc was the root of issue #656) can be ignored.

jaylorch:
OK, I changed it to ignore Destruct trait bounds instead of ~const ones.

tjhance (approved):

PR #728, opened 2023-07-29
Support substring match for --verify-function, fix --verify-root problem

jialin-li:
support unique substring match, beginning and ending wildcard matches for `--verify-function`
fixes the `--verify-root` bug where `--verify-root` will include both the root module and vstd's root module.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -591,7 +591,7 @@ impl Verifier {
         qid_map: &HashMap<String, vir::sst::BndInfo>,
         module: &vir::ast::Path,
         function_name: Option<&Fun>,
-        exact_match: bool,  // for --verify-function flag
+        exact_match: bool, // for --verify-function flag
         comment: &str,
         desc_prefix: Option<&str>,
     ) -> bool {
@@ -977,7 +977,8 @@ impl Verifier {
                     .join("\n");
                     return Err(error(msg));
                 }
-            } else { // wildcard match
+            } else {
+                // wildcard match
                 let left_wildcard = verify_function.starts_with("*");
                 let right_wildcard = verify_function.ends_with("*");
                 let mut wildcard_mismatch = false;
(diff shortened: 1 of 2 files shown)


source/rust_verify/src/verifier.rs
```diff
@@ -913,14 +930,21 @@ impl Verifier {
             funs.insert(function.x.name.clone(), (function.clone(), vis_abs));
         }
 
+        let mut exact_match = false;
```
jaybosamiya, line 933: Might be good to call this `verify_function_exact_match` or similar?

source/rust_verify/src/verifier.rs
```diff
@@ -934,6 +958,55 @@ impl Verifier {
...
+                let right_wildcard = verify_function.ends_with("*");
+                let mut wildcard_mismatch = false;
+
+                if left_wildcard && !right_wildcard {
+                    wildcard_mismatch =
+                        !filtered_functions.iter().any(|f| f.ends_with(clean_verify_function));
+                } else if !left_wildcard && right_wildcard {
+                    wildcard_mismatch =
+                        !filtered_functions.iter().any(|f| f.starts_with(clean_verify_function));
+                }
```
jaybosamiya, line 992: Alternative way of writing this code that eliminates the `let mut`
```suggestion
                let wildcard_mismatch = match (verify_function.starts_with("*"), verify_function.ends_with("*")) {
                  (true, false) => !filtered_functions.iter().any(|f| f.ends_with(clean_verify_function)),
                  (false, true) => !filtered_functions.iter().any(|f| f.starts_with(clean_verify_function)),
                  _ => false,
                };
```

You can also instead do it as a `let wildcard_mismatch = if ...` if you prefer that

jaybosamiya (approved):
Thanks!  This PR makes `--verify-function` a lot easier to work with for users.

Some minor comments inline.

Also, we probably want to update the `verus --help` message which currently says:
```console
$ verus --help
...
        --verify-function MODULE
                        Verify just one function (e.g. 'foo' or 'foo::bar')
                        within the one module specified by verify-module or
                        verify-root
...
```

source/rust_verify/src/verifier.rs
jialin-li: Both addressed in comments below!

utaal (approved):
Thank you!

I pushed a commit with a couple of fixes for issues that were already present, but that I only noticed when giving this a try.

Issue #727, opened 2023-07-28
--profile without --verify-function silently does nothing


jonhnet:
For some reason when we added --verify-function, we got "exceeded rlimit", and then it emitted profiler output. If the profiler requires verify-function, it should quit with an error if the flag is not supplied. If it's not emitting output because of not exceeding rlimit, why? We want to debug slow things too.

cc: @jialin-li

parno:
When you run with `--profile`, you only get analysis for functions that exceed the rlimit (on the theory that those are the functions of interest for profiling).  If you always want profiler output, e.g., to debug slow things, you should use `--profile-all`.

jonhnet:
Ah! Got it. We should add an explanatory message in the case the user asked for profiling and got nothing.

jonhnet:
@jialin-li I assigned you to add this message. You can make me do it if you want

jialin-li:
@jonhnet  the current help page for `--profile` says "collect and report prover performance data when resource limits are hit". Do we want to always remind users of this fact whenever they use this flag, or wait till verification succeeds and then suggest `--profile-all` as an alternative?

jonhnet:
I think the second thing. When we used it today we were surprised that we got no output. That's the perfect time to clarify what happened and what to do about it.

PR #729, opened 2023-07-29
suggest --profile-all when --profile is used without exceeding rlimits

jialin-li:
address #727


utaal (approved):
Thank you! Please rebase when merging.

jonhnet (approved):
thanks!

PR #730, opened 2023-07-30
Singular update

yizhou7:
Extend the expressiveness of `integer_ring`
* allow `a % m == b % m` in the ensures
* allow multiple ensures 

Bug fixes:
* avoid writing to a Singular process that has been terminated
* ask for acknowledgement when a query is done


jaybosamiya:
Some of the examples and explanation in PR #597 can be made cleaner after this PR is merged.

source/air/src/singular_manager.rs
```diff
@@ -22,29 +21,52 @@ impl SingularManager {
...
 }
 
+fn singular_writer_thread(requests: Receiver<Vec<u8>>, mut singular_pipe_stdin: ChildStdin) {
+    while let Ok(req) = requests.recv() {
+        singular_pipe_stdin
+            .write_all(&req)
+            .and_then(|_| writeln!(&singular_pipe_stdin))
+            .and_then(|_| singular_pipe_stdin.flush())
+            // The Singular process could die unexpectedly. In that case, we die too:
+            .expect("IO error: failure when sending data to Singular process across pipe");
```
utaal, line 48: Can you elaborate on when you've seen this happen? Does it appear to be in response to a command we send?

yizhou7: The old version was using `smt_process::writer_thread`, which sends an `echo` to Z3 after each command(s) for confirmation. Singular does not have `echo`, so sending that to a live Singular process will give an error. Further, Verus stops Singular by issuing `quit`. The process might not end immediately, so Verus might still send that `echo` to a process that is about to die, ending up with a broken pipe error message.

source/rust_verify/src/singular.rs
```diff
@@ -339,67 +336,66 @@ pub fn check_singular_valid(
...
-            "{} \ngenerated singular query: {}",
-            res[0].as_str(),
-            query
-        )))
-    } else {
-        ValidityResult::Invalid(
-            None,
-            air::messages::error(
-                format!(
-                    "postcondition not satisfied: Ensures polynomial failed to be reduced to zero, reduced polynomial is {}\n generated singular query: {} ",
```
utaal, line 397: My understanding is that with this change we are now only emitting "Failed to translate this expression into a singular query" as the failure message. That may not be ideal, as it doesn't indicate to the user if what they're trying to prove is not true, or if they wrote an invalid query.

Can "Failed to translate this expression into a singular query" indicate that the requires/ensures is invalid? Or is it always the equivalent of an "assertion violation" message?

source/rust_verify/src/singular.rs
```diff
@@ -339,67 +336,66 @@ pub fn check_singular_valid(
...
-                &ens.1.spans[0],
-            ),
-        )
+
+    for stmt in &*stmts {
+        if let air::ast::StmtX::Assert(err, expr) = &**stmt {
+            assert_eq!(err.labels.len(), 1);
+            let span = &err.spans[0];
+            let label = &err.labels[0].note;
+            if label.contains("require") {
```
utaal, line 357: We should not be using the error message label string to direct encoding choices, we should distinguish between them in the vir ast.

I think here https://github.com/verus-lang/verus/blob/main/source/vir/src/sst_to_air.rs#L2554-L2559 we are constructing the assumptions for the z3 query, and then we throw them away if it's a singular query. We may want to do something similar to what the z3 query does where the assumptions are passed in to the `QueryX` as locals.

source/rust_verify_test/tests/integer_ring.rs
utaal, line 1: All of the failing tests in this file do not check what error message is produced (it's ignored with `Err(_)`): can you add that check?

source/rust_verify/src/singular.rs
```diff
@@ -339,67 +336,66 @@ pub fn check_singular_valid(
...
+                            )));
+                        } else {
+                            return ValidityResult::Invalid(
+                                None,
+                                err.clone().secondary_label(
+                                    span,
+                                    format!("Singular returned:\n {}", res.join("\n")),
+                                ),
+                            );
+                        }
```
utaal, line 391: Can you please add a comment explaining how the output you are matching against indicates the outcome? (Why does "redefining" still allow for success? Why does the question mark indicate failure?)

utaal:
Can you add a few tests that expect failure for queries that are not supported by singular? (E.g. inequality.)
It would be good to have at least a bit of a smoke test against letting through queries that are unsupported / can cause unexpected results.

utaal:
Discussed with @yizhou7. Marking this as draft for now.

source/rust_verify/src/singular.rs
yizhou7: I have made some updates and the error messages are different depending on: 1. invalid input 2. internal encoding error 3. failed proof

source/rust_verify/src/singular.rs
yizhou7: I updated how air asserts are added, so that the requires and ensures are separated into two blocks:
https://github.com/verus-lang/verus/blob/d86cf9f6ca18185ea12cf74b2acefef98d20866e/source/vir/src/sst_to_air.rs#L2473
although still implicit about the first block being the requires and second block being the ensures

source/rust_verify_test/tests/integer_ring.rs
yizhou7: It is added to the checks now.

source/rust_verify/src/singular.rs
yizhou7: I added some comments on those.

yizhou7:
> Can you add a few tests that expect failure for queries that are not supported by singular? (E.g. inequality.) It would be good to have at least a bit of a smoke test against letting through queries that are unsupported / can cause unexpected results.

I added a few of those, and updated the names of the tests so it is clearer what they are for.

source/rust_verify/src/singular.rs
```diff
@@ -348,56 +363,81 @@ pub fn check_singular_valid(
...
+    for (query, err) in queries {
+        air::singular_manager::log_singular(context, &query, &func_span.as_string);
+        let res = singular_process.send_commands(query.as_bytes().to_vec());
+        if (res.len() == 1) && (res[0] == "0") {
+            // this query is ok (poly reduced to 0)
+            continue;
+        } else if (res.len() == 2) && (res[1] == "0") {
+            // this query is switching from the previous one
+            // this will redefine the ring
+            // ignore the first line of the output
```
utaal-b, line 423: I don't understand what this means: is this a line that appears as a "trailer" to a previous query? Please describe the expected interaction in more detail.

source/rust_verify/src/singular.rs
```diff
@@ -309,35 +307,52 @@ impl SingularEncoder {
...
+        stmts
     } else {
         panic!("internal error: integer_ring")
     };
 
+    let enss = if let air::ast::StmtX::Block(stmts) = &*reqs_enss[1] {
+        stmts
+    } else {
+        panic!("internal error: integer_ring")
+    };
```
utaal-b, line 352: This sort of code is what I'd hope to avoid through my suggestion below to use a `Query` or `StmtX` type specific to Singular.

source/rust_verify/src/singular.rs
```diff
@@ -5,6 +5,7 @@ use air::singular_manager::SingularManager;
 use sise::Node;
```
utaal-b, line 5: For all the error messages in this file, and `well_formed` can you clarify whether a certain expression or feature is unsupported. For example:

`usage of the keyword {} is not supported when using the integer_ring solver`
instead of
"usage of reserved keyword at variable name: {} in integer_ring",

source/vir/src/sst_to_air.rs
```diff
@@ -2439,39 +2439,44 @@ pub(crate) fn body_stm_to_air(
...
-            local: Arc::new(singular_vars),
-            assertion: Arc::new(air::ast::StmtX::Block(Arc::new(singular_stmts))),
-        });
+        // split the requires and ensures into separate blocks
+        let assertion = Arc::new(StmtX::Block(Arc::new(vec![
+            air::ast::StmtX::Block(Arc::new(singular_req_stmts)).into(),
+            air::ast::StmtX::Block(Arc::new(singular_ens_stmts)).into(),
+        ])));
+
+        let query = Arc::new(QueryX { local: Arc::new(singular_vars), assertion });
```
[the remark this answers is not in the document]
yizhou7: I added a struct `SingularQueryX`, instead of using `QueryX`. 
I tried making `QueryX` an enum that has two variants, but the change is more invasive.
I introduced a `CheckSingular` variant in `CommandX`.

source/rust_verify/src/singular.rs
yizhou7: I added some more comments on this. Basically, we redefine the ideal we work with, so we expect the redefinition  comment.

source/rust_verify/src/singular.rs
```diff
@@ -348,56 +363,81 @@ pub fn check_singular_valid(
...
+            // this query is ok (poly reduced to 0)
+            continue;
+        } else if (res.len() == 2) && (res[1] == "0") {
+            // this query is switching from the previous one
+            // this will redefine the ring
+            // ignore the first line of the output
+            assert!(res[0].contains("// ** redefining"));
+            continue;
+        } else if res[0].contains("?") {
+            // this query does not parse
```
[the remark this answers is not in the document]
yizhou7: I added some comments on this. The bottom line here is that "?" is not an expected character in the usual input/output. If the query parses  and works, we expect 0. If the query parses but reduction fails, we expect a polynomial. The parsing errors are prompted with error messages starting with "    ?".  I don't really want to match how many spaces there are in terms of indentations.

source/rust_verify/src/singular.rs
```diff
@@ -348,56 +363,81 @@ pub fn check_singular_valid(
...
+            return ValidityResult::UnexpectedOutput(String::from(format!(
+                "{} \ngenerated singular query: {}",
+                res[0].as_str(),
+                query
+            )));
+        } else {
+            let err = error(
```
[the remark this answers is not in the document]
yizhou7: I added some comments. It produces a non-zero polynomial that is the result of the reduction.

source/rust_verify/src/singular.rs
yizhou7: Yeah this part is now eliminated.

source/rust_verify/src/singular.rs
yizhou7: I made some adjustments to the error message.

source/air/src/focus.rs
```diff
@@ -18,6 +18,10 @@ pub fn focus_command_on_assert_id(command: &Command, assert_id: &AssertId) -> Op
             let query = Arc::new(QueryX { local: query.local.clone(), assertion });
             Some(Arc::new(CommandX::CheckValid(query)))
         }
+        CommandX::CheckSingular(..) => {
+            // TODO: what should we do here?
```
yizhou7, line 22: I am not exactly sure what to do with assertion ids in Singular queries.

utaal: This is part of the new `expand-errors` @tjhance wrote. I imagine we don't support expanding through singular queries, maybe @tjhance can comment on whether returning `None` is correct here?

source/air/src/ast.rs
```diff
@@ -202,13 +202,23 @@ pub struct QueryX {
...
-    Pop,                     // pop temporary global declarations
-    SetOption(Ident, Ident), // set-option option value (no colon on the option)
-    Global(Decl),            // global declarations
-    CheckValid(Query), // SMT check-sat (reporting validity rather than satisfiability), Possibly singular checks
+    Push,                         // push space for temporary global declarations
+    Pop,                          // pop temporary global declarations
+    SetOption(Ident, Ident),      // set-option option value (no colon on the option)
+    Global(Decl),                 // global declarations
+    CheckValid(Query),            // SMT check-sat (reporting validity rather than satisfiability)
+    CheckSingular(SingularQuery), // Singular check
```
utaal, line 223: It would be good, but not critical, if you could make the existence of this variant dependent on whether the `singular` feature is enabled. (with `#[cfg(feature = "singular")]`).
The same applies to the code in other files that handle it.

source/air/src/singular_manager.rs
```diff
@@ -2,28 +2,31 @@ use std::io::{BufRead, BufReader, Write};
...
-            .and_then(|_| smt_pipe_stdin.flush())
-            // The Singular process could die unexpectedly.  In that case, we die too:
-            // TODO: https://github.com/verus-lang/verus/pull/730
+            // ask for acknowledgement
+            .and_then(|_| writeln!(&singular_pipe_stdin, "print (\"{}\");", DONE))
+            .and_then(|_| singular_pipe_stdin.flush())
             .expect("IO error: failure when sending data to Singular process across pipe");
     }
-    // Exit when the other side closes the channel
+    // singular_pipe_stdin.write_all(b"quit;\n").expect("IO error: failure quitting Singular process");
```
utaal, line 29: Why is this commented out?

source/rust_verify/src/singular.rs
utaal: Great, thanks!

source/rust_verify_test/tests/integer_ring.rs
```diff
@@ -7,9 +7,9 @@ use common::*;
 test_verify_one_file! {
```
utaal, line 7: It would be nice, but not critical if some of the tests checked for unsupported features and how we emit the error message for those (I only see new tests with `assert_one_fails`).

source/rust_verify/src/singular.rs
utaal: 👍

source/air/src/ast.rs
yizhou7: I think that is a good point. Given that there is some discussion of lean or cvc5 integration, and potentially lean's ring tactics or cvc5's finite field solver may supersede Singular, so it is a good practice to put the current code under feature flag. I have made some changes accordingly.

source/air/src/singular_manager.rs
yizhou7: Ah good catch, I think I was debugging some odd IO error, and forgot to turn this back on. Thanks for noticing.

source/rust_verify_test/tests/integer_ring.rs
yizhou7: There are a few tests on unsupported features, e.g. https://github.com/verus-lang/verus/blob/edc5773030dd9b621fd3a17271fff22d33e53a75/source/rust_verify_test/tests/integer_ring.rs#L266
But they are written in a way similar to other expected-verification-failure tests. 
Is there a standard way to test and differentiate unsupported error vs. verification error?

utaal: We generally assert that what we're getting is a `VIR` error message, e.g. in https://github.com/verus-lang/verus/blob/eb988fe4a8600e012822fa0d6b0b5d94a14f0c23/source/rust_verify_test/tests/unions.rs#L77
and that the message matches what we expect.

utaal: You're also welcome to write a variation of the `assert_one_fails` function that accounts for `not supported` and does the appropriate matching.

source/air/src/focus.rs
utaal: I'd like to see this resolved, but I'd be okay with that happening after this is merged (so the branch doesn't drift).

utaal (approved):
Thank you for working through the review items, appreciated!

Note some of the additional comments -- they're-nice-to-haves but not blockers for merging.

Please squash when merging, and provide a descriptive commit message (you can use the PR description and any other major changes that have been made in the meantime). Thanks!

yizhou7:
> Thank you for working through the review items, appreciated!
> 
> Note some of the additional comments -- they're-nice-to-haves but not blockers for merging.
> 
> Please squash when merging, and provide a descriptive commit message (you can use the PR description and any other major changes that have been made in the meantime). Thanks!

Thank you for spending the time reviewing the pull request. Sorry that I was not able to find time to work on this consistently, given it is a rarely used feature. But hopefully this update would make it easier to develop verified cypography with Verus, if there is such a need in the future.

source/rust_verify/example/integer_ring/integer_ring.rs
```diff
@@ -116,6 +117,114 @@ pub proof fn multiple_offsed_mod_gt_0(a: nat, b: nat, c: nat) by(nonlinear_arith
...
+pub proof fn LemmaSmallMod(x: int, m: int) by(nonlinear_arith)
+    requires
+        x >= 0,
+        m > 0,
+        x < m,
+    ensures
+        x % m == x
+{}
+
+// can't use Singular for this proof because of bound
```
tjhance, line 141: there are multiple ensures - which one doesn't work?

yizhou7: > there are multiple ensures - which one doesn't work?

Then encoding for Singular currently does not allow requires/ensures with inequations such as `0 <= x % m`.

PR #734, opened 2023-08-02
add Zulip sign-up link to the readme

utaal:


tjhance:
Should we update the 'Status' section too?

> If you want to try Verus, please be prepared to ask for help by filing GitHub issues.

We could suggest they ask zulip questions instead.

PR #744, opened 2023-08-13
Support for ensures in spec fns

jonhnet:
A common pattern in our Dafny predicates is to build up a definition recursively, then use an ensures to export its meaning under a more-powerful quantifier. For example, with receipts we recursively construct a sequence of values where each is related to the one before it by some predicate; users want to know this relation is true ∀ entry pairs in the receipt.

In Dafny, a single 'ensures' line expressing the fact acts as the proof (because it's the necessary inductive property to complete the proof) as well as broadcasting the result via trigger to anyone who mentions the definition name: exactly the automation we want. The absence of this feature has led to a bunch of tedium in porting Dafny specs into Verus.

Ultimately completing the feature requires:

- The ability to use the ensures as an induction hypothesis (not yet supported in this PR)
- Appropriate interaction with recommends (which we want to just treat as requires, in practice)
- The present PR disallows `return` statements inside spec fns to keep the implementation easy, but that's an unnecessary restriction

### DIFF
--- a/source/air/src/messages.rs
+++ b/source/air/src/messages.rs
@@ -195,6 +195,13 @@ impl MessageX {
         Arc::new(e)
     }
 
+    /// Add a label: if no primary label exists, this will be the primary label; otherwise it will
+    /// be a secondary label.
+    pub fn append_label<S: Into<String>>(&self, span: &Span, label: S) -> Message {
+        if self.spans.len() == 0 { self.primary_label(span, label) }
+        else { self.secondary_label(span, label) }
+    }
+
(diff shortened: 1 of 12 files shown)


jonhnet:
This PR includes tests of stuff that works plus three ignored tests for stuff that doesn't, most conspicuously `test_spec_fn_ensures_induction`. Handing off to Chris for now.

utaal:
[retreat 2025] This will need a different implementation in the new architecture: closing in favor of a feature request: https://github.com/verus-lang/verus/discussions/1712

PR #746, opened 2023-08-15
preserve the type decoration in the type for Tracked<&mut T> params

utaal:
This (I believe) makes the treatment of `Tracked<&mut T>` more in line with the treatment of `Tracked<T>`.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -25,7 +25,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use vir::ast::{
     Fun, FunX, FunctionAttrsX, FunctionKind, FunctionX, KrateX, MaskSpec, Mode, ParamX,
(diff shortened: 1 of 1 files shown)


source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -470,14 +470,21 @@ pub(crate) fn check_item_fn<'tcx>(
...
-            is_ref_mut.map(|(t, _)| t).unwrap_or(input),
-            false,
-        )?;
+        let typ = {
+            let typ = mid_ty_to_vir(
+                ctxt.tcx,
+                &ctxt.verus_items,
+                id,
+                span,
+                is_ref_mut.map(|(t, _)| t).unwrap_or(input),
```
utaal, line 479: This strips the decoration if `is_ref_mut.is_some()`. The proposed change keeps track of it in the return value for `is_mut_ty` it and re-introduces it as a decoration of the non-`&mut` type.

Chris-Hawblitzel (approved):

PR #748, opened 2023-08-15
when expanding errors, if a function does not have a body, do not report it closed

utaal:
addresses https://github.com/verus-lang/verus/discussions/387#discussioncomment-5679193


chanheec (approved):
Thanks a lot for fixing this issue!

PR #749, opened 2023-08-15
fix hygiene of temp_vars (use a Span::mixed_site located at the param span)

utaal:
Re-attempt 177130db, 29c07fd8


utaal:
Merged as https://github.com/verus-lang/verus/commit/96882b698c14315cc39050e0f1ed132b3807ed31

PR #752, opened 2023-08-16
Improvements to multiset, map pervasives

jonhnet:
multiset: removed finite assumption (and axiom!), replaced with comment explaining that multiset can have infinite support, but not infinite multiplicity. Based::new comprehension on a function (consistent with other pervasive types), reworked other definitions to use it, and removed now-unnecessary axioms.

map: moved values from map.rs (should be just _t axioms) to map_lib (should be _v).

### DIFF
--- a/source/pervasive/map.rs
+++ b/source/pervasive/map.rs
@@ -61,12 +61,6 @@ impl<K, V> Map<K, V> {
 
     pub spec fn dom(self) -> Set<K>;
 
-    /// The set of keys mapped to by the domain of the map
- 
-    pub open spec fn values(self) -> Set<V> {
-        Set::<V>::new(|v: V| self.contains_value(v))
-    }
-
     /// Gets the value that the given key `key` maps to.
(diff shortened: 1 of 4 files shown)


jonhnet:
(removing tjhance as I forgot he was on vacation)

utaal (approved):
This is technically a breaking change, ~~but these have existed for only a few days, so I'm okay with landing without annoucements/deprecation~~.

Please squash when merging.

tjhance:
as I said in the slack, the change to multiset is unsound unless `#[verifier::accept_recursive_types(V)]` is changed

utaal:
> as I said in the slack, the change to multiset is unsound unless #[verifier::accept_recursive_types(V)] is changed

Thank you for catching this (and sorry for missing it). It's reverted for now: https://github.com/verus-lang/verus/commit/56edc0169da959a3c74c3f9034cecdcdb9f6987b

PR #759, opened 2023-08-17
handle impl function with different param names than the trait fn declaration

### LINKED ISSUE #278: Panic when trait implementation uses a different argument name
The panic only happens when the argument is mentioned in the `requires` or `ensures`.

```rust
use builtin_macros::*;
mod pervasive;

verus! {

trait T {
    fn f(&self, a: usize) -> (res: usize)
        ensures res == a;
}

struct S { }

impl T for S {
    fn f(&self, b: usize) -> usize {
        b
    }
}

fn main() {
}
}
```

```
fhs m: source 1 ν ./tools/rust-verify.sh /st/verus/test.rs
note: verifying root module

thread 'rustc' panicked at 'internal error: variable not in rename_map: a', vir/src/ast_to_sst.rs:123:21
```

---

* [ ] un-ignore regression test introduced in https://github.com/verus-lang/verus/commit/47caeb412537689ef8eb73bb30248fe064a40e35

utaal:
fixes #278

### DIFF
--- a/source/vir/src/func_to_air.rs
+++ b/source/vir/src/func_to_air.rs
@@ -3,6 +3,7 @@ use crate::ast::{
     TypX, Typs, VirErr,
 };
 use crate::ast_util::QUANT_FORALL;
+use crate::ast_visitor;
 use crate::context::Ctx;
 use crate::def::{
     new_internal_qid, prefix_ensures, prefix_fuel_id, prefix_fuel_nat, prefix_pre_var,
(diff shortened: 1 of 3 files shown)


source/vir/src/func_to_air.rs
```diff
@@ -877,3 +894,31 @@ pub fn func_def_to_air(
...
+
+fn map_expr_rename_vars(
+    e: &Arc<SpannedTyped<crate::ast::ExprX>>,
+    req_ens_e_rename: &HashMap<Arc<String>, Arc<String>>,
+) -> Result<Arc<SpannedTyped<crate::ast::ExprX>>, Arc<air::messages::MessageX>> {
+    ast_visitor::map_expr_visitor_env(
+        e,
+        &mut air::scope_map::ScopeMap::new(),
+        &mut (),
+        &|_state, _, expr| {
```
Chris-Hawblitzel, line 906: Should we worry about variable shadowing like this:
```
        &|_state, map, expr| {
            use crate::ast::ExprX;
            Ok(match &expr.x {
                ExprX::Var(i) if !map.contains_key(i) => {
                    expr.new_x(ExprX::Var(req_ens_e_rename.get(i).unwrap_or(i).clone()))
                }
                ExprX::VarLoc(i) if !map.contains_key(i) => {
                    expr.new_x(ExprX::VarLoc(req_ens_e_rename.get(i).unwrap_or(i).clone()))
                }
                ExprX::VarAt(i, at) if !map.contains_key(i) => {
                    expr.new_x(ExprX::VarAt(req_ens_e_rename.get(i).unwrap_or(i).clone(), *at))
                }
                _ => expr.clone(),
            })
        },
```
or should we just stop worrying and let `ItemLocalId` and `unique_local_name` take care of scoping?

Chris-Hawblitzel (approved):

PR #760, opened 2023-08-19
add a `closed` attribute and require open/closed on impls of trait spec functions

### LINKED ISSUE #382: open/closed should be required for trait spec fns
Right now we require every public spec fn to specify whether it's open or closed. Trait fns are all implicitly public, but we currently aren't doing this check for such functions.

utaal:
fixes #382

### DIFF
--- a/source/builtin_macros/src/is_variant.rs
+++ b/source/builtin_macros/src/is_variant.rs
@@ -15,7 +15,7 @@ pub fn attribute_is_variant(
         quote! {}
     } else {
         quote! {
-            #[verus::internal(publish)]
+            #[verus::internal(open)]
         }
     };
     let is_impls = s
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -272,8 +272,8 @@ impl Visitor {
 
         let publish_attrs = match &sig.publish {
             Publish::Default => vec![],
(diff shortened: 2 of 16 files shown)


Chris-Hawblitzel:
Looks good.  It looks like spec functions of `pub(crate)` trait impls are currently accepted without `open`/`closed`, though:

```
pub(crate) trait T {
    spec fn f() -> int;
}
pub(crate) struct S;
impl T for S {
    spec fn f() -> int { 5 }
}
```

utaal:
> Looks good. It looks like spec functions of `pub(crate)` trait impls are currently accepted without `open`/`closed`, though:
> 
> ```
> pub(crate) trait T {
>     spec fn f() -> int;
> }
> pub(crate) struct S;
> impl T for S {
>     spec fn f() -> int { 5 }
> }
> ```

That is true, but that's only the case for traits defined at the crate root, whose visibility is the same as the crate (i.e. the same as if `T` was declared `trait T`, without a visibility modifier). Inside modules:
```
    mod m1 {
        use vstd::prelude::*;

        pub(crate) trait T {
            spec fn f() -> int;
        }
        pub(crate) struct S;
        impl T for S {
            spec fn f() -> int { 5 }
        }
    }
```
`open/closed` is required. I added a test to that effect.

Chris-Hawblitzel (approved):

PR #763, opened 2023-08-23
Add node_is_in_cycle to scc Graph

Chris-Hawblitzel:
We were using the SCC graph to detect mutual recursion but not single recursion.  This pull request uses the SCC graph for both mutual recursion and single recursion, which is simpler and faster than the previous approach.  I tested on rust_verify_test with the following assertions and it computes the same results as the previous approach:

```
pub(crate) fn is_recursive_exp(ctx: &Ctx, name: &Fun, body: &Exp) -> bool {
let r = {
    if ctx.global.func_call_graph.get_scc_size(&Node::Fun(name.clone())) > 1 {
        // This function is part of a mutually recursive component
        true
    } else {
        let mut scope_map = ScopeMap::new();
        // Check for self-recursion, which SCC computation does not account for
        match exp_visitor_dfs(body, &mut scope_map, &mut |exp, _scope_map| match &exp.x {
            ExpX::Call(CallFun::Fun(x, resolved_method), _, _)
                if is_self_call(ctx, x, resolved_method, name) =>
            {
                VisitorControlFlow::Stop(())
            }
            _ => VisitorControlFlow::Recurse,
        }) {
            VisitorControlFlow::Stop(()) => true,
            _ => false,
        }
    }
};
assert!(r == ctx.global.func_call_graph.node_is_in_cycle(&Node::Fun(name.clone())));
r
}

pub(crate) fn is_recursive_stm(ctx: &Ctx, name: &Fun, body: &Stm) -> bool {
let r = {
    if ctx.global.func_call_graph.get_scc_size(&Node::Fun(name.clone())) > 1 {
        // This function is part of a mutually recursive component
        true
    } else {
        // Check for self-recursion, which SCC computation does not account for
        match stm_visitor_dfs(body, &mut |stm| match &stm.x {
            StmX::Call { fun, resolved_method, .. }
                if is_self_call(ctx, fun, resolved_method, name) =>
            {
                VisitorControlFlow::Stop(())
            }
            _ => VisitorControlFlow::Recurse,
        }) {
            VisitorControlFlow::Stop(()) => true,
            _ => false,
        }
    }
};
assert!(r == ctx.global.func_call_graph.node_is_in_cycle(&Node::Fun(name.clone())));
r
}
```

### DIFF
--- a/source/vir/src/func_to_air.rs
+++ b/source/vir/src/func_to_air.rs
@@ -375,7 +375,7 @@ pub fn req_ens_to_air(
 /// if the function is a spec function.
 pub fn func_name_to_air(
     ctx: &Ctx,
-    diagnostics: &impl Diagnostics,
+    _diagnostics: &impl Diagnostics,
     function: &Function,
 ) -> Result<Commands, VirErr> {
     let mut commands: Vec<Command> = Vec::new();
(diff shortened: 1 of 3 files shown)


parno (approved):
Looks like a nice clean up to me.  If we're likely to call `node_has_direct_edge_to_itself` a lot, we could cache the result, but I think most nodes will only have a few edges, so it may not give much (if any) win.

PR #766, opened 2023-08-23
Use Euclidean div/mod from num-bigint

jaybosamiya:
Now that the [PR](https://github.com/rust-num/num-bigint/pull/245) I had opened quite a while ago to `num-bigint` has been merged, we get to use these two functions from the library instead.


parno (approved):
Looks great -- congrats on landing the change in `num-bigint`!

PR #772, opened 2023-08-25
vstd: Add no-std build mode

zhaofengli:
This PR adds a `#![no_std]` build mode for vstd. It can be activated by passing `--vstd-no-std` and optionally `--vstd-no-alloc` to vargo.

### DIFF
--- a/source/pervasive/bytes.rs
+++ b/source/pervasive/bytes.rs
@@ -2,8 +2,6 @@
 
 #![allow(unused_imports)]
 
-extern crate alloc;
-
 use builtin::*;
 use builtin_macros::*;
 
--- a/source/pervasive/bytes.rs
+++ b/source/pervasive/bytes.rs
@@ -2,6 +2,8 @@
 
 #![allow(unused_imports)]
 
+extern crate alloc;
+
 use builtin::*;
 use builtin_macros::*;
 
@@ -85,7 +87,7 @@ pub exec fn u16_from_le_bytes(s: &[u8]) -> (x:u16)
 
 #[cfg(feature = "alloc")]
 #[verifier(external_body)]
-pub exec fn u16_to_le_bytes(x: u16) -> (s: Vec<u8>)
+pub exec fn u16_to_le_bytes(x: u16) -> (s: alloc::vec::Vec<u8>)
   ensures 
     s@ == spec_u16_to_le_bytes(x),
     s@.len() == 2,
@@ -177,7 +179,7 @@ pub exec fn u32_from_le_bytes(s: &[u8]) -> (x:u32)
 
 #[cfg(feature = "alloc")]
 #[verifier(external_body)]
-pub exec fn u32_to_le_bytes(x: u32) -> (s: Vec<u8>)
+pub exec fn u32_to_le_bytes(x: u32) -> (s: alloc::vec::Vec<u8>)
   ensures 
     s@ == spec_u32_to_le_bytes(x),
     s@.len() == 4,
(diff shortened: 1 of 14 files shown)


utaal:
Hi @zhaofengli, and thank you for the PR! I'll try and get to it as soon as I can.

utaal:
Thank you for rebasing, and sorry for the delay!

utaal:
One question: is it correct that with this PR there isn't a way when building to set `alloc` or `std` individually? If so, why are they two separate features? As something we can use in the future?

zhaofengli:
> One question: is it correct that with this PR there isn't a way when building to set alloc or std individually? If so, why are they two separate features? As something we can use in the future?

I just moved disabling alloc to a separate flag in vargo. Now both `--vstd-no-std` and `--vstd-no-alloc` are required for a no-std build without a global allocator available.

source/pervasive/bytes.rs
```diff
@@ -215,17 +217,16 @@ pub closed spec fn spec_u64_from_le_bytes(s: Seq<u8>) -> u64
 pub proof fn lemma_auto_spec_u64_to_from_le_bytes()
   ensures
     forall |x: u64|
-      #![trigger spec_u64_to_le_bytes(x)]
     {
-      &&& spec_u64_to_le_bytes(x).len() == 8
+      &&& (#[trigger] spec_u64_to_le_bytes(x)).len() == 8
```
utaal, line 221: Sorry for not asking earlier: why do you need these additional `#[trigger]` annotations? It doesn't look like these are affected by the exec functions being excluded if the flag is absent.

utaal:
Thank you! Looks good. Just one last check on the `#[trigger]` annotations (in a comment).

source/pervasive/bytes.rs
zhaofengli: I'm not sure, but without changing the `#[trigger]` in the assertion below (the change on this line is cosmetic) somehow the post conditions cannot be proven. Perhaps it's because of some `broadcast_forall` axioms that aren't enabled when `alloc` is disabled.

utaal: It seems the issue is instability (due to the change in other proofs in the same module), not the triggers. Marking the failing proofs `#[verifier::broadcast_forall]` seems to stabilize them.

utaal:
There seem to be a couple remaining fixes needed:
* making sure that `vstd` is re-built when the `--vstd-no-*` flags change;
* fixes the build for `--vstd-no-std` without `--vstd-no-alloc`;
* adding `#[verifier::spinoff_prover]` as explained in the comment.

I made these fixes and opened a PR against this branch: https://github.com/mars-research/verus/pull/1
I cannot otherwise make edits on this branch. If they look good to you, please merge them in, so that this PR gets updated.

zhaofengli:
Thanks for the fixes! I've merged the commits.

utaal:
@tjhance you added some of the `extern crate alloc;` imports, and made changes to use `alloc::*` types when necessary before. Can you take a look and check that my changes in https://github.com/verus-lang/verus/pull/772/commits/bb7bf7af03b54da99c8c6b347fa458fcde9f9782 seem reasonable?

utaal (approved):
Looks good to me. I'd like for @tjhance to have a look too, in particular at the changes that involve using the `alloc` crate.

source/pervasive/std_specs/vec.rs
```diff
@@ -222,6 +222,7 @@ pub fn ex_vec_as_slice<T, A: Allocator>(vec: &Vec<T, A>) -> (slice: &[T])
     vec.as_slice()
 }
 
+#[cfg(feature = "std")]
```
tjhance, line 225: why's this needed? is split_off not available with just `alloc`?

tjhance (approved):
looks fine to me

source/pervasive/std_specs/vec.rs
zhaofengli: Good catch! I went over vstd and did the following:

- Removed the gate in question, changed `std::clone::Clone` to `core::clone::Clone`
- Removed redundant `extern crate alloc`
- Made the gate on `string` more fine-grained (string slices can be used without `alloc`)

(Sorry for the late reply, was away yesterday)

utaal:
Merged as https://github.com/verus-lang/verus/commit/d2fd64204bf720bbc23a3917fb151b273e05cc6b

PR #773, opened 2023-08-25
Fix warning when building Verus

jaylorch:


tjhance:
This isn't a function we wrote, and I'd prefer to not modify any third-party code more than we have to. If the warning is annoying, should we just disable warnings instead?

tjhance (approved):
eh, actually, it probably doesn't matter very much.

PR #775, opened 2023-08-27
Move the verus-specific error message definitions out of air, and clean-up error reporting utilities in vir

utaal:
* parametrizes AIR structs by a `M: Message` type;
* moves the definition of Verus' Message to VIR;
* defines a `air_ast` module in VIR which contains type aliases for the AIR nodes with `M` set to Verus' `Message`;
* replaces `msg_error` and `error` in VIR with a single `error(span, note)` to make the usage consistent where errors are constructed with builder methods on `Message`; reorder arguments for `note`, `warning` to match the `(span, note)` order

This is (ultimately) to more cleanly support adding additional fields to `Message` that are only relevant to Verus/VIR, e.g. to avoid string matching here: https://github.com/verus-lang/verus/blob/main/source/vir/src/split_expression.rs#L474-L494


tjhance:
With spans, we were able to avoid adding a zillion template parameters by using `dyn std::any::Any`, and it seems to have worked well. Should we consider something similar here?

utaal:
The type parameters only need to be added everywhere in `air`, but can be hidden in upper layers with a module that defines type aliases (https://github.com/verus-lang/verus/pull/775/files#diff-f6fd9491c44d91e51fdd0fdb5c067660b5c307d6d0c6bf53034ab9642dac84af), so I think the syntactic noise is not too bad.
There's also a difference with `Span`: we never needed `air` to look into/construct `Span`s, but `air` needs to manipulate `Message`s (in the ways that the `trait Message` supports), including constructing new concrete `Message`s. I think it'd still be possible to use `dyn` for this (maybe `dyn Message` rather than `dyn Any`), but I'm not sure it would be cleaner.

Chris-Hawblitzel:
Can we try `dyn Message`, then?  I had the same reaction as @tjhance .  I like the new trait `Message`, and I think moving the definition of `Span` out of `air` is an improvement, but I think the `<M>` adds a lot of clutter to `air/src/ast.rs`.  I wish Rust made it easier to parameterize whole modules or crates over type variables, rather than having to parameterize every type and function.

utaal:
> Can we try `dyn Message`, then?

I have tried and it's maybe doable but quite messy: `air`'s consumer (`vir`) needs to know the concrete `Message` type to report the error diagnostic:

```rust
                ValidityResult::Invalid(air_model, error) => {
                ... 
                        reporter.report_as(&error, level);
```

This `error` comes back from `air`, so if it was a `dyn Message` we'd need to down-cast it to the concrete type here, but we cannot unless it's a `dyn Any`. But if it's `dyn Any` it cannot also be `dyn Message`, and AFAIK one cannot down-cast a `dyn Any` to a trait (i.e. to a non-concrete type).

Additionally, in `air`, this branch uses static functions defined on the `Message` trait to construct certain messages directly. That is not possible with a `dyn Message` object (where you need a `&self` for dispatch), so we'd need to thread through an additional `dyn MessageBuilder` for these calls.

While I agree that the clutter in `air` is unfortunate, I think a regular trait with type arguments is the right form of polymorphism here.

utaal:
It seems to have worked. The trick was to move the `Message` construction/manipulation functionality that `air` needs into a separate `dyn Trait` (`MessageInterface`). Implementing `Serialize` and `Deserialize` would have been tricky, but fortunately we don't actually ever need to serialize and deserialize messages.

source/air/src/ast.rs
```diff
@@ -156,16 +144,16 @@ pub enum ExprX {
     Bind(Bind, Expr),
     // Sometimes an axiom will have additional error messages. If an assert fails
     // and this axiom was relevant, then we append the error labels to the Message.
-    LabeledAxiom(MessageLabels, Expr),
-    LabeledAssertion(Message, Expr),
+    LabeledAxiom(Vec<Arc<dyn Any + Send + Sync>>, Expr),
```
Chris-Hawblitzel, line 147: Can we make a type abbreviation or a struct wrapper for `dyn Any + Send + Sync`?

source/air/src/context.rs
```diff
@@ -59,7 +60,8 @@ impl<'a, 'b: 'a> Default for QueryContext<'a, 'b> {
     }
 }
 
-pub struct Context {
+pub struct Context<'a> {
```
Chris-Hawblitzel, line 63: Could we use `Arc<dyn MessageInterface>` or `Rc<dyn MessageInterface>` to avoid the `<'a>` everywhere?

source/air/src/typecheck.rs
```diff
@@ -89,6 +89,7 @@ fn check_typs(typing: &Typing, typs: &[Typ]) -> Result<(), TypeError> {
 }
 
 fn check_exprs(
```
Chris-Hawblitzel, line 91: Could we add a copy of `message_interface` to `Typing` to avoid passing an extra argument?

source/vir/src/messages.rs
```diff
@@ -0,0 +1,329 @@
...
+                    .cloned()
+                    .collect(),
+                help: None,
+            })
+        }
+    }
+}
+
+// /// Construct an Error and wrap it in Err.
+// /// For more complex Error objects, use the builder functions in air::errors
```
Chris-Hawblitzel, line 187: These seem to be leftover comments.

Chris-Hawblitzel (approved):

PR #780, opened 2023-08-28
Remove dead code from seq.rs

Chris-Hawblitzel:
This was an unused, non-public function.  Even if it were used, it would only be a one-liner:

```
spec fn fill(len: nat, val: A) -> Seq<A> { Self::new(len, |i: int| val) }
```


tjhance:
This seems like a useful function. Should we just mark it public, instead?

parno (approved):
I'm okay with deleting it, given that no other lemmas seem to talk about it.  It's only mildly shorter than directly calling `Seq::new`, so it seems like there's value in using `Seq::new` consistently, both from a style and from a trigger perspective.

PR #781, opened 2023-08-28
User filter cleanup (--verify-function, --verify-module)

tjhance:
Factors out code related to --verify-function and --verify-module into a new file.

(Rationale: There was some consensus it would be a good idea to prune more aggressively for functions marked spinoff_prover. However, much of the code in verifier.rs is tied tightly to module-based grouping, which makes it difficult to prune along boundaries not defined by modules. Making this change alleviates this problem somewhat.)


utaal (approved):
Nice!

PR #785, opened 2023-08-29
use impl_paths from statically resolved function target, fixes #784

### LINKED ISSUE #784: missing ImplPaths edges
Verus fails to detect the following cyclic definitions.

The line (2) should have an edge to both impls (`impl&%0` and `impl&%1`). However, it only has an edge to `impl&%1`.

```rust
#![allow(unused_imports)]

use builtin_macros::*;
use builtin::*;
use vstd::*;

verus!{

trait Tr {
    spec fn stuff() -> bool;
}

struct X { }

// impl&%0
impl Tr for X
{
    spec fn stuff() -> bool {
        alpaca()                                // (1)
    }
}

spec fn alpaca() -> bool {
    // depends on the bound `X: Tr`
    // which depends on the above trait impl
    // which in turn depends on `alpaca`
    (P::<X> { t: None }).orange()               // (2)
}

struct P<T: Tr> {
    t: Option<T>,
}

trait Zr {
    spec fn orange(&self) -> bool;
}

// impl&%1
impl<T: Tr> Zr for P<T> {
    spec fn orange(&self) -> bool {
        !T::stuff()                             // (3)
    }
}

proof fn paradox() {
    assert(alpaca() == !alpaca());
    assert(false);
}

fn main() { }

}
```

tjhance:


tjhance:
I think this isn't the right fix. It works in this case, but it doesn't solve the fundamental problem that the impl_path list of the non-statically-resolved function was imcomplete.

PR #792, opened 2023-08-31
allow DSTs in ghost code (prototype)

tjhance:
## Context

Since we use Rust's typechecker, we have to obey its requirements regarding `Sized`. However, these requirements stem from the need to compile the code, so in principle, it should be fine to remove such requirements in ghost code.

Currently, ghost code _does_ invoke the usual rustc Sized checks, after which the Sized checks are then ignored in VIR. The result is a bit of a mismatch between the source code and the underlying logic. More precisely, there are a number of problems:

 * If the user wants to make their own pointer type, e.g., `MyRc<V>` with `spec fn view(&self) -> V`, it will be impossible to instantiate the pointer type with a DST.
 * Since `Sized` is used everywhere in ghost code, even where it's not necessary, it would be wasteful to encode the `Sized` trait bound - and therefore, currently, we don't. However, this means that we can't write any proof lemmas or axioms that rely on a `Sized` bound (like `V: Sized` ==> `size_of<&V> == word_size`). The currently accepted workaround proposes that we create a _new_ concept to be mean "type is sized" that _is_ encoded. However, the result is confusing to work with.
 * You have to add `Sized` bounds for trait params, which isn't terrible but is slightly annoying when the trait is only for spec purposes.

There's also some more motivation in this discussion by @jaybosamiya: https://github.com/verus-lang/verus/discussions/377

**This PR allows DSTs in ghost code.** I propose we make ?Sized the default (e.g., through the macro) in ghost code.

## Implementation

Unfortunately, even with all the unstable features currently available, there's no way to turn off all the Sized requirements (.e.g, that return values cannot be Sized). I considered a number of hacks that involved query-overriding, but they didn't pan out.

Soooo ... I'm proposing to fork the rustc_hir_typeck crate, as done in this PR. I mean, we've already forked 2 crates, what's 1 more?

Due to rustc's query framework, it's pretty easy to inject the modified verus_rustc_hir_typeck. The only modification I have to make to the crate is deleting the lines of code related to Sized checks. We don't even have to distinguish ghost from exec code here, because we'll end up using the unmodified typechecker when compiling the code anyway.

<sub>(also this would let us fix that annoying `as nat` error message)</sub>

### Todo

- [ ] make ?Sized the default in ghost code (by adding ?Sized bounds with the macro)
- [ ] ensure the providers are injected for the non-erased pass and the lifetime pass, but not for the erased executable pass
- [ ] tests


utaal:
At a high level, I'm very reluctant to fork `rust_hir_typeck`: we would become even more dependent on its internals, and on the way the relevant queries are defined (which may change); it would also mean that updating to newer rustc's requires a significant rebase (where the diff we need may be significantly different as a result of refactors in `rust_hir_typeck`).

> we've already forked 2 crates, 

These are small crates, and it was in part a way to speed up the move to unmodified rustc: ideally I'd prefer we found a way to remove the need for these forks too, rather than forking more.

utaal-b:
@tjhance are you okay with me closing this for now?

PR #795, opened 2023-09-01
Add `#![all_triggers]` attribute to output all minimal cover triggers to Z3

GenericMonkey:
**Changes**
- added new `#![all_triggers]` attribute for use in quantifiers, and fixed subsequent plumbing
- Extended `compute_triggers` in `triggers_auto` to -- on a boolean flag `all_triggers` -- compute either 
   1. the original best trigger as determined by `Score`
   2. simply return a full list of all minimal triggers
- Renamed `best_so_far` to `computed_triggers` for clarity

**Purpose**
While experimenting recently with how much automation we can get from verus, we found ourselves manually computing this set and explicitly adding all triggers. The feature gives us the ability to tag it more conveniently for these experiments.

**Testing**
Added a new test case to `rust_verify_test` that illustrates this change. the default `auto` will choose `res(i,j)` as the trigger, so the empty proof will not go through unless `res(3,4)` is asserted. With the `all_triggers` attribute, we get an empty proof.

### DIFF
--- a/source/vir/src/triggers_auto.rs
+++ b/source/vir/src/triggers_auto.rs
@@ -29,6 +29,13 @@ and then programmers can use manual triggers to make the quantifiers more libera
 rather than the defaults causing timeouts,
 and programmers having to use manual triggers to eliminate the timeouts.
 */
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum AutoType {
+    Regular,
+    All,
+    None
+}
+
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 enum App {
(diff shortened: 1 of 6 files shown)


source/vir/src/triggers_auto.rs
```diff
@@ -514,34 +522,72 @@ fn trigger_score(ctxt: &Ctxt, trigger: &Trigger) -> Score {
 
 // Find the best trigger that covers all the trigger variables.
 // This is a variant of minimum-set-cover, which is NP-complete.
```
utaal, line 524: Can you please update this docstring given the new functionality? (It also needs three slashes (`///`) to conform with the rustdoc convention.)

utaal (approved):
Looks good to me, but please wait for @Chris-Hawblitzel review. (And fix that comment I commented on.)

Chris-Hawblitzel (approved):

source/vir/src/triggers_auto.rs
GenericMonkey: Done

PR #611, opened 2023-06-15
Use a separate AST node, and generalised traits to support array, slice

utaal:
Replacement for https://github.com/verus-lang/verus/pull/398 based on current `main`

PR #803, opened 2023-09-06
Conversion between u16/u32/u128 and le bytes

jaylorch:


jaybosamiya (approved):
Overall, this looks good to me.

Just a few notes, regarding the `exec` parts of this PR, connecting this PR up to other PRs and features in Verus:
1. When #453 introduced u64 <-> bytes, we didn't yet have support for Rust fixed-size arrays, and thus could not support type signatures that looked like [from Rust's standard library](https://doc.rust-lang.org/std/primitive.u64.html#method.from_le_bytes); with #611 merged though, it appears that we might be able to move to using Rust fixed-size arrays, and eliminate the unnecessary allocation that occurs.
2. Now that we have support for adding specifications to Rust `std` (with #550 and others), it may even be possible for us to add these as specifications to the functions in the plain Rust types.

Given that both of the following points are true, it appears that we won't actually need to reuse these names when we move to the allocation-free fixed-array ones, which is good news, since it will allow for a smooth migration if/when we deprecate these.  In any case, any deprecation needed in the future pertains solely to the `exec` parts of the code being added; all the `spec` and `proof` parts would remain the same anyways, so nothing would need to be changed when we do that later on.

Thus overall, I am in favor of merging this as is, and when we eventually add the specifications to `std` for the equivalent functions, we can deprecate these additions in support of the newer ones that add specs to `u16::from_le_bytes` and such.

TLDR: Looks good to me, thanks! :)

PR #805, opened 2023-09-07
Support vec![value; length]

jaylorch:


tjhance:
Unfortunately I don't think this is sound. The requirement is `T: Clone`, but there is no guarantee that `elem.clone()` returns an element that is == to `elem`.

jaylorch:
Ah, interesting!

jaylorch:
Is there a way to restrict it to types that are primitive integers?

Issue #408, opened 2023-03-01
Missing recommends checks for `choose`, is-variant/get-variant functions


utaal:
[Edited, as the initial report was inaccurate.]

Something like doesn't verify in a specific context.

```rust
proof fn choose_contains_set(m: Set<nat>)
    requires m.finite(), m.len() > 0
{
    let c = m.choose();
    assert(m.contains(c)); // FAILS
}
```

---

* [ ] minimize the failing case

tjhance:
Is this a bug? I think it needs the requirement that m be nonempty

achreto:
Yes, the following does verify for me:

```rust
proof fn choose_contains_set2(m: Set<nat>)
        requires m.finite() && m.len() > 0
{
    let c = m.choose();
    assert(m.contains(c)); // Succeeds
}
```

@utaal I think the issue we encountered was with the interaction of map and its domain when removing an entry. I tried to extract a minimum example, but somehow the proof goes through in that case. Maybe something else is going on?

utaal:
Oh you're right.

tjhance:
perhaps `choose` should have a corresponding `recommends` check?

HaoYang670:
The verus-analyzer now fails to parse the `choose` function (https://github.com/verus-lang/verus-analyzer/issues/20). Are these two issues related?

Marti2203:
I guess this issue can be closed?

PR #807, opened 2023-09-08
Refactor and generalize recommends checking

Chris-Hawblitzel:
Refactor and generalize recommends checking, making it more comprehensive and potentially preparing to use the same mechanisms for other purposes, like decreases checking or requires checking.

The main refactorings are:
- split `disable_recommends` into `disable_recommends` and `only_generate_pure_exp`, where the latter can, in the future, handle things that aren't recommends checking, such as decreases checking or requires checking
- rename various functions as "_skip_checks" and avoid these "_skip_checks" functions by default, so that the default behavior in the AST->SST translation is to check recommends (and potentially other things, like decreases checking or requires checking), and any other behavior is clearly documented
 
This also adds recommends for choose ( https://github.com/verus-lang/verus/issues/408, https://github.com/verus-lang/verus/discussions/802 ).


tjhance (approved):
great! I think a lot of stuff in ast_to_sst is clearer now.

utaal (approved):
Awesome.

Issue #789, opened 2023-08-29
bad (latent-unsound) ordering of AIR commands for termination checking


tjhance:
In the AIR that results from this code, the function-axiom for `impossible_fun` (which is roughly, `impossible_fun() == !impossible_fun()`), appears before `bad_lemma()`. This could potentially help `bad_lemma()` prove false. (However, because of triggers, this does not seem to happen. But this isn't something we want to rely on.)

```rust
spec fn impossible_fun() -> bool
    decreases 0int
      via f_decreases
{
    !impossible_fun()
}

#[verifier::decreases_by]
proof fn f_decreases() {
    bad_lemma();
}

proof fn bad_lemma()
    ensures false,
{
    assert(false);
}
```

tjhance:
This was fixed in 84aca0c39d75bcbce47d1d3a494025913b0fecfe

PR #816, opened 2023-09-15
refactor the command handling and ordering in verifier.rs

tjhance:
Collecting AIR commands and spitting them out in the right order is something that's likely going to change a lot soon. There's issue #789, similar problems related to #565, upcoming broadcast_forall with proof bodies, upcoming requires on spec functions.

Currently, verifier.rs generates AIR in two phases; first it handles spec functions (mostly generating axioms) and then it handles proof and exec functions (mostly generating proof obligations). The proof obligations require substantially richer logic, due to prover options and other flags. But as issue #789 shows, we can't keep it collected into two phases. In fact, the two-phase approach is already causing problems, because 'spec termination' proof obligations are handled in the first phase, and thus they miss out on the richer option-handling of the second phase. (For example, it's currently not possible to use assert-by-bitvector in a decreases-by lemma.)

This PR doesn't (intentionally) change any behavior or fix any issues, but it does completely rewrite these 2 main loops, so I think they'll be a lot more approachable. All the logic that decides what AIR to generate and in what order, now goes into a 'generator' object that can returns 'Ops'. Each 'op' is either a set of axioms or a set of queries. The main loop in verifier.rs simply pops off the queue, handles appropriately, and repeats. To handle expand-errors or recommends-checking, verifier.rs can push new tasks into the queue so that they'll be handled in the next iteration.

### DIFF
--- a/source/rust_verify/src/commands.rs
+++ b/source/rust_verify/src/commands.rs
@@ -2,6 +2,7 @@ use crate::buckets::Bucket;
 use air::ast::Command;
 use air::messages::Diagnostics;
 use std::collections::HashMap;
+use std::collections::VecDeque;
 use std::sync::Arc;
 use vir::ast::Visibility;
 use vir::ast::{Fun, Function, Krate, Mode, VirErr};
@@ -13,7 +14,6 @@ use vir::func_to_air::SstMap;
 use vir::messages::Message;
 use vir::recursion::Node;
 use vir::update_cell::UpdateCell;
-use std::collections::VecDeque;
 
 #[derive(Clone, Copy, Debug)]
 pub enum ContextOp {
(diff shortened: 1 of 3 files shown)


source/rust_verify/src/commands.rs
```diff
@@ -0,0 +1,414 @@
...
+        std::mem::swap(&mut new_sst_map, &mut self.sst_map);
+
+        self.ctx.fun = None;
+        self.ctx.expand_flag = false;
+
+        Ok(vec![Op::query(QueryOp::Body(style), commands, &function)])
+    }
+}
+
+fn mk_fun_ctx(f: &Function, checking_recommends: bool) -> Option<FunctionCtx> {
```
Chris-Hawblitzel, line 347: The old mk_fun_ctx is still in verifier.rs.  Can we delete the old one?

source/rust_verify/src/commands.rs
```diff
@@ -0,0 +1,414 @@
...
+    pub ctx: &'a mut vir::context::Ctx,
+    krate: Krate,
+    bucket: Bucket,
+    reporter: &'a D,
+
+    sst_map: SstMap,
+    func_map: HashMap<Fun, (Function, Visibility)>,
+
+    current_queue: Vec<Op>,
+    queue_idx: usize,
```
Chris-Hawblitzel, line 70: Would it make sense to use `VecDeque` instead of `Vec` + `queue_idx`?

Chris-Hawblitzel (approved):
At some point (before we generalize broadcast_forall), we might also want to handle `traits_to_air` with the queue, so that each trait bound axiom is emitted only after the trait obligations have been satisfied.

PR #818, opened 2023-09-15
max should return max, not min

jonhnet:


parno:
You're a purist huh?

utaal (approved):
WOOOOO!

PR #820, opened 2023-09-15
include math in prelude

jonhnet:
math::max is a nice thing to be able to type without another special `use` line.

### DIFF
--- a/source/pervasive/prelude.rs
+++ b/source/pervasive/prelude.rs
@@ -2,6 +2,7 @@ pub use builtin::*;
 pub use builtin_macros::*;
 
 pub use super::view::*;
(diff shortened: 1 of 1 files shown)


utaal (commented):
I'd rather we didn't do this, it would be the only case where we reexport a module name (rather than all/certain members).

If you'd like this in your project, you can make a custom `prelude` by making a `prelude` module:
```rust
// src/prelude.rs
pub use vstd::prelude::*;
pub use vstd::math;
```
```rust
// src/lib.rs
pub mod prelude;
```
and import `use crate::prelude::*;` instead of `use vstd::prelude::*;`.

PR #832, opened 2023-09-25
initial support for array literals

tjhance:
To do the encoding, we need to create expressions involving `array.view().index(i)`. This means more hardcoded vstd paths, but there doesn't really seem to be a way around it.

This encoding uses AIR 'assume' statements, so it doesn't work in spec functions or other places that require pure expressions. It would be nice to support those use cases, since then we could use array literals to implement `seq!`, which is something we've talked about & which has been on the backlog for a while. It would probably be best to get chris's opinion on that, though, so I'm not handling that in this PR.

### DIFF
--- a/source/pervasive/array.rs
+++ b/source/pervasive/array.rs
@@ -49,5 +49,18 @@ pub proof fn array_len_matches_n<T, const N: usize>(ar: &[T; N])
 {
 }
 
+// Referenced by Verus' internal encoding for array literals
+#[doc(hidden)]
+#[cfg_attr(verus_keep_ghost, rustc_diagnostic_item = "vstd::array::array_index")]
+pub open spec fn array_index<T, const N: usize>(ar: &[T; N], i: int) -> T {
+    ar.view().index(i)
+}
+
+// Referenced by Verus' internal encoding for array literals
+#[doc(hidden)]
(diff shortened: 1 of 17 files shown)


utaal:
Could we do something similar to the encoding for string literals, for consistency and to avoid the hardcoded paths? I.e. 

* VIR nodes for primitives, like https://github.com/verus-lang/verus/blob/main/source/vir/src/ast.rs#L376-L377

and
* axioms to connect the encoding and the view, like https://github.com/verus-lang/verus/blob/main/source/pervasive/string.rs#L164-L183

utaal:
I have looked into it and it is tricker than I expected, because of the polymorphism. Given that, and the fact that we don't have a better mechanism to handle these internal calls right now, I'm happy with it.

utaal:
I've rebased with the intent to merge, but found a panic while writing a couple more tests:

The following panics:
```rust
        exec fn test() {
            let ghost a = [3u64, 4, 5];
        }
```

with `generated ill-typed AIR code: error 'use of undeclared function array.array_len.?' in expression '(array.array_len.? $ (UINT 64) $ (CONST_INT 3) tmp%1@)`.

Can you take a look? (Shall we disallow this?)

utaal (approved):

PR #843, opened 2023-10-07
Support exec-mode const declarations

Chris-Hawblitzel:
In Verus, const declarations are a lot like function declarations.  In fact, at the VIR level, const declarations are 0-parameter function declarations, and references to constants are desugared into 0-argument function calls.  Functions can have mode `exec`, `proof`, or `spec`.  Constants were intended to support modes `exec`, `proof`, and `spec` as well.  However, for simple constants like integers, it's more useful for a constant to have dual `exec/spec` mode, so that, for example, a constant like `MAX_U8 = 255` can be used in both specifications and executable code.  For this reason, the default mode for a const declaration has always been dual `exec/spec`.  Unfortunately, this dual mode overshadowed some of the other modes (`exec` and `proof`), which were never implemented for const declarations on their own.  Right now, the only modes implemented for const declarations are `spec`, dual `exec/spec`, and the rather useless dual `proof/spec`.  Here's a summary of the current status of modes supported by function declarations and const declarations:

| syntax | meaning for fn | meaning for const |
|-|-|-|
| (default) | exec | dual exec/spec |
| exec | exec | dual exec/spec |
| proof | proof | dual proof/spec |
| spec | spec | spec |

This was not intentional; rather, the exec and proof modes were never implemented, and they ended up meaning dual `exec/spec` and dual `proof/spec` by accident.

This pull request adds `exec`-mode const declarations (also `proof`-mode const declarations, although these might not be useful).  It changes the meaning of `exec const` and `proof const` to mean `exec` mode and `proof` mode, as one would expect:

| syntax | meaning for fn | meaning for const |
|-|-|-|
| (default) | exec | dual exec/spec |
| exec | exec | exec |
| proof | proof | proof |
| spec | spec | spec |

There's still no dedicated syntax for dual `exec/spec`; it's just the default for const.  This could be added later if we want.  There's also still no support for dual `exec/spec` for `fn` functions.  This could also be added later, although `when_used_as_spec` already provides much of the functionality one would want from dual `exec/spec` functions, so it might not be necessary.

This pull request would have been just a few lines of changes in `rust_to_vir_func.rs` if it only needed the changes to meaning shown in the tables above.  However, since `exec` and `proof` modes weren't previously implemented for const declarations, there was a large missing feature that had to be implemented, namely ensures clauses for `exec` and `proof` const declarations.  Without an ensures clause, it's not possible to prove anything useful about the "body" of an `exec` const declaration, just like with `exec` functions.  Adding ensures clauses to const declarations required adding new syntax.  The most direct syntax, with an ensures in front of the `=` would be too ambiguous and hard to read:

```
exec const C: int ensures C == 7 = 7;
```

We could put the ensures at the end, but this would be inconsistent with the function syntax and could become strange if the "body" expression of the const declaration was large enough to span multiple lines:

```
exec const C: int = 7 ensures C == 7;
```

In the end, I thought using a more function-like syntax would be clearest in the case were the const declaration has an ensures clause, so this is what I implemented for the pull request:

```
spec const B: int = 7;
exec const C: int ensures C == 7 { 7 }
```

Here's an example with an `exec` const:

```
        spec fn f() -> int { 1 }
        const C: u64 = 3 + 5;
        spec const S: int = C as int + f();
        const fn e() -> (u: u64) ensures u == 1 { 1 }
        exec const E: u64 ensures E == 2 { 1 + e() }

        fn test1() {
            let x = C;
            assert(x == 8);
            assert(S == 9);
            let y = E;
            assert(y == 2);
        }
```

Here's an example combining an `exec` const with `when_used_as_spec` (note that this example can still be done more easily with dual `exec/spec` mode; this example is just for illustration):

```
        spec const SPEC_E: u64 = 7;

        #[verifier::when_used_as_spec(SPEC_E)]
        exec const E: u64 ensures E == SPEC_E { 7 }

        fn test1() {
            let y = E;
            assert(y == E);
        }
```

### DIFF
--- a/dependencies/prettyplease/src/expr.rs
+++ b/dependencies/prettyplease/src/expr.rs
@@ -1016,7 +1016,7 @@ impl Printer {
         }
     }
 
-    fn small_block(&mut self, block: &Block, attrs: &[Attribute]) {
+    pub(crate) fn small_block(&mut self, block: &Block, attrs: &[Attribute]) {
         self.word("{");
         if attr::has_inner(attrs) || !block.stmts.is_empty() {
             self.space();
--- a/dependencies/prettyplease/src/item.rs
+++ b/dependencies/prettyplease/src/item.rs
@@ -46,7 +46,12 @@ impl Printer {
         self.ty(&item.ty);
         self.word(" = ");
         self.neverbreak();
-        self.expr(&item.expr);
+        if let Some(expr) = &item.expr {
+            self.expr(expr);
+        }
+        if let Some(block) = &item.block {
+            self.small_block(block, &[]);
+        }
         self.word(";");
         self.end();
         self.hardbreak();
--- a/dependencies/syn/src/gen/clone.rs
+++ b/dependencies/syn/src/gen/clone.rs
@@ -1289,7 +1289,9 @@ impl Clone for ItemConst {
             ident: self.ident.clone(),
             colon_token: self.colon_token.clone(),
             ty: self.ty.clone(),
+            ensures: self.ensures.clone(),
             eq_token: self.eq_token.clone(),
+            block: self.block.clone(),
             expr: self.expr.clone(),
             semi_token: self.semi_token.clone(),
         }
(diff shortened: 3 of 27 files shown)


tjhance (approved):

tjhance:
Glad to see this, it should be helpful for implementing 'static'

utaal (approved):
The curly braces
```rust
exec const E: u64 ensures E == SPEC_E { 7 }
```
may be a little hard to discover for someone familiar with Rust, but I cannot think of anything better.

PR #846, opened 2023-10-09
Time profile fixes

GenericMonkey:
* delete profile files from previous run to avoid reading stale traces
* do not panic on unexpected / unparseable traces, emit an error instead
* skip profiling for AIR contexts that do not run commands
* report verification time for individual functions
* remove restricting to top 3 modules when time-expanded and output-json are specified.


parno (approved):
Looks reasonable to me.

PR #847, opened 2023-10-10
Support 'static' globals

tjhance:
The basic idea is like consts ... except if you reference the same static more than once, you always get equal values.

~Currently some tests are failing, but most are related to the mode-checking issue we discussed on slack~ Now fixed

### DIFF
--- a/dependencies/prettyplease/src/item.rs
+++ b/dependencies/prettyplease/src/item.rs
@@ -210,7 +210,12 @@ impl Printer {
         self.ty(&item.ty);
         self.word(" = ");
         self.neverbreak();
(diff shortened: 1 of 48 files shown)


tjhance:
OK, this should be good to go now.

Chris-Hawblitzel:
Do you have any examples of using static, and verifying something about it?

tjhance:
I added an example of verifying something similar to lazy_cell. There's currently an issue related to lifetime-checking, so it doesn't pass ... I'll look into it tomorrow.

Chris-Hawblitzel (approved):

PR #850, opened 2023-10-11
Generate SMT commands in SCC order

Chris-Hawblitzel:
Generate SMT commands in SCC (strongly connected component) order, including proof/exec function body checks, so that axioms relying on function body checks appear after the function body checks.  For example, broadcast_forall axioms and trait impl axioms now appear after the necessary function body obligations have been satisfied.

Previously, some SMT commands were generated in SCC order, but others were generated in a separate (later) phase in function order.  Also, trait implementation axioms were generated in yet another (earlier) separate phase.  This pull request eliminates these separate phases so that all function checking is in SCC order.  This will enable these features in the future:
- circular dependencies between spec functions with ensures and proof functions with ensures
- untrusted broadcast_forall functions (especially broadcast_forall functions with trait bounds)

Here's an example of where this matters:

```
trait T { proof fn q() ensures false; }

spec fn f<A>(i: int) -> bool { false }

#[verifier::external_body]
#[verifier::broadcast_forall]
proof fn p<A: T>(i: int)
    ensures f::<A>(i)
{
}

struct S;
impl T for S {
    proof fn q() {
        assert(f::<S>(7)); // FAILS
        // fails because we can't use the broadcast_forall with A = S
        // until after S: T has been established,
        // which can't happen until after S::q has been proven.
    }
}

proof fn test() {
    S::q();
    assert(false);
}
```

Previously, the axiom saying that S implements the trait T was generated in an early phase, before the proof of `S::q` had actually finished the implementation of S as T.  Now, the axiom for S implementing T is generated at the end of the SCC for `S::q`, so that the checking of `S::q` can't circularly rely on this axiom.  (This wasn't a soundness issue yet, because we don't yet support untrusted broadcast_forall, but it is necessary for soundness when we do support untrusted broadcast_forall.)  In general, the pattern of SMT commands is: for each SCC, in sorted SCC order, first generate SMT queries for the obligations for that SCC, then generate SMT assumptions for the axioms enabled by that SCC (e.g. spec function definition axioms, broadcast_forall axioms, trait impl axioms).


source/vir/src/context.rs
```diff
@@ -198,6 +198,25 @@ impl GlobalCtx {
         for t in &krate.traits {
             crate::recursive_types::add_trait_to_graph(&mut func_call_graph, t);
         }
+        for f in &krate.functions {
+            // Heuristic: add all external_body functions first.
+            // This is currently needed because:
+            // - external_body broadcast_forall functions are currently implicitly imported
+            // - open_atomic_invariant_begin/open_local_invariant_begin/open_invariant_end
```
tjhance, line 205: `open_atomic_invariant_begin` and friends are all marked external, so they should be completely ignored, right? Why do we need to account for them?

source/vir/src/context.rs
```diff
@@ -198,6 +198,25 @@ impl GlobalCtx {
...
+            }
+        }
+        for t in &krate.trait_impls {
+            // Heuristic: put trait impls first, because functions don't necessarily have
+            // explicit dependencies on all the trait impls when they are implicitly
+            // used to satisfy broadcast_forall trait bounds.
+            // (Otherwise, we'd need the programmer to put declarations in order,
+            // is in Coq or F*.)
+            func_call_graph.add_node(Node::TraitImpl(t.x.impl_path.clone()));
+        }
```
tjhance, line 219: what exactly are these heuristics for? I assume it's something like "trying to get the final AIR to come out in a nice-ish order relative to its hard constraints" but can you make that clearer?

tjhance (approved):

source/vir/src/context.rs
Chris-Hawblitzel: Oops, my comment was inaccurate.  I misdiagnosed what was going on.

source/vir/src/context.rs
Chris-Hawblitzel: I updated the comment here.

utaal (approved):

PR #853, opened 2023-10-12
Report diagnostics in line order

utaal:
As this has come up a few times, and it turned out to not be particularly difficult, this PR buffers verification errors and warnings (from recommends checks) and reports them in source order, to make them easier to find for the user. However, if a function is taking too long (2 seconds currently) it switches modes, and starts reporting the diagnostics immediately to preserve interactivity.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -499,6 +499,7 @@ impl Verifier {
         bucket_id: &BucketId,
         reporter: &impl air::messages::Diagnostics,
         source_map: Option<&SourceMap>,
+        diagnostics_to_report: &std::cell::RefCell<Option<Vec<(Message, MessageLevel)>>>,
         level: Option<MessageLevel>,
         air_context: &mut air::context::Context,
         assign_map: &HashMap<*const vir::messages::Span, HashSet<Arc<std::string::String>>>,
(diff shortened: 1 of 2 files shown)


source/rust_verify/src/verifier.rs
```diff
@@ -512,8 +513,15 @@ impl Verifier {
         let report_long_running = || {
             let mut counter = 0;
             let report_fn: Box<dyn FnMut(std::time::Duration) -> ()> = Box::new(move |elapsed| {
-                let msg =
+                let mut msg =
                     format!("{} has been running for {} seconds", context.1, elapsed.as_secs());
+                if let Some(in_line_order) = reports_in_line_order.take() {
```
Chris-Hawblitzel, line 518: Should we also sort the existing reports here (i.e. sort the current contents of `in_line_order`)?  I realize that there may still be reports printed later out of order, but it seems like the more often we sort, the closer we'll be to source order.

source/rust_verify/src/verifier.rs
```diff
@@ -499,6 +499,7 @@ impl Verifier {
         bucket_id: &BucketId,
         reporter: &impl air::messages::Diagnostics,
         source_map: Option<&SourceMap>,
+        reports_in_line_order: &std::cell::RefCell<Option<Vec<(Message, MessageLevel)>>>,
```
Chris-Hawblitzel, line 502: For consistency, can we use `*_source_order` rather than `*_line_order`?

Chris-Hawblitzel (approved):

source/rust_verify/src/verifier.rs
utaal: Also based on our conversation offline, I renamed these variables to `diagnostics_to_report`, to (hopefully) better communicate the intent.

PR #856, opened 2023-10-16
Support bounds on traits and associated types

utaal:
This did not require changes in the encoding. Most of the changes handle termination checking for functions and well-founded-ness checking for datatypes when trait and associated type bounds are allowed.

The F*/Coq-style encoding described in `recursive_types.rs` has been extended to include the encoding for these bounds.

Concretely,
* in VIR, traits' typ_bounds include bounds on `Self` (except for the trivial bound Self: T for a trait T added by rustc);
* in VIR, traits now carry `GenericBounds` for their associated types;
* in VIR, associated type impls have `impl_paths` for the impls that are used to satisfy the bounds on the associated type;
* in the function call graph, we add an edge between impls of traits A and B (for the same type) if A: B (see `recursive_types.rs`);
* in the function call graph, we add edges for associated type bounds, similarly to bounds on generic arguments;
* in the type-only graph, we add edges for associate type implementations that use an impl to satisfy an associated type bound (the associated bound is not a predicate that becomes an `impl_path` of the concrete datatype itself).

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -857,6 +857,93 @@ pub(crate) fn param_ty_to_vir_name(param: &rustc_middle::ty::ParamTy) -> String 
     }
 }
 
+pub(crate) fn process_predicate_bounds<'tcx, 'a>(
+    tcx: TyCtxt<'tcx>,
+    param_env_src: DefId,
+    verus_items: &crate::verus_items::VerusItems,
+    predicates: impl Iterator<Item = &'a (rustc_middle::ty::Clause<'tcx>, Span)>,
+) -> Result<Vec<vir::ast::GenericBound>, VirErr>
+where
+    'tcx: 'a,
+{
+    let mut bounds: Vec<vir::ast::GenericBound> = Vec::new();
+    for (predicate, span) in predicates {
+        // REVIEW: rustc docs say that skip_binder is "dangerous"
+        match predicate.kind().skip_binder() {
+            ClauseKind::RegionOutlives(_) | ClauseKind::TypeOutlives(_) => {
+                // can ignore lifetime bounds
+            }
+            ClauseKind::Trait(TraitPredicate { trait_ref, polarity: ImplPolarity::Positive }) => {
+                let substs = trait_ref.args;
+
+                // For a bound like `T: SomeTrait<X, Y, Z>`, then:
+                // T should be index 0,
+                // X, Y, Z, should be the rest
+                // The SomeTrait is given by the def_id
+
+                // Note: I _think_ rustc organizes it this way because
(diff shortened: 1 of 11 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -857,56 +860,17 @@ pub(crate) fn param_ty_to_vir_name(param: &rustc_middle::ty::ParamTy) -> String
     }
 }
 
-pub(crate) fn check_generics_bounds<'tcx>(
+pub(crate) fn process_predicate_bounds<'tcx, 'a>(
```
utaal, line 863: Note that the diff algorithm considered this a rename, where it's in fact factoring out common code and `check_generic_bounds` is otherwise unchanged (you can see it being re-added later).

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -799,7 +799,7 @@ pub(crate) fn check_generic_bound<'tcx>(
     param_env_src: DefId,
     span: Span,
     trait_def_id: DefId,
-    args: &Vec<rustc_middle::ty::Ty<'tcx>>,
+    args: Option<&Vec<rustc_middle::ty::Ty<'tcx>>>, // None if it's a `trait A: B` generic bound on Self
```
Chris-Hawblitzel, line 802: Is the `Option` still used?  It looks like it's always `Some`.

source/vir/src/recursive_types.rs
```diff
@@ -587,6 +593,39 @@ pub fn check_traits(krate: &Krate, ctx: &GlobalCtx) -> Result<(), VirErr> {
...
+
+    // To handle bounds on associated types:
+    //   trait T {
+    //     type Y: U;
+    //     fn f(x: Self, y: Y) -> bool;
+    //     fn g(x: Self, y: Y) -> Self { requires(f(x, y)); };
+    //   }
+    // We can store a Dictionary_U inside Dictionary_T:
+    //   struct Dictionary_T<Self, Y> {
+    //     a: Dictionary_U<Self::Y>,
```
Chris-Hawblitzel, line 620: This is kind of my fault, because I forgot to comment the encoding of associated types, but I think `Y` should be a field of `Dictionary_T` rather than a field of `Self` or a type parameter to `Dictionary_T`.  This way, each trait gets to add its own associated types without disturbing `Self`, and users of `Dictionary_T` can read `Y` from the dictionary instead of having to figure out what `Y` is supposed to be as a type argument to the dictionary.

Chris-Hawblitzel (approved):

source/rust_verify/src/rust_to_vir_base.rs
utaal: Good catch, thanks! Forgot that the only user of None was the one I deleted.

PR #863, opened 2023-10-18
Experimental Profile Flags and Serialization tools

GenericMonkey:
- adding experimental flags `spinoff-all` and `capture-profiles`
  - `spinoff-all` runs each query in its own z3 instance
  - `capture-profiles` silences the output to profile-all generation and emits a `.graph` files, serialized with a format defined in `tool_facade`, for consumption with external tools (one such tool will be added in an upcoming PR)
- adding `tool_facade` crate to help with serialization of data to be exported from rust_verify
- add plumbing in verus to export instantiation graph information using new crate
- collect quantifier span information for non-user-level quantifiers, to be analyzed via `--capture-profiles`

When both `-V spinoff-all` and `-V capture-profiles` are omitted, these changes should have no effect.


Chris-Hawblitzel:
What does the name "facade" mean here?

Chris-Hawblitzel (approved):

utaal:
In offline discussions, we have settled on `internals_interface` as a better name.

utaal:
@parno, as I know you're currently away, I'm going to land this to unblock another PR. If -- when you're back -- you see issues, please bring them up and we'll address them.

source/air/src/profiler.rs
```diff
@@ -89,7 +104,91 @@ impl Profiler {
...
 
-        Ok(Profiler { message_interface, quantifier_stats: user_quant_costs })
+        Ok(Profiler { message_interface, quantifier_stats: user_quant_costs, instantiation_graph })
+    }
+
+    pub fn instantiation_graph(&self) -> Option<&InstantiationGraph> {
+        self.instantiation_graph.as_ref()
+    }
+
+    fn make_instantiation_graph(model: &Model) -> InstantiationGraph {
```
parno, line 114: I'm just looking at this now, but `make_instantiation_graph` seems a bit duplicative with the graph construction in the smt2utils crate.  Should we consolidate the logic in one place or the other?

PR #877, opened 2023-11-02
vargo: Decouple from rustup

zhaofengli:
This PR makes vargo aware of a new environment variable, `VARGO_TOOLCHAIN`, which can be one of the following:

1. Unset: Uses rustup with the toolchain set in `rust-toolchain.toml` (current behavior)
2. Any valid rustup toolchain specification: Uses rustup with the specified toolchain
3. `host`: Uses the Rust toolchain from `PATH`

(3) allows Verus to be built without rustup, which is useful when the Rust toolchain is managed by something else like Nix + [rust-overlay](https://github.com/oxalica/rust-overlay). Furthermore, (2) can be used to temporarily override the rustup toolchain to test Rust updates.

We've been using this downstream for a while since we use Nix to manage everything. We are submitting this upstream to see if this is something you would like and to stop bumping into merge conflicts :smile:

### DIFF
--- a/source/vstd_build/src/main.rs
+++ b/source/vstd_build/src/main.rs
@@ -95,6 +95,11 @@ fn main() {
         "--out-dir".to_string(),
         verus_target_path.to_str().expect("invalid path").to_string(),
     ];
+    if let Ok(sysroot) = std::env::var("RUST_SYSROOT") {
+        child_args.push("--sysroot".to_string());
+        child_args.push(sysroot);
+    }
+
     if release {
         child_args.push("-C".to_string());
         child_args.push("opt-level=3".to_string());
--- a/source/verus/build.rs
+++ b/source/verus/build.rs
@@ -1,6 +1,13 @@
 use regex::Regex;
 
 fn main() {
+    let vargo_toolchain = std::env::var("VARGO_TOOLCHAIN").ok();
+
+    if let Some("host") = vargo_toolchain.as_deref() {
+        println!("cargo:rustc-env=VERUS_TOOLCHAIN=host");
+        return;
+    }
+
     let output = match std::process::Command::new("rustup")
         .arg("show")
         .arg("active-toolchain")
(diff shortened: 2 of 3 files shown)


utaal (commented):
Hi! Thank you for the PR!
This all seems reasonable, but let me think through the implications. One small question in the meantime (Edit: now I've actually posted the question :).

source/vstd_build/src/main.rs
```diff
@@ -95,6 +95,11 @@ fn main() {
         "--out-dir".to_string(),
         verus_target_path.to_str().expect("invalid path").to_string(),
     ];
+    if let Ok(sysroot) = std::env::var("RUST_SYSROOT") {
+        child_args.push("--sysroot".to_string());
+        child_args.push(sysroot);
+    }
```
utaal, line 101: Why do you need this? Is it to set a sysroot manually, that would otherwise be set by `rustup`?

zhaofengli: (Somehow missed the notification, my apologies)

Yes, `rustup` wrappers automatically [set the correct `LD_LIBRARY_PATH`](https://github.com/rust-lang/rustup/blob/5af4bc4a0d4bc69ea9091a7935fb3783c5fb508e/src/toolchain/toolchain.rs#L117-L118) (or equivalents on other platforms) when executing the toolchain binaries, and binaries distributed by `rust-overlay` and `fenix` [have their runpaths patched](https://github.com/oxalica/rust-overlay/blob/58240e1ac627cef3ea30c7732fedfb4f51afd8e7/mk-aggregated.nix#L49-L54) so the default sysroot detection logic works.

utaal: Thanks! Do you know if there's a way to set the sysroot for rustc with an environment variable? Or is it just the `--sysroot` flag?

zhaofengli:
Sorry all, it's been a while but I rebased the PR on top of the latest changes.

source/vstd_build/src/main.rs
zhaofengli: (Sorry again)

I quickly looked, and I couldn't find an environment variable to do this override.

utaal:
[retreat 2025 triage] @zhaofengli now that there are binary distributions of Verus, is this still relevant?

stephen-huan:
This looks good. On the more hacky side, I have a source build [flake](https://github.com/stephen-huan/verus-flake) (indeed, using rust-overlay) which gets around the rustup issue by providing a [fake rustup](https://github.com/stephen-huan/verus-flake/blob/master/src/rustup.py) (verus doesn't actually do that much with it...).

Binary distributions are orthogonal to this. On Nix binaries aren't even executable without patching (https://github.com/NixOS/nixpkgs/pull/452386) and the goal here is a build from source anyways. In particular, we want to avoid rustup and needing to install the right toolchain at runtime.

PR #878, opened 2023-11-02
vstd/bytes: Stabilize proofs

zhaofengli:
Follow up to #772 to stabilize another proof function with `spinoff_prover`.


zhaofengli:
Yes, that's correct. The instability only seems to happen with `--vstd-no-std`.

utaal (approved):

PR #880, opened 2023-11-02
rustup to 1.72.1

utaal:
Most things were obvious. The one thing I am not entirely clear on is the bound regions in the predicate when computing `ImplPaths`. I'm conservatively (and consistently) replacing them with free regions, but it may be okay to just erase them?

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -7,10 +7,10 @@ use rustc_ast::{ByRef, Mutability};
 use rustc_hir::definitions::DefPath;
 use rustc_hir::{GenericParam, GenericParamKind, Generics, HirId, QPath, Ty};
 use rustc_infer::infer::TyCtxtInferExt;
-use rustc_middle::ty::GenericParamDefKind;
 use rustc_middle::ty::Visibility;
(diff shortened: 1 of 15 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -224,14 +224,21 @@ pub(crate) fn get_impl_paths<'tcx>(
...
+
+            // REVIEW: is erasing all regions okay?
+            let delegate = rustc_middle::ty::fold::FnMutDelegate {
+                regions: &mut |_| tcx.lifetimes.re_erased,
+                types: &mut |b| panic!("unexpected bound ty in binder: {:?}", b),
+                consts: &mut |b, ty| panic!("unexpected bound ct in binder: {:?} {}", b, ty),
+            };
+            let poly_trait_refs =
+                tcx.replace_escaping_bound_vars_uncached(poly_trait_refs, delegate);
+            let poly_trait_refs = poly_trait_refs.skip_binder();
```
utaal, line 240: There are, apparently, non-late-bound regions in things like:
```
    pub fn as_ref(&self) -> (a: Option<&A>)
```
(the `Option<&a>`)

`codegen_select_candidate` now accepts a `TraitRef`, without a binder, and panics if it has escaping bound vars.

Chris-Hawblitzel (approved):

PR #884, opened 2023-11-03
In "assert(...);", in positive positions, auto-promote == to =~=.

Chris-Hawblitzel:
// In ordinary asserts,
    // in positive positions,
    // for == on types explicitly supporting ext_eq,
    // replace == with =!=
    // Example:
    //   assert(b ==> c && x == y); // x: S and y: S where S is ext_eq
    // -->
    //   assert(b ==> c && x =!= y);
    // Rationale: assert and assert-by are sort of tactics for various proof techniques
    // (e.g. nonlinear_arith, bit_vector, compute),
    // and in this spirit,
    // the ordinary assert can supply a default "tactic" with some basic heuristics.
    // (To opt out of such heuristics, we could support something like "assert(expr) by()"
    // with an empty by().)


source/vir/src/datatype_to_air.rs
```diff
@@ -74,7 +74,8 @@ fn uses_ext_equal(ctx: &Ctx, typ: &Typ) -> bool {
         TypX::Air(_) => panic!("internal error: uses_ext_equal of Air"),
         TypX::StrSlice => false,
         TypX::Char => false,
-        TypX::Primitive(_, _) => true,
+        TypX::Primitive(crate::ast::Primitive::Array, _) => true,
+        TypX::Primitive(crate::ast::Primitive::Slice, _) => true,
```
utaal, line 78: Is this to cause a type check failure if we add more primitives?

utaal (approved):

source/vir/src/datatype_to_air.rs
Chris-Hawblitzel: Yes, in case there's a new `Primitive` for which `uses_ext_equal` should return a different result.

source/vir/src/heuristics.rs
```diff
@@ -0,0 +1,120 @@
...
+        TypX::Primitive(crate::ast::Primitive::Array, _) => true,
+        TypX::Primitive(crate::ast::Primitive::Slice, _) => true,
+    }
+}
+
+pub(crate) fn insert_ext_eq_in_assert(ctx: &Ctx, exp: &Exp) -> Exp {
+    // In ordinary asserts,
+    // in positive positions,
+    // for == on types explicitly supporting ext_eq,
+    // replace == with =!=
```
tjhance, line 33: why's this written as `=!=` instead of `=~=`?

source/vir/src/heuristics.rs
```diff
@@ -0,0 +1,120 @@
...
+            BinaryOp::Eq(Mode::Spec)
+                if auto_ext_equal_typ(ctx, &e1.typ)
+                    && crate::ast_util::types_equal(&e1.typ, &e2.typ) =>
+            {
+                let op = BinaryOpr::ExtEq(false, e1.typ.clone());
+                let e1 = crate::poly::coerce_exp_to_poly(ctx, e1);
+                let e2 = crate::poly::coerce_exp_to_poly(ctx, e2);
+                exp.new_x(ExpX::BinaryOpr(op, e1, e2))
+            }
+            BinaryOp::And | BinaryOp::Or | BinaryOp::Xor => {
```
tjhance, line 73: the arguments to Xor are not 'positive' positions

tjhance (changes requested):

tjhance (approved):

PR #888, opened 2023-11-04
Replace --arch-word-bits with an in-source directive to set the size of `usize`

utaal:
E.g.
```rust
global size_of usize == 8;
```

This decision also gets recorded as part of the serialized crate, guaranteeing that crates compiled with different assumptions cannot be used together.

This is in part in preparation for broader support for size_of.

### DIFF
--- a/source/vir/src/ast_simplify.rs
+++ b/source/vir/src/ast_simplify.rs
@@ -895,6 +895,7 @@ pub fn simplify_krate(ctx: &mut GlobalCtx, krate: &Krate) -> Result<Krate, VirEr
         external_fns,
         external_types,
         path_as_rust_names,
+        arch,
     } = &**krate;
     let mut state = State::new();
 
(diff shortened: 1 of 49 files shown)


source/vir/src/ast_simplify.rs
```diff
@@ -1039,6 +1031,19 @@ pub fn merge_krates(krates: Vec<Krate>) -> Result<Krate, VirErr> {
...
+                if l != r {
+                    return Err(crate::messages::error_bare(
+                        "all crates must have compatible arch_word_bits (set via `global size_of usize`",
+                    ));
+                } else {
+                    crate::ast::ArchWordBits::Exactly(l)
+                }
+            }
+            (crate::ast::ArchWordBits::Either32Or64, other) => other,
+            (other, crate::ast::ArchWordBits::Either32Or64) => other,
```
tjhance, line 1045: can you check that `other == 32 || other == 64` here

tjhance (approved):

PR #889, opened 2023-11-05
Add enum field access (`->`) and `matches` syntax

utaal:
This is based on conversations at the verus all hands, and on the patterns I saw scanning `verus_systems_code`.

This adds two new syntax extensions:
* `v->f` accesses field `f` in enum `v`, if that field name is used only by one variant, and produces (hopefully) readable error messages for unsupported cases, pointing to:
* `v matches Some(x) ==> x == 3`  (as suggested by @Chris-Hawblitzel), or in general `<v> matches <pattern> ==> <expr>` which expand to `(if let <pattern> = <v> { <expr> } else { true })`.

See the tests for error messages and more usage info.

---

#### `v->f` valid for any field that has the same type in all variants where it appears

This can be made to type check, but cannot be encoded directly to SMT: field accessors in the datatype theory are per-variant, and cannot be aliased. We could encode this to a `match` over all the variants that use the field, but that seems counter to our intent to keep the encoding lean, and make more expensive choices explicit in the syntax.

#### Only `matches`

This would work, but there are enough cases in `verus_systems_code` where field names are unique across variants. `->` allows for less syntactic noise when this is the case, without additional encoding cost.

Note that because we need to do this at the macro level, the `matches` syntax expects a regular pattern (so, unless there's a `use`, one needs the fully qualified `EnumName::VariantName` in the pattern. This contributes to higher syntactic noise with `matches`.

This case:

```rust
enum A<T> {
  This(T),
  That,
}
```

also allows the shorthand `v->0` (if only one variant has a tuple constructor).

### DIFF
--- a/dependencies/syn/src/expr.rs
+++ b/dependencies/syn/src/expr.rs
@@ -235,6 +235,7 @@ ast_enum_of_structs! {
         BigOr(BigOr),
         Is(ExprIs),
         Has(ExprHas),
+        Matches(ExprMatches),
         GetField(ExprGetField),
 
         // Not public API.
@@ -866,7 +867,8 @@ impl Expr {
             | Expr::Is(ExprIs { attrs, .. })
             | Expr::Has(ExprHas { attrs, .. })
             | Expr::Yield(ExprYield { attrs, .. })
-            | Expr::GetField(ExprGetField { attrs, .. }) => mem::replace(attrs, new),
+            | Expr::GetField(ExprGetField { attrs, .. })
+            | Expr::Matches(ExprMatches { attrs, .. }) => mem::replace(attrs, new),
             Expr::Verbatim(_) => Vec::new(),
             Expr::BigAnd(_) => Vec::new(),
             Expr::BigOr(_) => Vec::new(),
@@ -1174,7 +1176,7 @@ pub(crate) mod parsing {
         Arithmetic,
         Term,
         Cast,
-        HasIs,
+        HasIsMatches,
     }
 
     #[derive(PartialEq, Eq, Clone, Copy)]
@@ -1244,7 +1246,7 @@ pub(crate) mod parsing {
                 | Precedence::Arithmetic
                 | Precedence::Term
                 | Precedence::Cast
-                | Precedence::HasIs => Associativity::Left,
+                | Precedence::HasIsMatches => Associativity::Left,
             }
         }
     }
@@ -1494,7 +1496,7 @@ pub(crate) mod parsing {
                     colon_token,
                     ty: Box::new(ty),
                 });
-            } else if Precedence::HasIs >= base && input.peek(Token![is]) {
+            } else if Precedence::HasIsMatches >= base && input.peek(Token![is]) {
                 let is_token: Token![is] = input.parse()?;
                 let variant_ident = input.parse()?;
                 lhs = Expr::Is(ExprIs {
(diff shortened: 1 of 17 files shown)


utaal:
(Test pass on my machine. I don't know why they don't pass on GitHub's runner, I'll take a look soon.)

Chris-Hawblitzel:
Cool!  This looks like nice syntax for specifications.

dependencies/syn/src/expr.rs
```diff
@@ -1507,6 +1511,27 @@ pub(crate) mod parsing {
...
                     rhs: Box::new(rhs),
                 });
+            } else if Precedence::HasIsMatches >= base && input.peek(Token![matches]) {
+                let matches_token: Token![matches] = input.parse()?;
+                let pat = input.parse()?;
+                let implies_token = input.parse()?;
+                let mut rhs = unary_expr(input, allow_struct)?;
+                loop {
+                    let next = peek_precedence(input);
+                    if next >= Precedence::Assign {
```
Chris-Hawblitzel, line 1521: Should this say `>= Precedence::Imply`?  Right, now:

```
enum E { A, B }
...
    assert(E::A matches E::B ==> true <==> false);
```

gets parsed like:

```
    assert(E::A matches E::B ==> (true <==> false));
```

whereas I'd expect it to get parsed like:

```
    assert((E::A matches E::B ==> true) <==> false);
```

Chris-Hawblitzel: A similar issue, but with a different cause, is that:

```
    assert(false && E::A matches E::A ==> true);
```

gets parsed like:

```
    assert(false && (E::A matches E::A ==> true));
```

whereas I'd expect it to get parsed like:

```
    assert((false && E::A matches E::A) ==> true);
```

Chris-Hawblitzel:
Should we also support `matches` without a right-hand side, as in:

```
assert(Some(true) matches Some(_));
```

utaal:
Just discussed @tjhance's point offline with him: we expect to still have (many) cases that benefit from `#[is_variant]`'s `get_X_y()` without a corresponding `matches`, so we won't deprecate `#[is_variant]` for now.

dependencies/syn/src/expr.rs
utaal: Those are good points. Thank you. (The latter will take a bit longer to fix, as you might expect.)

utaal:
@tjhance also suggests `<x> matches <Pat> && ...`

Chris-Hawblitzel:
> @tjhance also suggests `<x> matches <Pat> && ...`

`&&& <x> matches <Pat> &&& ...` may also be something to consider.

utaal:
(This seems to have caused a failure in the singular test suite, I'll investigate.)

dependencies/syn/src/verus.rs
```diff
@@ -1196,22 +1258,96 @@ pub(crate) fn disallow_prefix_binop(input: crate::parse::ParseStream) -> crate::
...
+            MatchesOpToken::Implies(input.parse()?)
+        } else if input.peek(Token![&&]) {
+            MatchesOpToken::AndAnd(input.parse()?)
+        } else {
+            unreachable!()
+        };
+        let mut rhs = expr::parsing::unary_expr(input, allow_struct)?;
+        loop {
+            let next = expr::parsing::peek_precedence(input);
+            if next >= expr::parsing::Precedence::Imply {
```
Chris-Hawblitzel, line 1295: For the `&&` case, this should probably be precedence And:

```
assert(E::A matches E::B && false ==> false);
```

Chris-Hawblitzel: (Alternately, if you want `a matches b && c1 && c2 && ... && cn` to parse the `c1 && ... && cn` left associatively, you could call back into the parser recursively instead looping.)

source/builtin_macros/src/syntax.rs
```diff
@@ -1339,6 +1534,20 @@ impl VisitMut for Visitor {
...
 
+        if let Expr::Binary(ExprBinary { right, .. }) = &expr {
+            if let Expr::Matches(ExprMatches {
+                op_expr: Some(MatchesOpExpr { op_token, .. }),
+                ..
+            }) = &**right
+            {
+                if matches!(op_token, MatchesOpToken::Implies(_)) {
+                    *expr = Expr::Verbatim(
+                        quote_spanned! { expr.span() => compile_error!("matches using ==> is currently not allowed on the right-hand-side of && and || (use parentheses)") },
```
Chris-Hawblitzel, line 1545: This check may need to be generalized.  For example:

```
assert(false == E::A matches E::B && false);
```

dependencies/syn/src/verus.rs
utaal-b: I preferred the former, to keep the matched variable available in further conjuncts, but let me know if you think it should be otherwise.

source/builtin_macros/src/syntax.rs
utaal-b: I generalized the check, and added a number of tests.

Chris-Hawblitzel (approved):
There are still a couple things marked "TODO" but nothing blocking a merge.

PR #890, opened 2023-11-06
fix some array issues with type-invariants

tjhance:
fix a couple of issues reported by @matthias-brun 

1. Array literals were not getting their type invariant.
2. Structs with array literals as fields were not getting marked as needing a type invariant

### DIFF
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -15,8 +15,8 @@ use crate::sst::{
     Pars, Stm, StmX, UniqueIdent,
 };
 use crate::sst_util::{
-    bitwidth_sst_from_typ, free_vars_exp, free_vars_stm, sst_array_index, sst_array_len,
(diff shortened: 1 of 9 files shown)


source/vir/src/ast_to_sst.rs
```diff
@@ -1462,13 +1462,10 @@ pub(crate) fn expr_to_stm_opt(
...
 
-            // assume v.len() == len
-            let len_eq = sst_equal(
-                &expr.span,
-                &sst_array_len(ctx, &expr.span, &v),
-                &sst_int_literal(&expr.span, elems.len() as i128),
-            );
-            stms.push(Spanned::new(expr.span.clone(), StmX::Assume(len_eq)));
+            // assume the type invariant
+            // (this implies that v.len() == len because of a vstd axiom)
```
utaal, line 1466: Which axiom?

tjhance: It's `array_len_matches_n` in pervasive/arrays.rs. I made the comment more specific.

utaal (approved):

PR #891, opened 2023-11-07
add is_empty, is_full to map_lib

jonhnet:
it's painful to write "m == Map::empty()" or "m == map![]" due to holes in type inference. So Maps should have `is_empty` and `is_full` just as Sets do.


jonhnet:
Me too; somehow I got that on my other commit but missed it here.

utaal (approved):

PR #900, opened 2023-11-13
in fn signature, support specifying which invariant namespaces are opened

tjhance:
TODO: tests

### DIFF
--- a/dependencies/syn/dev/main.rs
+++ b/dependencies/syn/dev/main.rs
@@ -8,7 +8,7 @@ syn_dev::r#mod! {
 
     pub proof fn foo2()
         requires true,
-        opens_invariants none
+        opens_invariants [x, y]
     {
     }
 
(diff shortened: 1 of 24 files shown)


tjhance:
this should be ready now

tjhance:
There's a bit of a tangential fix in here. When I added the `opens_invariants` clause I added a check that it uses `old` for any mutable references, similar to the check for `requires` clauses. This led me to notice we didn't have a similar check for `decreases` clauses, so I added that in too.

It turns out there were already some tests that assumed the opposite behavior, that you could write `*a` in a decreases clause and it would always mean the old value. This was surprising to me since it diverged from the 'requires' clause behavior, so I changed those test cases too. I figure this probably doesn't come up much because it's not really common to use decreases clauses in exec functions anyway. Does that all seem ok?

Chris-Hawblitzel (approved):

PR #905, opened 2023-11-16
fix #885 and other ordering related incompleteness

### LINKED ISSUE #885: Axiom ordering seems may be missing an edge when using a spec definition to satisfy the postcondition on a proof fn in a trait
There's code like this:
```rust
        pub trait KeyTrait : Sized {
            fn zero() -> (z: Self) ensures z == Self::zero_spec(); // <-- should this force the definition for `zero_spec` to appear before the proof obligation for `zero` in the impl?

            spec fn zero_spec() -> Self where Self: std::marker::Sized;
        }
        
        pub struct KeyStruct { pub ukey: u64, }
        
        impl KeyTrait for KeyStruct {
            fn zero() -> (z: Self) { KeyStruct{ukey: 0} }

            open spec fn zero_spec() -> Self { KeyStruct{ukey: 0} }
        }
```

that fails to verify, in:

[2023-11-03-19-25-29.zip](https://github.com/verus-lang/verus/files/13253474/2023-11-03-19-25-29.zip)

ordering `zero()` after `zero_spec()` in the `impl KeyTrait for SHTKey` makes it pass.

utaal:
fix https://github.com/verus-lang/verus/issues/885, and non-deterministic incompleteness for cases like:

```rust
trait T {
    spec fn a() -> bool;
    proof fn p() ensures Self::a();
    proof fn q() ensures Self::a();
}

struct S { pub dummy: nat }
impl T for S {
    spec fn a() -> bool;
    proof fn p() { Self::q(); }
    proof fn q() { /* ... */  }
}
```

without this change, Self::p was sometimes lowered to AIR without having processed Self::q first, which was not found as a function with a postcondition https://github.com/verus-lang/verus/blob/825487c07f70a5413ac0a51471bc8fdd90036074/source/vir/src/sst_to_air.rs#L1921 which resulted in the encoding for `p()` to be an empty block instead of the expected `(assume <q's postcondition>)`.

### DIFF
--- a/source/vir/src/recursion.rs
+++ b/source/vir/src/recursion.rs
@@ -1,6 +1,6 @@
 use crate::ast::{
-    AutospecUsage, CallTarget, Constant, ExprX, Fun, Function, FunctionKind, GenericBoundX,
-    IntRange, MaskSpec, Path, SpannedTyped, Typ, TypX, Typs, UnaryOpr, VirErr,
+    AutospecUsage, CallTarget, CallTargetKind, Constant, ExprX, Fun, Function, FunctionKind,
+    GenericBoundX, IntRange, MaskSpec, Path, SpannedTyped, Typ, TypX, Typs, UnaryOpr, VirErr,
 };
 use crate::ast_to_sst::expr_to_exp_skip_checks;
 use crate::ast_util::typ_to_diagnostic_str;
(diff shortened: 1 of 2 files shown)


source/vir/src/recursion.rs
```diff
@@ -466,13 +518,17 @@ pub(crate) fn expand_call_graph(
                 }
 
                 // f --> f2
-                call_graph.add_edge(f_node.clone(), Node::Fun(callee.clone()))
+                call_graph.add_edge(f_node.clone(), Node::Fun(callee.clone()));
+                // Ordering (+a): also add an edge f --> f2' where f2' is the method implemented by f2
+                if let CallTargetKind::Method(Some((_, _, _))) = kind {
+                    call_graph.add_edge(f_node.clone(), Node::Fun(x.clone()));
```
Chris-Hawblitzel, line 524: With https://github.com/verus-lang/verus/pull/917, I think this should no longer be needed because we'd already have `x_resolved --> T --> x`.

utaal:
I believe this has been subsumed by a number of changes.

PR #914, opened 2023-11-21
make state_machine_macros compilable without nightly (also fix one issue in builtin_macros)

utaal:
* add cfg_attr(verus_keep_ghost to a number of attributes in macros on items that are not erased and that are not available outside verus
* remove box_patterns


source/state_machines_macros/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![feature(box_patterns)]
+// #![feature(box_patterns)]
```
tjhance, line 1: you can just delete this line

tjhance (approved):

source/state_machines_macros/src/lib.rs
utaal: ty!

PR #915, opened 2023-11-21
Fix "unwrap_or_else(|| ..." in record_history.rs so it builds

tenzinhl:
Commit [a158ba1281317ee83d71bf5f00d32a54fb1df1d7](https://github.com/verus-lang/verus/commit/a158ba1281317ee83d71bf5f00d32a54fb1df1d7) from a few days ago changed ".expect(&format!" with ".unwrap_or_else(|| panic!" across the repo (including in `record_history.rs`).

However this pattern was actually incorrect for `Result` types (since Result's expect the closure for `unwrap_or_else` to take a single argument). There was a follow-up in commit [5caba470ad99b2336d044866716799d93f3888c8](https://github.com/verus-lang/verus/commit/5caba470ad99b2336d044866716799d93f3888c8) to fix this in most `verus` files, however this commit missed `record_history.rs` (which as of time of writing won't compile). This went undetected as CI currently does not try to build `verus` with the `--feature record-history` flag.

This commit makes it so that the `record-history` feature at the very least builds (I haven't run any tests since I wasn't sure if there were any).

### DIFF
--- a/source/verus/src/record_history.rs
+++ b/source/verus/src/record_history.rs
@@ -81,16 +81,14 @@ pub fn record_history_commit(
                                 fs_path.display(),
                             )
                         })?;
-                        let content_oid = repo.blob(contents.as_bytes()).unwrap_or_else(|| {
+                        let content_oid = repo.blob(contents.as_bytes()).unwrap_or_else(|_| {
                             panic!(
                                 "failed to create a git blob for file {}",
                                 cur_path.join(&name).display()
                             )
-                            .as_str()
                         });
                         treebuilder.insert(name.as_os_str(), content_oid, 0o100644).unwrap_or_else(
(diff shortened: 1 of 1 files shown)


tenzinhl:
@utaal are there tests in the repo for the `record-history` feature? I'll try and just run the code rn to make sure it works, although the changes were typically in error-handling cases so I'm not sure if I'll be able to do a good job verifying these changes work.

utaal (approved):

utaal:
> @utaal are there tests in the repo for the record-history feature?

There aren't, unfortunately. I thought about adding them but they require infrastructure for running `verus` in tests that we don't currently have.

utaal:
I'm going to land this so I can look into https://github.com/verus-lang/verus/pull/916.

PR #917, opened 2023-11-21
Add Node::Trait -> Node::Fun dependency to simplify call graph

Chris-Hawblitzel:
For a `trait T { fn f; }`, add `T --> f` edge.  Also, for an `impl T for S { fn f... }`, add `f --> T` edge.  Together, these subsume some of the special case fine-grained per-call nodes that exist or are being proposed ( https://github.com/verus-lang/verus/pull/905 ).  For example, there used to be a fine-grained case for `trait T { fn f() requires f2() }` that added `T --> f2`; now, this is implied by `T --> f --> f2`.


utaal (approved):
I've also tested this on a few projects, to make sure it doesn't further restrict cases we currently accept and that are in use.

PR #931, opened 2023-12-11
impl<A: Copy> Copy for Tracked<A>

tjhance:


source/builtin/src/lib.rs
```diff
@@ -458,6 +458,16 @@ impl<A> Clone for Ghost<A> {
 
 impl<A> Copy for Ghost<A> {}
 
+impl<A: Copy> Clone for Tracked<A> {
```
Chris-Hawblitzel, line 461: Could this be `A: Clone`, to be more flexible?

Chris-Hawblitzel (approved):
I guess we could already do this indirectly anyway:

```
fn d<A: Copy>(Tracked(x): Tracked<A>) -> (Tracked<A>, Tracked<A>) {
    (Tracked(x), Tracked(x))
}
```

so there's no harm in also allowing it directly.

source/builtin/src/lib.rs
tjhance: We don't actually supported `Tracked::clone` at all right now. If we _did_ support it, how do you think it should work? We either treat it like a 'copy' (in which we case we do need the `A: Copy` bound here, as I wrote) or we treat it like `Tracked(a.borrow.clone())`. Even _defining_ Clone for a tracked object when it's not Copy is already a pretty unusual operation. My inclination is to continue not supporting `Tracked::clone`.

Chris-Hawblitzel: I guess `Clone::clone` wouldn't work on the `A` value anyway, because it's an `exec` function taking an `exec` argument, not a `proof` function taking a `tracked` argument.

tjhance:
If anyone is wondering about the status of this, I decided to wait until my other work on the Clone trait is done.

PR #932, opened 2023-12-12
In lifetime_generate, only select from all_impls(trait_id) on demand

Chris-Hawblitzel:
Previously, the use of `ctxt.tcx.all_impls(trait_id)` was pulling in many unrelated definitions from the Rust library, which led to various failures when using a Rust library trait (e.g. `IntoIterator`).  This pull request prunes away the unneeded definitions, omitting trait impls for types that aren't used.


source/rust_verify/src/lifetime_generate.rs
```diff
@@ -1874,58 +1988,20 @@ fn erase_trait<'tcx>(ctxt: &Context<'tcx>, state: &mut State, trait_id: DefId) {
...
+                for t in &datatypes {
+                    if !state.typ_to_trait_impls.contains_key(t) {
+                        state.typ_to_trait_impls.insert(*t, Vec::new());
                     }
-                    _ => (),
+                    state.typ_to_trait_impls.get_mut(&t).unwrap().push(impl_id);
                 }
             }
+            assert!(!state.impl_to_remaining_typs.contains_key(&impl_id));
+            state.impl_to_remaining_typs.insert(impl_id, (name.clone(), datatypes));
```
utaal-b, line 2003: I think you can just assert that `insert` returned `None`.

source/rust_verify/src/lifetime_generate.rs
```diff
@@ -88,6 +88,13 @@ pub(crate) struct State {
     pub(crate) datatype_decls: Vec<DatatypeDecl>,
     pub(crate) assoc_type_impls: HashMap<AssocTypeImpl, Vec<AssocTypeImplType>>,
     pub(crate) fun_decls: Vec<FunDecl>,
+    // For an impl "bounds ==> trait T(...t...)", point some or all of t to impl:
+    // (We add each impl for T to this when we process T)
+    // (To avoid importing unnecessary impls, we delay processing impl until all t are used)
+    // t1 -> impl, ..., tn -> impl
+    typ_to_trait_impls: HashMap<DefId, Vec<DefId>>,
+    // impl -> (t1, ..., tn) and process impl when t1...tn is empty
+    impl_to_remaining_typs: HashMap<DefId, (Id, Vec<DefId>)>,
```
utaal-b, line 97: `remaining_typs_to_process_impl`?

utaal-b (commented):
Can you also confirm my understanding that this will not consider impls for which _some_ of the type arguments are external?

Chris-Hawblitzel:
> Can you also confirm my understanding that this will not consider impls for which _some_ of the type arguments are external?

If some of the type arguments are unreached, then the impl itself will be unreached.  External types should be unreachable since they are supposed to be disallowed from the verified code altogether.  (There are some places we still need to check this, such as "impl InternalTrait for ExternalType", but that would be a separate well-formedness issue.)

utaal-b (approved):

PR #936, opened 2023-12-13
more specific error msg for assert-by-compute equalities

tjhance:
Example:

```
error: expression simplifies to `3 > 8`, which evaluates to false
  --> ty.rs:15:12
   |
15 |     assert(1 + 2 > 3 + 5) by(compute);
   |            ^^^^^^^^^^^^^

```

### DIFF
--- a/source/vir/src/interpreter.rs
+++ b/source/vir/src/interpreter.rs
@@ -1488,6 +1488,62 @@ fn eval_expr_internal(ctx: &Ctx, state: &mut State, exp: &Exp) -> Result<Exp, Vi
     Ok(res)
 }
 
+/// Restore the free variables we hid during interpretation
+/// and any sequence expressions we partially simplified during interpretation
+fn cleanup_exp(exp: &Exp) -> Result<Exp, VirErr> {
+    crate::sst_visitor::map_exp_visitor_result(exp, &mut |e| match &e.x {
+        ExpX::Interp(InterpExp::FreeVar(v)) => {
+            Ok(SpannedTyped::new(&e.span, &e.typ, ExpX::Var(v.clone())))
+        }
+        ExpX::Interp(InterpExp::Seq(v)) => {
+            match &*e.typ {
+                TypX::Datatype(_, typs, _) => {
+                    // Grab the type the sequence holds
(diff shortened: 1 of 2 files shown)


source/vir/src/interpreter.rs
```diff
@@ -1488,6 +1488,57 @@ fn eval_expr_internal(ctx: &Ctx, state: &mut State, exp: &Exp) -> Result<Exp, Vi
...
 
+enum SimplificationResult {
+    True,
+    False(Option<Exp>),
+    Complex(Exp),
+}
+
+/// Evaluate the result
+///
+/// This special case certain kinds of operations to present more specific error messages
```
parno, line 1499: "This special case [handles] ..."

source/vir/src/interpreter.rs
```diff
@@ -1521,15 +1572,26 @@ fn eval_expr_launch(
...
-        match mode {
+
+    match result {
+        SimplificationResult::True => {
+            return Ok((crate::sst_util::sst_bool(&exp.span, true), state.msgs));
+        }
+        SimplificationResult::False(None) => {
+            return Err(error(&exp.span, "expression simplifies to false"));
+        }
+        SimplificationResult::False(Some(small_exp)) => {
```
parno, line 1588: In the `SimplificationResult::Complex(res)` branch, we clean up some of the internal structures we may have created during interpretation.  Will those be leaked in this branch (e.g., what if one of the terms is a sequence expression)?  If so, we might want to hoist the clean-up into its own function and apply it here too.

tjhance: The way the code works pre-PR is that this clean-up happens in the `ComputeMode::Z3` branch, but it doesn't happen in the `ComputeMode::ComputeOnly` where the expression is only used in an error message. After this PR it works the same way. I don't do any cleanup on small_exp, but the only thing it's used for is the error message.

So I don't know if it's intentional, but it shouldn't 'leak' anything unless it was before.

parno: Hmm, maybe we should take the opportunity to clean up both error reporting paths, in that case.

parno (approved):
Looks great to me -- thanks!

PR #938, opened 2023-12-13
Support into_iter for Range

Chris-Hawblitzel:


Chris-Hawblitzel:
Note: this is based largely on [`rust/library/core/src/iter/range.rs`](https://github.com/rust-lang/rust/blob/master/library/core/src/iter/range.rs)

tjhance (approved):

source/rust_verify_test/tests/regression.rs
```diff
@@ -543,7 +543,7 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
-    #[test] test_for_loop_387_discussioncomment_5683342 verus_code! {
+    #[ignore] #[test] test_for_loop_387_discussioncomment_5683342 verus_code! {
```
utaal-b, line 546: Considering it will get superseded soon, it's not a big concern, but I'd prefer it didn't unnecessarily remain ignored when we merge support for `for` loops. Can you remove this as part of https://github.com/verus-lang/verus/pull/954/files ?

utaal-b (approved):

PR #941, opened 2023-12-14
Fix interpreter logging for multi-threaded verification. 

### LINKED ISSUE #940: Verus fails to write to the interpreter log when run with multiple threads
Given the program below, if we verify it with `verus --log interpreter test.rs`, Verus produces an empty log file in `.verus-log/crate.interp`.  The sample problem occurs if we use `--log-all`.  However, the interpreter logging works if Verus runs in single-threaded mode; i.e., if we run `verus --num-threads 1  --log interpreter test.rs`.  This doesn't appear to affect other forms of logging, e.g., for SMT logging.

```
use vstd::prelude::*;

verus! {
    fn test() {
        assert(40 + 2 == 42) by(compute_only);
    }

    fn main() {}
}
```

achreto:
When running verus with `--log interpreter` and multi-threading, the check whether we want to have interpreter logging was mistakenly using the `log_vir_simple` argument instead of the `log_interpreter`. 

Fixes #940

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1683,18 +1683,11 @@ impl Verifier {
         #[cfg(debug_assertions)]
         vir::check_ast_flavor::check_krate(&krate);
 
-        let interpreter_log_file = Arc::new(std::sync::Mutex::new(
-            if self.args.log_all || self.args.log_args.log_interpreter {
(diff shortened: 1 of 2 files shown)


parno:
Thanks for the quick fix!  I had stared at the code for a bit, but I failed to identify the `log_vir_simple` check as the culprit.

parno (approved):

PR #942, opened 2023-12-14
allow trait method implementations to add ensures clauses

tjhance:
This is completely sound since it lets the trait method implementation promise more than the generic method did.

This will be useful for adding more interesting postconditions when implementing traits from std.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -2352,6 +2352,7 @@ impl Verifier {
 
         let vir_crate = vir::traits::demote_foreign_traits(&path_to_well_known_item, &vir_crate)
             .map_err(map_err_diagnostics)?;
+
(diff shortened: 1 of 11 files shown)


Chris-Hawblitzel:
Sounds like a good idea.

This gives `ill-typed AIR code: error 'in call to ens%scratch!T.f., expected 5 arguments, found 7 arguments' in expression '(ens%scratch!T.f. $ TYPE%scratch!S. B&. B& B&. B& b~2@)'`:

```
trait T<A> {
    proof fn f(a: &A) ensures true;
}
struct S;
impl<B> T<B> for S {
    proof fn f(b: &B) ensures true {  }
}
```

tjhance:
OK, I had a mistaken assumption that the typ_args for the generic function always contained the typ_args for the non-generic function. This should be fixed now

Chris-Hawblitzel (approved):

PR #949, opened 2023-12-27
handle numeric fields in struct syntax

tjhance:
Did you know you can do:

```rust
enum Foo {
    Bar(u32, u32),
    Qux,
}

fn test() {
    let b = Foo::Bar { 1: 30, 0: 20 };
    assert(b.get_Bar_0() == 20);
    assert(b.get_Bar_1() == 30);
}
```

Well, I didn't! I only realized it because rustc's internal de-sugaring of '?' generates code that uses this form. Then I checked, and it turns out users can write it this way too. This PR handles the numeric identifiers correctly (previously, Verus would panic because it didn't translate the to VIR correctly).

VIR previously had this note for struct-syntax patterns and struct-syntax constructors:

```
/// For tuple-style variants, the field initializers appear in order and are named "_0", "_1", etc.
/// For struct-style variants, the field initializers may appear in any order.
```

The "appear in order" part is no longer true as of this PR. It seems nobody was relying on it, anyway, but it would be good for someone other than me to double-check.

### DIFF
--- a/source/rust_verify/src/lifetime_generate.rs
+++ b/source/rust_verify/src/lifetime_generate.rs
@@ -3,7 +3,7 @@ use crate::erase::{ErasureHints, ResolvedCall};
 use crate::rust_to_vir_base::{
     def_id_to_vir_path, local_to_var, mid_ty_const_to_vir, mid_ty_to_vir_datatype,
 };
-use crate::rust_to_vir_expr::get_adt_res;
+use crate::rust_to_vir_expr::{field_name_to_vir_ident, get_adt_res};
 use crate::verus_items::{PervasiveItem, RustItem, VerusItem, VerusItems};
 use crate::{lifetime_ast::*, verus_items};
 use air::ast_util::str_ident;
@@ -77,6 +77,7 @@ pub(crate) struct State {
     datatype_worklist: Vec<DefId>,
     imported_fun_worklist: Vec<DefId>,
     id_to_name: HashMap<String, Id>,
+    field_to_name: HashMap<String, Id>,
     typ_param_to_name: HashMap<(String, Option<u32>), Id>,
     lifetime_to_name: HashMap<(String, Option<u32>), Id>,
     fun_to_name: HashMap<Fun, Id>,
@@ -99,6 +100,7 @@ impl State {
             datatype_worklist: Vec::new(),
             imported_fun_worklist: Vec::new(),
             id_to_name: HashMap::new(),
+            field_to_name: HashMap::new(),
             typ_param_to_name: HashMap::new(),
             lifetime_to_name: HashMap::new(),
             fun_to_name: HashMap::new(),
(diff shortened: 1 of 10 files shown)


Chris-Hawblitzel (approved):

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2201,3 +2203,11 @@ pub(crate) fn closure_to_vir<'tcx>(
         panic!("closure_to_vir expects ExprKind::Closure");
     }
 }
+
+pub(crate) fn field_name_to_vir_ident(name: &str) -> vir::ast::Ident {
+    if name.bytes().nth(0).unwrap().is_ascii_digit() {
```
utaal-b, line 2208: Do we have to rely on the first character being a digit for this? Could rustc already have facilities for understanding this positional field assignment style?

tjhance: I'm not aware of one.

If we don't like it, it might be easier to just handle all fields - numeric and non-numeric - in the same way.

tjhance: right now we convert "0" -> "_0" but this is probably unnecessary. Nothing in our AIR encoding requires fields to start with a non-digit character.

utaal-b: > right now we convert "0" -> "_0" but this is probably unnecessary. Nothing in our AIR encoding requires fields to start with a non-digit character.

This might be because SMTLIB atoms (e.g. field accessors function names) need to start with a [_A-Za-z] ? But it may be they are always prefixed by the name of the datatype anyway.

source/rust_verify/src/lifetime_generate.rs
```diff
@@ -137,6 +139,12 @@ impl State {
         Self::id(&mut self.rename_count, &mut self.id_to_name, IdKind::Local, &raw_id, f)
     }
 
+    fn field<S: Into<String>>(&mut self, raw_id: S) -> Id {
+        let raw_id = raw_id.into();
+        let f = || raw_id.clone();
+        Self::id(&mut self.rename_count, &mut self.field_to_name, IdKind::Field, &raw_id, f)
```
utaal-b, line 145: Will this store the name in `key_to_name` even for numeric fields for which we are not going to use the `rename_count` in `lifetime_emit`?

tjhance: Yes

source/rust_verify/src/rust_to_vir_expr.rs
utaal-b: I'm actually okay with this transformation as (if I understand correctly) it maps names to names and it does not have to match other code.

utaal-b: Maybe a comment would be nice.

utaal-b: It seems brittle to have the condition here that's almost (but not exactly) supposed to match https://github.com/verus-lang/verus/blob/main/source/rust_verify/src/rust_to_vir_adts.rs#L76-L82

Considering that we don't have the distinction from `rustc_middle` there, and here we can get the corresponding names as the initializer names, maybe we should indeed do this:

> it might be easier to just handle all fields - numeric and non-numeric - in the same way

source/rust_verify/src/lifetime_generate.rs
utaal-b: Why is the distinction in `lifetime_emit` necessary?

tjhance: I guess it's not necessary, now that I think about it. I guess the original reason I did it was because I didn't think it made sense to use IdKind::Local for field names.

utaal-b (approved):

PR #950, opened 2023-12-27
Support ? for Option and Result

tjhance:
Internally, rustc de-sugars the 'x?' operation to something like this:

```
match x.branch() {
    ControlFlow::Break(b) => { return from_residual(x); }
    ControlFlow::Continue(c) => c,
}
```

To support '?', we just need to add support for all the content in the de-sugar:

 * The [`std::ops::ControlFlow`](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html) enum. This is easy with `external_type_specification`.
 * The [`std::convert::Infallible`](https://doc.rust-lang.org/stable/std/convert/enum.Infallible.html) enum. This is an _empty_ enum, which Verus doesn't support, but we can mark it `external_body`.
 * Option's implementation of the [`std::ops::Try`](https://doc.rust-lang.org/stable/std/ops/trait.Try.html) trait (specifically the `branch` function).
   * [Option's implementation](https://doc.rust-lang.org/std/option/enum.Option.html#impl-Try-for-Option%3CT%3E), [Source](https://doc.rust-lang.org/src/core/option.rs.html#2468)
 * Option's implementation of the [`FromResidual`](https://doc.rust-lang.org/stable/std/ops/trait.FromResidual.html) trait
   * [Option's implementation](https://doc.rust-lang.org/std/option/enum.Option.html#impl-FromResidual-for-Option%3CT%3E), [Source](https://doc.rust-lang.org/src/core/option.rs.html#2487)

~~Result should work the same way, but the problem is that the [Result's implementation of FromResidual](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromResidual%3CResult%3CInfallible,+E%3E%3E-for-Result%3CT,+F%3E) is more generic than we support right now. It's a similar problem to vec indexing, and we can work around it the same way, but this PR doesn't handle it.~~

UPDATE: Result is now supported.

### DIFF
--- a/source/pervasive/std_specs/control_flow.rs
+++ b/source/pervasive/std_specs/control_flow.rs
@@ -47,19 +47,26 @@ pub fn ex_option_from_residual<T>(option: Option<Infallible>) -> (option2: Optio
     Option::from_residual(option)
 }
 
+pub spec fn spec_from<S, T>(value: T, ret: S) -> bool;
 
-
-/*#[verifier::external_fn_specification]
-pub fn ex_result_from_residual<T, E>(result: Result<Convert::Infallible, E>)
-      -> (cf: ControlFlow<<Result<T, E> as Try>::Residual, <Result<T, E> as Try>::Output>)
+#[verifier::broadcast_forall]
+#[verifier::external_body]
+pub proof fn spec_from_blanket_identity<T>(t: T, s: T)
     ensures
-        cf === match result {
-            Ok(v) => ControlFlow::Continue(v),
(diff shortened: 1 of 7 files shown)


Chris-Hawblitzel (approved):

tjhance:
(merged this into the wrong branch ... I'll fix it manually)

tjhance:
I forgot that this is dependent on #949. So, I will merge this into main after #949 is in.

tjhance:
Now merged in, for real, via 49cd0a599b79e28180e8a18dd6541596366ac437

PR #951, opened 2024-01-02
support unions

tjhance:
Onions have a natural representation as enums, with one field per variant. This makes the translation a little more complicated (rustc represents them like structs) and of course we also need to have non-optional checks on field access.

I changed the Verus `is_variant` builtin to support onions, and made `get_union_field` to be an analogue of `get_variant_field`.

Not supported in this PR:
 * Onion patterns, assignment to onion fields, or &mut references to onion fields
 * #[is_variant] attribute (synstructure library does not support onions, so we basically have to re-do the is_variant implementation entirely)
 * Using onions with external_type_specification


Chris-Hawblitzel (approved):

source/vir/src/ast.rs
```diff
@@ -256,12 +256,20 @@ pub enum UnaryOp {
     CharToInt,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord, ToDebugSNode)]
+pub enum VariantCheck {
+    None,
+    //Recommends,
```
utaal-b, line 262: This is great in preparation for https://github.com/verus-lang/verus/pull/889. Thanks!

utaal-b (approved):

jaybosamiya:
A picture of onion varieties

PR #961, opened 2024-01-16
Refactor state in modes.rs

Chris-Hawblitzel:
Try to leverage ownership and borrow checking in modes.rs to consistently push and pop state during mode checking, even when there are errors.  With this, the mode checking routines no longer have to explicitly save and restore state when changing it.  Instead, pushing into the Typing state returns a new, owned Typing state handle that automatically pops the state changes when it goes out of scope.  Mutable borrow checking ensures that any shadowed Typing state handles are unusable until the new Typing state handle drops.  Since this touches almost all of modes.rs, I'm keeping it as a separate pull request from the main for-loop pull request.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -68,90 +68,149 @@ impl Ghost {
     }
 }
 
-struct Typing {
+struct Ctxt {
     pub(crate) funs: HashMap<Fun, Function>,
     pub(crate) datatypes: HashMap<Path, Datatype>,
     pub(crate) traits: HashSet<Path>,
+    pub(crate) check_ghost_blocks: bool,
+    pub(crate) fun_mode: Mode,
+}
+
+// Accumulated data recorded during mode checking
+struct Record {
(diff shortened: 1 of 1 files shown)


tjhance (approved):

source/vir/src/modes.rs
```diff
@@ -68,90 +68,149 @@ impl Ghost {
...
-    vars_scope_count: usize,
-    in_forall_stmt: bool,
-    block_ghostness: Ghost,
-    ret_mode: Option<Mode>,
-    atomic_insts: Option<AtomicInstCollector>,
+struct Typing<'a> {
+    // don't use these fields directly; use * and push_*
+    internal_state: &'a mut State,
+    internal_undo: Option<Box<dyn for<'b> FnOnce(&'b mut State)>>,
 }
```
utaal-b, line 105: If you place this in a submodule and `use` it at this level, the fields would actually be private to the current module, so there's no risk of someone accidentally accessing `internal_*` directly.

utaal-b (approved):

PR #967, opened 2024-01-24
expand-errors revamp

tjhance:
I've been experimenting with a re-worked expand-errors mechanism. This new method has 2 main goals:

 * Prioritize giving unambiguous information over performance.
 * Present information in a succinct but informative way.

Here's an example:

```
spec fn e() -> bool;
spec fn f() -> bool;
spec fn g() -> bool;
spec fn h() -> bool;

spec fn k(i: int) -> bool;

spec fn z() -> bool {
    (e() ==> h())
      && (forall |i: int| k(i))
}

spec fn stuff() -> bool {
    f() && g() && z()
}

proof fn test()
    requires f(),
{
    assert(stuff());
}
```

The output looks like this:

Screen Shot 2024-01-24 at 2 59 28 PM

Basically this means:
  * stuff() was expanded to f(), g(), and z(), which were each tried individually. f() succeeded, then the other 2 failed.
  * g() was not expanded because it's uninterpreted.
  * z() was expanded further into `e() ==> h()` and `forall |i| k(i)`. Then, `h()` and `k(i)` both failed.

This isn't in any way final - for one thing, we need to clean up the internal junk
(the expressions being printed are internal Exp nodes). We also need to provide source line information for the ones that fail, like in the current system.

Here's a more complicated example. In this run, I configured it to stop after the first failure, else it would have taken forever. Again, the output is a bit weird because of some internal nodes that needs to get cleaned up:

Screen Shot 2024-01-24 at 3 02 41 PM

### Implementation

The implementation is in several parts:

 * First, the addition of "assert_ids" to assertions in AIR and in SST.
 * A "focus" mechanism, basically a function `(AIR, assert_id) -> AIR` to that removes all asserts expect the one with the given assert_id. See `air/src/focus.rs`.
 * An "expand" function, `(sst::Stm, assert_id) -> sst::Stm`. This is similar to the current split_expression code in that it expands an assert into a bunch of other asserts, but it only expands one assert at a time, and it only branches through a single function. The assert_ids are hierarchical, so for example, if it expands the assert with id '42', then the new asserts will have ids '42_0', '42_1', etc. Also, it, produces an 'ExpansionTree', which explains exactly what expansion it did so it can be preented to the user. See `vir/src/expand_errors.rs`.
 * The expand_errors_driver, which iteratively applies the expand function and determines the queries to be run. This is part of the main query loop. See `vir/src/expand_errors_driver.rs`.


jaybosamiya:
Overall, I really like the [proposed output](https://private-user-images.githubusercontent.com/592054/299442454-a283dd5d-3ffc-4eed-a9fe-39b6b0c029f8.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDYxMjc5MzMsIm5iZiI6MTcwNjEyNzYzMywicGF0aCI6Ii81OTIwNTQvMjk5NDQyNDU0LWEyODNkZDVkLTNmZmMtNGVlZC1hOWZlLTM5YjZiMGMwMjlmOC5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMTI0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDEyNFQyMDIwMzNaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0wZTc2Zjk4MGFjMjkwZDg1ZGQzYWRkMjEwNzI1NzI3NGQ5N2JlZjBiOGZmODZjZGNlNTNkMDM1YzFlNjdjNmVlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.4crR58RQpeWOwNePnE0Nf2SiY97jHMc8PCGwTxA4rJY), but have a few minor points:
* none of `stuff`,`f`,... take any arguments, but the output passes `0` to them, which is confusing
* more generally, is the intention that the expanded arguments would be shown (I would like this)
  - if yes, we might also consider include the formal parameter name (so something like `k(i: 2)`, or if it gets `k(i: i)` then collapse to `k(i)` for succinctness); would help for easy readability
* just making a note here that for IDE integration (either verus-analyzer or verus-mode.el), we need to confirm that things look parse-able to what they're reading; we can look at this closer to when this is ready to merge, but yeah, both of them assume `--expand-errors`, and since this touches output, would possibly need a bit of coordination to not break them

tjhance:
> none of stuff,f,... take any arguments, but the output passes 0 to them, which is confusing

Yeah these are part of the 'internal junk' I mentioned. The plan is to clean all that up.

> more generally, is the intention that the expanded arguments would be shown? (I would like this, fwiw)

Yes. Right now it also substitutes the arguments into the expanded function calls, I'm not sure if that's good or not.

> if yes, we might also consider include the formal parameter name (so something like foo(i: 2)

That's a good idea

> just making a note here that for IDE integration 

This is a good point. I don't know what the needs of the IDE integration are. I imagine that I'll eventually add additional 'notes' that point to source lines, similar to the notes in the existing system.

jaylorch:
Nice! I'd suggest removing the string "[uninterpreted]", though, to make the output more streamlined and easier to understand.

parno:
> I'd suggest removing the string "[uninterpreted]", though, to make the output more streamlined and easier to understand.

I think that would make sense for the passing case, but in the failure case (`g()`), it seems like it might be helpful to leave it there, so the user knows why `g` wasn't expanded.  Although perhaps there's a better way of communicating that.  Something similar could possibly apply to `opaque` functions too.

Chris-Hawblitzel:
This seems very useful, and the output looks nice.  One thing I noticed is that timeouts during the expansion get reported as ExpandErrorsResult::Pass.  Would it be clearer to mark these as something like ExpandErrorsResult::Timeout, print them differently, and have them count towards MAX_FAILURES, so the user doesn't wait too long for them?

Here's a test case that reproduces this with --no-vstd and --rlimit 3:

```
spec fn u(i: int) -> bool;

spec fn r(i: int, j: int) -> bool
    decreases i
{
    if i > 0 {
        r(i - 1, 7 * i + 3 * j - 1) || r(i - 1, 11 * i + 5 * j - 3)
    } else {
        u(j)
    }
}

spec fn g1() -> bool;
spec fn g2(i: int) -> bool { g1() && r(12, i) }
spec fn f1() -> bool;
spec fn f2() -> bool { f1() }
spec fn f3() -> bool { f2() }
spec fn f4(i: int) -> bool { f3() && g2(10 * i) && r(13, 21 * i) }
spec fn f5() -> bool { f4(3) && f4(4) && f4(5) && g2(11) && r(14, 22) }
spec fn f6() -> bool { f5() && f5() && r(15, 23) }

proof fn test() {
    reveal_with_fuel(r, 100);
    assert(f6());
}
```

tjhance:
> One thing I noticed is that timeouts during the expansion get reported as ExpandErrorsResult::Pass.

oh good catch. This is a mistake.

utaal-b (approved):

Chris-Hawblitzel (approved):

jaybosamiya:
Hmm, looks like I missed the notification when this was changed from draft to ready. It would've been nice to have checked against IDEs before merging as I'd indicated above, but oh well.

Turns out this _does_ break things on the IDE.

The most obvious is that the output is ugly when mousing over the expanded error:

image

For what is causing this issue, see the `message` and `rendered` fields of `--error-format=json` before and after this PR.

The much more problematic one is that fewer things are being reported. In particular, fewer spans are marked (see that only `h` got the green squiggly below) which means that the IDE will underline only one expanded thing, since it assigns one underline to each line in `--expand-errors=json` or `--expand-errors=short` (fwiw, the `=short` one is what the Emacs side uses, and `=json` is what the VSCode side uses; `=short` is much more human readable, so for improving this, it is good to look at that)
image
This one is very hard to fix on the IDE side, without parsing what is intended to be human-readable output, and should definitely be fixed on the Verus side.

jaybosamiya:
Messed around with things a little bit, and one possible fix to the first issue is something like this:
```
diff --git a/source/rust_verify/src/expand_errors_driver.rs b/source/rust_verify/src/expand_errors_driver.rs
index 7c8632b0..21cb5b80 100644
--- a/source/rust_verify/src/expand_errors_driver.rs
+++ b/source/rust_verify/src/expand_errors_driver.rs
@@ -557,10 +557,11 @@ of additional expressions in the triggering context in the expanded
 version."###
         }
 
-        let mut note = vir::messages::note_bare(s);
+        let mut note = vir::messages::note_bare("expanded errors");
         for (span, _) in self.spans.iter() {
             note = note.primary_span(span);
         }
+        note = note.help(s);
 
         use vir::messages::ToAny;
         note.to_any()
```

Unfortunately, this comes does make the output slightly uglier for the terminal (so I am not suggesting exactly this), but at least gives us a potential starting point

image

tjhance:
I can't find a way to re-open a PR, but I'll take a look at this when I get a chance

jaybosamiya:
Thanks!

Yeah, I don't think GitHub supports reopening PRs, might need to just open a new PR from the same branch, and link to this one to allow seeing the convo

PR #971, opened 2024-02-02
don't compile Regex in frequently called function get_rust_item

tjhance:
Savings are significant; on the mem-allocator project, rust-to-vir goes down from nearly 2s. to 0.3s.


utaal-b:
This is a tiny dependency, but just in case we can avoid it: would the `OnceCell`s introduced here be sufficient? https://github.com/rust-lang/rust/pull/105587

PR #972, opened 2024-02-03
Arithmetic library

jaylorch:
This pull request adds an arithmetic library to Verus. It was originally ported from the [Dafny standard library for
arithmetic](https://github.com/dafny-lang/dafny/tree/master/Source/DafnyStandardLibraries/src/Std/Arithmetic/).


source/pervasive/arithmetic/internals/div_internals_nonlinear.rs
```diff
@@ -0,0 +1,61 @@
...
+    ensures
+        forall|x: int, d: int|
+            0 <= x < d && d > 0 ==> #[trigger]
+            (x / d) == 0,
+{
+}
+
+// TODO: how to translate the `real` type?
+//       seems to be only used here
+/* the quotient of dividing a positive real number (not 0) by a smaller positive real number*/// proof fn lemma_real_div_gt(x:real, y:real)
```
parno, line 52: I think we should cut this, since Verus doesn't support `real`s.  I think this was only include in the Dafny std. lib. due to its use in Ironclad's DiffPriv db

source/pervasive/arithmetic/div_mod.rs
```diff
@@ -0,0 +1,2397 @@
...
+};
+#[cfg(verus_keep_ghost)]
+use crate::arithmetic::internals::general_internals::{is_le};
+#[cfg(verus_keep_ghost)]
+use crate::arithmetic::math::{add as add1, sub as sub1, div as div1};
+use crate::arithmetic::mul::*;
+
+/*****************************************************************************
+   * Division:
+   *****************************************************************************/
```
parno, line 52: Maybe it's just how GitHub is showing it, but the 2nd and 3rd lines of this comment (and the one further below) seem to be oddly indented.  FWIW, I'd also be inclined to cut the colon.

source/pervasive/arithmetic/math.rs
parno, line 1: I think many of these definitions duplicate versions in https://github.com/verus-lang/verus/blob/main/source/pervasive/math.rs.  We should consolidate them here or there.

source/pervasive/arithmetic/mul.rs
```diff
@@ -0,0 +1,757 @@
...
+    assert forall|x: int, y: int| 0 != x && 0 != y && x * y >= 0 implies #[trigger]
+    (x * y) >= x && #[trigger]
+    (x * y) >= y by {
+        lemma_mul_ordering(x, y);
+    };
+}
+
+/*
+    We don't port LemmaMulEquality or LemmaMulEqualityAuto from the
+    Dafny standard library for arithmetic, since they're never useful.
```
parno, line 185: This is a bit enigmatic.  Why are they never useful?

source/pervasive/arithmetic/mul.rs
```diff
@@ -0,0 +1,757 @@
...
+}
+
+/// Proof that multiplication distributes over subtraction when the
+/// subtraction happens in the multiplicand (i.e., in the left-hand
+/// argument to `*`). Specifically, `(y - z) * x == y * x - z * x`.
+/*
+    Note: This isn't in the Dafny standard library, but it's a logical
+    extension of [`lemma_mul_is_distributive_sub`] and
+    [`lemma_mul_is_distributive_add_other_way`].
+*/
```
parno, line 500: I'm not sure we necessarily need to call out places where we extend the Dafny std. lib.  Presumably we'll continue to deviate as their library and ours grow in different directions.  Maybe it would help to change the opening comment at the start of the file to something like, "It was originally based on ..."

source/pervasive/arithmetic/power.rs
```diff
@@ -0,0 +1,802 @@
...
+pub proof fn lemma_square_is_pow2(x: int)
+    ensures
+        pow(x, 2) == x * x,
+{
+    // the original dafny is just a reveal clause
+    // maybe I can do it with reveal_with_fuel? but I don't know how
+    // the following doesn't work
+    // assert(pow(x, 2) == x * x) by {
+    // reveal_with_fuel(pow, 2);
+    //  };
```
parno, line 164: This works with `reveal_with_fuel(pow, 3)`:
https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=5bc7836e15e5205a1d1dc9582901a774
You need one fuel to learn that `pow(x, 2) == x * pow(x, 1)`, another fuel to learn `pow(x, 1) == x * pow(x, 0)`, and one more to learn what `pow(x, 0)` is.

parno (commented):
Thanks very much for pushing this to completion!  Having this in place will already simplify a variety of our system-related proofs.  I've included a few mostly minor comments and suggestions.

In the long run, I think we should aim to keep the comment above a lemma and its auto version in sync.  However, I think the best approach would be something mechanistic, e.g., a macro similar to #[broadcast_forall] that creates a manually invokable, quantified version of the non-auto original.

source/pervasive/arithmetic/internals/div_internals_nonlinear.rs
jaylorch: OK, done.

source/pervasive/arithmetic/math.rs
jaylorch: Good idea. I put them all in `vstd::math`. That way we don't break any existing code that depends on them being there.

source/pervasive/arithmetic/mul.rs
jaylorch: I fleshed out the comment more to clarify, and got Chris's help with the wording.

source/pervasive/arithmetic/mul.rs
jaylorch: OK, I cut those comments.

source/pervasive/arithmetic/power.rs
jaylorch: Thanks, updated!

jaylorch:
Thanks for all the feedback, Bryan! I believe I've addressed all your suggestions, other than the `auto` one which sounded like something you were suggesting for the future. I've also merged to the latest version of Verus and made sure it works (most notably, that Z3 4.12.5 still verifies verything). So I believe it's ready for merging, once the CI tests complete successfully. What do you think?

parno (approved):
> other than the auto one which sounded like something you were suggesting for the future
Yes, that's definitely future work.

Otherwise, this looks great -- thanks so much for getting into a usable state!

PR #973, opened 2024-02-04
Use resolved impl_paths in recursion.rs

Chris-Hawblitzel:
To check for termination in the presence of traits, we model traits as typeclasses that are encoded as "dictionary" datatypes whose fields contain the methods of the trait.  For example, we model `trait T { fn f(); fn g(); }` as something like `struct T { f: Fn(), g: Fn(), }`.  An impl for a trait is modeled as (1) implementing the methods as functions and (2) constructing a dictionary datatype value that contains those functions in the datatype fields.  Note that the methods are constructed first, before the datatype for the impl is constructed.  Therefore, the methods do not recursively have access to their own impl.

This has been the overall design for a while, but Verus has not taken full advantage of this design -- the current cycle checking is overly conservative.  In particular, when there are direct calls to an impl's methods, we should be able to model those as direct calls to the modeled methods, not as indirect calls made by looking up the methods in the dictionary.

Currently, recursion.rs models method calls as lookups in a dictionary even when the call is statically resolved (except for one special case allowing direct intra-impl method calls).  As far as I can tell, this is a historical accident on my part: my commit https://github.com/verus-lang/verus/commit/5c37959340b64401d92e7e6ac3399b9eebeb4a00 introduced impl_paths as part of a bug fix, but used the unresolved impl_paths rather than the more precise resolved impl_paths, and no subsequent commit has revised this.

This pull request simply changes recursion.rs to use the resolved impl_paths when they are available.  This subsumes the previous special case for intra-impl method calls.  It also enables a couple tests that were marked as TODO items.  Note that the test_broadcast_forall_causes_cycle test is now a postcondition failure rather than a cycle; this is because the underlying issue for this test was addressed separately by https://github.com/verus-lang/verus/pull/850 .

### DIFF
--- a/source/rust_verify_test/tests/traits.rs
+++ b/source/rust_verify_test/tests/traits.rs
@@ -1665,6 +1665,33 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] test_broadcast_forall_causes_cycle_simple verus_code! {
+        pub trait Tr {
+            spec fn f() -> bool;
+
+            proof fn bad() ensures false; // FAILS
+        }
+
+        #[verifier::broadcast_forall]
+        #[verifier::external_body]
+        pub proof fn proves_false_requiring_trait_bound<T: Tr>()
+            ensures
+                #[trigger] T::f() == !T::f(),
+        { }
+
+        struct X { }
+
+        impl Tr for X {
+            open spec fn f() -> bool { true }
+
+            proof fn bad() {
+                assert(Self::f());
+            }
+        }
+    } => Err(err) => assert_one_fails(err)
+}
+
+test_verify_one_file! {
     #[test] test_decreases_trait_bound verus_code! {
         trait T {
             proof fn impossible()
--- a/source/rust_verify_test/tests/traits.rs
+++ b/source/rust_verify_test/tests/traits.rs
@@ -536,12 +536,9 @@ test_verify_one_file! {
             }
         }
     } => Err(err) => {
-        // TODO: we could make the recursion rules more precise to allow decreases checking in this example:
-        //assert_eq!(err.errors.len(), 2);
-        //assert!(relevant_error_span(&err.errors[0].spans).text.iter().find(|x| x.text.contains("FAILS")).is_some());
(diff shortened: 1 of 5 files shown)


tjhance (approved):

utaal-b:
> Spec functions are still encoded the same way. Is this a problem? i.e., is it possible that we:
> 
> * Have some trait `Tr` with spec function `f`
> * Have some AIR axioms that are generic over `f`
> * In some trait impl, our 'recursion' checks treat a call to `f` as statically resolved
> * But the proof of that same trait impl still has access to the generic AIR axioms about `f`

My understanding is that this test:
```rust
        mod M {
            pub trait Tr {
                spec fn f() -> bool;

                proof fn bad() ensures false;
            }

            // note the external_body isn't necessary here
            #[verifier::broadcast_forall]
            #[verifier::external_body]
            pub proof fn proves_false_requiring_trait_bound<T: Tr>()
                ensures #[trigger] T::f() == !T::f(),
            {
                T::bad();
            }
        }

        use M::*;

        struct X { }

        impl Tr for X {
            open spec fn f() -> bool
            {
                true
            }

            proof fn bad() {
                other_bad();
            }
        }

        pub proof fn other_bad()
            ensures false, // FAILS
        {
            // It is important that the Trait-Impl-Axiom axiom for "impl Tr for X"
            // (axiom (tr_bound%M.Tr. $ TYPE%X.)
            // appears after "impl Tr for X" and its dependencies.
            // Since "impl Tr for X" depends on other_bad,
            // this ensures that the the Trait-Impl-Axiom axiom for "impl Tr for X"
            // also appears after other_bad.
            // This ensures that the broadcast_forall for proves_false_requiring_trait_bound
            // isn't enabled (its precondition isn't satisfied) for "Tr for X" until
            // after "impl Tr for X" and other_bad are checked.
            // Otherwise, the axiom could be used here in other_bad to prove false.
            let t = X::f();
        }
```

checks for the potential issue that @tjhance is raising. Is that right?

tjhance:
@utaal-b I just realized we never looped you in, but chris and I discussed this offline and found a semi-related issue (#974) but otherwise agreed it's fine

utaal-b (approved):

PR #976, opened 2024-02-07
fix handling of rlimit considering z3's interpretation of a 0 rlimit

utaal-b:
make sure that the default is f32::MAX (e.g. for the vstd build) so that we produce a sensible compute budget for assert by compute

### DIFF
--- a/source/vir/src/interpreter.rs
+++ b/source/vir/src/interpreter.rs
@@ -1655,7 +1655,7 @@ fn eval_expr_launch(
         fun_calls: HashMap::new(),
     };
     // Don't run for too long
-    let max_iterations = (rlimit as f64 * RLIMIT_MULTIPLIER as f64) as u64 * RLIMIT_MULTIPLIER;
+    let max_iterations = (rlimit as f64 * RLIMIT_MULTIPLIER as f64) as u64;
     let ctx = Ctx { fun_ssts: &fun_ssts, max_iterations, arch, global };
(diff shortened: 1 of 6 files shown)


source/vir/src/interpreter.rs
```diff
@@ -1655,7 +1655,7 @@ fn eval_expr_launch(
         fun_calls: HashMap::new(),
     };
     // Don't run for too long
-    let max_iterations = (rlimit as f64 * RLIMIT_MULTIPLIER as f64) as u64 * RLIMIT_MULTIPLIER;
+    let max_iterations = (rlimit as f64 * RLIMIT_MULTIPLIER as f64) as u64;
```
Chris-Hawblitzel, line 1658: Does this need a check for overflow?

Chris-Hawblitzel (approved):
It sounds like `--rlimit infinity` and `#[verifier::rlimit(infinity)]` are popular choices to replace `0`.

source/vir/src/interpreter.rs
utaal-b: I don't think so, the float multiplication will clip at infinity which is turned into u64::MAX by the conversion

Chris-Hawblitzel: > I don't think so, the float multiplication will clip at infinity which is turned into u64::MAX by the conversion

Would the same reasoning apply to `(rlimit * RLIMIT_PER_SECOND).min(u32::MAX as f32) as u32`?

PR #978, opened 2024-02-08
upgrade to z3 4.12.5

utaal-b:
We'd want to announce this in advance, as it may break user code (due to instability).

The breakage on vstd was minimal, and there was no breakage in tests.

By disabling z3's pattern inference, this PR also ensures that we never rely on Z3 picking triggers for us.

### DIFF
--- a/.github/workflows/get-z3.sh
+++ b/.github/workflows/get-z3.sh
@@ -1,5 +1,5 @@
 #! /bin/bash
-z3_version="4.12.2"
+z3_version="4.12.5"
 
 filename=z3-$z3_version-x64-glibc-2.31
 wget https://github.com/Z3Prover/z3/releases/download/z3-$z3_version/$filename.zip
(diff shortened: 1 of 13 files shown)


tjhance:
I tried it out and got 4 failures on the mem alloc project. That's not _too_ terrible - it always seems like a few functions flake when I update verus.

Is there a particular impetus for this or is just a matter of generally keeping up-to-date?

utaal-b:
The specific impetus is support for `(:set-option pi.enabled false)` which we requested, and has now landed in the latest release.

`(:set-option pi.enabled false)` disables Z3's pattern (trigger) inference, so if you don't pass a `(:pattern` attribute, you really get no triggers. This is useful for example when verifying with mbqi instead (as part of the work here https://github.com/verus-lang/verus/tree/epr-mbqi which we hope to merge soon).

But it's also been 6 months since the last bump, and I'd say we don't want to fall too far behind.

tjhance (approved):

utaal-b:
And it looks like I have some AIR tests to take a look at, first.

Chris-Hawblitzel (approved):

PR #981, opened 2024-02-12
Trait default method implementations

Chris-Hawblitzel:
This pull request adds support for trait default method implementations.

In Rust, traits can have default implementations of methods.  For example:

```
    trait T {
        fn f();
        fn g() { Self::f(); Self::g(); }
        fn h();
    }
```

As usual with traits, the tricky part of handling default implementations is checking termination.  The simplest way to handle termination checking would be to make a Node::Fun for the default implementation, following rustc's internal approach, which:
- represents the default implementation as a function T::g
- represents g's calls Self::f() and Self::g() as dynamically dispatched through
  the Self parameter, via the "Self: T" bound.
- represents other calls to T::g as statically bound to T::g (assuming this isn't overridden
  by the impl that is being called through).

However, requiring the "Self: T" be available in T::g would make it hard for impls to use g:

```
    impl T for bool {
        fn f() {}
        fn h() { Self::g(); }
    }
```

In this simple encoding, the impl of h can't call it's own inherited copy of g, because it doesn't yet satisfy the "bool: T" bound (this bound isn't satisfied until after the impl's f and h are fully defined.)

Instead of using the simple approach described above, we make termination checking more flexible by creating a Node::Fun for each inherited copy of a default method, so that there are potentially many Node::Fun values for a single default method.  In the example above, there is one Node::Fun for the original T::g in T and one Node::Fun for the copy of T::g inherited by "impl T for bool".

As long as we check termination for all the copies, it is safe to make copies and to dispatch calls to specific copy for the impl we are calling.  In the example above, h's call to Self::g() calls "impl T for bool"'s copy of g, not the original g in T.

We treat the inherited copies of default methods just like ordinary methods implemented in the impl, so that the inherited copy does not require a "Self: T" bound.  This means that the example above is checked as if all the method were implemented from scratch inside the impl:

```
    impl T for bool {
        fn f() {}
        fn g() { Self::f(); Self::g(); } // this is our own *copy* of the method inherited from T
        fn h() { Self::g(); }
    }
```


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -192,23 +192,93 @@ pub(crate) fn qpath_to_ident<'tcx>(
...
+    span: Span,
+}
+
+fn instantiate_pred_clauses<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    mut def_id: DefId,
+    args: rustc_middle::ty::GenericArgsRef<'tcx>,
+) -> Vec<(Option<ClauseFrom<'tcx>>, Clause<'tcx>)> {
+    // We could get the information directly like this:
+    let direct_clauses = tcx.predicates_of(def_id).instantiate(tcx, args).predicates;
```
utaal-b, line 209: Why was this sufficient before, and now we need the additional information. Is it just to distinguish the `self` trait bound?

source/vir/src/recursion.rs
```diff
@@ -468,19 +469,50 @@ pub(crate) fn expand_call_graph(
 
     // Add f --> f2 edges where f calls f2
     // Add f --> D: T where one of f's expressions instantiates A: T with D: T
-    crate::ast_visitor::function_visitor_check::<VirErr, _>(function, &mut |expr| {
+    //
+    // When instantiating A: T with D: T, note that D: T could, from Rust's perspective,
+    // be the Self: T bound that we remove (see the comments in recursive_types.rs) and
+    // that therefore shouldn't be available here.  Fortunately, in this case,
+    // rustc gives us the concrete D: T bound for the actualy impl, so that
```
utaal-b, line 476:
```suggestion
    // rustc gives us the concrete D: T bound for the actual impl, so that
```

utaal-b (approved):
Thank you for the write-up in the comment.

source/vir/src/ast.rs
```diff
@@ -850,11 +850,14 @@ pub enum FunctionKind {
     },
     /// Method implementation inside an impl, implementing a trait method for a trait for a type
     TraitMethodImpl {
+        /// Fun declarted by trait for this method
```
tjhance, line 853: declarted -> declared

tjhance:
let me make sure I understand the whole get_impl_paths thing:

Ordinarily, when the user calls some trait method, say `Tr::foo`, we statically resolve to the definition of `foo` to a trait implementation method, get the DefId for that the implementation method, and then pass that DefId to `get_impl_paths`.

But in the case of a default trait method, there's no such DefId, so our only option is to use the default implementation in the trait decl. And then we have to 'patch' the trait-solving algorithm to ignore the 'Self<T, ...>' bound.

Is that basically right?

tjhance:
There's something called 'predicates_defined_on', which appears to be the same as predicates_on, minus the Self trait bound. Explanation in the rustc docs: https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.predicates_of

Could this be made useful? If we used this and then plugged the predicates into the original get_impl_paths implementation, would it do the right thing?

tjhance (approved):

source/rust_verify/src/rust_to_vir_base.rs
Chris-Hawblitzel: > Why was this sufficient before, and now we need the additional information. Is it just to distinguish the `self` trait bound?

Yes, that's right.  The only place that uses the additional information (`ClauseFrom.is_self_trait_bound`) is this:

```
                    match (&mut remove_self_trait_bound, inst_bound) {
                        (Some((expected_id, self_trait_impl_path)), Some(b))
                            if *expected_id == b.source_id && b.is_self_trait_bound =>
                        {
                            assert!(self_trait_impl_path.is_none());
                            **self_trait_impl_path = Some(impl_path);
                        }
                        _ => {
                            impl_paths.push(impl_path);
                        }
                    }
```

which only matters when `remove_self_trait_bound` is Some, which is only used when handling direct calls to default method implementations.

Chris-Hawblitzel:
> But in the case of a default trait method, there's no such DefId, so our only option is to use the default implementation in the trait decl. And then we have to 'patch' the trait-solving algorithm to ignore the 'Self<T, ...>' bound.

Right, we are conceptually making separate copies of the default trait method in each impl that inherits the default trait method, and there's no DefId that refers to our conceptual copy.  There's only the original DefId of the original default trait method that we're copying from.  So we have to patch `get_impl_paths` to remove the Self<T, ...> bound that would be required to call the original default trait method but would not be required to call the per-impl copies of the method.

PR #988, opened 2024-02-15
Change Seq to an enum and verify its axioms

### LINKED ISSUE #986: Replace fully axiomatic Seq with a model
The following datatype should be adequate as a model for all of the sequence axioms:

```rs
#[verifier::accept_recursive_types(A)]
pub enum Seq<A> {
    Nil,
    Cons(A, Box<Seq<A>>),
}
```

We should use this as the model and prove the current set of axioms against it.

tchajed:
Instead of making `Seq` an axiomatic type (with `verifier::external_body`), make it an enum that implements the usual inductive definition of a list:

pub enum Seq<A> {
    Nil,
    Cons(A, Box<Seq<A>>),
}

This PR verifies as lemmas everything that was previously an axiom about `Seq`. Due to current limitations of Verus, if these lemmas are annotated with `verifier::broadcast_forall`, they do not have the exact same behavior as the current axioms in that they must be `reveal`'d explicitly rather than being ambiently available, which would affect many downstream proofs. To maintain compatibility I've added the lemmas alongside the axioms, but the ultimate goal would be to remove this duplication.

Now that we have a model it's possible to remove some preconditions and show that this is sound (since the new lemmas are verified rather than assumed). I've done this where I saw opportunities, but didn't propagate the changes to the axioms for now.

This is an attempt to fix #986.

### DIFF
--- a/source/pervasive/seq.rs
+++ b/source/pervasive/seq.rs
@@ -28,28 +28,44 @@ verus! {
 /// To prove that two sequences are equal, it is usually easiest to use the
 /// extensional equality operator `=~=`.
 
-#[verifier::external_body]
-#[verifier::ext_equal]
 #[verifier::accept_recursive_types(A)]
-pub struct Seq<A> {
-    dummy: marker::PhantomData<A>,
+#[verifier::ext_equal]
+pub enum Seq<A> {
+    Nil,
+    Cons(A, Box<Seq<A>>),
 }
 
 impl<A> Seq<A> {
     /// An empty sequence (i.e., a sequence of length 0).
 
     #[rustc_diagnostic_item = "verus::pervasive::seq::Seq::empty"]
-    pub spec fn empty() -> Seq<A>;
+    pub closed spec fn empty() -> Seq<A> {
+        Seq::Nil
+    }
 
     /// Construct a sequence `s` of length `len` where entry `s[i]` is given by `f(i)`.
 
     #[rustc_diagnostic_item = "verus::pervasive::seq::Seq::new"]
-    pub spec fn new(len: nat, f: impl Fn(int) -> A) -> Seq<A>;
+    pub closed spec fn new(len: nat, f: spec_fn(int) -> A) -> Seq<A>
+        decreases len
+    {
+        if len == 0 { Seq::Nil }
+        else {
+            Self::new((len-1) as nat, f).push(f(len-1))
+        }
+    }
 
     /// The length of a sequence.
 
     #[rustc_diagnostic_item = "verus::pervasive::seq::Seq::len"]
-    pub spec fn len(self) -> nat;
+    pub closed spec fn len(self) -> nat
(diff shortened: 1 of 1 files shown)


tchajed:
My first attempt didn't work because as @tjhance points out the proofs are just using the broadcast_forall axioms. I've started to prove the axioms and found an unsoundness in `axiom_seq_add_index2`:

```rs
use vstd::prelude::*;

verus! {
    proof fn seq_bad()
        ensures false
    {
        let s1: Seq<int> = seq![1];
        let s1_2: Seq<int> = seq![1, 2];
        let s2: Seq<int> = seq![];
        assert(s1.add(s2)[0] == s2[-1]);
        assert(s1_2.add(s2)[1] == s2[-1]);
    }

    fn main() {}
}
```

tchajed:
The issue is that this:
```rs
pub proof fn axiom_seq_add_index2<A>(s1: Seq<A>, s2: Seq<A>, i: int)
    requires
        0 <= s1.len(),
        i < s1.len() as int + s2.len(),
    ensures
        s1.add(s2)[i] == s2[i - s1.len()],
```

should have a precondition `0 <= s1.len() <= i`.

parno:
This is very cool!  In theory, we could also use this to simplify the interpreter, which currently has a bunch of special logic for simulating sequences.  It's missing that logic for maps and sets, since no one has asked for it yet, but that could be avoided as well.  However, the interpreter's current policy is not to open `closed` function definitions, so that would be an issue here.  Would it be worth considering using `opaque` instead of `closed`?

Chris-Hawblitzel:
> In theory, we could also use this to simplify the interpreter

I think the plan is to hide the entire datatype definition as a private field of a struct, so the functions would have to be marked closed (opaque open pub functions aren't allowed to access private fields).  We don't have any equivalent of opaque datatypes or opaque public fields.

Also, right now the interpreter uses an "Optimized representation of intermediate sequence results": `Seq(Vector<Exp>)`.  We'd lose this if we interpreted the underlying model directly.  I'm pretty nervous about making the model available directly, even if it's marked opaque.  This depends on pruning to keep the model out of the client SMT context, whereas with `closed` we could erase the model entirely from the vstd library file and be sure it didn't show up in the SMT context.

utaal:
@tchajed @parno what's the state of this? Shall we keep it as an open experiment for now?
Do we know what the performance and verification impact of this is on existing code?

utaal:
Feel free to reopen if it becomes relevant again.

Chris-Hawblitzel:
I think the `broadcast` feature is mature now, so the earlier issues encountered with broadcast should no longer be issues.  My only remaining comment is that `pub enum Seq` should be `pub struct Seq` that contains the `enum` as a private internal field, so other modules don't see the `enum` directly.

utaal:
@tchajed expressed interest in revisiting this.

PR #994, opened 2024-02-16
add a way to inline AIR for testing in Verus source, and add a test for #974, closes #974

### LINKED ISSUE #974: spec functions need tr_bound predicates
broadcast_forall functions are guarded by 'tr_bound' predicates. Spec functions, currently, are not, but spec functions can be used to effectively broadcast information. We need to add 'tr_bound' predicates to spec functions too. (It might be OK to restrict this to recursive functions.)

Example:

```rust
pub trait Tr {
    spec fn bool_function(b: bool) -> bool;

    proof fn bool_function_true_is_false()
        ensures Self::bool_function(true) == false;
}

#[verifier::decreases_by]
proof fn dby<T: Tr>(b: bool) {
    T::bool_function_true_is_false();
}

pub open spec fn spec_fn_with_implications<T: Tr>(b: bool) -> bool
    decreases (if b { 1nat } else { 0 })
        via dby::<T>
{
    if b == false {
        false
    } else {
        !spec_fn_with_implications::<T>(T::bool_function(b))
    }
}

struct X { }

impl Tr for X {
    open spec fn bool_function(b: bool) -> bool { true }

    proof fn bool_function_true_is_false()
    {
    }
}
```

Observe that the spec function of `spec_fn_with_implications` effectively broadcasts the fact that `T::bool_function(true) == false`. To make use of this fact, you need some triggers. I'm not sure if it's possible to observe it right now, it might be possible with the experiments to use other trigger strategies?

As an example, see the attached AIR file:

[ttrait.air.txt](https://github.com/verus-lang/verus/files/14170660/ttrait.air.txt)

This is modified from the original AIR with the addition of the following:

```
  (assert
   ("assertion failed" "assertion failed")
   (ttrait!spec_fn_with_implications.? $ TYPE%ttrait!X. (B true))
  )
  (assume
   (ttrait!spec_fn_with_implications.? $ TYPE%ttrait!X. (B true))
  )
```

utaal-b:
Currently based on https://github.com/verus-lang/verus/pull/1003.

This adds a diagnostics/experimental only feature to inline AIR statement in the Verus source (when the flag `-V allow-inline-air` flag is present).

I'm using it to (indirectly) test that https://github.com/verus-lang/verus/commit/cf45c95b07b7dc44ba84ba2c00327be0891da7ff fixes the latent unsoundness in https://github.com/verus-lang/verus/issues/974

---

* [x] figure out what to do about the state_machine test that fails because verus overflows the stack

### DIFF
--- a/source/air/src/parser.rs
+++ b/source/air/src/parser.rs
@@ -474,7 +474,7 @@ impl Parser {
         Ok(Arc::new(ExprX::Bind(Arc::new(bind), body)))
     }
 
-    pub(crate) fn node_to_stmt(&self, node: &Node) -> Result<Stmt, String> {
+    pub fn node_to_stmt(&self, node: &Node) -> Result<Stmt, String> {
         match node {
             Node::List(nodes) => match &nodes[..] {
                 [Node::Atom(s), e] if s.to_string() == "assume" => {
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -1322,3 +1322,10 @@ pub fn infer_spec_for_loop_iter<A>(_: A, _print_hint: bool) -> Option<A> {
 pub const fn global_size_of<T>(_bytes: usize) {
     unimplemented!()
 }
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::builtin::inline_air_stmt"]
+#[verifier::proof]
+pub fn inline_air_stmt(_s: &str) {
+    unimplemented!()
+}
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -73,6 +73,7 @@ pub struct ArgsX {
     pub log_args: LogArgs,
     pub show_triggers: ShowTriggers,
     pub ignore_unexpected_smt: bool,
+    pub allow_inline_air: bool,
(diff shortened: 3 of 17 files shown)


Chris-Hawblitzel:
I've been wondering for a while whether we should make a pass to change things like `i~2` to just `i` when there are no conflicting variables.  I bring this up because I'm worried that referring to `i~2` won't be stable.

utaal-b:
I was thinking about stability, and was hoping that the logic would change rarely enough to not matter.
Do you think it's worth giving it a shot, and if it turns out to be unstable we can add that pass?

utaal-b:
(I'd honestly really like having this inline AIR tool with reliably named variables, but I don't know if it's a priority.)

tjhance:
Yes, I think it's a good idea to clean those up too. It's caused instability in the past.

utaal-b:
> we should make a pass to change things like i~2 to just i when there are no conflicting variables

This PR's tests are now more stable thanks to https://github.com/verus-lang/verus/pull/997 (which is temporarily the base, until it gets merged).

utaal-b:
FYI, this is now rebased on `main` with the more stable identifiers.

Chris-Hawblitzel (approved):

tjhance (approved):

PR #996, opened 2024-02-17
Update actions/checkout and upload-artifact

tchajed:
This avoids some deprecation warnings in CI about using an older version of Node.js


utaal-b (approved):
ty

PR #1003, opened 2024-02-20
Structure VIR-level identifiers with an explicit disambiguator, and rename rust-provided disambiguators in a stable manner

utaal-b:
Co-authored with @Chris-Hawblitzel.

As discussed for https://github.com/verus-lang/verus/pull/994. Stable variable name identifiers in AIR can help reduce query instability (due to removing Rust's numbering as a source of randomness).

This introduces a `VarIdent` type that maintains an `Ident`, a disambiguator, and a list of suffixes that have been appended to the identifier (instead of performing string concatenation immediately, as before).

For:

```
proof fn test() {
    let x = { let k: int = 12; k + 2 };
    let x = { let k: int = 13; k + 2 };
}
```

it produces

```
(check-valid
 (declare-const no%param Int)
 (declare-const x@ Int)
 (declare-const x$1@ Int)
 (axiom fuel_defaults)
 (block
  (assume
   (= x@ (let
     ((k$ 12))
     (Add k$ 2)
  )))
  (assume
   (= x$1@ (let
     ((k$ 13))
     (Add k$ 2)
)))))
```

---

The variable naming uses a combined approach of retaining the rust disambiguator ids (`local_id`, encoded as `VarIdentDisambiguate ::RustcId(usize)`) until a (scope-based) rename pass converts them into stable identifiers (`VarIdentDisambiguate:: VirRenumbered(u64)`).

`TypParam`s and `Field`s are normally just `Idents` (`Arc<String>`) but sometimes we mix them into lists of `VarIdents`, so their `Ident`s have also been turned into `VarIdents` with `VarIdentDisambiguate::TypParam` or `VarIdentDisambiguate::Field`.

Other identifiers generated internally use the other variants of the `VarIdentDisambiguate` enum.

---

* [ ] double-check the identifier sanitization in Singular


source/rust_verify/src/singular.rs
```diff
@@ -54,6 +54,8 @@ fn sanitize_var_name(name: String) -> String {
             res.push(c);
         } else if c == '_' {
             res.push_str("__");
+        } else if c == '!' {
+            res.push_str("____");
```
utaal-b, line 58: This is probably not quite right yet. I'll come back to it in a bit.

source/vir/src/def.rs
```diff
@@ -686,15 +688,56 @@ pub fn strslice_defn_path(vstd_crate_name: &Ident) -> Path {
...
+    match uniq_id {
+        VarIdentDisambiguate::AirLocal => {}
+        VarIdentDisambiguate::NoBodyParam => {
+            out.push_str(SUFFIX_PARAM);
+        }
+        VarIdentDisambiguate::Field => {
+            out.push_str(SUFFIX_PARAM);
+        }
+        VarIdentDisambiguate::TypParam => {}
+        VarIdentDisambiguate::RustcId(id) | VarIdentDisambiguate::ClosureReturnValue(id) => {
```
tjhance, line 711: why do multiple cases go to the same body?

source/vir/src/def.rs
```diff
@@ -686,15 +688,56 @@ pub fn strslice_defn_path(vstd_crate_name: &Ident) -> Path {
...
+        }
+        VarIdentDisambiguate::VirExprNoNumber => {
+            out.push_str(SUFFIX_LOCAL_EXPR);
+        }
+        VarIdentDisambiguate::VirParam => {
+            out.push_str(SUFFIX_PARAM);
+        }
+        VarIdentDisambiguate::AstSimplifyTemp(id) | VarIdentDisambiguate::AstToSstTemp(id) => {
+            write!(&mut out, "{}", id).unwrap();
+        }
```
tjhance, line 735: Same question

tjhance (changes requested):
Can we remove the suffixes and fold the functionality into the disambiguator? It would be nice to make VarIdentDisambiguate a small Copy type and remove the extra Arc.

source/vir/src/def.rs
```diff
@@ -243,81 +243,77 @@ pub fn suffix_global_id(ident: &Ident) -> Ident {
...
-    match name.rfind(SUFFIX_LOCAL_EXPR) {
-        None => {}
-        Some(i) => {
-            name = name[..i].to_string();
-        }
-    }
-    Arc::new(name)
+pub fn subst_rename_ident(x: &VarIdent, n: u64) -> VarIdent {
+    // In principle, we could get renamed more than once, so accumulate n as a suffix
+    x.push_suffix(&format!("{SUBST_RENAME_SEPARATOR}{n}"))
```
tjhance, line 269: I don't think this is necessary. `subst_rename_binders` already checks that the new subst index won't trample any existing 'free_vars'. And I don't see why nesting the idents would help anyway.

Chris-Hawblitzel:
Ok, I think the last few commits should address the comments so far.

tjhance (approved):

PR #1004, opened 2024-02-21
Apply verusfmt to vstd for the first time

parno:


utaal-b:
Thank you and @jaybosamiya for working on this!

We may want to separately review if there are style changes that we'd want to tweak, but I'd have three procedural requests before we go ahead and merge this:
* [ ] if we're using verusfmt on the standard library, we're now taking a dependency on it from verus; shall we move it into the `verus-lang` org so contributors who make verus changes have access to make `verusfmt` fixes if necessary?
* [ ] can we set up a CI check and the related machinery to let users check verusfmt on the standard library, so that it remains formatted?
* [ ] (minor) I generally use an empty author for commits that make broad changes applied with a tool, or by importing external code (with `--author "Nobody <nobody>"), so that we maintain (indirectly) the original author of a line of code.

I'd in particular like to see CI, so that we make sure committed code remains in-line with `verusfmt`, which also preserves the `git blame`, which turns out often useful to determine when/how a change was made. (If there's no CI, then the code tends to drift, and then gets reformatted occasionally, which uninformative attributions in git blame again.)

parno:
> shall we move it into the verus-lang org so contributors who make verus changes have access to make verusfmt fixes if necessary?

Yes, if you're comfortable with that, we're happy to make the move.  

> can we set up a CI check and the related machinery to let users check verusfmt on the standard library, so that it remains formatted?

Yes, definitely, this should be part of CI.  Would it be reasonable to set that up as a separate PR, so the changes are distinct from the massive but mechanical changes in this PR?

> I generally use an empty author for commits that make broad changes applied with a tool, or by importing external code (with --author "Nobody <nobody>"), 

Is that something we can do to a commit retroactively, or do we need to restart this PR?

jaybosamiya:
Will coordinate with you @utaal-b on Slack to make the move (since stuff might need updates due to permissions and such).

As for the CI PR, I recommend we use the (automatically) prebuilt binaries that I've set up for verusfmt.  However, I don't know how GitHub will handle things upon the move of verusfmt to the verus-lang org (I assume they'll URL-rewrite, but I haven't tested that with release artifacts in particular), so to keep things simple, it might be best for us to finish migrating it to the verus-lang and _then_ setup the CI PR.

Regarding author-nobody: agreed on rewritten history if it seems ok to @parno; I'm bringing this up mostly to also point out that we should also additionally add the commit to the `.git-blame-ignore-revs` file, so that [GitHub will automatically ignore such commits in its blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view); the addition to the `.git-blame-ignore-revs` prob is best added as part of this PR itself.

parno:
If you're able to remove the author for 69f9fee and add it to a `.git-blame-ignore-revs` file, that sounds fine to me.   I'm glad we can do it retroactively. 

I can work with Jay to get verusfmt added to the CI once we migrate the verusfmt repo to `verus-lang`

utaal-b:
Done. We must not squash or rebase this branch when marging through GitHub's UI as that would invalidate the git hash we added to `.git-blame-ignore-revs`. This will have to be a merge commit.

parno:
We're now in sync with `main`.  Can we merge this in at this point?  It seems simpler than rebasing, since I don't know if the custom edits in `4072213` will replay cleanly when doing the rebase, and they were tedious to do by hand.

utaal-b:
> since I don't know if the custom edits in 4072213 will replay cleanly when doing the rebase, and they were tedious to do by hand.

From my offline chat with @jaybosamiya I got the impression that there were no custom edits in that commit, which is why I suggested the rebase. Now knowing that's not the case, I absolutely agree with merging instead of rebasing.

I don't know if you're just planning ahead, but I'd very much prefer if we only merged this once the CI check is ready to be merged as well, as I'd rather not have `vstd` re-diverge from "approved" verusfmt.

parno:
> I'd very much prefer if we only merged this once the CI check is ready to be merged as well, as I'd rather not have vstd re-diverge from "approved" verusfmt.

I think the CI changes should be ready this week, but if we hold off merging, every edit to vstd will cause a conflict with this PR and require manual intervention.  If we merge now, those edits might cause a small drift, but that seems easy to fix mechanically by simply running `verusfmt`.  Are there additional costs associated with temporary divergence?

tjhance:
Since verusfmt is frequently updated, is it going to cause problems? Like a CI that succeeds one day and then fails because verusfmt is updated? Is there a plan to avoid this kind of annoyance?

utaal-b:
That's a fair point. As long as you're confident we will have CI support sometime this week, I personally think you can go ahead and merge.

However, I would also like to hear from @tjhance and @Chris-Hawblitzel that they don't have any other concerns/questions first.

utaal-b (approved):
(Conditional on @tjhance and @Chris-Hawblitzel's approval.)

Chris-Hawblitzel:
I'm fine with merging and waiting on CI.  To address @tjhance 's point about verusfmt itself being updated, we might need the CI to point to a particular version of verusfmt at any given time.

utaal-b:
> I'm fine with merging and waiting on CI. To address @tjhance 's point about verusfmt itself being updated, we might need the CI to point to a particular version of verusfmt at any given time.

We will also need to make sure that the verus dev has the correct version installed at the time, otherwise their local check won't match with what CI expects. Perhaps a task for vargo?

/cc @jaybosamiya

Chris-Hawblitzel (approved):

jaybosamiya:
> Since verusfmt is frequently updated, is it going to cause problems? Like a CI that succeeds one day and then fails because verusfmt is updated? Is there a plan to avoid this kind of annoyance?

I don't believe that "CI succeeds one day, fails the next" is something to be particularly worried about due to a collection of factors:
1. We expect and strive to have very few changes to its output on things that is already supported. Basically, verusfmt is regularly updated, but almost all of its updates are to improve support for things (new Verus syntax features, etc.), rather than change style on existing features.
2. Verusfmt _itself_ runs a CI job which performs snapshot checks to prevent divergence (for example, see the [Verus `syntax.rs` unchanged test](https://github.com/verus-lang/verusfmt/blob/1508d14bf8490b511ab6309d717a500b70e2b8e9/tests/syntax-rs-unchanged.rs), or any of our [consistency tests](https://github.com/verus-lang/verusfmt/blob/1508d14bf8490b511ab6309d717a500b70e2b8e9/tests/verus-consistency.rs)), thereby preventing unintentional divergence. We can additionally add more tests to the snapshot if there are particular files that folks want extra assurance on. We already have some files planned to be added there (soon) for extra assurance.
3. In the rare case of intentional changes to output, that would be a situation that requires some coordination, but I think we can handle this on a case-by-case basis, since we expect such changes are to be primarily hyper-localized.

>  To address @tjhance 's point about verusfmt itself being updated, we might need the CI to point to a particular version of verusfmt at any given time.
One approach to this is to have two separate jobs in the CI, one pinned (this is the primary formatting check, and thus PRs must satisfy this) and one floating at latest verusfmt release (this is a secondary check, and thus won't fail or block merging or anything; will only act as warning upon divergence). 

Given the intended stability goals for verusfmt, I suspect that we don't need to be particularly paranoid about this: I think we should perform pinning only if we see serious breakages being an actual issue. For now, I think it is reasonable for us to just use the latest verusfmt release (which again is intended to be stable).

> We will also need to make sure that the verus dev has the correct version installed at the time, otherwise their local check won't match with what CI expects. Perhaps a task for vargo?

I understand this could be a possible longer-term concern, but for now, I don't think we should block on this. Adding it as part of `vargo fmt` might be good as a separate PR though to improve convenience for verus devs.

parno:
In light of the discussion above, I'm going to go ahead and merge this.  We'll update the CI soon thereafter.

PR #1010, opened 2024-02-22
Support checking loops without spinning off a separate query

Chris-Hawblitzel:
Currently, Verus always spins off a separate SMT query to verify a loop.  This pull request adds the option of not spinning off a query, and instead verifying the loop within the original query.  Both approaches are useful in different situations:
- Verifying the loop in the original query allows the loop verification to share the context with the original query, so that fewer loop invariants are needed.  However, the query may become slow for medium and large functions.
- Spinning off a separate query gives the loop its own private context, allowing for faster, more modular verification.  However, more loop invariants may be needed to share information between the outer query and the loop query.

In general, verifying the loop in the original query is more convenient for small functions, while spinning off a separate query is more practical for medium and large functions.

A new attribute `verifier::spinoff_loop` controls which of these two options is chosen, with `verifier::spinoff_loop` and `verifier::spinoff_loop(true)` turning on the separate query and `verifier::spinoff_loop(false)` turning it off.  The default behavior is still to spin off a separate query, so that by default programmers don't run into trouble with timeouts.  However, this default can be changed by placing `verifier::spinoff_loop(false)` on a function, a module, or the entire crate.  For example, if someone wants the default to be `verifier::spinoff_loop(false)` for the whole crate, they can add the `#![verifier::spinoff_loop(false)]` attribute to the root of the crate, and then if needed use `verifier::spinoff_loop(true)` on individual modules, functions, or loops.


tjhance (approved):

utaal-b (approved):

jonhnet:
Aawww yiss, livin' the dream. This is great. Thank you. I just fought this pain for the last time. In the function I'm working on, I could remove (count 'em) _four_ obvious invariants. I look forward to never having to stumble through that again.

jonhnet:
Naming suggestion:
[verifier::spinoff_prover] at the function level, from a user perspective, doesn't change anything about what I "have to do", it just changes the performance of the verifier (and thus can tickle stability issues).
In the loop context, `spinoff` now has a very different meaning: context is being explicitly forgotten where otherwise (inside a function) context generally accumulates.
For the while loops, I'd suggest an attribute name that says more about how it affects the user, rather than how it's implemented. Perhaps `isolate_loop`?

Chris-Hawblitzel:
Note that the name ended up being `verifier::loop_isolation(false)` (rather than `verifier::spinoff_loop(false)`).

jonhnet:
Hah! That explains why my crate-level use of `spinoff_loop(false)` feature stopped working, driving me batty for a while. :vD

Issue #608, opened 2023-06-14
Verus not instantiating deeply nested `forall`s in functions


pratapsingh1729:
We are attempting to work with a recursive datatype where the branching is controlled by a `FnSpec`. We want to write our own extensional equality predicate on these datatypes, but we are unable to do any proofs with it, since it seems that Verus will not instantiate the `forall` quantifier that we need for the extensionality. As a minimal example:

```
#[is_variant]
pub enum T {
    Base(nat),
    F(FnSpec(nat) -> T)
}

pub open spec fn is_equal (x : T, y : T) -> bool 
    decreases x 
{ 
    match x {
        T::Base(b) => x == y,
        T::F(ff) => match y {
            T::F(gg) => forall|v| #[trigger] is_equal(ff(v), gg(v)),
            _ => false
        }
    }
}

pub proof fn test(ff : FnSpec(nat) -> T, gg : FnSpec(nat) -> T) {
    assert ((forall |v| #[trigger] is_equal(ff(v), gg(v))) ==> is_equal(T::F(ff), T::F(gg))); // 1
    assert (is_equal(T::F(ff), T::F(gg)) ==> (forall |v| #[trigger] is_equal(ff(v), gg(v)))); // 2
}
```

Both of the assertions in `test` fail. It seems like both assertions should follow directly from "reducing" the definition of `is_equal`, so we aren't sure why this isn't going through.

pratapsingh1729:
Looking at the generated smt, we have the following which I think maps to assertion 1:
```
      (=>
       (forall ((v~22$ Poly)) (!
         (=>
          (has_type v~22$ NAT)
          (equality!is_equal.? (%%apply%%0 ff~2@ v~22$) (%%apply%%0 gg~4@ v~22$))
         )
         :pattern ((equality!is_equal.? (%%apply%%0 ff~2@ v~22$) (%%apply%%0 gg~4@ v~22$)))
         :qid user_equality__test_5
         :skolemid skolem_user_equality__test_5
       ))
       (equality!is_equal.? (Poly%equality!T. (equality!T./F (%Poly%fun%1. (Poly%fun%1. ff~2@))))
        (Poly%equality!T. (equality!T./F (%Poly%fun%1. (Poly%fun%1. gg~4@))))
     ))
```
But the axiom for the recursive case of `is_equal` has a pattern that uses `equality!rec%is_equal.?`:
```
(assert
[...]

PR #1017, opened 2024-02-29
Refactor sst_visitor.rs

Chris-Hawblitzel:
I was working on https://github.com/verus-lang/verus/issues/608 and found that none of the visitors in sst_visitor were quite right for the bug fix I had in mind.  Rather than add yet another visitor or hack in another feature to an existing visitor, I figured it was time to overhaul and generalize the existing visitors.  I think the original visitors in sst_visitor were too specialized to particular tasks, and as a result, we've had to add more and more visitors over time.  At the moment, sst_visitor contains 3 separate passes over `Exp` and 6 separate passes over `Stm`.  This pull request generalizes all of this into 1 pass over `Exp` and 1 pass over `Stm`.  The pull request leaves the old interfaces in place, so no code outside sst_visitor sees any difference, but all of the existing interfaces are now just simple wrappers around a more general trait-based visitor.  Hopefully, some of the same code and techniques can also apply to ast_visitor.

### DIFF
--- a/source/vir/src/sst_visitor.rs
+++ b/source/vir/src/sst_visitor.rs
@@ -42,6 +42,31 @@ impl Scoper for ScopeMap<VarIdent, bool> {
     }
 }
 
+/*
+The central general-purpose SST visitor trait is `vir::sst_visitor::Visitor`.
+A specific visitor specializes `Visitor` by providing the following:
+- An error type `Err` that is returned when a visit exits early
+- A normal return type, specified by instantiating `R: Returner`
+  with either `R = Walk` or `R = Rewrite`:
+  - `Walk` defines the return type `R::Ret<A> = ()`, so that `visit_typ`, `visit_exp`,
+    and `visit_stm` return type `()`.
+  - `Rewrite` defines the return type `R::Ret<A> = A`, so that `visit_typ` returns `Typ`,
+    `visit_exp` returns `Exp`, and `visit_stm` returns `Stm`.
+- An optional scoped variable tracker specified by type `Scope` and data supplied by `scoper`.
+  This could be anything, but the current visitors either use `Scope = NoScoper`
+  or `Scope = VisitorScopeMap`.
+- Overridden `visit_typ`, `visit_exp`, and `visit_stm` methods.
+  (For convenience, these have no-op default implementations,
+  so if you don't override these you get a visitor that returns immediately without recursing.)
+
+The `visit_typ`, `visit_exp`, and `visit_stm` methods could do anything they want, but typically
+at least one of them will call the built-in recursive traversal methods provided by `visit_exp_rec`
+and `visit_stm_rec`.  These built-in recursive traversal methods recursively call back into the
+user-supplied `visit_typ`, `visit_exp`, and `visit_stm`.
+Typically, the user-supplied `visit_typ`, `visit_exp`, and `visit_stm`
+will do some preprocessing before recursing, and/or postprocessing after recursing,
+and/or filtering to decide whether to recurse or not.
+*/
 pub(crate) trait Visitor<Err, R: Returner<Err>, Scope: Scoper> {
     // These methods are often overridden to make a specific sort of visit
 
--- a/source/vir/src/sst_visitor.rs
+++ b/source/vir/src/sst_visitor.rs
@@ -67,7 +67,7 @@ Typically, the user-supplied `visit_typ`, `visit_exp`, and `visit_stm`
 will do some preprocessing before recursing, and/or postprocessing after recursing,
 and/or filtering to decide whether to recurse or not.
 */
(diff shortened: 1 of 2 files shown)


tjhance:
Could you detail the high-level picture a bit?

I recently to implement a weird ad hoc visitor for expand-errors. https://github.com/verus-lang/verus/blob/16cd2764a8b18848e673700aa6bbd78521022eed/source/vir/src/sst_visitor.rs#L685C15-L685C50
Would it be easy to implement this visitor with your system?

Chris-Hawblitzel:
> Could you detail the high-level picture a bit?

Sure.  I can write some documentation here and add it as comments to the source code.

The central general-purpose SST visitor trait is `vir::sst_visitor::Visitor`:

```
pub(crate) trait Visitor<Err, R: Returner<Err>, Scope: Scoper> {
    fn visit_typ(&mut self, typ: &Typ) -> Result<R::Ret<Typ>, Err>;
    fn visit_exp(&mut self, exp: &Exp) -> Result<R::Ret<Exp>, Err>;
    fn visit_stm(&mut self, stm: &Stm) -> Result<R::Ret<Stm>, Err>;
    fn scoper(&mut self) -> Option<&mut Scope>;
...
}
```

A specific visitor specializes `Visitor` by providing the following:
- An error type `Err` that is returned when a visit exits early
- A normal return type, specified by instantiating `R: Returner` with either `R = Walk` or `R = Rewrite`:
  - `Walk` defines the return type `R::Ret<A> = ()`, so that `visit_typ`, `visit_exp`, and `visit_stm` return type `()`.
  - `Rewrite` defines the return type `R::Ret<A> = A`, so that `visit_typ` returns `Typ`, `visit_exp` returns `Exp`, and `visit_stm` returns `Stm`.
- An optional scoped variable tracker specified by type `Scope` and data supplied by `scoper`.  This could be anything, but the current visitors either use `Scope = NoScoper` or `Scope = VisitorScopeMap`.
- Overridden `visit_typ`, `visit_exp`, and `visit_stm` methods.  (For convenience, these have no-op default implementations, so if you don't override these you get a visitor that returns immediately without recursing.)

The `visit_typ`, `visit_exp`, and `visit_stm` could do anything they want, but typically at least one of them will call the built-in recursive traversal methods provided by `visit_exp_rec` and `visit_stm_rec`.  These built-in recursive traversal methods recursively call back into the user-supplied `visit_typ`, `visit_exp`, and `visit_stm`.  Typically, the user-supplied `visit_typ`, `visit_exp`, and `visit_stm` will do some preprocessing before recursing, and/or postprocessing after recursing, and/or filtering to decide whether to recurse or not.

> I recently to implement a weird ad hoc visitor for expand-errors. https://github.com/verus-lang/verus/blob/16cd2764a8b18848e673700aa6bbd78521022eed/source/vir/src/sst_visitor.rs#L685C15-L685C50 Would it be easy to implement this visitor with your system?

Yes, this sort of visitor can be written as a recursive function `visit_stm_prev` that provides its own handling of `StmX::Block` and sends the other cases to `visit_stm_rec` (which recursively calls into `visit_stm`, which recursively calls into `visit_stm_prev`).  Here's a simplified example:

```
struct MapStmPrevVisitor<'a, F> {
    fs: &'a mut F,
}

impl<'a, F> MapStmPrevVisitor<'a, F>
where
    F: FnMut(&Stm, Option<&Stm>) -> Result<Stm, VirErr>,
{
    fn visit_stm_prev(&mut self, stm: &Stm, prev: Option<&Stm>) -> Result<Stm, VirErr> {
        let stm = if let StmX::Block(ss) = &stm.x {
            let mut stms: Vec<Stm> = Vec::new();
            for (i, s) in ss.iter().enumerate() {
                let prev = if i == 0 { None } else { Some(&ss[i - 1]) };
                stms.push(self.visit_stm_prev(s, prev)?);
            }
            Spanned::new(stm.span.clone(), StmX::Block(Arc::new(stms)))
        } else {
            self.visit_stm_rec(stm)?
        };
        (self.fs)(&stm, prev)
    }
}

impl<'a, F> Visitor<VirErr, Rewrite, NoScoper> for MapStmPrevVisitor<'a, F>
where
    F: FnMut(&Stm, Option<&Stm>) -> Result<Stm, VirErr>,
{
    fn visit_stm(&mut self, stm: &Stm) -> Result<Stm, VirErr> {
        self.visit_stm_prev(stm, None)
    }
}

pub(crate) fn map_stm_prev_visitor<F>(stm: &Stm, fs: &mut F) -> Result<Stm, VirErr>
where
    F: FnMut(&Stm, Option<&Stm>) -> Result<Stm, VirErr>,
{
    let mut visitor = MapStmPrevVisitor { fs };
    visitor.visit_stm(stm)
}
```

tjhance (approved):
I think I get it! Thanks.

PR #1018, opened 2024-02-29
add a concept of 'prophecy-dependent functions'

tjhance:
This is pretty minimal, but it should be enough that Andrea's mutable-references can plug right into it.

---

The key idea is to distinguish functions that are "prophecy dependent" vs "prophecy independent". The vast majority of functions will be prophecy independent. An example of a prophecy-dependent function would be getting the future value of a prophecy variable or the future value of a mutable reference.

In this PR, we are very restrictive on where prophecy-dependence can be used. Right now, prophecy dependence can only be used:

 * in spec functions marked `#[prophetic]`
 * in asserts, requires, ensures, invariants (places where it's very obvious that values cannot "escape")

However, we can add more smarts later. For example, we should be able to call proof functions with prophecy-dependent arguments as long as it doesn't mutate any tracked state and we treat any of its outputs as prophecy-dependent.

It's important that users won't have to worry about prophecy rules when they aren't making use of them. Luckily, this system is all nice and parametric, so functions like `spec fn identity<T>(t: T) -> T { t }` are prophecy-independent like you expect, even if `T` is instantiated as `&mut S` or `Prophecy<S>`. One way to think about this is to represent prophecy values as functions over the future-state. In this framework, prophecy-dependent functions take an implicit parameter representing the future-state, whereas prophecy-independent ones don't. So you can pass parametrically pass around these parameterized values as long as you don't instantiate them.

### DIFF
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -283,6 +283,8 @@ pub(crate) enum Attr {
     // Marks a trait as "sealed", i.e. not implementable in Verus code
     // requires it to also be marked `unsafe`
     Sealed,
+    // Marks spec functions that depend on resolved prophecies
+    ProphecyDependent,
 }
 
 fn get_trigger_arg(span: Span, attr_tree: &AttrTree) -> Result<u64, VirErr> {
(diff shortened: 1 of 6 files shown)


Chris-Hawblitzel (approved):

utaal-b (approved):

tjhance:
Thanks for the reminder!

I kind of wanted to let it stew on my brain to see if I still thought it's a good idea a couple weeks later. I didn't realize how much time already passed 😅

PR #1019, opened 2024-02-29
Add `verusfmt --check` to CI

jaybosamiya:
This PR adds `verusfmt --check` to the CI for vstd

With this commit, we should prevent drifts from happening from the initial reformat PR (#1004).

There were some minor changes to verusfmt since then (primarily fixes of indentation of code _after_ a `verus!` macro), which causes a minor amount of whitespace churn in this PR, but I've marked that commit as authored by `verusfmt` (reminder to myself to add to `.git-blame-ignore-revs` right before merging; also to use a merge commit, rather than a squashing)


.github/workflows/ci.yml
```diff
@@ -29,6 +29,19 @@ jobs:
...
           cargo fmt -- --check
 
+  verusfmt:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+      - name: setup verusfmt
+        run: |
+          curl --proto '=https' --tlsv1.2 -LsSf https://github.com/verus-lang/verusfmt/releases/latest/download/verusfmt-installer.sh | sh
```
mmcloughlin, line 39: Pin `verusfmt` version? Otherwise this check could start failing on main branch?

jaybosamiya: Thanks for bringing this up. I've written more about this in https://github.com/verus-lang/verus/pull/1004#issuecomment-1965148204

In summary, it is an explicit decision to _not_ pin it.

jaybosamiya:
Added documentation as a separate section, since verusfmt will soon be applied to other files too (such as `syntax.rs`)---will have those changes in a separate PR, to not slow this particular PR down.

I also did a rebase onto `main` to update the reformatting commit, since it seems like `main`'s vstd has moved on since #1004 while waiting on the CI being activated, and thus vstd had diverged from being properly formatted.

utaal-b:
Thank you. Looks good, but the current instructions would only work on *nix and we have Windows users (and developers).

I added the equivalent file search to `vargo`, and a check that the user has a recent-enough verusfmt that we can bump occasionally in #1024. If that looks good, merge it in here, and then I think we can merge this.

jaybosamiya:
I looked over that PR, and it indeed would be useful to add verusfmt to `vstd fmt`; however, some updates might be necessary there.

Just so that this PR is not blocked on that, I added instructions for Windows to the docs in addition to clarifying that the prior instructions were for *nix.

Additionally, I just did a rebase of this PR against latest `main`, to re-run verusfmt against all the updates to vstd on `main`.

utaal-b (approved):
I'm happy to get this merged, but I'd like to get the `vargo fmt` support in soon so that the dev process doesn't require more manual steps.

Thanks!

jaybosamiya:
Thanks! Did a last rebase and added the commit to `.git-blame-ignore-revs`. As soon as the CI goes green, will merge this in.

PR #1021, opened 2024-03-01
Support Clone trait

tjhance:
* Extend external_fn_specification with general support specifying trait function impls
* Hardcode support for Clone and the generic Clone::clone
* Add support for a few types, including Vec

Note that Clone::clone_from is completely ignored. It has a default implementation which causes all kinds of problems. We pretend it doesn't exist and error if the user tries to use it.

### DIFF
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -1441,6 +1441,7 @@ dependencies = [
  "bincode",
  "console",
  "getopts 0.2.21 (git+https://github.com/utaal/getopts.git?branch=parse-partial)",
+ "indexmap 1.9.3",
  "indicatif 0.17.7",
  "internals_interface",
  "num-bigint 0.4.4",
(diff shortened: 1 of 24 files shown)


parno:
Can we strengthen the post-condition for `clone` in some way so that this example doesn't require a manual proof (except perhaps the `assert(w@ =~= v@)`)?
```
        fn test_bool_vec(v: Vec<bool>) {
            let w = v.clone();
            proof {
                assert_seqs_equal!(w@, v@, i => {
                    assert(call_ensures(<bool>::clone, (&v@[i],), w@[i]));
                    assert(w@[i] == v@[i]);
                });
            }
            assert(w@ =~= v@);
        }
```
This seems rather heavyweight (and unintuitive) for a developer who expects clone to "obviously" produce a copy of the original object.  (unless the example above is spelled out more than necessary for didactic purposes?)

tjhance:
It should work without the proof{} block

parno:
Excellent!

Chris-Hawblitzel:
> Can we strengthen the post-condition for `clone` in some way so that this example doesn't require a manual proof (except perhaps the `assert(w@ =~= v@)`)?

We can probably do even better by adding the postcondition `vec@ =~= res@ ==> vec@ == res@,`.  Then we don't even need the assert.

Chris-Hawblitzel (approved):

parno:
> We can probably do even better by adding the postcondition vec@ =~= res@ ==> vec@ == res@,. Then we don't even need the assert

Ooh, very cool!

PR #1024, opened 2024-03-08
add verusfmt formatting for vstd to vargo

utaal-b:
This will run verusfmt as part of `vargo fmt` if `verusfmt` is available, or print a warning instead.

### DIFF
--- a/tools/vargo/src/main.rs
+++ b/tools/vargo/src/main.rs
@@ -8,6 +8,8 @@
 #[path = "../../common/consts.rs"]
 mod consts;
 
+const MINIMUM_VERUSFMT_VERSION: [u64; 3] = [0, 2, 7];
+
 mod util;
 
 use filetime::FileTime as FFileTime;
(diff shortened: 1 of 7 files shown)


tools/vargo/src/main.rs
```diff
@@ -679,6 +683,111 @@ fn run() -> Result<(), String> {
...
+                                    .extract::<3>()
+                                    .1
+                                    .iter()
+                                    .map(|v| v.parse::<u64>().unwrap())
+                                    .collect::<Vec<u64>>();
+                                for (cv, ev) in
+                                    verusfmt_version.iter().zip(MINIMUM_VERUSFMT_VERSION.iter())
+                                {
+                                    if ev > cv {
+                                        return Err(format!("expected `verusfmt` version to be at least {}.{}.{}, found {}.{}.{}; refer to CONTRIBUTING.md for installation instructions",
```
jaybosamiya, line 725: `CONTRIBUTING.md` points to GitHub for installation instructions anyways, we can just point folks there directly instead?

```suggestion
                                        return Err(format!("expected `verusfmt` version to be at least {}.{}.{}, found {}.{}.{}; refer to https://github.com/verus-lang/verusfmt for installation instructions",
```

tools/vargo/src/main.rs
```diff
@@ -679,6 +683,111 @@ fn run() -> Result<(), String> {
...
+                                }
+                                let original_args = original_args
+                                    .iter()
+                                    .map(|x| x.as_str())
+                                    .collect::<std::collections::HashSet<&str>>();
+                                if !original_args.is_subset(&std::collections::HashSet::from([
+                                    "--",
+                                    "--check",
+                                    "fmt",
+                                    "--color=always",
```
jaybosamiya, line 740: I'm not sure why this is here, since:
```console
$ vargo fmt -- --color=always
Option 'color' given more than once
Option 'color' given more than once
error: `cargo fmt` returned status code Some(1)

$ vargo fmt --color=always
error: unexpected argument '--color' found

  tip: to pass '--color' as a value, use '-- --color'

Usage: cargo fmt [OPTIONS] [-- <rustfmt_options>...]

For more information, try '--help'.
error: `cargo fmt` returned status code Some(2)
```

tools/vargo/src/main.rs
```diff
@@ -679,6 +683,111 @@ fn run() -> Result<(), String> {
...
+                                    format!(
+                                        "verusfmt returned error code {:?} on file {}, exiting",
+                                        verusfmt_status.code(),
+                                        vstd_file_path.display()
+                                    )
+                                    .as_str(),
+                                );
+                                std::process::exit(verusfmt_status.code().unwrap_or(1))
+                            }
+                        }
```
jaybosamiya, line 787: `verusfmt` explicitly continues checking files even upon failure (see https://github.com/verus-lang/verusfmt/pull/37), since that leads to better user experience (when there is more than one file that has an issue, it is best to show them all at once, rather than quitting on the first). The easiest way to achieve this would be to simply pass all the arguments to verusfmt directly (so `verusfmt.args(walk_vstd)` or similar).

jaybosamiya (commented):
Thanks for adding verusfmt to `vargo fmt`. I've added some inline comments.

One high-level comment that I am concerned about is complexity/repetition when we start to include other files (such as `syntax.rs`) into the fold.

The other concerning fact is that for some reason, vargo does something that disagrees with running verusfmt normally wrt line wrapping---this seems to be in the the `rustfmt` portion of `verusfmt` (i.e., stuff outside `verus!{ }`), so my guess is this is caused due to differing versions of rustfmt somehow(?). But yeah, I think we should get to the bottom of that, before including this into `vargo fmt`. Do you have any guesses as to what could be causing this?

tools/vargo/src/main.rs
utaal-b: This was to let the reader skip a step. `CONTRIBUTING.md` points to `verusfmt` for installation instruction.

tools/vargo/src/main.rs
utaal-b: Because `--color=always` gets added earlier by `vargo` to `args`, so this would fail otherwise.

tools/vargo/src/main.rs
utaal-b: Makes sense.

utaal-b:
> One high-level comment that I am concerned about is complexity/repetition when we start to include other files (such as syntax.rs) into the fold.

Can you elaborate? My preference is that there should be _one command_ the user can run to fix all the formatting errors that the CI would complain about, to avoid a longer cycle of the user making a change and having to wait for CI to see if they need to run formatting on something. I just run `vargo fmt` before committing when I develop, and I know CI won't fail.

utaal-b:
> The other concerning fact is that for some reason, vargo does something that disagrees with running verusfmt normally wrt line wrapping---this seems to be in the the rustfmt portion of verusfmt (i.e., stuff outside verus!{ }), so my guess is this is caused due to differing versions of rustfmt somehow(?). But yeah, I think we should get to the bottom of that, before including this into vargo fmt. Do you have any guesses as to what could be causing this?

I suspect you're right this may have to do with which version of rustfmt gets picked up based on the toolchain, but I think it's going to be easier to debug once the CI PR is merged so I can check against a "blessed" formatting of vstd.

tools/vargo/src/main.rs
jaybosamiya: Fwiw, I was suggesting that we change it to skip the step; I see you've already changed that manually, so marking this as resolved.

tools/vargo/src/main.rs
jaybosamiya: Ah, thanks for the explanation!

jaybosamiya:
Apologies, I should've clarified better as "not a blocking concern"---I meant complexity for maintaining consistency across the CI script, and the documentation, and in vargo. I agree 100% that for users, simply running a single command is the right move.

The wrapping differences were a blocker though, but indeed, we should probably figure that out once we have the CI PR merged (doing that shortly)

jaybosamiya (approved):
Things overall look good to me.  One minor comment: should we add `--no-verusfmt` as an argument, which disables all verusfmt for `vargo fmt`?  That way, we can possibly aid with debugging and also looking at CI logs would be a bit nicer, since they won't then show the verusfmt-not-available warning for the `cargo fmt` step (which can possibly lead to user confusion)

The CI thing _does_ lead to a separate question too btw: should we consider installing `verusfmt` as part of the activation step (verusfmt installation scripts support installing to specific paths, so we can also have it just be similar to vargo in terms of being added to PATH), and thus no longer needing a separate verusfmt CI step (instead the `vargo fmt --check` step is sufficient).

None of the above is blocking btw, although I do think it would be nice to add in the `--no-verusfmt` sooner rather than later.

utaal-b:
> That way, we can possibly aid with debugging and also looking at CI logs would be a bit nicer, since they won't then show the verusfmt-not-available warning for the cargo fmt step (which can possibly lead to user confusion).

This is a good point. I (partly, perhaps) addressed this by performing the verusfmt check as part of the `vargo fmt -- --check`, by installing verusfmt first.

PR #1025, opened 2024-03-08
Fix the link to z3 release

marshtompsxd:
The z3 version before the fix is not found here: https://github.com/Z3Prover/z3/releases.


utaal-b:
Thanks!

PR #1027, opened 2024-03-09
fix #979, support casting from T: Integer to int and nat

### LINKED ISSUE #979: builtin::Integer isn't enough to make an `as int` cast
If you verify this example as it appears:
```rust
use vstd::prelude::*;

verus!{

pub trait Obligations<T> {
    spec fn reveal(t: T) -> T
        ;
}

struct S<T> {
    t: T
}

impl<T> Obligations<T> for S<T> {
    open spec fn reveal(t: T) -> T
    {
        t
    }
}

impl<T/*: builtin::Integer*/> S<T> {
    spec fn t_int(t: T) -> int {
        Self::reveal(t) as int
    }
}

}
```

You get this helpful message:
```rust
verus --time --crate-type=lib ./simple-bug.rs
error[E0277]: the trait bound `T: builtin::Integer` is not satisfied
   --> ./simple-bug.rs:23:9
    |
23  |         Self::reveal(t) as int
    |         ^^^^^^^^^^^^^^^-------
    |         |
    |         the trait `builtin::Integer` is not implemented for `T`
    |         required by a bound introduced by this call
    |
note: required by a bound in `builtin::spec_cast_integer`
   --> /home/jonh/verus/source/builtin/src/lib.rs:797:32
    |
797 | pub fn spec_cast_integer<From: Integer, To: Integer>(_from: From) -> To {
    |                                ^^^^^^^ required by this bound in `spec_cast_integer`
help: consider restricting type parameter `T`
    |
21  | impl<T: builtin::Integer/*: builtin::Integer*/> S<T> {
    |       ++++++++++++++++++
```

However, adding the proposed type restriction (the commented text in the example) doesn't work; it gives this error:
```rust
error: Verus currently only supports casts from integer types and `char` to integer types
  --> ./simple-bug.rs:23:9
   |
23 |         Self::reveal(t) as int
   |         ^^^^^^^^^^^^^^^^^^^^^^
```

utaal-b:


Chris-Hawblitzel:
Looks good.  A couple comments:

I think `impl Integer for S {}` should be an error:

```
pub open spec fn plus_three<T: Integer>(t: T) -> nat {
    t as nat + 3
}

struct S;
impl Integer for S {}

proof fn test() {
    assert(plus_three(S) + 1 == 1 + plus_three(S));
}
```

Can we put `CastToInteger` in `UnaryOp` instead of `UnaryOpr`?

utaal-b:
Good catch. Thanks!

PR #1030, opened 2024-03-11
Add DeepView

Chris-Hawblitzel:


source/vstd/std_specs/vec.rs
```diff
@@ -207,16 +215,37 @@ pub fn ex_vec_split_off<T, A: Allocator + core::clone::Clone>(
...
+            #![all_triggers]
+            0 <= i < vec.len() ==> call_ensures(T::clone, (&vec[i],), res[i]),
+        vec_clone_trigger(*vec, res),
+        vec@ =~= res@ ==> vec@ == res@,
 {
     vec.clone()
 }
 
+/*
+//TODO: improve pruning so that this is pruned away unless vec_clone_trigger is used
```
utaal-b, line 236: Is this commented out because it would provide equality of deep view if we can prove extensional equality of the deep view of the vectors in the presence of the right trigger on `clone`, but you don't want it broadcasted everywhere?

Chris-Hawblitzel: It's commented out because it would bloat the SMT encoding by bringing in everything about all the types implementing DeepView, even if the code never uses DeepView.  I would like to improve pruning of broadcasts soon so that it's based it on the triggers in each broadcast rather than always bringing in all broadcasts in the module.

utaal-b (approved):

PR #1034, opened 2024-03-13
Unwind

tjhance:
pretty much as we discussed yesterday.

This might break some code out there, if anyone is using AtomicInvariant/LocalInvariant. Maybe we should make an announcement before rolling this out.

I was thinking maybe we could also have a 'global' option (similar to the options about size_of that andrea added) to be used with -C panic=abort. If panic=abort, then all the unwinding checks can be disabled.

**current status:**

I realized there are a couple of issues, places where unwinding could happen that this PR doesn't handle:

 - [x] Box::new, Rc::new. These operations are all ignored in VIR right now because they're basically no-ops, but I think they can unwind. (Well, I'm not completely positive, but I'm not aware of a guarantee that they won't unwind.)
 - [x] drops

### DIFF
--- a/dependencies/syn/dev/main.rs
+++ b/dependencies/syn/dev/main.rs
@@ -1,29 +1,15 @@
 syn_dev::r#mod! {
     // Write Rust code here and run `cargo check` to have Syn parse it.
 
-    pub proof fn foo<T: FnOnce(u8) -> ()>()
-    {
-    }
-
-    pub proof fn foo2(f: fn(u8) -> ())
-    {
-    }
-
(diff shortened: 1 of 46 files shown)


Chris-Hawblitzel (approved):

jaybosamiya:
@tjhance The latest verusfmt release now includes the PR you'd made. I kicked off a re-run of this PR's CI for formatting. The failure that shows up now will automatically be fixed if you rebase onto main due to #1062

utaal:
Marking as draft, given the current issues in the PR description. Feel free to un-mark and merge if I misinterpreted those.

PR #1036, opened 2024-03-18
Make types more precise after adjustment and REF

utaal-b:
Decorations on the types of expression don't generally matter, but we are a bit imprecise in what we compute at the moment. The decorations matter for trait dispatch via the solver, which we believe to be sound as long as the type arguments to function calls are obtained from rustc and not from the types of the expressions of the arguments.

This PR makes the decorations of types of expressions more precise nonetheless. It helped me reason about next steps for prophecy encoding. *It also produces a correct type for the expression of calling an external body function (which had been hardcoded to `Bool`*).

### DIFF
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -70,7 +70,7 @@ pub(crate) fn fn_call_to_vir<'tcx>(
     {
         return Ok(bctx.spanned_typed_new(
             expr.span,
-            &Arc::new(TypX::Bool),
+            &expr_typ()?,
             ExprX::Block(Arc::new(vec![]), None),
         ));
     }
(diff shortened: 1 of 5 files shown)


tjhance (commented):
Why is it necessary to add a separate 'deref' flag?

Would it not be possible to perform all the relevant type-decoration logic at the 'deref' operation itself? i.e., when you encounter a non-mut deref *X, first descend into X like normal, then strip the Ref decoration off of the type of the resulting expression.

Is there some reason we need to remember that we're inside a deref, which I'm missing?

utaal-b:
> Why is it necessary to add a separate 'deref' flag?

It isn't. Good catch. (I'll blame that original choice on the fact that I was still not feeling great. :)

I reworked things roughly the way you described.

tjhance (approved):

Chris-Hawblitzel (approved):

PR #1039, opened 2024-03-19
update edition to 2021 and allow setting edition via command line

tjhance:
Right now there's no way to change the edition, and it's hardcoded to 2018.

(Did we start verus before edition 2021...? Dang)

Anyway, this both updates the edition to 2021 and also lets you set it on the command line.

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -263,8 +263,10 @@ where
         + Send
         + Sync,
 {
-    rustc_args.push(format!("--edition"));
-    rustc_args.push(format!("2018"));
+    if !rustc_args.contains(&"--edition".to_string()) {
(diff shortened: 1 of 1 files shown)


tjhance:
updating the default edition isn't really necessary, I just figured I'd do it while I'm at it. The important thing is allowing the user to set the edition

utaal-b:
One question: I believe 2018->2021 had primarily syntax implications, but I think some type-checking implications too. I imagine they probably don't affect us, but did you check if there's anything newly allowed in 2021 that we might not handle correctly?

tjhance:
I went through the list [here](https://doc.rust-lang.org/edition-guide/rust-2021/index.html) and didn't see anything concerning.

utaal-b (approved):

PR #1041, opened 2024-03-19
update to 1.76

tjhance:
This one was kind of challenging - 1.76 introduces these "host_effect" params as extra generic-const params. This seems to have something to do with const generics, so I think we just want to ignore them.


utaal-b:
Can you check why `fmt` is unhappy?

tjhance:
fmt is fixed

Chris-Hawblitzel (approved):

utaal-b (approved):
Please announce this on Zulip in advance, before merging, so folks can prepare.

PR #1042, opened 2024-03-21
Require open-invariant credits

jaylorch:
Future planned Verus features, such as `dyn` and trait methods that open invariants, may make Verus unsound. This issue is described in the "Logical Paradoxes" section of the Iris 4.1 Reference (`https://plv.mpi-sws.org/iris/appendix-4.1.pdf`). This pull request applies a solution to this, described in the paper "Later Credits: Resourceful Reasoning for the Later Modality" by Spies et al. (available at `https://plv.mpi-sws.org/later-credits/paper-later-credits.pdf`). As such, this pull request will make Verus "safe" when adding certain new features.

With this pull request, we now require the expenditure of an open-invariant credit, only obtainable in `exec` mode, when opening an invariant. Most code opening invariants doesn't have to change, since we change the `open_atomic_invariant!` and `open_local_invariant!` macros to automatically take care of credits. Each of these macros first creates an open-invariant credit and then uses it to open the invariant.

However, if the developer needs to open an invariant from `proof` mode (with `open_atomic_invariant_in_proof!` or `open_local_invariant_in_proof!`), they need to explicitly deal with open-invariant credits. This is because each of these macros takes a `tracked` `OpenInvariantCredit` as its first parameter. The developer can get one with `let Tracked(credit) = create_open_invariant_credit();`, but this call must be made from `exec` mode.


jaylorch:
An alternative way to deal with nested invariants is to hardcode special consideration for `create_open_invariant_credit`, treating it as non-exec when counting non-atomic and atomic steps in an atomic context. But this seemed hacky to me. What do you think?

jaylorch:
I renamed "later credit" to "open-invariant credit" to make more sense in the Verus context. But maybe I should have just called it a "permission to open an invariant" (i.e., `PermissionToOpenInvariant` instead of `OpenInvariantCredit`, and `obtain_permission_to_open_invariant` instead of `create_open_invariant_credit`). What do you think?

tjhance:
Should we just call 'em later credits?

upamanyus:
It is still possible to allocate `OpenInvariantCredit`s in proof mode by using `Tracked::new(OpenInvariantCredit{})`. E.g. [the proof of false](
https://github.com/upamanyus/verus-examples/blob/706f2164478fa35300f324a04215ef41db7ba31f/false.rs) using an axiomatized model for `dyn` still goes through with these changes. Is there a way to ensure that the *only* place one could get an `OpenInvariantCredit` is via `create_open_invariant_credit()`?

One way I can think is to add an uninterpreted predicate `is_valid(self) -> bool` on `OpenInvariantCredit` and axiomatize that the postcondition of `create_open_invariant_credit` is that `is_valid()` holds on the returned credit. Or maybe just give it a private field? I think that would be comparable to why `PointsTo`s can't be forged.

tjhance:
That's a good catch. The `OpenInvariantCredit` struct needs be marked `#[verifier::external_body]` to prevent the user from constructing them by calling the constructor.

jaylorch:
Yes, thanks, @upamanyus, for catching that, and thanks, @tjhance, for the fix. I've applied it.

jaylorch:
> Should we just call 'em later credits?

I figured that would be fairly confusing to Verus users, since you have to understand a lot about Iris to understand what the word "later" has to do with it.

jaylorch:
Originally, I had thought that by using a semicolon after the definition of `struct OpenInvariantCredit` instead of `{}` that I was preventing anyone from knowing what the fields were. But I guess that's just equivalent to an empty list of fields?

tjhance:
yes it's not very common, but it's just a struct with no fields.

jaylorch:
> An alternative way to deal with nested invariants is to hardcode special consideration for `create_open_invariant_credit`, treating it as non-exec when counting non-atomic and atomic steps in an atomic context. But this seemed hacky to me. What do you think?

I also wasn't sure if such special consideration would be sound. The more we treat `create_open_invariant_credit` differently from a normal `exec` function, the easier it is to imagine someone convincing Verus to ignore it.

source/vstd/invariant.rs
```diff
@@ -233,6 +233,42 @@ declare_invariant_impl!(LocalInvariant);
...
+
+#[doc(hidden)]
+#[cfg_attr(verus_keep_ghost, verifier::proof)]
+#[verifier::external_body]
+pub struct OpenInvariantCredit {}
+
+#[cfg(verus_keep_ghost)]
+#[verifier::external_body]
+#[inline(always)]
+pub fn create_open_invariant_credit() -> Tracked<OpenInvariantCredit>
```
tjhance, line 258: Add a note that it's intentional that this is marked 'exec'.

tjhance (changes requested):

tjhance:
> I also wasn't sure if such special consideration would be sound. The more we treat create_open_invariant_credit differently from a normal exec function, the easier it is to imagine someone convincing Verus to ignore it.

The biggest worry I had was that you could open an atomic invariant block, and then write non-terminating code that generates an infinite number of credits. This issue isn't even specific to credits though - we _always_ have to make sure that an atomic block won't allow non-terminating exec code.

Since termination of exec code is ... hard, we have dead-simple rules. You can't use recursion or loops in an atomic block at all. So it's obvious that you can only generate a finite number of credits.

jaylorch:
> > An alternative way to deal with nested invariants is to hardcode special consideration for create_open_invariant_credit, treating it as non-exec when counting non-atomic and atomic steps in an atomic context. But this seemed hacky to me. What do you think?
> 
> I would prefer this option to avoid the complexity with `open_nested_atomic_invariant!`. The atomicity check is really about heap-modifying atomic operations. All no-op functions, including 'ghost code' and pseudo-ghost code like the credit function, can be ignored.

OK, good point. I've updated it to no longer treat `create_open_invariant_credit` as non-atomic, and thereby to obviate the nested variants of the macros.

source/vstd/invariant.rs
jaylorch: Done.

source/vstd/invariant.rs
```diff
@@ -233,6 +233,44 @@ declare_invariant_impl!(LocalInvariant);
...
+    opens_invariants none
+{
+    Tracked::<OpenInvariantCredit>::assume_new()
+}
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::vstd::invariant::spend_open_invariant_credit"]
+#[doc(hidden)]
+#[inline(always)]
+pub proof fn spend_open_invariant_credit(credit: Tracked<OpenInvariantCredit>) {
```
Chris-Hawblitzel, line 270: It looks like this is a ghost (duplicable) parameter at the moment, because it's a `proof` function paramater that doesn't say `tracked`.  I'd expect either `pub proof fn spend_open_invariant_credit(tracked credit: OpenInvariantCredit)` or (slightly redundandly) `pub proof fn spend_open_invariant_credit(tracked credit: Tracked<OpenInvariantCredit>)`.

jaylorch: Ah, good catch. Fixed.

tjhance: Is there a test case for this? Something that checks the same credit object can't be used twice.

source/rust_verify/example/invariants.rs
```diff
@@ -1,46 +1,46 @@
-#[allow(unused_imports)]
```
utaal-b, line 1: This file appears (mostly?) unchanged but GitHub shows it as completely rewritten. Are these whitespace changes? Could you maintain the existing formatting, so we can keep track of exactly what change was made as result of this?

source/rust_verify_test/tests/atomics.rs
```diff
@@ -1,344 +1,380 @@
-#![feature(rustc_private)]
```
utaal-b, line 1: Same here. It looks like this file was rewritten? Is it again perhaps different newlines?

source/vir/src/modes.rs
```diff
@@ -699,7 +699,16 @@ fn check_expr_handle_mut_arg(
...
                             ai.add_atomic(&expr.span);
                         } else {
-                            ai.add_non_atomic(&expr.span);
+                            // A call to `create_open_invariant_credit` is a no-op, so
+                            // it's fine to include in an atomic block. And it's useful
+                            // to be able to do so, so that we can nest an opening of
+                            // an invariant inside an opening of another invariant. So
+                            // we special-case this call to not treat it as non-atomic.
+                            if function.x.name
+                                != crate::fun!("vstd" => "invariant", "create_open_invariant_credit")
```
utaal-b, line 708: What's the `crate::fun!` macro?

Could this use `verus_items` instead to match this function?

source/vstd/invariant.rs
```diff
@@ -233,6 +233,44 @@ declare_invariant_impl!(LocalInvariant);
...
+#[cfg_attr(verus_keep_ghost, verifier::proof)]
+#[verifier::external_body]
+pub struct OpenInvariantCredit {}
+
+// It's intentional that `create_open_invariant_credit` uses `exec` mode. This prevents
+// creation of an infinite number of credits to open invariants infinitely often.
+#[cfg(verus_keep_ghost)]
+#[verifier::external_body]
+#[inline(always)]
+pub fn create_open_invariant_credit() -> Tracked<OpenInvariantCredit>
```
utaal-b, line 260: You may want a `rustc_diagnostic_item` on this too, to match with it in `modes.rs` (see the other comment).

utaal-b (changes requested):

source/rust_verify/example/invariants.rs
jaylorch: > This file appears (mostly?) unchanged but GitHub shows it as completely rewritten. Are these whitespace changes (or perhaps a line terminator changes)? Could you maintain the existing formatting, so we can keep track of exactly what change was made as result of this?

Fixed.

source/rust_verify_test/tests/atomics.rs
jaylorch: Fixed.

source/vir/src/modes.rs
jaylorch: I don't seem to be able to use `verus_items` from `vir`. This is consistent with what I thought I learned in the architecture part of the Verus all-hands meeting: that `vir` is at a lower layer than `rust_verify`. Is there a trick to using `rust_verify` functions from `vir`?

utaal-b: That's correct, but I'd rather we'd avoid further ad-hoc string-matching in VIR. The common strategy is to still add them to `verus_items`, in the `Vstd` group, and give them a path (as required by the `Vstd` constructor): they should get encoded with that path, which you can then refer to like in this case: https://github.com/verus-lang/verus/blob/f2ed4b35dcd7fd7ad5876d44fa4f1f778deb405e/source/vir/src/def.rs#L782-L790

This way name changes in vstd are less likely to change how things are mapped in VIR.

Let me know if I can help / clarify further.

jaylorch: In `vir`, where can I get the `vstd_crate_name` to pass to this? The `ctxt` I have doesn't seem to have a `vstd_crate_name` field.

utaal-b: It's in the VIR `GlobalCtx`, you may need to add it to the `ctxt` for prune if it's not already reachable from there. I have little time today so I only gave a cursory look at that, let me know if it doesn't make sense / I can assist more.

Chris-Hawblitzel (approved):
Could you also add a test to make sure the credit expression is getting verified, something like:

```
open_atomic_invariant_in_proof!(vstd::pervasive::proof_from_false() => &i => inner => { ... }) // FAILS
```

source/vir/src/modes.rs
jaylorch: OK, thanks for the tip. I've addressed this issue in the latest commit.

jaylorch:
> Could you also add a test to make sure the credit expression is getting verified, something like:
> 
> ```
> open_atomic_invariant_in_proof!(vstd::pervasive::proof_from_false() => &i => inner => { ... }) // FAILS
> ```

Good idea. I've added the test, and it worked.

jaylorch:
Thanks for all the suggestions and fixes, @tjhance, @upamanyus, @utaal-b, and @Chris-Hawblitzel! I think I've addressed all of the issues you raised. Please take a final look and see what you think.

utaal-b (approved):
Consider squashing when merging.

PR #1043, opened 2024-03-21
check internal "rustc" attrs

tjhance:
The rust stdlib has a prevalance of internal "rustc" attributes, which vary greatly in how relevant they are to verification.

This PR includes an initial whitelist of "safe" rustc attributes, and errors if it encounters an attribute that it doesn't recognize. As we verify the stdlib, we'll likely come across some attributes that require special support. For example, `rustc_never_returns_null_ptr` and `rustc_nowind` will require additional proof obligations.

### DIFF
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -1,4 +1,4 @@
-use crate::attributes::{get_ghost_block_opt, get_verifier_attrs, GhostBlockAttr};
+use crate::attributes::{get_ghost_block_opt, GhostBlockAttr};
 use crate::context::BodyCtxt;
 use crate::erase::{CompilableOperator, ResolvedCall};
 use crate::reveal_hide::RevealHideResult;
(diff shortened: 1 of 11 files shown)


source/rust_verify/src/fn_call_to_vir.rs
```diff
@@ -143,8 +143,10 @@ pub(crate) fn fn_call_to_vir<'tcx>(
     }
 
     let f_attrs = bctx.ctxt.tcx.get_attrs_unchecked(f);
-    let f_vattrs = get_verifier_attrs(f_attrs, Some(&mut *bctx.ctxt.diagnostics.borrow_mut()))?;
-    if f_vattrs.internal_get_field_many_variants {
+    if crate::attributes::is_get_field_many_variants(
+        f_attrs,
+        Some(&mut *bctx.ctxt.diagnostics.borrow_mut()),
+    )? {
         return Err(vir::messages::error(
```
utaal-b, line 150: Can you explain why this change is necessary: i.e. why you cannot `get_verifier_attrs`  via `bctx.ctxt` here?

tjhance: Sometimes it would examine a function in the rust standard lib that we support via external_fn_specification.

utaal-b: Can you please add a comment to explain that?

utaal (approved):

PR #1045, opened 2024-03-22
Enable generic trait methods

Chris-Hawblitzel:


source/vir/src/recursive_types.rs
```diff
@@ -722,6 +722,21 @@ pub fn check_traits(krate: &Krate, ctx: &GlobalCtx) -> Result<(), VirErr> {
...
+    //     fn f<A: U>(x: Self, y: Self) -> bool;
+    //     fn g(x: Self, y: Self) -> Self { requires(f(x, y)); };
+    //   }
+    // We take a Dictionary_U as a parameter:
+    //   struct Dictionary_T<Self, A> {
+    //     f: Fn<A>(a: Dictionary_U<A>, x: Self, y: Self) -> bool,
+    //     g: Fn(x: Self, y: Self) -> Self { requires(f(x, y)); },
+    //   }
+    // This adds an edge:
+    //   - T --> U
```
utaal-b, line 736: Where is this already handled?

Chris-Hawblitzel: I just pushed an update that clarifies this comment.  In `expand_call_graph`, the `Add f --> T for any function f with "where ...: T(...)"` creates `f --> U`.

utaal-b (approved):

tjhance (approved):

PR #1052, opened 2024-03-28
support actual raw ptrs 4real

tjhance:
Initial support for `*mut T` and `*const T`.

I have put off supporting these types for a long time because it requires figuring out What Actually Are Pointers In Rust, and leaving behind the kinda-idealized "a pointer is an address" model of PPtr. This was sort of philosophical because I like the idealized model and think it's closer to the Right Thing for greenfield verified code. However, the lack of support for the 'actual' Rust pointer types is increasingly a bottleneck for verifying the standard library, so here we are.

This PR adds VIR support for the `*mut` and `*const`, a model for pointers, and a few elementary operations. Similar to slices and arrays, pointer types are implemented as a `Primitive` type in VIR, with the actual 'view' model and operations defined in vstd. Both `*mut T` and `*const T` are encoded by the same type, and they are distinguished via a decoration. You can cast freely between them. (I confirmed with the Rust opsem team that there isn't actually any semantic difference between the two.)

### The model

OK, I made it sound really intimidating, but when you really dig into it, it kind of isn't.

A pointer consists of three things:

 * An address
 * The "metadata" (the length for DSTs, and the vtable for a dyn type)
 * The _provenance_ (See [the docs](https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance) or [the RFC](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html).)

To handle provenance, we're just going to treat it as an abstract type. Maybe we'll add more axioms about it later, but I think it won't matter that much. I think for a lot of applications you can just use the `*mut T`, treating it like some abstract identifier without really worrying about what its model is.

One mistake I made with PPtr was specifying everything in terms of the address; for example, the model of `PointsTo` used a `usize` address rather than the `PPtr` itself, which means you were _always_ having to get the address of the pointer to do anything with it. For the new API, I rectified this. In the new `PointsTo` object, the pointer is just, you know, a _pointer_, `*mut T`, as it should be.

I didn't actually modify PPtr itself, partly because I wanted to start from a fresh slate, and partly because a lot of code already depends on PPtr. I think once `*mut` is fleshed out, we can probably just verify PPtr in terms of `*mut` anyway.

The only features I added were the basic pointer read and write functionality, same as usual. In the future, we'll implement:

 * Allocation and deallocation
 * PointsToRaw equivalent
 * DST support
 * Pointer arithmetic and address-related operations
 * Strict provenance API

### DIFF
--- a/source/rust_verify_test/tests/raw_ptrs.rs
+++ b/source/rust_verify_test/tests/raw_ptrs.rs
@@ -1,0 +1,191 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] basics verus_code! {
+        use vstd::prelude::*;
+        use vstd::raw_ptr::*;
+
+        fn test_mut_const_casts(x: *mut u8) {
+            let y = x as *const u8;
+            let z = y as *const u8;
+            assert(z == x);
+            assert(z == y);
+        }
+
+        fn test_addr_doesnt_imply_eq(x: *mut u8, y: *mut u8) {
+            assume(x@.addr == y@.addr);
+            assert(x == y); // FAILS
+        }
+
+        fn test_view_does_imply_eq(x: *mut u8, y: *mut u8) {
+            assume(x@ == y@);
+            assert(x == y);
+        }
+
+        fn test_view_does_imply_eq_const(x: *const u8, y: *const u8) {
+            assume(x@ == y@);
+            assert(x == y);
+        }
+
+        fn test_null() {
+            let x: *const u8 = core::ptr::null();
+            let y: *mut u8 = core::ptr::null_mut();
+            assert(x == y);
+            assert(x@.addr == 0);
+        }
+
+        fn test_manipulating(x: *mut u8, Tracked(pt): Tracked<PointsTo<u8>>) {
+            let tracked mut pt = pt;
+
+            assume(x == pt.ptr());
+            assume(pt.is_uninit());
+
+            ptr_mut_write(x, Tracked(&mut pt), 20);
+
+            assert(x == pt.ptr());
+            assert(pt.is_init());
+            assert(pt.value() == 20);
+
+            let val = ptr_ref(x, Tracked(&pt));
+
+            assert(val == 20);
+
+            let val = ptr_mut_read(x, Tracked(&mut pt));
+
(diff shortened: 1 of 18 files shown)


source/vstd/raw_ptr.rs
```diff
@@ -0,0 +1,296 @@
...
+pub broadcast proof fn axiom_ptr_mut_from_data<T>(data: PtrData)
+    ensures
+        (#[trigger] ptr_mut_from_data::<T>(data))@ == data,
+{
+}
+
+#[verifier::external_body]
+pub broadcast proof fn ptrs_mut_eq<T>(a: *mut T, b: *mut T)
+    requires
+        #[trigger] a@ == #[trigger] b@,
```
Chris-Hawblitzel, line 182: This leads to a quadratic number of quantifier instantiations.  There's a standard trick to reduce this to a linear number of instantiations based on SMT uninterpreted function equality:

```
pub spec fn reverse_view_mut<T>(p: PtrData) -> *mut T;
pub spec fn reverse_view_const<T>(p: PtrData) -> *const T;

#[verifier::external_body]
pub broadcast proof fn ptrs_mut_eq<T>(a: *mut T)
    ensures
        reverse_view_mut::<T>(#[trigger] (a@)) == a,
{
}

#[verifier::external_body]
pub broadcast proof fn ptrs_eq<T>(a: *const T)
    ensures
        reverse_view_const::<T>(#[trigger] (a@)) == a,
{
}
```

source/rust_verify_test/tests/raw_ptrs.rs
```diff
@@ -0,0 +1,170 @@
...
+            let y = x as *const u8;
+            let z = y as *const u8;
+            assert(z == x);
+            assert(z == y);
+        }
+
+        fn test_addr_doesnt_imply_eq(x: *mut u8, y: *mut u8) {
+            assume(x@.addr == y@.addr);
+            assert(x == y); // FAILS
+        }
```
Chris-Hawblitzel, line 21: If I understand correctly, the following test should fail as well (and it does):

```
fn test(x: *const u8, y: *const u8) {
    if x == y {
        assert(x == y);
    }
}
```

Could this be added to the tests?

Chris-Hawblitzel (approved):

PR #1053, opened 2024-03-30
Discuss configuring VS Code for Verus development

### LINKED ISSUE #1050: Updating a test requires rebuilding everything
If I run Verus tests, then make a small change to a single test, then test again, it takes minutes. It looks like nearly all of Verus is getting rebuilt even though all that's changed is the test file.

For instance, if I run `vargo test -p rust_verify_test --test open_invariant`, then edit `rust_verify_test/tests/open_invariant.rs`, then run `vargo test -p rust_verify_test --test open_invariant` again, it takes several minutes to start the test, during which it displays informational messages suggesting it's compiling a huge variety of crates.

jaylorch:
Fixes #1050

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,10 +37,11 @@ To automatically apply these suggestions to the source code, type:
 vargo fmt
 ```
 
-If you want to edit the source code of Verus in VS Code, copy
-`.vscode/settings.json.template` to `.vscode/settings.json` and then edit that
-file to customize it for your local setup. For instance, delete `[.exe]` if
-you're not using Windows and change it to `.exe` if you are.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -37,7 +37,13 @@ To automatically apply these suggestions to the source code, type:
 vargo fmt
 ```
 
-### Running verusfmt manually
+If you want to use VS Code to edit the source code of Verus, copy
```
utaal, line 40:
```suggestion
If you want to use VS Code with `rust-analyzer` to edit the source code of Verus, copy
```

utaal (approved):
Thanks! I added a small clarification as a suggested change.

PR #1056, opened 2024-04-03
PCM updates

jaylorch:
Slight naming changes in `pcm.rs`, and a new library of utility functions in `pcm_lib.rs`

### DIFF
--- a/source/vstd/pcm.rs
+++ b/source/vstd/pcm.rs
@@ -6,7 +6,7 @@ verus! {
 /// Interface for ghost state that is consistent with the common
 /// presentations of partially commutative monoids (PCMs) / resource algebras.
 ///
-/// For applications, I generally advise using the
+/// For applications, the general advice is to use the
 /// [`tokenized_state_machine!` system](https://verus-lang.github.io/verus/state_machines/),
 /// which lets you focus on updates and invariants rather than composition.
 ///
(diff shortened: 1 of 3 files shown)


jaylorch:
As I've been using it, I've thought it might be simpler to not have partially commutative semigroups at all and just have partially commutative monoids. It's not a big deal to add a unit to any partially commutative semigroup to make it a PCM. If you agree, I'll update this to just use PCM.

jaylorch:
> As I've been using it, I've thought it might be simpler to not have partially commutative semigroups at all and just have partially commutative monoids. It's not a big deal to add a unit to any partially commutative semigroup to make it a PCM. If you agree, I'll update this to just use PCM.

I included some examples, and then committed a change showing how they get simplified by removing `PCSemigroup`.

tjhance (approved):
Iris uses RAs without necessarily a unit. I'm not entirely sure why, I thought it might be a good idea to do the same, but I don't feel particularly strongly about it.

upamanyus:
I'm not sure why Iris has RAs without unit either, other than the small convenience of not needing to declare a unit element on each RA.

RAs without unit should not be necessary. At a theoretical level, RAs without units can be systematically extended to have units nicely with a free-forgetful adjunction (Iris proof of this [here](https://github.com/mit-pdos/perennial/blob/master/src/program_proof/verus/ucmra_cmra_adjunction.v)). Practically, this can be done by wrapping a non-unital RA with an "option" RA transformer, which uses None as a unit.

Iris also does add a unit to all the RAs when constructing the global resource algebra that actually gets plugged into the base logic (see `ResF` in section 7.5 of https://plv.mpi-sws.org/iris/appendix-4.2.pdf).

More practically, proofs of programs using Iris typically only use the existing library of RAs and RA combinators. At that level of abstraction, (e.g. using a key-value map RA or taking the product of two other RAs), the burden of having to include a unit in every RA would not make an appearance. It seems reasonable to keep Verus conceptually simpler by excluding "partial commutative semigroups".

PR #1058, opened 2024-04-04
Support specifications for external traits

Chris-Hawblitzel:


Chris-Hawblitzel:
For example, suppose there is an external trait T:

```
#[verifier::external]
trait T {
    fn f(&self, q: &Self, b: bool) -> usize;
    type X;
}
```

We can add a specification to T as follows:

```
trait ExT {

    fn f(&self, q: &Self, b: bool) -> (r: usize)
        requires b
        ensures r > 7
        ;
    type X;
}
```

Here, the specially named associated type `ExternalTraitSpecificationFor` specifies which trait is being specified.  With this specification in place, verified code can use the trait T and the members f and X of T, and any uses of T::f will be verified based on the specification provided by ExT::f.  For example:

```
        fn test<A: T>(a: &A) {
            let i = a.f(a, true);
            assert(i > 7);
            let i = a.f(a, false); // FAILS
        }
```

The external trait specification is not required to include all members of the trait.  Members that are not included are not accessible to verified code.

I considered using inheritance for this instead of the specially named `ExternalTraitSpecificationFor` associated type, but then it's not clear which of multiple inherited traits (e.g. Sized) is the trait being specified.  I also considered using individual `#[verifier::external_fn_specification]` specifications for each function member of the trait, but then it's not clear that the functions will have the appropriate context (e.g. the right Self and self declarations and the right associated types and available references to other members in the specs).

utaal:
Are _all_ reachable implementations of the trait assumed to uphold the trait specification?
For example:
* crate `A` (unverified) defines trait `T`
* crate `B` (unverified) defines `S` which implements `T`
* crate `C` verified, imports `A` and `B`: does this assume that `S` implements `T` according to `T`'s external specification?

Chris-Hawblitzel:
> Are _all_ reachable implementations of the trait assumed to uphold the trait specification?

It's an even bigger assumption than that.  All reachable and unreachable implementations are assumed to uphold the trait specifications.  For example, if we verify a crate with `pub fn test<A: T>(...)`, we assume that whatever type instantiates A will uphold the specification for T, even if this type comes from an unverified crate that hasn't been written yet.  In other words, this is a contract on both current and future unverified code that must be satisfied to correctly link unverified code with verified code.

Chris-Hawblitzel:
> this is a contract on both current and future unverified code that must be satisfied to correctly link unverified code with verified code.

This is true not just for trait specifications applied to external traits, but also to public traits declared directly in verified code, since we have no way of stopping future unverified code in other crates from also implementing these traits.

tjhance (approved):

tjhance:
Thanks for taking care of this!

utaal (approved):
Sorry for the latency. Considering the breadth of the assumption made when one writes this, I was thinking how we could nudge users towards using this sparingly, or if we could put other safeguards in place. Those should not be blockers at this stage though.

What traits are we planning to specify this way in the near term? (In addition to those already present in the PR.)

Chris-Hawblitzel:
> What traits are we planning to specify this way in the near term?

The main short-term use is just to declare the existence of traits that are used in bounds, without declaring any of the members of the trait.  In particular, we need to declare marker traits like Copy, Sized, Sync, etc.

The only current specifications of trait *members* are for Clone, for which we declare the function clone to have no requires and no ensures (although specific implementations of clone can add more ensures) and for `core::ptr::Pointee`, which has an associated type used in raw_ptr.rs.  Iterator might be the next trait to specify members of.

Also, we currently have specifications for specific implementations of trait members that we haven't declared; it's not clear whether we should allow this without first declaring the trait and trait member:
  - Try for Result::branch
  - Try for Option::branch
  - FromResidual for Result::from_residual
  - FromResidual for Option::from_residual
  - CheckedAdd, etc. for num.rs
  - Index for Vec
  - Iterator for Range::next

There are also our own traits, like Structural, SpecOrd, SpecAdd, etc. to look at.

PR #1059, opened 2024-04-04
Avoid duplicated qids/skolemids in datatype height axioms

yizhou7:
I was doing some experiments with QI, and I noticed that all datatype height axioms share the same `:qid prelude_datatype_height`, and `:skolemid skolem_prelude_datatype_height`. Having duplicated qids for different quantifiers complicates the analysis a bit, so I am proposing an update here.

### DIFF
--- a/source/vir/src/prelude.rs
+++ b/source/vir/src/prelude.rs
@@ -686,6 +686,8 @@ pub(crate) fn datatype_height_axiom(
     let height = str_to_node(HEIGHT);
     let height_lt = str_to_node(HEIGHT_LT);
     let height_rec_fun = str_to_node(HEIGHT_REC_FUN);
+    let qid = str_to_node(format!("prelude_datatype_height_{}", field).as_str());
+    let skolem_id = str_to_node(format!("skolem_prelude_datatype_height_{}", field).as_str());
     let field = str_to_node(field.as_str());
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel (approved):
This looks fine.  Do you also want to include the type name in case multiple types have the same field?

yizhou7:
> This looks fine. Do you also want to include the type name in case multiple types have the same field?

@Chris-Hawblitzel Thanks for bringing that up. I am not sure if I am reading it correctly, but it seems like the datatype name is pre-pended to the field at when calling `datatype_height_axioms`?

https://github.com/verus-lang/verus/blob/eb988fe4a8600e012822fa0d6b0b5d94a14f0c23/source/vir/src/datatype_to_air.rs#L463

As a result the current quantifier (on this branch) looks like the following:

```
(assert
 (forall ((x adts!Vehicle2.)) (!
   (=>
    (is-adts!Vehicle2./Car x)
    (height_lt (height (Poly%adts!Car. (adts!Vehicle2./Car/0 x))) (height (Poly%adts!Vehicle2.
       x
   ))))
   :pattern ((height (Poly%adts!Car. (adts!Vehicle2./Car/0 x))))
   :qid prelude_datatype_height_adts!Vehicle2./Car/0
   :skolemid skolem_prelude_datatype_height_adts!Vehicle2./Car/0
)))
```

PR #1062, opened 2024-04-09
Broadcast-use formatting style change

jaybosamiya:
Minor formatting-only style change, that adds newlines to `broadcast use` groups. See https://github.com/verus-lang/verusfmt/pull/55 for details on this upcoming change in verusfmt.

The total actual changes in Verus end up being just 2 new newlines, and thus was considered a reasonable breakage for better style. This PR makes the necessary updates, as well as bumps the minimum verusfmt version number.

This PR is blocked by the new release of verusfmt, but should be merged in immediately after the new version of verusfmt releases, in order to prevent spurious formatting CI issues in other PRs. This is also why the CI is _expected_ to fail on this, until the release happens, at which point the CI will start to succeed and it can be merged.

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -971,6 +971,7 @@ impl<A> Seq<Seq<A>> {
             self.len() == 1 ==> self.flatten() == self.first(),
     {
         broadcast use Seq::add_empty_right;
+
         if self.len() == 1 {
             assert(self.flatten() =~= self.first().add(self.drop_first().flatten()));
         }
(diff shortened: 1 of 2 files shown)


jaybosamiya:
@utaal-b (GitHub wouldn't let me request the `-b` as reviewer for the PR fwiw), I'll wait for you to confirm that the PR looks reasonable, once you say it does, then I'll do the verusfmt release and then merge this in, so that I can take care of the coordination locally.

utaal:
Thank you for working on this. I'm a bit uneasy about verusfmt making changes to add/remove empty lines in blocks (I believe rustfmt never does this). Also see https://github.com/verus-lang/verusfmt/pull/55#issuecomment-2046793855

jaybosamiya:
I agree that rustfmt (with some exceptions) does not mess with empty lines in blocks. 

However, the design of verusfmt essentially means that we don't have a clean way to handle those empty blocks (more details in that thread). This is not related to this specific change, but generally how verusfmt itself works right now. This particular just changes verusfmt's opinions on `broadcast use`, but otherwise leaves the rest of its decisions regarding lines unchanged.

utaal (approved):

PR #1063, opened 2024-04-10
vstd: bitwise mask properties

mmcloughlin:
This PR extends `vstd::arithmetic::power2` and `vstd::bits` modules with proofs
relating to bitwise masks.

A spec for the low bits mask is added to the `power2` module, along with proofs of
basic properties.

In addition, the `vstd::bits` module now contains a proof that bitwise and with a mask
is equivalent to modulo by the corresponding power of 2.


source/vstd/arithmetic/power2.rs
```diff
@@ -90,6 +94,27 @@ pub proof fn lemma_pow2_auto()
     }
 }
 
+/// Proof relating 2^e to 2^(e-1) for a specific e.
+pub proof fn lemma_pow2_unfold(e: nat)
```
mmcloughlin, line 98: Is "unfold" a good name for this?

I borrowed it from:
https://github.com/verus-lang/verus/blob/1604eeabe5e625b3386733e7649f4914b533f410/source/rust_verify/example/power_of_2.rs#L46

source/vstd/arithmetic/power2.rs
```diff
@@ -133,33 +158,6 @@ pub proof fn lemma_pow2_strictly_increases_auto()
     }
 }
 
-/// Proof that, for the given positive number `e`, `(2^e - 1) / 2 == 2^(e - 1) - 1`
-pub proof fn lemma_pow2_mask_div2(e: nat)
```
mmcloughlin, line 137: I checked that this is not used in the `verus` repository itself, but it just occurred to me that it might be by an external verus user. How large is the verus ecosystem at this point, can we check all/most of the projects using it? What sort of stability guarantees does verus offer at this point, if any?

My thought was that if we knew it's unused, the same lemma might be more naturally stated in terms of the `mask(n)` spec. See `lemma_mask_div2{,_auto}` below.

source/vstd/arithmetic/power2.rs
```diff
@@ -239,4 +237,168 @@ pub proof fn lemma2_to64()
...
 }
 
+/// Mask with low n bits set.
+pub open spec fn mask(n: nat) -> nat {
+    (pow2(n) - 1) as nat
+}
+
+/// Proof relating the n-bit mask to a function of the (n-1)-bit mask, for given
+/// n.
+pub proof fn lemma_mask_unfold(n: nat)
```
mmcloughlin, line 247: Same question about "unfold" terminology here.

source/vstd/arithmetic/power2.rs
```diff
@@ -239,4 +237,168 @@ pub proof fn lemma2_to64()
     ) by(compute_only);
 }
 
+/// Mask with low n bits set.
+pub open spec fn mask(n: nat) -> nat {
```
mmcloughlin, line 241: Should we be more explicit here? `low_bits_mask`?

Should all these be in a separate `mask.rs` file?

source/vstd/arithmetic/power2.rs
```diff
@@ -239,4 +237,168 @@ pub proof fn lemma2_to64()
...
+        forall|n: nat| #![trigger mask(n)] n > 0 ==> mask(n) / 2 == mask((n - 1) as nat),
+{
+    assert forall|n: nat| n > 0 implies #[trigger] mask(n) / 2 == mask((n - 1) as nat) by {
+        lemma_mask_div2(n);
+    }
+}
+
+/// Proof establishing the concrete values of all masks of bit sizes from 0 to
+/// 32, and 64.
+pub proof fn lemma_mask_values()
```
mmcloughlin, line 327: Here I copied the 0-32 and 64 range from `lemma2_to64()`.

I wonder if we should include all `0-64`? I also wondered if we could macro-generate a lemma like this with something like https://docs.rs/seq-macro/latest/seq_macro/, though this would require incurring a dependency or writing our own proc macro. (In practice I just wrote a quick Python script and copied it in.)

I'm also not sure about naming: `lemma2_to64` is an unusual name, and `lemma_mask_values` is maybe too vague.

source/vstd/bits.rs
```diff
@@ -204,3 +208,112 @@ lemma_shl_is_mul!(lemma_u64_shl_is_mul, lemma_u64_shl_is_mul_auto, lemma_u64_pow
...
+                }
+            }
+        }
+
+        #[doc = "Proof that for all natural n and x of type "]
+        #[doc = stringify!($uN)]
+        #[doc = ", and with the low n-bit mask is equivalent to modulo 2^n."]
+        pub proof fn $name_auto()
+            ensures
+                forall|x: $uN, n: nat| n < <$uN>::BITS ==> #[trigger] (x & (mask(n) as $uN)) == (x % (pow2(n) as $uN)),
```
mmcloughlin, line 274: I was unsure how to set triggers for this. At first I tried both sides of the equality, but that was disallowed due to an error about not being able to configure triggers for both arithmetic and bitwise operations on `x`.

mmcloughlin:
I wonder whether it would be better to replace these proofs (and those added in #1054) with proof-by-cases instead?

I quickly threw together an [example of what it might look like](https://gist.github.com/mmcloughlin/96128733df61adbb1d95603ed21220eb). I would imagine we'd prefer not to use a Python script in practice, but we could replace it with a macro-based approach or just a Rust version of the Python code-generator.
@jaybosamiya has already [shown how we might use macros](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=342c5128425154cd055303fd795b14ea).

I guess this is somewhat a matter of taste. I'm just curious what people think.

source/vstd/arithmetic/power2.rs
parno: This lemma seems very specific.  In the place where it's used, would revealing `pow2` do the trick instead?

source/vstd/arithmetic/power2.rs
parno: > How large is the verus ecosystem at this point, can we check all/most of the projects using it?

We don't yet have a system for that, and even when it's up and running, it will only have partial coverage.

> What sort of stability guarantees does verus offer at this point, if any?

No official guarantees yet.

source/vstd/arithmetic/power2.rs
parno: I'd be inclined to put them in `bits.rs`, since it seems like the primary use case is reasoning about bit-manipulation.  And yes, I would use something more explicit, since there are many kinds of mask one might want.  I think `low_bits_mask` or `lsb_mask` could work.

source/vstd/arithmetic/power2.rs
parno: unfold seems fine to me.

source/vstd/arithmetic/power2.rs
parno: If you think people will want mask values in the 33-63 range, feel free to add the full set.  It does seem like something that should be done with a macro, however.  I'd suggest asking on #verus about taking a dependency on another crate.  For name, I'd suggested appending `_to64` to whatever name we end up with for `mask`.

source/vstd/bits.rs
parno: I think triggering on the LHS is preferable.  It seems unlikely that you'll care about this relation just because you perform a mod operation w.r.t. pow2, but it seems quite likely that you'll care if you're doing the bitwise operation.

source/vstd/bits.rs
```diff
@@ -204,3 +208,112 @@ lemma_shl_is_mul!(lemma_u64_shl_is_mul, lemma_u64_shl_is_mul_auto, lemma_u64_pow
...
+        {
+            assert forall|x: $uN, n: nat| n < <$uN>::BITS implies #[trigger] (x & (mask(n) as $uN)) == (x % (pow2(n) as $uN)) by {
+                $name(x, n);
+            }
+        }
+
+        // Helper lemma breaking an and operation into the low bit and the rest.
```
parno, line 281: Can we replace "an and operation" with "a bitwise-and"?  I find the former hard to parse.

utaal:
Please feel free to continue to iterate on it, but can I please ask you to hold off on merging this for the time being?
I have a somewhat pervasive vstd change that I am trying to land but it's being delayed by (I believe unrelated) CI issues.
Thanks!

mmcloughlin:
> Please feel free to continue to iterate on it, but can I please ask you to hold off on merging this for the time being?

Yes of course. There isn't any urgency to land this, and I can work on @parno's comments in the meantime.

parno:
@utaal I believe the changes to vstd are complete at this point, right?

utaal:
They are. Thanks for checking!

source/vstd/arithmetic/power2.rs
mmcloughlin: I tried to extend this to all values 0-64, however I appear to have triggered a bug in the process #1097 #1096. Perhaps leave this as-is for now?

source/vstd/bits.rs
mmcloughlin: Okay, sounds good thanks.

source/vstd/bits.rs
mmcloughlin: Yes, that's clearer. Done.

mmcloughlin:
@parno I've updated this to account for your comments. Please take another look (not urgent). Thanks!

I would still be interested in your thoughts on taking the alternative approach described in https://github.com/verus-lang/verus/pull/1063#issuecomment-2049644890. It seems to me that could be more verbose but ultimately simpler.

source/vstd/arithmetic/power2.rs
parno: Okay, that's fine.

parno:
Regarding proof style, I'm not sure we want to have a big collection of lemmas with a gigantic list of constants.  If we can do it via macros (ideally with one or two generic ones), then I'm not opposed to using "brute force" proofs if they save human effort.

mmcloughlin:
> Regarding proof style, I'm not sure we want to have a big collection of lemmas with a gigantic list of constants. If we can do it via macros (ideally with one or two generic ones), then I'm not opposed to using "brute force" proofs if they save human effort.

Yes, agreed that macros or some other code-generation would be necessary if we were to do it that way instead.

What do you think about landing this PR as-is for now?

parno (approved):
Yes, looks good to me.

mmcloughlin:
> Yes, looks good to me.

Great! Assuming it's good to go, please could you squash merge it?

PR #1071, opened 2024-04-13
disable z3 literal sorting by default

yizhou7:
Following up on our discussion on Friday, 
this PR intends to improve stability by disabling `rewriter.sort_disjunctions` in z3. 

Currently, this is also done in F*:
https://github.com/FStarLang/FStar/blob/ae3c85a3ef7aedb15b8be0e04054b3c34c3c5e73/src/smtencoding/FStar.SMTEncoding.Z3.fst#L518

We had some discussions on this flag previously:
https://github.com/verus-lang/verus/discussions/147#discussioncomment-6630517

I have not tried this on the existing Verus projects. 
If we think more data is needed, please feel free to convert this into a draft. 

Update:

Looks like the change would cause some of `nonlinear_arith` (z3) to fail:

`error: function body check: Resource limit (rlimit) exceeded; consider rerunning with --profile for more details`

I have updated the test cases, which likely were in an unstable state to begin with.

### DIFF
--- a/source/air/src/context.rs
+++ b/source/air/src/context.rs
@@ -233,6 +233,7 @@ impl Context {
             self.set_z3_param_u32("smt.arith.solver", 2, true);
             self.set_z3_param_bool("smt.arith.nl", false, true);
             self.set_z3_param_bool("pi.enabled", false, true);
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (approved):
Thanks for this!

PR #1074, opened 2024-04-16
For external items, check def_id_to_vir_path_option before proceeding

Chris-Hawblitzel:


utaal (approved):
@tjhance is particularly busy, and this is a small change that can unblock a user, so unless you specifically want to wait for him, please go ahead and merge.

Issue #1070, opened 2024-04-12
Deep view unsupported for &[u8]


gancherj:
This works:
```
spec fn test(xs : &[u8]) -> Seq<u8> { 
    xs.view()
}
```
but this doesn't:
```
spec fn test(xs : &[u8]) -> Seq<u8> { 
    xs.deep_view()
}
```
as it returns the error 
```
  ^^^^^^^^^ method cannot be called on `&'a [u8]` due to unsatisfied trait bounds
  
   = note: the following trait bounds were not satisfied:
           `[u8]: std::marker::Sized`
           which is required by `&[u8]: vstd::view::DeepView`
           `[u8]: vstd::view::DeepView`
           which is required by `&[u8]: vstd::view::DeepView`
```

Is this simply a matter of adding the function to `SliceAdditionalSpecFns`?

jaybosamiya:
Adding this to `source/vstd/view.rs` works:
```diff
modified   source/vstd/view.rs
@@ -1,4 +1,6 @@
 #[allow(unused_imports)]
+use crate::prelude::*;
+#[allow(unused_imports)]
 use builtin::*;
 #[allow(unused_imports)]
 use builtin_macros::*;
@@ -40,6 +42,16 @@ impl<A: DeepView> DeepView for &A {
     }
 }
 
+impl<T: DeepView> DeepView for [T] {
+    type V = Seq<T::V>;
+
+    #[verifier::inline]
+    open spec fn deep_view(&self) -> Self::V {
+        let v = self.view();
+        Seq::new(v.len(), |i: int| v[i].deep_view())
+    }
+}
+
 #[cfg(feature = "alloc")]
 impl<A: View> View for alloc::boxed::Box<A> {
     type V = A::V;
```

Chris-Hawblitzel:
@jaybosamiya Thanks!  I turned this into pull request https://github.com/verus-lang/verus/pull/1078 .

Chris-Hawblitzel:
Implemented by https://github.com/verus-lang/verus/pull/1078

PR #1078, opened 2024-04-22
Implement View and DeepView in slice and array

Chris-Hawblitzel:
This is related to https://github.com/verus-lang/verus/issues/1070 .


Chris-Hawblitzel:
At the moment, verusfmt complains about the trait syntax `pub trait ArrayAdditionalSpecFns<T>: View<V = Seq<T>> { ... }`.

source/vstd/slice.rs
```diff
@@ -10,18 +10,30 @@ pub use super::std_specs::vec::VecAdditionalSpecFns;
...
+impl<T: DeepView> DeepView for [T] {
+    type V = Seq<T::V>;
+
+    open spec fn deep_view(&self) -> Seq<T::V> {
+        let v = self.view();
+        Seq::new(v.len(), |i: int| v[i].deep_view())
+    }
+}
 
+#[verusfmt::skip] // this doesn't help
```
jaybosamiya: Regarding this line: the `#[verusfmt::skip]` attribute doesn't work if a file is not parseable by verusfmt; I've opened an issue to track supporting full-file skips at https://github.com/verus-lang/verusfmt/issues/59 

As for the blocker on formatting, that fix (https://github.com/verus-lang/verusfmt/pull/58) has been merged and released. I gave the fmt CI on this PR a kick, and its current complaint is the number of spaces on this line 🙃  If you remove the skip, the formatting succeeds:

```suggestion
```

source/vstd/slice.rs
```diff
@@ -10,18 +10,30 @@ pub use super::std_specs::vec::VecAdditionalSpecFns;
...
     spec fn view(&self) -> Seq<T>;
+}
+
+impl<T: DeepView> DeepView for [T] {
+    type V = Seq<T::V>;
+
+    open spec fn deep_view(&self) -> Seq<T::V> {
+        let v = self.view();
+        Seq::new(v.len(), |i: int| v[i].deep_view())
+    }
```
jaybosamiya, line 25: Just making a quick note here that @y1ca1 had some concerns (during a separate discussion) about this adding more triggers than explicitly reimplementing `deep_view` on each of the vstd data structures individually.

Personally I think that implementing `deep_view` separately on each of (say) `[u8]`, `[u32]`, `[u64]` and so on is not really a good idea (because it doesn't scale and would prevent usage on user-defined types without them manually _also_ implementing deepview on `[Foo]`), and that we should just go with the current approach.

However, @y1ca1, if I've misunderstood what you meant, please do comment here.

jaybosamiya (approved):
Overall looks good to me, thanks!  Minor comment added about a possible concern (I don't think it should be a blocker though).  Also, I see you've already removed the now-unnecessary `verusfmt::skip`, thanks!

source/vstd/slice.rs
y1ca1: Thanks for mentioning this. What I meant by "explicitly reimplementing" `deep_view` is to provide separate implementation on each of the methods for the generic slice type`[T]` whose specs are described in `deep_view`. This would arguably save a lot of triggers if one uses `deep_view` ubiquitously, but yeah, definitely less ergonomic and harder to maintain.

source/vstd/slice.rs
```diff
@@ -39,6 +50,27 @@ pub exec fn slice_index_get<T>(slice: &[T], i: usize) -> (out: &T)
     &slice[i]
 }
 
+////// Len (with autospec)
+pub open spec fn spec_slice_len<T>(slice: &[T]) -> usize;
+
+// This axiom is slightly better than defining spec_slice_len to just be `slice@.len() as usize`
+// (the axiom also shows that slice@.len() is in-bounds for usize)
+#[verifier(external_body)]
+pub broadcast proof fn axiom_spec_len<T>(slice: &[T])
```
utaal, line 59: This will now need to be added to the `vstd_default` broadcast group to be visible to users.

utaal (approved):

gancherj:
Hi! Is it possible to get this PR merged this week? Thanks

PR #1085, opened 2024-04-25
Support embedding 'builtin' and 'vstd' as modules in order to support verifying core

tjhance:
* Add a 'cfg' option, `verus_verify_core` for vstd and builtin to use in order to know how to import things
 * Rewrite most imports to be relative (i.e., with `super`)
 * In builtin_macros, use the new cfg option to determine whether to use `crate::builtin` or `::builtin`
 * Add a verus flag `--is-core`, which then sets the `verus_verify_core` cfg option
   * Though somewhat orthogonal, I also added an `--is-vstd` flag in order to clarify the distinction between "i am vstd" and "no vstd". I also replaced the 'vstd' boolean with an enum.
 * Fix an issue in hir_hide_reveal_rewrite.rs (see comment)


utaal:
Can you provide a little more context on the general strategy for this?
How will the crates be split, loaded, exported, and verified?

What is the intended effect of `verus_verify_core`, conceptually?

tjhance:
Currently, to verify a crate, we need to import 'builtin' and 'vstd'; however, the crate 'core' cannot import these crates since they themselves depend on 'core'. Therefore, to verify core, we either need to:

 * Create a copy of core, call it 'core2', and we have to carefully distinguish between core and core2 at all times. And we wouldn't be able to do anything about the fact that all the lang_items are already defined in core, so we couldn't redefine them in core2. Therefore, we would have to make a bunch of changes to core2 to replace certain parts with the ones already defined in core.
 * Find a way to get builtin and vstd into core. Since we can't import them as crates, we have to embed them as modules.

I think the first one is just untenable, the number of changes would be significant and they would be difficult to keep track of. Hence, this PR allows a special mode where builtin and vstd are embedded as modules. This is a mode that should only be necessary for core. With any other crate (e.g., alloc or std) it should be possible to just import builtin and vstd like normal.

When it comes to the _exporting_ of verified core and all its specifications, I think there will be more things to figure out, but at this point I'd rather cross that bridge when we get to it.

utaal:
One more question: mechanically, how is the import of `builtin` and `vstd` into `core` accomplished, at a high level?

tjhance:
I think this test should answer your question: https://github.com/verus-lang/verus/blob/bc67fb7e8e09cead14933b61ce6a331fd4fb1ca6/source/rust_verify_test/tests/basic.rs#L645

utaal (approved):
Thanks for the pointer!

PR #1091, opened 2024-04-29
support 'spec const int'

tjhance:
I couldn't think of another way to do this.


tjhance:
closing as chris has a differnt soln

PR #1092, opened 2024-04-30
Implement integer spec ops to help --compile on spec const items

Chris-Hawblitzel:
This allows using `--compile` with spec const declarations like:

```
pub spec const Y: int = 20 - 10;
```

Previously, this wasn't supported because the integer operations were implemented as `unimplemented!()`.  Now, the operations have dummy implementations (whose exact definitions don't matter, since we don't actually care about the compiled bodies of spec constants).  This uses the `impl const` nightly feature to allow the trait-based operations to be used in consts.


utaal:
Just to check, why did the operations being `unimplemented!()` result in failed compilation?

source/builtin/src/lib.rs
```diff
@@ -11,6 +11,8 @@
     register_tool(verus),
     register_tool(verifier)
 )]
+#![cfg_attr(verus_keep_ghost, feature(const_trait_impl))]
+#![cfg_attr(verus_keep_ghost, feature(effects))]
```
utaal, line 15: What's this feature?

Chris-Hawblitzel:
> Just to check, why did the operations being `unimplemented!()` result in compilation failure?

For the purpose of type checking, `pub spec const` is treated as an actual Rust `const` when using `--compile` (e.g. when generating libraries like vstd that client code will link to a type check against).  Rust complains if a `const` declaration can't be compiled.  For example, it complains if the const declaration calls a function whose implementation is `unimplemented!()`.

source/builtin/src/lib.rs
Chris-Hawblitzel: > What's this feature? Rather, why is it needed for this PR?

In the spirit of Rust's regular arithmetic ops, the spec arithmetic ops are defined as trait members (e.g. trait `SpecAdd` has member `spec_add`).  To support declarations like `pub spec const Y: int = 20 - 10;`, implementations of these members need to be declared `const`.  Otherwise, Rust complains that they can't be used in a const declaration like `pub spec const Y: int = 20 - 10;`.  As far as I can tell, Rust does not make it easy to declare implementations of trait functions as `const`.  The best I could find was to use the two features `const_trait_impl` and `effects` (in which you declare traits `#[const_trait]` and impls `impl const`) even though all I need is to declare particular member function implementations const; I don't actually need the generality of `const_trait_impl` and `effects`.  If there's an easier way to do this, let me know.

tjhance (approved):

source/builtin/src/lib.rs
utaal: Thanks for the write-up: can you check the progress of those two features and see if it's likely they'll go forward? We don't want to pick up more nightly features if they may be experiments or don't have a path to stabilization.

PR #1113, opened 2024-05-10
handle integer range bounds for const type params

### LINKED ISSUE #1065: Const generic needs bounds
Hi all,

Is this a bug? 
```
proof fn test<const N : usize>() {
        assert (0 <= N); // fails to verify
        assert (N <= usize::MAX); // this one too
 }
```

The following variations also fail to verify:
- Adding `global size_of usize == 8;`
- Changing `usize` to `u64`

tjhance:
fixes #1065

### DIFF
--- a/source/rust_verify_test/tests/generics.rs
+++ b/source/rust_verify_test/tests/generics.rs
@@ -51,3 +51,41 @@ test_verify_one_file! {
         }
     } => Err(e) => assert_one_fails(e)
 }
+
+test_verify_one_file! {
+    #[test] const_generics_int_ranges verus_code! {
+        proof fn test<const N : u8>() {
+            assert (0 <= N);
+            assert (N <= 255);
+        }
+
+        proof fn test2<const N : u8>() {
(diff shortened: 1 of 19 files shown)


source/rust_verify_test/tests/generics.rs
```diff
@@ -51,3 +51,65 @@ test_verify_one_file! {
...
+
+test_verify_one_file! {
+    #[test] const_generics_int_ranges verus_code! {
+        proof fn test<const N : u8>() {
+            assert (0 <= N);
+            assert (N <= 255);
+        }
+
+        proof fn test2<const N : usize>() {
+            assert (0 <= N); // fails to verify
```
Chris-Hawblitzel, line 63: The comment still says it fails, but it succeeds now.

Chris-Hawblitzel: Can we also have a simple failing test, like `assert(N < 255)` for `N: u8`?

Chris-Hawblitzel (approved):

PR #1115, opened 2024-05-15
Ensure is_sized in global layout lemma

ouuan:
This could be useful for satisfying the precondition of `PointsToRaw::into_typed`, saving a `layout_for_type_is_valid`/`size_of`.

Now users can write things like `global size_of int == 1` to make ghost/unsized types `is_sized`, but that's another topic. We may need to constraint the types that can use `global size_of`, but it's not a problem to ensure `is_sized` here. Anyhow, `assert(size_of::<int>() == 1)` is already a mistake.


utaal:
> This leaves only the problem that global size_of int == 0 works and we can have is_sized::<int>(). Not a big problem IMO 🤔️

Hi, sorry for the very high latency on this. Does this mean that you think this change is now unnecessary?

ouuan:
> Does this mean that you think this change is now unnecessary?

No. Maybe my previous explanation was not clear enough. I was describing a potential unsoundness introduced by this change, where a user can write `global size_of int == 0` (`int` or any other ghost types), and then `is_sized::<int>()` will be true.

tjhance:
I don't see any problem. There's no issue with ghost types being sized; it's just irrelevant.

`global` already emits a compile-time check that the provided size is correct, which implicitly depends on that type being Sized. So there's no issue emitting the `is_sized` predicate as a Verus axiom.

utaal (approved):

PR #1116, opened 2024-05-15
Use bit_vector instead of external_body verifier for set_bit64_proof

ouuan:
Now the bit_vector verifier is enough to automatically prove this.


utaal (approved):

utaal:
Thank you!

PR #1119, opened 2024-05-16
vstd: broadcast bits lemmas

### LINKED ISSUE #1118: Convert `_auto` lemmas to `broadcast` lemmas in bitwise mask properties
https://github.com/verus-lang/verus/commit/78438925a7ffc9b6c0ac65c0f66c0be3ea981b78 introduced a number of `_auto` lemmas that should probably be `broadcast` lemmas instead. We may want to make that change soon, to avoid breaking more dependencies on the `_auto` lemmas if more users reference these.

cc @mmcloughlin @parno

mmcloughlin:
This PR replaces the `*_auto` lemmas in the vstd bits module with broadcast
lemmas.

Fixes #1118

### DIFF
--- a/source/vstd/bits.rs
+++ b/source/vstd/bits.rs
@@ -29,7 +29,7 @@ use crate::arithmetic::mul::{
 use crate::calc_macro::*;
 
 } // verus!
-  // Proofs that shift right is equivalent to division by power of 2.
+// Proofs that shift right is equivalent to division by power of 2.
 macro_rules! lemma_shr_is_div {
     ($name:ident, $uN:ty) => {
         #[cfg(verus_keep_ghost)]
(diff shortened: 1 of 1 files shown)


mmcloughlin:
I ran `vargo fmt` before but I think I've now fallen fowl of #1106, so I'll do it again :)

mmcloughlin:
Actually, I don't think it could have been caused by #1106 landing, since I didn't rebase against it.

I suspect the reason is that `verusfmt` released 0.3.2, and Verus CI pulls down latest version.

I still think it would be preferable for Verus to pin its dependency on `verusfmt`. https://github.com/verus-lang/verus/pull/1019#discussion_r1508450189

PR #1120, opened 2024-05-16
vstd/arithmetic: broadcast power2 lemmas

mmcloughlin:
Replace the `*_auto` lemmas in `vstd::arithmetic::power2` with `broadcast`
versions.

Updates #1118


utaal (approved):
Thank you for the quick turn-around on this!

PR #1122, opened 2024-05-17
Add no-std and no-alloc build tests in CI

ouuan:


utaal (approved):
Thanks! This will increase the CI time a bit, which isn't ideal, but it's probably worth it, so I'm happy to merge it, at least tentatively.

I renamed the job again to something shorter, hope that's okay with you.

ouuan:
> This will increase the CI time a bit

It will take more resources, but the turnaround time is nearly the same as they are running in parallel. It would take more time only if there are too many concurrent runs that the queue is full.

> I renamed the job again to something shorter, hope that's okay with you.

Fine.

PR #1128, opened 2024-05-19
Add missing span-locations feature for proc-macro2 in line_count

ouuan:
I faced compilation errors without this feature flag. This feature has been required since [proc-macro2 1.0.25](https://docs.rs/proc-macro2/1.0.25/proc_macro2/struct.LineColumn.html), so this should be already broken when the line_count tool was first added. I'm wondering why this was not discovered earlier 🤔️

### DIFF
--- a/source/tools/line_count/Cargo.toml
+++ b/source/tools/line_count/Cargo.toml
@@ -11,7 +11,7 @@ prettyplease_verus = { path = "../../../dependencies/prettyplease" }
 getopts = "*"
 toml = "0.8"
 serde = { version = "1.0", features = ["std", "derive", "rc"] }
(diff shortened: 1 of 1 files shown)


ouuan:
Oh, I just found https://github.com/verus-lang/verus/blob/8e7e368d7be270c49461c940375e1b770d95851a/source/tools/line_count/.cargo/config.toml#L1-L2

Somehow my cargo is not reading it.

ouuan:
Oh, I have the following in my `~/.cargo/config.toml` to use mold. The target-specific setting overrode the project local setting 🥲

```toml
rustflags = [
    "-Clink-arg=-fuse-ld=/usr/bin/mold",
]
```

PR #1139, opened 2024-05-24
Use &str and String from std lib

tjhance:
I made a best effort to be backwards compatible but this was impossible in general because some function names conflicts with the std library function names. e.g., `as_bytes` returns a Vec instead of &[u8]

### DIFF
--- a/source/vstd/string.rs
+++ b/source/vstd/string.rs
@@ -2,7 +2,7 @@
 #![allow(unused_imports)]
 
 #[cfg(feature = "alloc")]
-use alloc::string::{self, ToString};
+use alloc::string::{self, String, ToString};
 
 use super::prelude::*;
 use super::seq::Seq;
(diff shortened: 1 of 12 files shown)


parno:
Seems like it simplifies the Verus internals, which is nice!

source/vstd/string.rs
```diff
@@ -10,55 +10,97 @@ use super::view::*;
 
 verus! {
 
+//#[deprecated = "Use String from the standard library instead"]
```
utaal, line 13: Why are all these commented out?

utaal:
This is great, and I think it's okay to break compatibility a bit.

We should put out an announcement with the specific expected breakage though: I know of at least one project that uses a bunch of the functionality from the wrappers.

utaal:
Are you up for writing up a draft announcement with the specific changes we expect users to have to make?

tjhance:
It'd be pretty short, something like:

```
Hi everyone, we're deprecating our String and StrSlice types and supporting the std library String and &str types instead. StrSlice is now a type alias for &str, but it's deprecated, and we expect people to just start just using &str instead. Likewise, new_strlit is deprecated, and now you can just write a string literal without having to coerce it into a StrSlice.

Other breaking changes:
- [list changes]
```

utaal (approved):
Please post the announcement when you merge.

Thank you!

Issue #1137, opened 2024-05-23
Some `impl View`s seem to disappear when compiling Verus code with `cargo`


pratapsingh1729:
Suppose I have the following code in `src/main.rs` in some crate:
```
use vstd::prelude::*;
use std::rc::Rc;

verus! {

#[verifier(external_body)]
pub exec fn rc_clone<T:View>(t: &Rc<T>) -> (r: Rc<T>)
    ensures r.view() == t.view()
{ Rc::clone(t) }

pub exec fn foo(x: Rc<Vec<u8>>) -> (y: Rc<Vec<u8>>)
    ensures y.view() == x.view()
{
    rc_clone(&x)
}

}

fn main() {}
```
This verifies just fine with Verus, run as `verus main.rs`. Indeed, if I run `verus main.rs --compile`, it compiles just fine as well.

However, suppose my `Cargo.toml` then contains the following:
```
[dependencies]
builtin = { git = "https://github.com/verus-lang/verus", branch = "main" }
builtin_macros = { git = "https://github.com/verus-lang/verus", branch = "main" }
vstd = { git = "https://github.com/verus-lang/verus", branch = "main" }
```
I then compile the crate using `cargo build`. This gives me the following bizarre error:
```
   Compiling verus-vec-view-issue v0.1.0 (/usr0/home/prataps/src/verus-vec-view-issue)
error[E0277]: the trait bound `Vec<u8>: vstd::view::View` is not satisfied
  --> src/main.rs:14:14
   |
14 |     rc_clone(&x)
   |     -------- ^^ the trait `vstd::view::View` is not implemented for `Vec<u8>`
   |     |
   |     required by a bound introduced by this call
   |
   = help: the following other types implement trait `vstd::view::View`:
             bool
             isize
             i8
             i16
             i32
             i64
             i128
             usize
           and 19 others
note: required by a bound in `rc_clone`
  --> src/main.rs:7:24
   |
7  | pub exec fn rc_clone<T:View>(t: &Rc<T>) -> (r: Rc<T>)
   |                        ^^^^ required by this bound in `rc_clone`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `verus-vec-view-issue` (bin "verus-vec-view-issue") due to 1 previous error
```
[...]

PR #1140, opened 2024-05-24
Move View for Vec and add feature = "allocator"

Chris-Hawblitzel:
This is a fix for https://github.com/verus-lang/verus/issues/1137 .  The fix looked straightforward initially: we just have to move the declaration of View for Vec out of std_specs (where it gets erased during an ordinary rustc build) and into view.rs, where it is kept in all builds.  However, Vec is generic over Allocator, and we can't express this without using feature(allocator_api).  I didn't want to add feature(allocator_api) to all vstd builds, but some people might want the version that is generic over Allocator, so I added another feature flag `feature = "allocator"` to control this.

### DIFF
--- a/source/vstd/std_specs/vec.rs
+++ b/source/vstd/std_specs/vec.rs
@@ -19,21 +19,6 @@ pub struct ExVec<T, A: Allocator>(Vec<T, A>);
 #[verifier::external_body]
 pub struct ExGlobal(alloc::alloc::Global);
 
-impl<T, A: Allocator> View for Vec<T, A> {
-    type V = Seq<T>;
-
-    spec fn view(&self) -> Seq<T>;
-}
-
-impl<T: DeepView, A: Allocator> DeepView for Vec<T, A> {
-    type V = Seq<T::V>;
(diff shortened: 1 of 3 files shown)


utaal:
How does one set this new `feature = "allocator"`? Do we need to add it to `vargo`?

Chris-Hawblitzel:
> How does one set this new `feature = "allocator"`? Do we need to add it to `vargo`?

For vargo, we've already got `verus_keep_ghost` enabled, so hopefully `allocator` won't matter in this case.  Mainly, `allocator` is for direct compilation with cargo/rustc when `verus_keep_ghost` is false.

utaal:
Can we document it somewhere? (I realize we may be also missing docs for `alloc` and `std`).

Chris-Hawblitzel:
> Can we document this feature somewhere? (I realize we may be also missing docs for `alloc` and `std`).

I added a discussion for this here: https://github.com/verus-lang/verus/discussions/1162

utaal (approved):

PR #1145, opened 2024-05-28
[ci] use macos runner for main ci run, emit binary releases

utaal:
Screenshot 2024-05-28 at 16 26 10

---

TODO:

* [x] check for `rustup` in the `verus` driver, instruct the user to install rustup with the expected toolchain if absent
* [x] is there a reasonable way to preserve executable-ness in zip files?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   fmt:
-    runs-on: ubuntu-latest
+    runs-on: macos-14
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -32,27 +32,26 @@ jobs:
         run: |
           cargo fmt -- --check
 
-  test:
-    runs-on: ubuntu-latest
+  test-and-release-macos:
+    runs-on: macos-14
     strategy:
       matrix:
-        features: ['', 'singular', 'record-history', 'no-std', 'no-alloc']
+        features: ['', 'record-history', 'no-std', 'no-alloc', 'singular']
     steps:
       - name: checkout
         uses: actions/checkout@v4
       - name: get z3
         working-directory: ./source
         run: |
-          ../.github/workflows/get-z3.sh
+          ./tools/get-z3.sh
           echo z3 version `./z3 --version`
       - name: setup rust
         run: |
           curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL "https://sh.rustup.rs" | sh -s -- --default-toolchain none -y
-      - name: setup singular
+      - name: download singular
         if: matrix.features == 'singular'
         run: |
-          sudo apt-get update
-          sudo apt-get install -y singular
+          curl -LO https://www.singular.uni-kl.de/ftp/pub/Math/Singular/UNIX/Singular-4-4-0_M1.dmg
       - name: build and test
         working-directory: ./source
         run: |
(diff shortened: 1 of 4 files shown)


utaal:
@verus-lang/verus-devs anyone interested in picking up the review for this? I can give a quick overview at the verus meeting later.

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   fmt:
-    runs-on: ubuntu-latest
+    runs-on: macos-14
     steps:
```
jaybosamiya, line 16: Any reason that the format job has been changed to MacOS? Unless I'm mistaken, having it be ubuntu _should_ give us _slightly_ better perf, so if not crucial, it would be nice for this to just stay Ubuntu.

.github/workflows/ci.yml
```diff
@@ -90,19 +90,33 @@ jobs:
...
+          . ../tools/activate
+          vargo clean
+          vargo build --release
+          ./target-verus/release/verus --version --output-json > ./target-verus/release/version.json
+      - name: upload verus release artifact
+        uses: actions/upload-artifact@v4
+        if: matrix.features == ''
+        with:
+            name: verus-arm64-macos
+            path: source/target-verus/release
```
jaybosamiya, line 106: Might be good to add a quick test before this to confirm that it _is_ actually on arm64. We don't want to accidentally get x86-64 binaries being called arm64 (Rosetta will prevent it from being _immediately_ obvious, but it will end up with folks getting slower Verus runs, which is not a good experience)

.github/workflows/ci.yml
```diff
@@ -130,3 +144,113 @@ jobs:
           $env:VERUS_Z3_PATH = "$(Get-Location)/z3"; vargo test -p rust_verify_test --test basic
           $env:VERUS_Z3_PATH = "$(Get-Location)/z3"; vargo test -p rust_verify_test --test layout
         shell: powershell
+      - name: build verus release
+        working-directory: .\source
+        run: |
+          ../tools/activate.ps1
+          $env:VERUS_Z3_PATH = "$(Get-Location)/z3"; vargo build --release
+          # $env:VERUS_Z3_PATH = "$(Get-Location)/z3"; ./target-verus/release/verus.exe --version --output-json > ./target-verus/release/version.json
```
jaybosamiya, line 152: Why is this commented out?

.github/workflows/ci.yml
```diff
@@ -130,3 +144,113 @@ jobs:
...
+          vargo clean
+          vargo build
+          VERUS_Z3_PATH="$(pwd)/z3" vargo test -p rust_verify_test --test basic
+      - name: build verus release
+        working-directory: ./source
+        run: |
+          . ../tools/activate
+          vargo clean
+          vargo build --release
+          # TODO ./target-verus/release/verus --version --output-json > ./target-verus/release/version.json
```
jaybosamiya, line 186: Similarly, any reason this is commented out (or a TODO)?

tools/vargo/src/main.rs
```diff
@@ -1061,6 +1084,23 @@ fn run() -> Result<(), String> {
 
                     std::fs::copy(&from_f, &to_f)
                         .map_err(|x| format!("could not copy file ({})", x))?;
+
+                    if is_exe {
+                        writeln!(
+                            &mut macos_prepare_script,
+                            "xattr -d com.apple.quarantine {}",
```
jaybosamiya, line 1091: While it is _probably_ fine (who uses directory names with spaces?!), it probably is a good idea for these file names (here and below, in all the `writeln`s) to have quoted file names)

tools/vargo/src/main.rs
```diff
@@ -1186,6 +1250,41 @@ fn run() -> Result<(), String> {
...
+                    format!("could not set permissions on linux prepare script ({})", x)
+                })?;
+            }
+
+            if let Some(version_info) = verus_version {
+                let version_info_path = target_verus_dir.join("version.txt");
+                std::fs::write(&version_info_path, version_info.version)
+                    .map_err(|x| format!("could not write to version file ({})", x))?;
+            }
+
```
jaybosamiya, line 1287: Is this the reason that the version-writing commands are commented out from the CI portion?

jaybosamiya (approved):
I wasn't at the Verus meeting this week (so maybe someone else should review this PR too if you think it is crucial @utaal), but overall this PR looks good to me. And yay, binary releases!!! Thanks for implementing this.

A few minor comments inline.

Regarding the zip file executable-bit preservation: zip files are supposed to preserve that (and do), so I don't think that the `chmod` stuff is actually necessary, but clearly you seem to have hit some issue with it before, which is why you have it implemented? EDIT: My guess is that this is happening due to the artifact upload and zipping; if you zip it locally, and _then_ upload, rather than having the zip phase being after the upload, I _think_ it should work correctly.

I suspect the `com.apple.quarantine` one is going to be necessary simply because of it being a downloaded binary, so that probably could not be completely eliminated.

.github/workflows/ci.yml
utaal: It turns out `macos-14` runners have three cores instead of two: this may help for `fmt` too, as we do verusfmt in parallel, but I admittedly just changed it to be consistent with the primary test job.

.github/workflows/ci.yml
utaal: Do you have a command in mind to do this?

.github/workflows/ci.yml
utaal: Because I couldn't get it to work, and I don't know why. It's not critical. I'll add a TODO.

tools/vargo/src/main.rs
utaal: They are just file names (no other path segments) and they all don't have any special characters. I'm more worried about getting the quoting wrong, than to get a failure if we somehow introduce an artifact with a weird name.

tools/vargo/src/main.rs
utaal: Because I couldn't get the windows one to work, and they are redundant with this.

utaal:
Thanks for the review!

> EDIT: My guess is that this is happening due to the artifact upload and zipping; if you zip it locally, and then upload, rather than having the zip phase being after the upload, I think it should work correctly.

I think this is right. I tried to address it by zipping myself but didn't consider that the first artifact upload would have already stripped the `+x`. I'm now trying to zip before uploading the first time.

.github/workflows/ci.yml
jaybosamiya: Ah, I didn't know the number of cores thing was different, TIL!

.github/workflows/ci.yml
jaybosamiya: `uname -m` is portable, and I believe should work (am not an an arm machine right now to test though)

PR #1159, opened 2024-06-10
Make the get-z3 script use curl instead of wget

lincolnq:
Wget is not installed by default on Macs, and my Linux also didn't have it. But curl is built-in to most systems, so I switched the script to use curl. I also made a few other small drive-by changes:

- better bash quoting
- use bash -eu to stop if there's an error
- display the URL being downloaded to the user for easier debugging


utaal (approved):
These all make sense to me. Thanks!

PR #1165, opened 2024-06-14
Specify behavior of std::collections::HashMap

jaylorch:
This code adds specifications for the standard-library type `std::collections::HashMap`. It also adds utility structures `HashMapWithView` and `StringHashMap`.

Most of the specification for `HashMap` only applies if you use `HashMap<Key, Value>`. If you use some custom build hasher, e.g., with`HashMap<Key, Value, CustomBuildHasher>`, the specification won't specify much.

Likewise, the specification is only meaningful when you know that `Key::hash` is deterministic and that only identical keys satisfy the executable `==` operator. We have an axiom that all primitive types and `Box`es thereof meet these requirements. But if you want to use some other key type `MyKey`, you need to explicitly state your assumption that it satisfies these requirements with `assume(vstd::std_specs::hash::obeys_key_model::<MyKey>());`. In the future, we plan to devise a way for you to prove that it satisfies these requirements so you don't have to make such an assumption.

To make most use of the specification, you should use `broadcast use vstd::std_specs::hash::group_hash_axioms;`. This will bring various useful axioms about the behavior of a `HashMap` into the ambient reasoning context. In the future, if we find that having these axioms in scope doesn't impact performance, we may put them into the global ambient context so you don't have to explicitly `broadcast use` them.

For cases where you want the view of a `HashMap` to be `Map<<Key as View>::V, Value>` instead of `Map<Key, Value>`, use `vstd::hash_map::HashMapWithView`. For the specific case that you want the key to be `String`, use `vstd::hash_map::StringHashMap`.


source/vstd/std_specs/hash.rs
```diff
@@ -0,0 +1,383 @@
...
+#[verifier::when_used_as_spec(spec_hash_map_len)]
+pub fn ex_hash_map_len<Key, Value, S>(m: &HashMap<Key, Value, S>) -> (len: usize)
+    ensures
+        len == spec_hash_map_len(m),
+{
+    m.len()
+}
+
+#[verifier::external_fn_specification]
+pub fn ex_hash_map_new<Key, Value>() -> (m: HashMap<Key, Value, RandomState>)
```
parno, line 211: Here and in `ex_hash_map_with_capacity`, the third generic argument to `HashMap` swaps from `S` (used above) to `RandomState`.  Is that intentional?

source/vstd/std_specs/hash.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+#[verifier::external_fn_specification]
+pub fn ex_hash_map_insert<Key, Value, S>(
+    m: &mut HashMap<Key, Value, S>,
+    k: Key,
+    v: Value,
+) -> (result: Option<Value>) where Key: Eq + Hash, S: BuildHasher
+    ensures
+        m@ == old(m)@.insert(k, v),
```
parno, line 248: Should this equality also be gated on `does_type_have_deterministic_hash::<Key>() && does_type_conform_to_build_hasher_model::<S>()`?

parno: Also, does it account for this bit of the [documentation](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert:~:text=The%20key%20is%20not%20updated%2C%20though%3B%20this%20matters%20for%20types%20that%20can%20be%20%3D%3D%20without%20being%20identical)?
> The key is not updated, though; this matters for types that can be == without being identical

source/vstd/std_specs/hash.rs
```diff
@@ -0,0 +1,383 @@
...
+// `&Q` to a certain value. And the postcondition of `get` says that
+// its result matches the output of that specification function. (It
+// also says that its result corresponds to the output of
+// `map_contains_borrowed_key`, discussed above.) But this isn't very
+// helpful by itself, since there's no body to that specification
+// function. So we have special-case axioms that say what this means
+// in two important circumstances: (1) `Key = Q` and (2) `Key =
+// Box<Q>`. (TODO: Include another axiom for the case `Key = String, Q
+// = str`.)
+pub spec fn map_maps_borrowed_key_to_value<Key, Value, Q: ?Sized>(
```
parno, line 318: Can `map_maps_borrowed_key_to_value` be shortened to `maps_borrowed_key_to_value`?

source/rust_verify_test/tests/hash.rs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                self.i.hash(state);
+                self.j.hash(state);
+            }
+        }
+
+        fn test()
+        {
+            broadcast use vstd::std_specs::hash::group_hash_axioms;
+            assume(vstd::std_specs::hash::does_type_have_deterministic_hash::<MyStruct>());
```
parno, line 124: Would it be worthwhile to add a negative test, similar to this one, where you don't include the assumption about `MyStruct`?

source/vstd/std_specs/hash.rs
```diff
@@ -0,0 +1,383 @@
...
+/// with`HashMap<Key, Value, CustomBuildHasher>`, the specification
+/// won't specify much.
+///
+/// Likewise, the specification is only meaningful when you know that
+/// the `Key` has a deterministic hash. We have an axiom that all
+/// primitive types and `Box`es thereof have a deterministic hash. But
+/// if you want to use some other key type `MyKey`, you need to
+/// explicitly state your assumption that it has a deterministic hash
+/// with
+/// `assume(vstd::std_specs::hash::does_type_have_deterministic_hash::<MyKey>());`.
```
parno, line 15: Can we shorten `does_type_have_deterministic_hash` to `deterministic_hash` and `does_type_conform_to_build_hasher_model` to something like `obeys_hasher_model`?

parno (commented):
Thanks very much for putting this together!  It will be very handy to have (even partial) support for reasoning about HashMaps.

source/vstd/std_specs/hash.rs
jaylorch: Yes, this is intentional. The `new` function in the Rust standard library doesn't give you a choice of `S`. This is the literal signature from https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.new.

parno: Ah, interesting.  Thanks for clarifying!

source/vstd/std_specs/hash.rs
jaylorch: I figured the inequality didn't need to be gated on those, because it only affects the abstract view of the map, which we can consider to have changed even if there's no way to get concrete access to that abstract view. I guess this is a pretty subtle point, though, and I might as well gate it.

As for that bit of documentation, no, I hadn't taken it into account! I'll have to change `does_type_have_deterministic_hash` to something stronger implying that it not only has a deterministic hash but is suitable for this model of HashMap. Specifically, it can't do those shenanigans with two non-identical keys satisfying `==`.

Eventually, I also plan to write a wrapper named `HashMapWithView` or something that represents a `HashMap` whose `Key` satisfies `View<KeyView>`. Then, it'll be abstracted as a `Map<KeyView, Value>`. But I think `HashMap` should stick with only modeling "normal" keys.

source/vstd/std_specs/hash.rs
jaylorch: Sure, I'll do that.

source/rust_verify_test/tests/hash.rs
jaylorch: Sure. Though hopefully eventually we won't need such an assume, at which point we can remove the test.

source/vstd/std_specs/hash.rs
jaylorch: Given what you said above, we need something stronger than that it has a deterministic hash. It has to do that and also require keys satisfying `==` to be identical. How about `satisfies_hash_table_key_model`? As for `obeys_hasher_model`, it's not that the type has to obey our model of `Hasher`, it's that it has to construct objects that obey that model. How about `satisfies_build_hasher_model`?

jaylorch:
Is `verifier::prune_unless_this_module_is_used` effective enough that we can just add these axioms into the global ambient context, without worrying about performance implications?

source/vstd/std_specs/hash.rs
jaylorch: I've made the changes.

source/vstd/std_specs/hash.rs
parno: > also require keys satisfying == to be identical

Would it be enough to require identical views, rather than identical keys?

>  How about `obeys_hash_table_key_model`?

Since we're already in the HashMap namespace, perhaps `obeys_key_model` would suffice?

> How about `obeys_build_hasher_model`?

That could work.  What do you think of `builds_valid_hashers`?

source/vstd/std_specs/hash.rs
parno: Thanks very much!

source/vstd/std_specs/hash.rs
jaylorch: I've updated the names to `obeys_key_model` and `builds_valid_hashers`.

parno: Cool, sounds good.

panda2134:
> Sorry for the long delay on this.
> 
> We discussed hash map specs for a while at the retreat, and IIRC, the decision then was to add a new trait and add the capability to write specs dependent on this trait bound. However, it might be some time before we get that working.
> 
> Meanwhile, this PR is available now, and the feature is a long-requested one. Should we proceed with this, with the expectation of later moving to what we decided on at the retreat?

BTW this PR is now outdated, not compiling with latest changes on the main branch (complaining about generics-related errors).

panda2134:
> > Sorry for the long delay on this.
> > We discussed hash map specs for a while at the retreat, and IIRC, the decision then was to add a new trait and add the capability to write specs dependent on this trait bound. However, it might be some time before we get that working.
> > Meanwhile, this PR is available now, and the feature is a long-requested one. Should we proceed with this, with the expectation of later moving to what we decided on at the retreat?
> 
> BTW this PR is now outdated, not compiling with latest changes on the main branch (complaining about generics-related errors).

The problem is caused by aab953d515624c66232da1066586f0b9441273e1. Script for reproduction:

```
#!/bin/bash
set +xe
cd /tmp
git clone git@github.com:verus-lang/verus.git
cd verus
git checkout aab953d515624c66232da1066586f0b9441273e1
curl https://github.com/verus-lang/verus/pull/1165.patch | git am
cd source
./tools/get-z3.sh
source ../tools/activate
vargo build
```

panda2134:
Error message:

```
note: could not find span associated with error message; printing raw error message instead

trait T2547_BuildHasher {
    type A2548_Hasher : T2546_Hasher+?Sized;
}

error[E0046]: not all trait items implemented, missing: `A2548_Hasher`
    --> dummyrs.rs:2857:1
     |
277  |     type A2548_Hasher : T2546_Hasher+?Sized;
     |     --------------------------------------- `A2548_Hasher` from trait
...
2857 | impl T2547_BuildHasher for D2398_RandomState {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `A2548_Hasher` in implementation


error: not all trait items implemented, missing: `A2548_Hasher`
  --> /home/panda2134/.rustup/toolchains/1.76.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/hash/random.rs:80:1
   |
80 | / impl BuildHasher for RandomState {
81 | |     type Hasher = DefaultHasher;
   | |_________^

note: This error was found in Verus pass: ownership checking of tracked code

thread 'main' panicked at vstd_build/src/main.rs:140:9:
vstd build failed with exit code Some(1)
error: vstd_build returned status code Some(101)
```

jaylorch:
Thanks, @panda2134! @Chris-Hawblitzel is working on a fix for that.

jaylorch:
OK, this PR is now up to date with the latest version of the main branch.

parno (approved):

tjhance (approved):

Chris-Hawblitzel (approved):

Chris-Hawblitzel:
I just noticed that `HashMap` is marked `#[verifier::reject_recursive_types(Key)]` and `#[verifier::reject_recursive_types(Value)]`.  Is there any objection to changing these to `#[verifier::accept_recursive_types(Key)]` and `#[verifier::accept_recursive_types(Value)]`?  It's always safe to say "reject", but this seems overly cautious in this case.  Since this is a finite map, it should be in principle implementable as a `Vec<(Key, Value)>`, which would allow "accept" on both the keys and values.

tjhance:
I agree. This point came up recently on the zulip as well.

jaylorch:
OK, I just pushed a change to allow recursive types for `Key` and `Value` (but not `S`). By the way, could you please point me to the thread on zulip that discusses this?

tjhance:
It was here: https://verus-lang.zulipchat.com/#narrow/channel/399078-help/topic/mutual.20recursive.20type

PR #1168, opened 2024-06-20
Fix missing type annotation in examples in vstd/seq.rs.

XuhengLi:
Some examples in vstd/seq.rs throw types errors due to missing type annotations over integer literals when initializing a `Seq`.

It is also worth noting that it only happens when the sequence only contains positive numbers, in which case the type checker cannot determine whether the number is an `int` or `nat`.


utaal (approved):
Thanks for the fixes. I prefer we suggest literal type annotations, and I removed the negative discussion, as it may be out of context for someone reading the docs on rustic (verusdoc).

PR #1173, opened 2024-06-22
Rustc 1.79.0

utaal:
This has taken a *while* but I think it's almost there.

~~After merging `main` there's this issue which I think is related to treating the marker trait specially for erasure:~~
(solved, see below)

```
trait T3254_Pointee {
    type A3255_Metadata : T3256_Freeze+T3249_Debug+T3176_Hash+T3253_Ord+T3252_PartialOrd<<Self as T3254_Pointee>::A3255_Metadata, >+T3251_PartialEq<<Self as T3254_Pointee>::A3255_Metadata, >+T3175_Eq+T3245_Clone+?Sized;
}

error[E0405]: cannot find trait `T3256_Freeze` in this scope
   --> dummyrs.rs:263:27
    |
263 |     type A3255_Metadata : T3256_Freeze+T3249_Debug+T3176_Hash+T3253_Ord+T3252_PartialOrd<<Self as T3254_Pointee>::A3255_Metadata, >+T3251...
    |                           ^^^^^^^^^^^^ not found in this scope
```


ziqiaozhou:
I tried the 1.79 change and it works with the all tests in verus repo after adding Freeze trait in vstd/std_spec/core.rs. However, I have encountered some errors in my code, I think it is due to the check on calling non-const inside a const function. I noticed that your chance also marked some functions like reveal_hide_ to const, but that is not enough since I may also use other verus item ensures/requires/returns/closure_to_fn_spec in a const function. 
```
error[E0015]: cannot call non-const fn `builtin::ensures::<builtin::FnSpec<(address::VirtAddr,), bool>>` in constant functions
   --> kernel/src/address.rs:257:5
    |
257 |     #[ensures(|ret: VirtAddr| ret.new_ensures(addr))]
```

I have no idea how to solve this similar to reveal_hide_, since those functions takes generic types without restricting it with Copy, and so cannot directly be marked as const. I am wondering why it does not trigger this error before, since it should be a strict check in rust even before 1.79.

ziqiaozhou:
@utaal , I tried this change. I think it works well after a small change to vstd to support Freeze trait.



> I tried the 1.79 change and it works with the all tests in verus repo after adding Freeze trait in vstd/std_spec/core.rs. However, I have encountered some errors in my code, I think it is due to the check on calling non-const inside a const function. I noticed that your chance also marked some functions like reveal_hide_ to const, but that is not enough since I may also use other verus item ensures/requires/returns/closure_to_fn_spec in a const function.
> 
> ```
> error[E0015]: cannot call non-const fn `builtin::ensures::<builtin::FnSpec<(address::VirtAddr,), bool>>` in constant functions
>    --> kernel/src/address.rs:257:5
>     |
> 257 |     #[ensures(|ret: VirtAddr| ret.new_ensures(addr))]
> ```
> 
> I have no idea how to solve this similar to reveal_hide_, since those functions takes generic types without restricting it with Copy, and so cannot directly be marked as const. I am wondering why it does not trigger this error before, since it should be a strict check in rust even before 1.79.

NVM, it seems not due to this reason. I think it might be other issue in my code since calling ensures/requires does work for const functions in another clean project.

ziqiaozhou:
> This has taken a _while_ but I think it's almost there.
> 
> After merging `main` there's this issue which I think is related to treating the marker trait specially for erasure:
> 
> ```
> trait T3254_Pointee {
>     type A3255_Metadata : T3256_Freeze+T3249_Debug+T3176_Hash+T3253_Ord+T3252_PartialOrd<<Self as T3254_Pointee>::A3255_Metadata, >+T3251_PartialEq<<Self as T3254_Pointee>::A3255_Metadata, >+T3175_Eq+T3245_Clone+?Sized;
> }
> 
> error[E0405]: cannot find trait `T3256_Freeze` in this scope
>    --> dummyrs.rs:263:27
>     |
> 263 |     type A3255_Metadata : T3256_Freeze+T3249_Debug+T3176_Hash+T3253_Ord+T3252_PartialOrd<<Self as T3254_Pointee>::A3255_Metadata, >+T3251...
>     |                           ^^^^^^^^^^^^ not found in this scope
> ```

@utaal, the freeze issue could be fixed by adding the external specification for Freeze trait in vstd.  https://github.com/ziqiaozhou/verus/commit/7e824a27e49f47f2fa85e76aee5c6d9b97dd63c7

utaal:
Thank you for the pointer @ziqiaozhou!

ziqiaozhou:
I deleted my prior comments since I was thinking that error is due to other reasons, since the error only happens in a large project where the code may use some new features. The error disappears if I extract that specific chunk of code out. Later on, I realized that it is fixable by const_header_wrapper.

To support that in the new macro, I need to change the rewrite similar to the change in syntax.rs. I just lazily set that const_header_wrapper for all spec function calls. I think we can just set it by default no matter whether the function is const or non-const. I also noticed that some macros rely on syntax::rewrite_expr or proof_block or others will need to set is_const: true as default.

https://github.com/ziqiaozhou/verus/commit/ae239c617eaf276cbffa7a5229cd8b537fcc3ce2

utaal:
@Chris-Hawblitzel ran veritas and found that my fixes for the new const rules are not sufficient, specifically for this case, from https://github.com/utaal/verified-nrkernel

```rust
pub exec const x86_arch_exec: ArchExec ensures x86_arch_exec@ == x86_arch_spec {
    let layers = [
        ArchLayerExec { entry_size: L0_ENTRY_SIZE, num_entries: 512 },
        ArchLayerExec { entry_size: L1_ENTRY_SIZE, num_entries: 512 },
        ArchLayerExec { entry_size: L2_ENTRY_SIZE, num_entries: 512 },
        ArchLayerExec { entry_size: L3_ENTRY_SIZE, num_entries: 512 },
    ];
    assert(x86_arch_spec.layers =~= layers@.map(|n,e:ArchLayerExec| e@));
    ArchExec { layers }
}
```

where there is a mix of proof code, which should be allowed to call anything, and exec code which should conform to the const rules. At the moment I don't know how we can address this.

source/rust_verify/src/verifier.rs
```diff
@@ -2772,6 +2770,24 @@ impl rustc_driver::Callbacks for VerifierCallbacksEraseMacro {
     fn config(&mut self, config: &mut rustc_interface::interface::Config) {
         config.override_queries = Some(|_session, providers| {
             providers.hir_crate = hir_crate;
+
+            // Do not actually evaluate consts if we are not compiling, as doing so triggers the
+            // constness checker, which is more restrictive than necessary for verification.
+            // Doing this will delay some const-ness errors to when verus is run with `--compile`.
+            providers.eval_to_const_value_raw =
+                |_tcx, _key| Ok(rustc_middle::mir::ConstValue::ZeroSized);
```
utaal, line 2778: @Chris-Hawblitzel I did this, which is quite heavy-handed, but seems to work.

Chris-Hawblitzel (approved):

PR #1175, opened 2024-06-25
Prune broadcast functions based on their triggers

Chris-Hawblitzel:
The old way of pruning broadcast functions was to keep all the broadcasts from a module if the module itself was reachable.  This pull request introduces a more precise pruning policy that mostly replaces the old policy.  In the new policy, a broadcast function is kept (not pruned) if any of its triggers are reachable.  (If none of its triggers are reachable, then there would be no way to trigger the broadcast.)

There are a few exceptional cases that don't fall under this policy: arithmetic triggers, extensional equality triggers, and height triggers.  For these, the old module-based policy is used.  In the future, we could address these with other policies, such as type-based pruning.

The trigger-based policy relies on broadcast functions having explicit triggers.  Therefore, this pull request adds a warning for any broadcast function that does not provide explicit triggers.


Chris-Hawblitzel:
> There are a few exceptional cases that don't fall under this policy: arithmetic triggers, extensional equality triggers, and height triggers. For these, the old module-based policy is used. In the future, we could address these with other policies, such as type-based pruning.

I just added type-based pruning and removed the old module-based policy.  The new policy, based on the actual functions and types in the triggers, is more precise than the old module-based policy, which was kind of a hack.

utaal (approved):

PR #1177, opened 2024-06-25
Fix veritas config for verified-storage

hayley-leblanc:
This PR updates the Veritas config file `run_configuration_all.toml` to account for changes in the build steps for the `verified-storage` project -- mainly, adding a step to build the new crate `pmsafe` that the main `storage_node` crate depends on. I've also added a call to `apt update`, which seems to be necessary to get the required LLVM/clang-related dependencies and removed the toolchain option when building `deps_hack` because it now has a `rust-toolchain.toml` file.


utaal:
Try now. Thanks! (Please use "Rebase and merge")

PR #1178, opened 2024-06-25
Implement repeat for the seq macro

jaybosamiya-ms:
Finishes off one of the tasks on https://github.com/verus-lang/verus/discussions/1176?sort=top#discussioncomment-9875969

### DIFF
--- a/source/rust_verify_test/tests/seqs.rs
+++ b/source/rust_verify_test/tests/seqs.rs
@@ -29,6 +29,9 @@ test_verify_one_file! {
             let s7 = seq![true ==> false, false ==> true];
             assert(!s7.index(0));
             assert(s7.index(1));
+            let s8: Seq<int> = seq![5; 10];
+            assert(s8[0] == 5);
+            assert(s8.len() == 10);
 
             assert(s1.to_set().finite()) by {
                 seq_to_set_is_finite(s1);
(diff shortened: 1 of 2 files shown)


utaal (approved):
Yay! The shortest development cycle of all time. :D

https://github.com/verus-lang/verus/discussions/1176?sort=top#discussioncomment-9875969

utaal:
Can you briefly summarize what was the test failure we were seeing and how it was fixed? Thanks!

jaybosamiya-ms:
It was an issue of functions/closures/funspec.  In particular, throwing away the passed-in-argument via `_` is considered an advanced pattern by Verus (in the style of general parameter pattern binds, such as `(a, b)` or such), and thus not available in spec land, but we _can_ just bind it to an unused variable (`_x`), which lets it go through.

Essentially, the code change:
```diff
- $crate::vstd::seq::Seq::new($n, |_| $elem)
+ $crate::vstd::prelude::closure_to_fn_spec(|_x: _| $elem),
```

The error message was confusing for a while, but once I realized that the issue was the `_`, and switched it to `_x`, it was much more straightforward to realize that I also needed to wrap it with the `closure_to_fn_spec`

utaal:
Aha! Thanks for the explanation.

PR #1185, opened 2024-06-26
Add issue template for bug reports

jaybosamiya-ms:
Reminds users to look attach the `--record` output.


utaal (approved):
LGTM. Thank you!

PR #1188, opened 2024-06-26
Add documentation on type sizes

hayley-leblanc:
This PR adds a new page to the docs with info on setting the size of `Sized` types for use in ghost code, including how to set the size of `usize`/`isize`. There was a small amount of documentation on `usize`/`isize` in the the [Integers](https://verus-lang.github.io/verus/guide/integers.html) part of the tutorial, so to make it easier to keep the documentation in sync I moved this to the new file. I'm not sure whether it would be better to have this information in the tutorial or the reference, though, so I'm happy to move some/all of the documentation back to the tutorial if folks think that would be a better location for it.

### DIFF
--- a/source/docs/guide/src/SUMMARY.md
+++ b/source/docs/guide/src/SUMMARY.md
@@ -147,6 +147,7 @@
 - [Misc. Rust features]()
   - [Statics](./static.md)
   - [char](./char.md)
+  - [Sized types](./sized.md)
 - [Command line]()
   - [--record](./reference-flag-record.md)
 - [Planned future work]()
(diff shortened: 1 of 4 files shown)


tjhance:
This is a bit redundant with existing layout documentation (https://verus-lang.github.io/verus/guide/reference-global.html) However, the existing page lacks examples, which this PR provides.

I realized it's been a few months, but if you're still interested in merging this, I think the best course is to rebase and consolidate the new examples into the existing `"global" directive` reference section.

utaal:
@tjhance are you up for making those changes? Otherwise I can do a pass at some point soon.

utaal:
This is up for grabs, if someone wants to pick it up. We expect this to be relatively easy to complete.

PR #1190, opened 2024-06-26
Add for loop for Verus Documentation 

FeizaiYiHao:
This PR adds for loop example in the Verus Documentation, including a short intro for for loops, a code example, and short explanation for how to use `iter` in a for loop.


utaal:
Thank you!

PR #1192, opened 2024-06-26
Documentation for arithmetic and mixed triggers

matthias-brun:
I marked this as draft because I still need to extract the examples into example/test files but I'm waiting for @utaal's changes to the test harness to support multiple expecting-to-fail tests in a single file.

@Chris-Hawblitzel Feel free to adjust the text (or let me know what you'd like changed) if you think my text doesn't sufficiently dissuade people from using arithmetic/mixed triggers.

@jopolzin Please also have a read through and let me know if this is understandable.


matthias-brun:
There is now full support for mixed triggers: https://github.com/verus-lang/verus/commit/b6344b668dc13ff4f658852c0d45c9b00ea649d8

PR #1193, opened 2024-06-26
add struct, enum section to guide

jonhnet:
* field accessors removed from operators section; replaced with forward reference

### DIFF
--- a/source/docs/guide/src/datatypes_enum.md
+++ b/source/docs/guide/src/datatypes_enum.md
@@ -38,7 +38,7 @@ If an `enum` field reuses a name, you can qualify the field access:
 ```
 
 For variants like `Shape` declared with round parentheses `()`,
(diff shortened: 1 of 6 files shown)


source/docs/guide/src/datatypes_enum.md
```diff
@@ -0,0 +1,66 @@
...
+{{#include ../../../rust_verify/example/guide/datatypes.rs:life}}
+```
+
+`match` works as in Rust.
+```rust
+{{#include ../../../rust_verify/example/guide/datatypes.rs:shape}}
+```
+
+For variants like `Shape` declared with round parentheses `()`,
+you can use Verus' `->' tuple-like syntax to acces a single field
```
utaal, line 41:
```suggestion
you can use Verus' `->' tuple-like syntax to access a single field
```

utaal (approved):
<3

PR #1198, opened 2024-06-28
Add support for array literals

parno:
This introduces an AIR-level array literal, which we then connect all the way up to the rust_verify level.  This includes updating the interpreter to handle the new array literals.

The AIR-level approach allows an efficient encoding to associate array indices to values.  A future PR will take advantage of this new encoding to optimize sequence literals by changing how `seq!` behaves.

Commit 3779b06d86b9da is small, but it should be carefully reviewed, since it changes the mode checking for tuples, arrays, datatypes.  The change is consistent with our formalization, but it's worth double checking.  @Chris-Hawblitzel plans to add a few additional tests as a sanity check.


y1ca1:
This is very cool! I played with it and it really enables a seamless experience of spec array and exec array interplays.

source/vstd/array.rs
```diff
@@ -9,7 +9,9 @@ verus! {
 impl<T, const N: usize> View for [T; N] {
     type V = Seq<T>;
```
y1ca1, line 10: Should we mark the view of arrays to be just `[T; N]` (like those primitive integer types)?

parno: My guess is that a sequence view will be more useful, since the spec writer can then make use of all the features in the sequence library.

tjhance (approved):

parno:
@Chris-Hawblitzel Thanks for adding the mode test.  Did you get a chance to review the rest of the code?  No hurry on my side; I'm just not sure if we're ready to merge.

Chris-Hawblitzel (approved):

Chris-Hawblitzel:
> Did you get a chance to review the rest of the code? No hurry on my side; I'm just not sure if we're ready to merge.

I think you can go ahead and squash and merge.

PR #1199, opened 2024-06-28
remove tuples from the pool of potential auto triggers

### LINKED ISSUE #695: Imprecise trigger report for requires clause
Consider the following:
```rust
use vstd::prelude::*;

verus! {

exec fn foo(f: impl FnOnce(u32) -> u32) -> (r: u32)
  requires forall |x| f.requires((x,))
{
   f(5)
}

}

fn main() {}
```

Running Verus on this prints the warning:
```console
$ verus a.rs
note: verifying root module

note: automatically chose triggers for this expression:
 --> a.rs:6:12
  |
6 |   requires forall |x| f.requires((x,))
  |            ^^^^^^

note:   trigger 1 of 1:
 --> a.rs:6:34
  |
6 |   requires forall |x| f.requires((x,))
  |                                  ^^^^

note: Verus printed one or more automatically chosen quantifier triggers
      because it had low confidence in the chosen triggers.
      To suppress these messages, do one of the following:
        (1) manually annotate a single desired trigger using #[trigger]
            (example: forall|i: int, j: int| f(i) && #[trigger] g(i) && #[trigger] h(j)),
        (2) manually annotate multiple desired triggers using #![trigger ...]
            (example: forall|i: int| #![trigger f(i)] #![trigger g(i)] f(i) && g(i)),
        (3) accept the automatically chosen trigger using #![auto]
            (example: forall|i: int, j: int| #![auto] f(i) && g(i) && h(j))
        (4) use the --triggers-silent command-line option to suppress all printing of triggers.
      (Note: triggers are used by the underlying SMT theorem prover to instantiate quantifiers;
      the theorem prover instantiates a quantifier whenever some expression matches the
      pattern specified by one of the quantifier's triggers.)
 --> a.rs:6:12
  |
6 |   requires forall |x| f.requires((x,))
  |            ^^^^^^

verification results:: verified: 2 errors: 0
```

Notice that running Verus _appears_ to have chosen `(x,)` as a trigger, but this is not a valid point to annotate as a `#[trigger]` (and trying to manually put that as a trigger point won't work). Indeed, looking at the smt2 file (after `#![auto]` to silence the warning), Verus has chosen `f.requires((x,)` as the full trigger, rather than `(x,)` as it claims.

Also, this `f.requires((x,))` is the _only_ valid point that a trigger can be placed for this expression. In fact, Verus is even kinda aware of this, since it doesn't report a warning if you instead have this (notice rather than `(x,)`, it is just `x`):
```rs
exec fn foo(f: impl FnOnce(u32) -> u32) -> (r: u32)
  requires forall |x| f.requires(x)
  ensures f.ensures((5,), r)
{
   f(5)
}
```
This succeeds without any warning:
```console
$ verus a.rs
note: verifying root module

verification results:: verified: 2 errors: 0
```

If we ask if for triggers, it even prints the right thing:
```console
$ verus a.rs --triggers
note: verifying root module

note: automatically chose triggers for this expression:
 --> a.rs:6:12
  |
6 |   requires forall |x| f.requires(x)
  |            ^^^^^^

note:   trigger 1 of 1:
 --> a.rs:6:23
  |
6 |   requires forall |x| f.requires(x)
  |                       ^^^^^^^^^^^^^

verification results:: verified: 2 errors: 0
```

---

Tested on:
```
Verus
  Version: 0.2023.07.20.2dc6a17
  Profile: release
  Platform: macos_aarch64
  Toolchain: 1.68.0-aarch64-apple-darwin
```

JoPolzin:
Ideally Closes #695 , which fixed itself in the meantime, but raised the problem that tuples are chosen as auto-triggers, as they are handled as functions internally.

Checking if a typeconstructor is a Tuple-constructor and discarding it from the set of possible triggers resolves this.

### DIFF
--- a/source/vir/src/def.rs
+++ b/source/vir/src/def.rs
@@ -402,6 +402,16 @@ pub fn prefix_tuple_param(i: usize) -> Ident {
     Arc::new(format!("{}{}", PREFIX_TUPLE_PARAM, i))
 }
 
+pub fn is_tuple(path: &Path, variant: &Ident) -> bool {
+    let it = path.krate.is_none()
+        && path.segments.len() == 1
+        && path.segments[0].starts_with(PREFIX_TUPLE_TYPE);
+    if it {
+        assert!(variant.starts_with(PREFIX_TUPLE_TYPE));
+    }
+    it
+}
+
 pub fn prefix_spec_fn_type(i: usize) -> Path {
     let ident = Arc::new(format!("{}{}", PREFIX_SPEC_FN_TYPE, i));
     Arc::new(PathX { krate: None, segments: Arc::new(vec![ident]) })
--- a/source/vir/src/triggers_auto.rs
+++ b/source/vir/src/triggers_auto.rs
@@ -323,7 +323,11 @@ fn gather_terms(ctxt: &mut Ctxt, ctx: &Ctx, exp: &Exp, depth: u64) -> (bool, Ter
             let (is_pures, terms): (Vec<bool>, Vec<Term>) =
                 args.map(|e| gather_terms(ctxt, ctx, &e.a, depth + 1)).unzip();
             let is_pure = is_pures.into_iter().all(|b| b);
-            (is_pure, Arc::new(TermX::App(App::Ctor(path.clone(), variant), Arc::new(terms))))
+            if crate::def::is_tuple(path, &variant) {
+                (false, Arc::new(TermX::App(ctxt.other(), Arc::new(terms))))
+            } else {
+                (is_pure, Arc::new(TermX::App(App::Ctor(path.clone(), variant), Arc::new(terms))))
+            }
         }
         ExpX::NullaryOpr(_) => (false, Arc::new(TermX::App(ctxt.other(), Arc::new(vec![])))),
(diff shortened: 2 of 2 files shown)


JoPolzin:
This is the new output for the issue #695 :)
```
error: Could not automatically infer triggers for this quantifer.  Use #[trigger] annotations to manually mark trigger terms instead.
 --> tuple.rs:6:12
  |
6 |   requires forall |x| f.requires((x,))
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

JoPolzin:
@utaal @Chris-Hawblitzel can you take a look at the failing tests? It looks like requires and ensures clauses for  closures depend on triggering on tuples. Does that change anything?

utaal:
It does change things. I don't yet know in what way exactly. Perhaps we can allow them just for closures.

@Chris-Hawblitzel are you already looking into it? (I saw you self-assigned as reviewer)

Chris-Hawblitzel:
> @Chris-Hawblitzel are you already looking into it? (I saw you self-assigned as reviewer)

I haven't actually started.  I'm happy to work on it, but feel free to take it if you want.

PR #1204, opened 2024-07-01
run the verus binary we produce to check it doesn't fail trivially

utaal:


utaal:
Thank you @jaylorch with the help with the Windows command line.

PR #1206, opened 2024-07-03
Faster proof of `lemma_hoist_over_denominator`

jaylorch:
Every time we build `vstd` as part of building Verus, the verifier complains that `lemma_hoist_over_denominator` takes well over 2 seconds to verify. This fixes that issue.


parno (approved):
Very elegant!  Thanks for doing this -- that complaint has been bugging me for a while.

PR #1207, opened 2024-07-04
user-defined type invariants (Part I)

tjhance:
This adds the ability to specify 'type invariants' on structs & enums. A type invariant is a boolean predicate that is enforced for every exec-mode and tracked-mode instantiation of the datatype. There are 3 parts to this feature, and this PR implements 2 of them:

- [x] Ability to specify a type invariant
- [x] Add assertions that the type invariant is maintained on every update and Ctor node
- [ ] Allow the user to assume the type invariant holds true anywhere they want

Despite the feature being incomplete (and therefore useless) this PR is ready for review. Inserting the assertions is the most important part from a soundness perspective, and this aspect should be completely handled by this PR.

This PR introduces a new pass, defined in `user_defined_type_invariants.rs` which adds assertions for every relevant Ctor node and assignment operation. The assertions are added as `AssertAssumeUserDefinedTypeInvariant` nodes. This transformation uses information from the mode-checking pass, and it is called directly from modes.rs.

I originally intended to insert these obligations at a much lower level, but since it uses information from mode-checking, I think keeping it near the mode-checking is a good idea. Also, putting this information into the AST early helps for recursion checking and pruning, which we can handle easily by adding edges that point from the `AssertAssumeUserDefinedTypeInvariant` nodes to the invariant predicate fn.

A user-defined type invariant is required to have the same generic args and trait bounds as the type it applies to. In recursion-checking, we also add edges from the `AssertAssumeUserDefinedTypeInvariant` to the relevant trait bounds.

### DIFF
--- a/source/rust_verify_test/tests/common/mod.rs
+++ b/source/rust_verify_test/tests/common/mod.rs
@@ -458,10 +458,7 @@ pub fn relevant_error_span(err: &Vec<DiagnosticSpan>) -> &DiagnosticSpan {
         return e;
     }
     err.iter()
-        .filter(|e| {
-            e.label != Some(vir::def::THIS_PRE_FAILED.to_string())
-                && e.label != Some("type invariant declared here".to_string())
-        })
+        .filter(|e| e.label != Some(vir::def::THIS_PRE_FAILED.to_string()))
         .next()
         .expect("span")
 }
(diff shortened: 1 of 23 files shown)


source/rust_verify_test/tests/common/mod.rs
```diff
@@ -457,7 +457,10 @@ pub fn relevant_error_span(err: &Vec<DiagnosticSpan>) -> &DiagnosticSpan {
         return e;
     }
     err.iter()
-        .filter(|e| e.label != Some(vir::def::THIS_PRE_FAILED.to_string()))
+        .filter(|e| {
+            e.label != Some(vir::def::THIS_PRE_FAILED.to_string())
+                && e.label != Some("type invariant declared here".to_string())
```
utaal, line 462: This special case should probably not belong here.
(I'm posting this as a single comment because I chanced upon it while trying to look at tests.)

tjhance: do you have a suggested alternative?

Chris-Hawblitzel (approved):

source/rust_verify_test/tests/common/mod.rs
tjhance: I took another look and your suggestion made complete sense this time. Not sure what I was thinking earlier.

utaal (approved):
I got delayed on this, and it doesn't make sense to hold it up. Sorry for the wait. I'll look at it later, but feel free to merge.

PR #1211, opened 2024-07-06
Split ast-to-sst code out of func_to_air and into ast_to_sst_func

Chris-Hawblitzel:
This just moves some code around so that the AST-to-SST code is in a separate file (ast_to_sst_func.rs) from the SST-to-AIR code in func_to_air.rs.  I think this makes func_to_air.rs easier to understand, since it now operates just on vir::sst::Exp and air::ast::Expr, without involving vir::ast::Expr.

I'm thinking of this as an initial step towards a simpler, more cohesive AST-to-SST translation, but I'd like to merge this initial step first so that the later steps are clearer and easier to review.

### DIFF
--- a/source/rust_verify/src/commands.rs
+++ b/source/rust_verify/src/commands.rs
@@ -10,11 +10,11 @@ use vir::ast::Visibility;
 use vir::ast::{
     Fun, Function, FunctionKind, ImplPath, ItemKind, Krate, Mode, Path, TraitImpl, VirErr,
 };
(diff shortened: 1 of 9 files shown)


Chris-Hawblitzel:
I'd also like to rename `func_to_air.rs` to `sst_to_air_func.rs`, but that can be done separately.

utaal (approved):

tjhance (approved):

PR #1215, opened 2024-07-08
Expand documentation on performance measurement.  Also cleanup the ToC a bit.

parno:


source/docs/guide/src/SUMMARY.md
```diff
@@ -65,14 +66,18 @@
 - [Traits]()
 - [Ghost and tracked variables]()
 - [Concurrency and Unsafe Code](concurrency_and_unsafe.md)
+- [Attributes and directives]()
+    - [external and external_body]()
+    - [inline]()
+    - [opaque]()
+    - [decreases_by]()
+    - [when_used_as_spec]()
```
tjhance, line 74: Do you have a plan for this section? I removed it intentionally when I added the attributes chapter to the reference. These attributes don't really have a unified theme, so doesn't really make sense for a tutorial chapter.

parno: No, I didn't add those.  I think they got pulled in when I merged `main` into this branch.  If you previously removed them, I'll remove them here.

PR #1224, opened 2024-07-13
Add cvc5 support to vargo

parno:
This involved factoring out functionality related to: (1) finding the solver binary, (2) checking the solver's version, and (3) copying the solver to Verus's build directory.  The Z3 functionality remains the same.  For cvc5, we don't warn the user if the cvc5 binary is missing, but if it's present, we do check that it's the expected version.


utaal (approved):
Thank you.
I made some defensive changes to the `vargo` code, and fixed how the CVC5 binary is picked up by the verus driver when running from a verusroot (i.e. `target-verus/release/verus`): please check the changes look good, and then feel free to merge (without squashing).

PR #1226, opened 2024-07-14
Implement FunctionSst and KrateSst, for whole-crate AST -> SST translation

Chris-Hawblitzel:
This pull request introduces a complete SST-level crate representation, `vir::sst::KrateSst`, which contains a vector of complete SST-level functions, `vir::sst::FunctionSst`.  It adds `ast_to_sst_crate.rs` to translate the AST `Krate` to the SST `KrateSst`.  This translation splits the AST processing into multiple steps: first, translate the functions, then process triggers, inlining, and computation as separate post-translation elaborations (see `sst_elaborate.rs`).  This simplifies the translation, for example by removing SstMap from the translation.

This pull request also more clearly separates AST from SST, so that `verify_bucket` works entirely on `KrateSst`, not `Krate`, and `commands.rs` focuses exclusively on SST -> AIR translation without having to worry about AST -> SST.  In the future, this should help to implement `poly.rs` and/or monomorphization at the SST level.

This continues https://github.com/verus-lang/verus/pull/1211 .

I've checked the verification performance using Veritas and it is indistinguishable from before the pull request.


tjhance:
Does this mean that the AST -> SST pipeline is run all functions, even if only a single module is verified?

I assume veritas mostly checks full crates. Did you check if there's a performance difference on single-module verification?

Chris-Hawblitzel:
> Does this mean that the AST -> SST pipeline is run all functions, even if only a single module is verified?

The AST -> SST pipeline runs on exactly the same functions as before; this hasn't changed.  This is because `ast_to_sst_krate` is only called from inside `verify_bucket_outer`, after per-bucket pruning of the AST `Krate`.  So if only a single module is verified, only the functions for that module will go through the AST -> SST pipeline, just as before.

> I assume veritas mostly checks full crates. Did you check if there's a performance difference on single-module verification?

I checked; the performance is the same for a small single module.

tjhance (approved):

utaal:
This is great and I'm sorry I never got around to reviewing it.

PR #1230, opened 2024-07-18
Remove extraneous sentence in map_values docs

pratapsingh1729:
Tiny tweak to the docs for this function: the `f` function in `map_values` doesn't take an `int` parameter.


parno (approved):
Good catch!

PR #1244, opened 2024-08-15
Ptrs

tjhance:
Bunch more work on ptrs

 * casting ptrs to int types with `as` (same as calling .addr())
 * allocation & deallocation
 * Implement a new version of PPtr that resembles the original version we presented in the oopsla paper. This one is *verified* in terms of the raw ptr API
 * tweak the way ZSTs work. Rust docs guarantee pointer operations with ZSTs are always allowed for any pointer, so our operations on PointsTo<V> for 0-sized types now reflects this
 * verus-mimalloc has been updated to verify against this branch (see https://github.com/verus-lang/verified-memory-allocator/tree/new-ptr)
 * Support `AtomicPtr` (this wasn't as useful as I hoped. I updated verus-mimalloc to use AtomicPtr but because comparing an address isn't the same thing as comparing a pointer, the 'compare_exchange' method isn't as useful as you'd want it to be, so it doesn't really give you any capabilities that AtomicUsize doesn't have)

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -892,7 +892,7 @@ pub const fn mul<IntegerType: Integer>(_left: IntegerType, _right: IntegerType) 
 #[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::builtin::spec_cast_integer"]
 #[verifier::spec]
-pub const fn spec_cast_integer<From: Integer, To: Integer>(_from: From) -> To {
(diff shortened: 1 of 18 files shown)


tjhance:
In combination with upcoming type-invariant features, we can now encapsulate simpl_pptr's wf() definition and remove it from all the specs:

https://github.com/verus-lang/verus/tree/simple-pptr-type-inv

Chris-Hawblitzel (approved):

PR #1246, opened 2024-08-21
Add is_empty and is_singleton to multiset

rikosellic:
I'm migrating code from Set to MultiSet and found that there is no spec function `is_empty` and `is_singleton` for MultiSet. So I add these two functions in `multiset.rs `and prove several lemmas about them imitating those in `set_lib.rs`.


source/vstd/multiset.rs
```diff
@@ -701,3 +701,70 @@ pub use assert_multisets_equal_internal;
...
+    {
+        broadcast use group_multiset_axioms;
+
+        self.lemma_contain_elem_equal_singleton(self.choose());
+    }
+
+    /// A multiset has exactly one element, if and only if, it has at least one element with multiplicity 1 and any two elements are equal.
+    pub proof fn lemma_is_singleton(s: Multiset<A>)
+        ensures
+            s.is_singleton() == (s.len() == 1),
```
tjhance, line 751: We usually use `<==>` when comparing bools like this.

tjhance (approved):

source/vstd/multiset.rs
rikosellic: Thanks for the suggestion! It is fixed now.

source/vstd/multiset.rs
```diff
@@ -701,3 +701,70 @@ pub use assert_multisets_equal_internal;
 pub use assert_multisets_equal;
 
 } // verus!
+verus! {
+
+impl<A> Multiset<A> {
```
utaal, line 706: Hi! Thanks for the PR!

Why did you place these in a separate `verus!` macro invocation and in a separate `impl` block for `Multiset`?
They either belong in the main `impl` block, or - more likely, to keep with the convention we have for `seq` and `set` - they should probably go in a new file:  `multiset_lib.rs`.

rikosellic: Thanks! I have moved the code to `multiset_lib.rs`.

utaal (approved):

PR #1252, opened 2024-09-03
guide: organize ToC a bit

tjhance:
move chapters around and add major headings to break it up a bit

### DIFF
--- a/source/docs/guide/src/develop_proofs.md
+++ b/source/docs/guide/src/develop_proofs.md
@@ -1,6 +1,6 @@
 # Using assert and assume to develop proofs
 
-The previous section started with an outline of proof:
+In [an earlier chapter](./spec_lib.md), we started with an outline of a proof:
 
 ```rust
 {{#include ../../../rust_verify/example/guide/lib_examples.rs:lemma_len_intersect_fail}}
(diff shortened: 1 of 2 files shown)


parno:
This looks like a nice improvement.  While we're at it, can you move the "[INTERLUDE: using assert and assume to develop proofs](develop_proofs.md)" bit to be in the "Understanding the prover" section?

Chris-Hawblitzel:
> This looks like a nice improvement. While we're at it, can you move the "[INTERLUDE: using assert and assume to develop proofs](develop_proofs.md)" bit to be in the "Understanding the prover" section?

It might work well being the very first subsection of "Understanding the prover" (and dropping the word "INTERLUDE").

tjhance:
I considered it, given what we discussed yesterday, but the content of that chapter relates to the previous chapter. I thought it would require a bit more thought.

tjhance:
ok since both chris & bryan suggested it, I moved the chapter and adjusted its introductory text.

source/docs/guide/src/develop_proofs.md
```diff
@@ -1,6 +1,6 @@
 # Using assert and assume to develop proofs
 
-The previous section started with an outline of proof:
+In [an earlier chapter](./spec_lib.md), we started with an outline of proof:
```
parno, line 3: I think we want "of [a] proof" (I know that was there before you started, but I just noticed it now).

PR #1253, opened 2024-09-04
Make the version check in build.rs a bit more liberal

parno:
When running Verus via VS Code on macOS, I've been seeing a panic in the `build.rs` script:
```
process didn't exit successfully: `/Users/parno/git/verus/source/target/debug/build/verus-9dd4b2228424913b/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at verus/build.rs:39:9:
  unexpected output from `rustup show active-toolchain`
  expected a toolchain override
  got: /Users/parno/.rustup/toolchains/1.76.0-aarch64-apple-darwin (environment override by RUSTUP_TOOLCHAIN)
```
It appears that the script expects the version to be at the very beginning of the string, while the `rustup` output includes a path at the beginning.  Tweaking the regex fixes the issue.

### DIFF
--- a/source/verus/build.rs
+++ b/source/verus/build.rs
@@ -15,7 +15,7 @@ fn main() {
         panic!("rustup failed");
     }
     let active_toolchain_re =
-        Regex::new(r"^(([A-Za-z0-9.-]+)-(?:aarch64|x86_64)-[A-Za-z0-9]+-[A-Za-z0-9-]+)").unwrap();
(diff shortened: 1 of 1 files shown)


utaal:
Do you by chance have a manual override set using a path? That's something we moved away from, and I'm not sure it's supported well. Can you please run `rustup override list`?

utaal:
Nevermind. Something seems to have changed, perhaps in `rustup`. I'm looking into it.

utaal:
I did something a bit different to maintain robustness here https://github.com/verus-lang/verus/commit/6a9c17ee2f6567c546da513d6f6f8b20d9c1a546 (and fixed a related new issue in the meantime).

PR #1260, opened 2024-09-09
refinement types for tracked values

zeldovich:
This seems like a simple and hopefully sound implementation of refinement types for tracked values.  This doesn't work for arbitrary non-tracked ghost values, because every Verus type is inhabited and there's the choose operator, but choose is mode spec, not mode proof, so it cannot synthesize tracked values.  In a cursory discussion, @jaylorch seemed to think this might be sound, so I'll tag him so he can look and chime in.

### DIFF
--- a/source/rust_verify/example/refined.rs
+++ b/source/rust_verify/example/refined.rs
@@ -1,0 +1,49 @@
+use vstd::prelude::*;
+use vstd::refined::*;
+
+verus! {
+    pub struct T {
+        pub a: u8,
+    }
+
+    pub struct TPred {
+    }
+
+    impl RefinedPred<T> for TPred {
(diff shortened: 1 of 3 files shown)


jaylorch:
Thanks! Looks sound to me. If we decide to add this to `vstd`, we'll need a test file in `rust_verify/tests/`.

jaylorch:
It seems like we could also have a refinement type that wraps a ghost value. The refinement wrapper would have to be tracked, but the element inside could be ghost.

zeldovich:
Good point -- the refined type needs to be tracked, but the underlying value could be either tracked or ghost.

I'll try to figure out the test framework thing at some point for this.

PR #1272, opened 2024-09-19
Fix `Vec::split_off` spec

zhengyao-lin:
Hello! Just a minor fix on the spec of `Vec::split_off` to avoid panics (per [Vec doc](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off))


parno:
Good spot!

PR #1276, opened 2024-09-20
add subrange_of_matching_take lemma to seq_lib

jonhnet:
Here's a lemma about Seq `take` and `subrange` I found valuable in verisplinter. I think it belongs in the seq library.

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -1707,6 +1707,23 @@ pub proof fn lemma_seq_take_index<A>(s: Seq<A>, n: int, j: int)
 {
 }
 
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
I was surprised by the sudden introduction of the additional newline (after the comment) to satisfy the formatter in your second commit (94676cf495fc1d7b995822dba24c21f3174998ad).  Turns out this was a subtle weirdness that was introduced likely when adding support for clauses/stanzas.  This is now fixed in https://github.com/verus-lang/verusfmt/pull/89 and should hopefully soon be part of the next verusfmt release.

Issue #1108, opened 2024-05-09
Invalid AIR generation when returning unit type


pratapsingh1729:
The following code:
```
use vstd::prelude::*;

verus! {

pub fn owl_unit() -> (res: ())
    ensures res == ()
{ () }

fn main() {
    owl_unit()
}

}
```
panics the verifier, as follows:
```
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:728:21:
internal error: generated ill-typed AIR code: error 'in call to ens%main!owl_unit., expected 2 arguments, found 1 arguments' in expression '(ens%main!owl_unit. 0)'
note: [run with `RUST_BACKTRACE=1` environment variable to display a backtrace](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021#)
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:339:17:
dropped, expected call to `into_inner` instead
stack backtrace:
...
```
[Minimal playground link with full backtrace](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=8a7fbc93d90dc20d79607c7d4957dae9)

On the other hand, removing the `ensures res == ()` on `owl_unit()` makes the panic go away ([link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=277c4b487712b6c31225a5caad150719)).

Thanks to parametricity, the `ensures res == ()` spec is not actually needed, but noting in case it points to a bigger bug.

Chris-Hawblitzel:
I'm adding other cases to this issue where using a unit return type can result in invalid AIR code.  For example (from https://github.com/verus-lang/verus/issues/1161 ):

```
trait T {
    spec fn f();
}

spec fn f<A: T>() {
    A::f()
}
```

tjhance:
Here's another one:

```rust
trait Tr : Sized {
    fn get() -> (r: Self)
        ensures r == r;
}

impl Tr for () {
    fn get() -> (r: Self)
    {
    }
}
```

tjhance:
With PR https://github.com/verus-lang/verus/pull/1280, all issues mentioned in this thread so far will be fixed, with the exception of:

```
trait T {
    spec fn f();
}

spec fn f<A: T>() {
    A::f()
}
```

This is a poly issue, so should be revisited after the upcoming poly refactor.

PR #1280, opened 2024-09-21
more consistent handling of unit-return functions

### LINKED ISSUE #1278: internal error: generated ill-typed AIR code, when using return type () in trait
Running the latest verus on the following code:

```
vstd::prelude::verus! {
    pub trait Trait<T> {
        spec fn ok(r: T) -> bool;
        proof fn apply() -> (result: T) ensures Self::ok(result);
    }

    pub struct S {
    }

    impl Trait<()> for S {
        open spec fn ok(r: ()) -> bool { true }
        proof fn apply() -> (result: ()) {}
    }
}
```

produces:

```
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:780:21:
internal error: generated ill-typed AIR code: error 'error 'error 'use of undeclared variable %return!' in expression '%return!'' in expression '(bug!Trait.ok.? $ TYPE%bug!S. $ TYPE%tuple%0. %return!)'' in expression '(%B (bug!Trait.ok.? $ TYPE%bug!S. $ TYPE%tuple%0. %return!))'
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:2110:29:
worker thread panicked
```

tjhance:
fixes #1278, fixes many sub-issues in #1108

 - handle traits properly: make sure every trait method impl has the same 'ens_has_return' status as its trait method decl
 - if the user names the unit-return value and references it in the ensures clause, substitute it out.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -29,7 +29,7 @@ use vir::ast::{
     Fun, FunX, FunctionAttrsX, FunctionKind, FunctionX, GenericBoundX, ItemKind, KrateX, Mode,
     ParamX, SpannedTyped, Typ, TypDecoration, TypX, VarIdent, VirErr,
 };
-use vir::ast_util::air_unique_var;
+use vir::ast_util::{air_unique_var, clean_ensures_for_unit_return};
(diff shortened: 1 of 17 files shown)


Chris-Hawblitzel (approved):
Thanks for fixing this

PR #1285, opened 2024-09-27
add pcm::Resource::validate_2_ref() that does not require &mut self

zeldovich:
The current `vstd::pcm::Resource::validate_2()` requires a `&mut self` to ensure that `self` and `other` are not refs to the same underlying `Resource` object.  If they were actually the same `Resource`, then asserting `P::op(self.value(), other.value()).valid()` would be unsound, as pointed out by @jaylorch.  However, this makes it inconvenient to implement agreement between two fractional-ownership resources.  For that use case, even if they are aliased, it's still true that their values are equal.

This PR tries to address this, by introducing `vstd::pcm::Resource::validate_2_ref()` that does not require a `&mut self`.  To account for the unsound possibility, the `ensures` clause of `validate_2_ref()` says that it might be that `self==other`.  I believe this makes it sound, and at the same time, useful for implementing agreement between two fractional resources, regardless of whether they are aliases to the same fractional ownership or not.

@jaylorch, does this look reasonable to you?

### DIFF
--- a/source/vstd/pcm.rs
+++ b/source/vstd/pcm.rs
@@ -208,6 +208,16 @@ impl<P: PCM> Resource<P> {
     }
 
     #[verifier::external_body]
+    pub proof fn validate_2_ref(tracked &self, tracked other: &Self)
+        requires
+            self.loc() == other.loc(),
+        ensures
+            P::op(self.value(), other.value()).valid() || self == other,
+    {
+        unimplemented!();
+    }
+
+    #[verifier::external_body]
     pub proof fn update_with_shared(
         tracked self,
         tracked other: &Self,


tjhance:
I don't believe this is sound, since you could have overlapping resources that are unequal. For example, using a map resource you could have:

 * `self = [0 => x, 1 => y]`
 * `other = [0 => z, 2 => w]`

These are unequal, but they overlap so their composition would be invalid.

For solving your problem with fractional agreement, I suspect you'll find the existing `join_shared` useful:

```rust
    pub proof fn join_shared<'a>(tracked &'a self, tracked other: &'a Self) -> (tracked out: &'a Self)
        requires
            self.loc() == other.loc(),
        ensures
            out.loc() == self.loc(),
            incl(self.value(), out.value()),
            incl(other.value(), out.value()),
```

tjhance:
By the way, the interactions between shared references and RA-based ghost state are a major topic of [my thesis](https://tjhance.github.io/travis_hance_thesis.pdf). Chapter 4 covers Verus's PCM trait & the intuition behind the less-standard rules that manipulate shared references.

zeldovich:
Thanks for the `join_shared` suggestion; that indeed seems to work nicely!  I don't have any remaining motivation for this PR, so I'll close it.

For the counterexample you mentioned, I think those conflicting `self` and `other` resources should not exist at the same time; if they do co-exist, that's already unsound..

tjhance:
> For the counterexample you mentioned, I think those conflicting self and other resources should not exist at the same time; if they do co-exist, that's already unsound..

It's possible to get them using the `weaken`. You start with `[0 => x, 1 => y, 2 => z]`, then use `weaken` to get:

 - `&[0 => x, 1 => y, 2 => z]` -> `&[0 => x, 1 => y]`
 - `&[0 => x, 1 => y, 2 => z]` -> `&[0 => x, 2 => z]`

However, I think it would be the case that your suggested rule would be sound if not for the presence of rules like `weaken` that can adjust the values of shared-references-of-ghost-state.

tjhance:
(Also the first version of the counterexample I posted was bad, and I edited it, so maybe you saw the original version.)

zeldovich:
Oh, interesting..  Reading your chapter 4 and your comment, I now realize that `weaken()` is doing something quite un-intuitive when coming from an Iris perspective..  Thanks for clarifying this!

PR #1290, opened 2024-09-30
Add missing dep state_machines_macros in vstd

ziqiaozhou:
I want to use a customized rustc to call verus for crates that needs verification in my project. I will use vstd as a dependency and built by cargo, since I will need to change build target. 

Everything works except for the missing dependencies state_machines_macros.

### DIFF
--- a/source/vstd/Cargo.toml
+++ b/source/vstd/Cargo.toml
@@ -15,6 +15,7 @@ path = "vstd.rs"
 [dependencies]
 builtin_macros = { path = "../builtin_macros" }
 builtin = { path = "../builtin" }
+state_machines_macros = { path = "../state_machines_macros" }
 
 [features]
 default = ["std"]


utaal:
I don't believe `state_machine_macros` is used in `vstd` (I think we have CI that checks that `vstd` builds successfully). If you're using `state_machine_macros`, can you add it as a dependency of those crates?

I may be missing something about what you're trying to do, though.

tjhance:
state_machines_macros is used in vstd/shared.rs

utaal:
Nevermind! I wonder why CI isn't catching that.

utaal (approved):

ziqiaozhou:
> I may be missing something about what you're trying to do, thou

It does not capture it because state_machine_macros crate is hard linked in verus default mode.

Issue #1255, opened 2024-09-05
Equality operators for an enum type are not supported well


zpzigi754:
I tried the below code where an equality operator is used for an enum type.

```
use vstd::prelude::*;

verus! {

#[derive(PartialEq)]
pub enum Fruits {
    APPLE,
    BANANA,
}

pub struct Meal {
    pub fruit: Fruits,
}

impl Meal {
    fn does_not_work(&self) {
        if self.fruit == Fruits::BANANA {
        }
    }
}

fn main() {
}

}
```

It failed with the following error.
```
error: The verifier does not yet support the following Rust feature: ==/!= for non smt equality types
  --> rust_verify/example/guide/enum.rs:17:12
   |
17 |         if self.fruit == Fruits::BANANA {
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 1 previous error
```

I've confirmed that using `match` or `if let` (the same semantic) works well. 
```
use vstd::prelude::*;

verus! {

pub enum Fruits {
    APPLE,
    BANANA,
}

pub struct Meal {
    pub fruit: Fruits,
}

impl Meal {
    fn work(&self) {
        match self.fruit {
            Fruits::BANANA => {},
            _ => {},
        }
    }

    fn work2(&self) {
        if let Fruits::BANANA = self.fruit {
        }
    }
}

fn main() {
}

}
```

Chris-Hawblitzel:
There is a feature called `Structural` that can be used to make this work:

```
#[derive(Structural, PartialEq, Clone, Copy)]
pub enum Fruits {
    APPLE,
    BANANA,
}

pub struct Meal {
    pub fruit: Fruits,
}

impl Meal {
    fn test(&self) {
        if self.fruit == Fruits::BANANA {
        }
    }
}
```

I guess we need to add documentation on this, and perhaps also update the error message to hint at using `Structural`.

zpzigi754:
Thank you for the suggestion of using `Structural`! I've also confirmed that the test code works with `Structural`.

Issue #178, opened 2022-06-15
Trusted pervasive (vstd) types should be `: Structural`


marshtompsxd:
Hi verus team, I was trying to check the equality of two structs that contain Option like this:
```
#[allow(unused_imports)]
use builtin_macros::*;
#[allow(unused_imports)]
use builtin::{ensures};

mod pervasive;
#[allow(unused_imports)]
use pervasive::{*, option::Option, map::*};

impl std::cmp::PartialEq for Option<u32> {
    #[verifier(external)]
    fn eq(&self, other: &Self) -> bool {
        match (self, other) {
            (Option::Some(a), Option::Some(b)) => a == b,
            (Option::None, Option::None) => true,
            _ => false,
        }
    }
}

impl std::cmp::Eq for Option<u32> {}

#[derive(Structural, PartialEq, Eq)]
pub struct MyStructA {
    pub content: Option<u32>,
}

fn MyFun() {
    let s1 = MyStructA{content:Option::Some(1)};
    let s2 = MyStructA{content:Option::Some(1)};
    assert(s1 == s2);
}

fn main() { }
```

However, verus reports the error:
```
error[E0277]: the trait bound `pervasive::option::Option<u32>: builtin::Structural` is not satisfied
   --> rust_verify/example/myexample.rs:23:10
    |
23  | #[derive(Structural, PartialEq, Eq)]
    |          ^^^^^^^^^^ the trait `builtin::Structural` is not implemented for `pervasive::option::Option<u32>`
    |
note: required by a bound in `builtin::AssertParamIsStructural`
   --> /Users/xudongs/verus/source/builtin/src/lib.rs:328:39
    |
328 | pub struct AssertParamIsStructural<T: Structural + ?Sized> {
    |                                       ^^^^^^^^^^ required by this bound in `builtin::AssertParamIsStructural`
    = note: this error originates in the derive macro `Structural` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
verification results:: verified: 0 errors: 0
```
My understanding about the equality check is:
1. #[derive(Structural)] is necessary for the struct we want to check equality for
2. if the struct has #[derive(Structural)], every field in this struct has to have #[derive(Structural)] as well

If both 1 and 2 are true, what would be the best way to check the equality of a struct that contains Option fields?

marshtompsxd:
[...]

PR #1293, opened 2024-10-01
add support for prophecy variables

zeldovich:
This design of prophecy variables follows the "Future is ours" paper: both allocation and resolution of a prophecy variable are exec-mode, and the prophecy value is an executable type.

### DIFF
--- a/source/rust_verify_test/tests/proph.rs
+++ b/source/rust_verify_test/tests/proph.rs
@@ -1,0 +1,68 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] prophecy_expected_use_1 verus_code! {
+        use vstd::proph::*;
+
+        fn test_bool() {
+            let ghost x: int = 1;
+            let p = Prophecy::<bool>::new();
+            proof {
+                if p@ {
+                    x = 2;
+                } else {
+                    x = 3;
+                }
+            }
+
+            p.resolve(&true);
+            assert(x == 2);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] prophecy_expected_use_2 verus_code! {
+        use vstd::proph::*;
+
+        #[derive(PartialEq, Eq, Structural)]
+        struct S {
+            a: u64,
+            b: u8,
+            c: bool,
+        }
+
+        fn test() {
+            let p = Prophecy::<S>::new();
+            p.resolve(&S{a: 1u64, b: 2u8, c: false});
+            assert(p@ == S{a: 1u64, b: 2u8, c: false});
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] prophecy_expected_use_3 verus_code! {
+        use vstd::proph::*;
+
+        fn test() {
+            let p = Prophecy::<bool>::new();
+            p.resolve(&true);
+            assert(p@ == true);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] prophecy_ghost_disallowed verus_code! {
+        use vstd::proph::*;
+
+        fn test() {
+            let p = Prophecy::<Ghost<bool>>::new();
+            p.resolve(&Ghost(!p.view().view()));
+            assert(false);
+        }
+    } => Err(err) => assert_rust_error_msg(err, "trait bounds were not satisfied")
+}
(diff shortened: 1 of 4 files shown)


tjhance:
Unfortunately the mode-restriction doesn't have the intended effect because ghost code can be embedded in exec code:

```
    fn main()
    {   
        let p = Prophecy::<Ghost<bool>>::alloc();
        p.resolve(Ghost(!p.view().view()));

        assert(false);
    }  
```

However, we have a not-documented feature that lets you mark spec functions as 'prophetic':

```
        #[verifier::prophetic]
        pub closed spec fn view(self) -> T { self.v@ }
```

This adds some additional restrictions that prevent time travel paradoxes. Nobody's actually _used_ the feature until now -- I originally implemented it in preparation for mutable reference support, which is still in-progress. However, it should be usable for standard prophecy variables as well, so I'd be happy to accept the PR if you add `#[verifier::prophetic]` to the view function.

There's also a discussion page for the feature here, though it's a little dated: https://github.com/verus-lang/verus/discussions/559

zeldovich:
That's a nice counter-example -- thanks!

I saw the prophetic attribute support, but it seems pretty infectious.  In the specific use case of prophecy variables that I wanted to achieve, this would cause every aspect of the specification to become prophetic, and as a result, hard to use higher up in the stack.  The Iris-style prophecy variables don't require tracking prophecy propagation in spec/proof, which is more ergonomic in some use cases (including mine).

Is there some type constraint that can require the type T for Prophecy<T> to be purely-exec?  As a strawman proposal, how about requiring the type `T` to be marshallable into bytes and back (with a requires/ensures spec that says it has to be idempotent)?

zeldovich:
Or even simpler, how about only allowing prophecies for byte slices?

Chris-Hawblitzel:
We could try `T: Structural`, using `builtin::Structural` ( https://github.com/verus-lang/verus/issues/178,  https://github.com/verus-lang/verus/issues/1255 ) , which guarantees exec equality matches spec equality.

tjhance:
Interesting idea. I think this should work.

tjhance (approved):

Chris-Hawblitzel:
We should probably add Travis's counterexample as a test to make sure it fails to verify.

jaybosamiya-ms:
Minor: when erased, `resolve` has the same signature as `drop`; basically, it consumes the value being resolved.  It is quite possible that people might just use this primarily with things which are freely `Copy`, so maybe this is not a major concern, but nonetheless, it might be worth considering:
```
    pub exec fn resolve(self, v: T) -> (r: T)
        ensures
            self@ == v,
            r == v,
    {
        v
    }
```
This is just so that the value `v` remains usable later on if you want to use it for something else. 

I will note that this _is_ a mostly backwards-compatible change to the signature (unless someone is _depending_ on it returning a `()`, users won't notice if a new return is added), so we can punt this until there is someone who really wants to use these prophecy variables with non-`Copy` types, at which point we can easily add these in.

jaybosamiya-ms:
Another minor comment: perhaps `new` is a more Rust-y idiomatic name than `alloc`?

Chris-Hawblitzel:
Another minor point: since these are `exec` no-ops, you might want to mark them with Rust's `#[inline(always)]` for efficiency.

jaybosamiya-ms:
I don't know if `Structural` always implies `Copy`. At least this example seems to indicate it doesn't? https://github.com/verus-lang/verus/blob/1a2fb586a88a9e1ced7532a14b6acf8b417baade/source/rust_verify/example/adts.rs#L14-L18

I don't have a strong preference on the non-mut-ref vs take-and-return-value approaches; the take-and-return-value has the niceness of not needing `&` at the callers in the common case, while the non-mut-ref is likely more idiomatic to see in the callee.

Not for this PR, but potential future improvement, we don't currently have a support for [`std::borrow::Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html), but that might be a way in the future to have the convenience of not needing the `&` at the caller (for `Copy` types), while also having the semantics of the borrow.  In that case, going with the non-mut-ref approach is the one that will smoothly switch over.

I'm likely over-thinking this :)

zeldovich:
In that ADT example, adding `Clone, Copy` to the `#[derive(...)]` gives `Copy` for `Vehicle`.  The base types that are `Structural` (https://github.com/verus-lang/verus/blob/main/source/builtin/src/lib.rs#L677) are all types that are also `Copy`.  The extra `&` is slightly odd to me, but I don't have a good internal sense of what's Rust-idiomatic here.

PR #1295, opened 2024-10-02
Add `extend` to HashMap and StringHashMap

marshtompsxd:
Add trusted `extend` to vstd HashMap and StringHashMap. The effect of the method is the same as Map's `union_prefer_right`.

### DIFF
--- a/source/vstd/hash_map.rs
+++ b/source/vstd/hash_map.rs
@@ -112,7 +112,7 @@ impl<Key, Value> HashMapWithView<Key, Value> where Key: View + Eq + Hash {
     }
 
     #[verifier::external_body]
-    pub fn extend(&mut self, other: Self)
+    pub fn union_prefer_right(&mut self, other: Self)
         ensures
             self@ == old(self)@.union_prefer_right(other@),
     {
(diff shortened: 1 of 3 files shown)


marshtompsxd:
I was trying to add external spec for HashMap's `extend` as well but then found it more complex than I thought...

source/vstd/hash_map.rs
```diff
@@ -110,6 +110,14 @@ impl<Key, Value> HashMapWithView<Key, Value> where Key: View + Eq + Hash {
     {
         self.m.clear()
     }
+
+    #[verifier::external_body]
+    pub fn extend(&mut self, other: Self)
+        ensures
+            self@ == old(self)@.union_prefer_right(other@),
+    {
+        self.m.extend(other.m)
```
utaal, line 119: Jay Lorch points out that there is no `<HashMap as _>::extend` method on `std::collection::HashMap`. I believe this works because it desugars to `<HashMap as std::tier::Extend>::extend(self.m, other.m)` where the signature of `extend` is `fn [extend](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend)<T>(&mut self, iter: T)
       where T: [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html)<Item = A>;` and `std::collection::HashMap` implements `IntoIterator<Item = (K, V)>`.
So, strictly speaking, this is a different API than the one offered by the regular `HashMap::extend`, so we may want to name it `union_prefer_right`, instead.

jaylorch (approved):
Thanks for your contribution!

source/vstd/hash_map.rs
marshtompsxd: Updated

PR #1297, opened 2024-10-02
Small macros to avoid use of verus!{} for executable codes.

ziqiaozhou:


jaybosamiya-ms:
Minor: `assert` being allowed _outside_ `proof { }` blocks in _regular_ Verus is _incredibly_ useful.  However, when using it in this particular context, it feels basically like a magic function (especially things like `assert (foo) by { bar }` and such will seem ultra magical.  Maybe when in this `#[verus_verify]` mode, we should require assertions to also be inside the `proof! { ... }` macros?

source/builtin_macros/src/lib.rs
```diff
@@ -90,6 +90,38 @@ pub fn verus(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+///   should still use `verus! {}`.
+///
+/// Usage:
+/// - To apply `requires`, `ensures`, `invariant`, or `proof` in `exec`
+///   functions annotated with `#[verus_verify]`, developers should call the
+///   corresponding macros at the beginning of the function or loop.
+///
+/// Rationale:
+/// - This approach avoids introducing new syntax into existing Rust executable
+///   code, allowing verification and non-verification developers collaborate
```
parno, line 109: Very minor: "developers collaborate" -> "developers to collaborate"

source/builtin_macros/src/lib.rs
```diff
@@ -90,6 +90,38 @@ pub fn verus(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+///   without affecting others.
+///   For developers who do not understand verification, they can easily ignore
+///   verus code via feature selection and use standard rust tools like
```
parno, line 112: Very minor: "rust" -> "Rust"

parno (commented):
This is a very interesting idea to try out!  I agree with Jay though that allowing a "bare" Verus-style `assert` outside of a proof block seems likely to be confusing in this setting.  It seems likely to be interpreted as some variant of a dynamic assertion, rather than a static one.

source/builtin_macros/src/lib.rs
```diff
@@ -90,6 +90,38 @@ pub fn verus(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+///   `rustfmt` and `rust-analyzer`.
+///
+/// Example:
+/// - Refer to the `test_my_funs_with_verus_verify` in `example/syntax.rs`.
+#[proc_macro_attribute]
+pub fn verus_verify(
+    attr: proc_macro::TokenStream,
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    syntax::rewrite_items(input, cfg_erase(), true)
```
Chris-Hawblitzel, line 122: This is related to the earlier comments about `assert` being allowed: I didn't expect `verus_verify` to rewrite the entire function.  I would have expected it to just add the necessary attributes around the unmodified function.

Chris-Hawblitzel: Can we just use the existing `#[verifier::verify]`, which doesn't rewrite the body?

ziqiaozhou: That should work. Is #[verifier::verify] supposed to be a stable macro I can directly use outside verus project or I need to define one in builtin_macro to replace it? I think it makes more sense to move assert into proof block proof!{assert(true);}

Chris-Hawblitzel: `#[verifier::verify]` is meant to be stable and visible to users.  I hope it works in regular non-Verus rustc/cargo builds, but I haven't actually tried it.

ziqiaozhou:
After discussing with Chris, I remove verus_verifier since it could be replaced by #[verifier::verify]. It will force users to use proof!{assert(xxx);}, as Jay suggested. I also updated the example to use mutable tracked variable. It is possible to pass mutable Tracked variable, but the way it is used is limited. However, this limitation may be acceptable as it introduces a guiding principle for developers: if they need to perform advanced operations with tracked variables, they should use verus macro to define proof functions, instead of adding complicated verus code into the executable code.

utaal:
This is interesting, but I think we need to be very careful, as we sometimes depend on the `Verus` macro to genderate things in a very specific pattern for us to pick it up.

For example, this won't work with the new way requires and ensures are encoded for consts, so they can still call non-const spec functions (they are wrapped in an additional closure).

Are we confident we can keep these in sync with the behavior of the primary `verus` macro?

ziqiaozhou:
> proc-macro attributes

I was discussing with Chris about whether to use proc-macro attributes, similar to Prusti's style. But we do not have strong preference at that moment. Since you also raised the same question, I feel it should be better to use proc-macro attributes for requires/ensures/invariants. 

I am okay to put it behind an experimental flag, since I have no idea how 1.79 works. 

BTW, I am also looking forward to your 1.79 change since the project I worked on now used 1.78 with inline_const feature.

ziqiaozhou:
I made the changes to use proc-macro attributes. I feel it is necessary to use  proc-macro attributes, especially for trait definition. 

The ideal way to use the new macros if using proc-macro attributes are
```
 #[requires(true)]
fn test() {
    proof!{
        lemma1();
        assert(true);
    }

    #[invariant(true)]
    loop {}
}

#[verus_verify]
trait TestTrait {
    #[requires(true)]
    fn test();
}

#[verus_verify]
impl Y {
    #[requires(true)]
    fn value(&self) -> u32 {
        self.val
    }
}
```

Here I have #[verus_verify] to add #[verifier::verify] if erase.keep() == true, so that I can still build code if I dierctly use rustc. 

I have two designs on mind:
1. #[requires]/#[ensures] proc-macro attributes parse and insert at the beginning of function body or insert additional trait method after the target trait method. #[invariant] insert at the beginning of loop expression block.
2. #[verus_verify] proc-macro attribute  put verifier::verify to all child items if needed and parse the #[requires]/#[ensures], and then do the work in design 1.

Design 1 is more lightweight and each attribute is local, but  I cannot guarantee the #[verifier::verify] is always placed on the top item when using spec attributes (e.g., requires/ensures) in an impl-methods, if the proc-macro attributes are working only on function body. 

Design 2 solved that issue but is more expensive, similar to verus!{} but is only for executable codes.

I prefer Design 1, but I am wondering whether it is possible to detect the misuse of verus annotation in rust_verify level instead of the builtin_macros. When calling requires/ensures/invariants/decreases, rust_verify should throw the errors if #[verifier::verify] is not on the item, instead of silently let it pass the verification.

ziqiaozhou:
I removed the recursive macros from the PR since we may want to keep it as simple as possible in this PR. The only exception is for trait methods. For a trait method, the rewrite is still local on the top item, but the requires and ensures need to be handled at the same time in order to insert them into a new function VERUS_SPEC__x.
In following example, the attribute requires will trigger the rewrite process. ensures are consumed by require attribute.
```
trait T{
#[requires(true)]
#[ensures(true)]
fn x();
}
```

In other cases, each attribute will trigger its own rewrite process.

Chris-Hawblitzel (approved):

jaybosamiya-ms:
Potential suggested syntax?
```rs
#[verus_verify(
  requires
    foo,
    bar,
  ensures |r: u64|
    baz,
    baz2,
)]
```

Chris-Hawblitzel:
I like @jaybosamiya-ms's proposed syntax.  I'm going to go ahead and merge the existing PR first (since it's an easy first step), and then let's see if we can move towards the proposed syntax.

PR #1299, opened 2024-10-07
keep create_open_invariant_credit available in exec mode

zeldovich:
Calls to `create_open_invariant_credit()` are exec-mode, so the function needs to be available for compilation by rustc.  Prior to this fix, compiling a Rust program that used `create_open_invariant_credit()` using rustc/cargo would lead to errors that `create_open_invariant_credit` is missing.

### DIFF
--- a/source/tools/line_count/src/main.rs
+++ b/source/tools/line_count/src/main.rs
@@ -1305,19 +1305,11 @@ impl<'f> Visitor<'f> {
     }
 
     fn fn_code_kind(&self, kind: CodeKind) -> CodeKind {
-        if self.in_state_machine_macro > 0 {
-            kind.join_prefer_left(CodeKind::Spec)
-        } else {
-            kind
-        }
+        if self.in_state_machine_macro > 0 { kind.join_prefer_left(CodeKind::Spec) } else { kind }
     }
 
     fn mode_or_trusted(&self, kind: CodeKind) -> CodeKind {
-        if self.trusted > 0 {
-            CodeKind::Trusted
-        } else {
-            kind
-        }
+        if self.trusted > 0 { CodeKind::Trusted } else { kind }
     }
 
     fn handle_signature(
--- a/source/vstd/invariant.rs
+++ b/source/vstd/invariant.rs
@@ -272,8 +272,7 @@ pub struct OpenInvariantCredit {}
 
 // It's intentional that `create_open_invariant_credit` uses `exec` mode. This prevents
 // creation of an infinite number of credits to open invariants infinitely often.
-#[cfg(verus_keep_ghost)]
-#[rustc_diagnostic_item = "verus::vstd::invariant::create_open_invariant_credit"]
+#[cfg_attr(verus_keep_ghost, rustc_diagnostic_item = "verus::vstd::invariant::create_open_invariant_credit")]
 #[verifier::external_body]
 #[inline(always)]
 pub fn create_open_invariant_credit() -> Tracked<OpenInvariantCredit>
@@ -290,11 +289,13 @@ pub fn create_open_invariant_credit() -> Tracked<OpenInvariantCredit>
 pub proof fn spend_open_invariant_credit_in_proof(tracked credit: OpenInvariantCredit) {
 }
 
-#[cfg(verus_keep_ghost)]
-#[rustc_diagnostic_item = "verus::vstd::invariant::spend_open_invariant_credit"]
+#[cfg_attr(verus_keep_ghost, rustc_diagnostic_item = "verus::vstd::invariant::spend_open_invariant_credit")]
 #[doc(hidden)]
 #[inline(always)]
-pub fn spend_open_invariant_credit(credit: Tracked<OpenInvariantCredit>)
+pub fn spend_open_invariant_credit(
+    #[allow(unused_variables)]
+    credit: Tracked<OpenInvariantCredit>,
+)
     opens_invariants none
     no_unwind
 {


utaal:
It would be nice if we could erase the calls to `create_open_invariant_credit`, but would probably require either special syntax, as otherwise we'd have no guarantee that a call to `create_open_invariant_credit` as seen by the macro is actually referring to this builtin.

@zeldovich do you expect you'll need to create _a lot_ of credits, i.e. do you expect this function to be called often, in what you're working on?

source/vstd/invariant.rs
utaal, line 1: If we use this approach, I suspect it'd be useful to apply the same change to `spend_open_invariant_credit`.

Chris-Hawblitzel: Hopefully `spend_open_invariant_credit` is already erased because it's a proof function and the whole proof block that calls it gets erased.

zeldovich:
I believe the erasure works out fine with this change (but there are some unexpected runtime overheads unrelated to this change).

Specifically, calling `create_open_invariant_credit()`, and using it to open an invariant, both result in no generated executable code in the binary.

But there is an additional two instructions generated for every Tracked argument passed to a function (even for, say, `Tracked<u8>`), e.g.:

```
    #[inline(never)]
    pub fn test4(
        Tracked(i1): Tracked<u8>,
        Tracked(i2): Tracked<u8>,
        Tracked(i3): Tracked<u8>,
        Tracked(i4): Tracked<u8>) {
    }

    test4(Tracked(0u8), Tracked(0u8), Tracked(0u8), Tracked(0u8));
```

produces:

```
0000000000007460 <test::test4::haa48793bf326f54b>:
    7460:       c3                      ret
    7461:       66 2e 0f 1f 84 00 00    cs nopw 0x0(%rax,%rax,1)
    7468:       00 00 00
    746b:       0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)

...
    74b0:       48 8d 05 49 01 00 00    lea    0x149(%rip),%rax        # 7600 <core::ops::function::FnOnce::call_once::h8eb2dea95327446a>
    74b7:       48 89 44 24 40          mov    %rax,0x40(%rsp)
    74bc:       48 8d 05 6d 00 00 00    lea    0x6d(%rip),%rax        # 7530 <core::ops::function::FnOnce::call_once::h159e9cd64cfc67ae>
    74c3:       48 89 44 24 38          mov    %rax,0x38(%rsp)
    74c8:       48 8d 05 a1 01 00 00    lea    0x1a1(%rip),%rax        # 7670 <core::ops::function::FnOnce::call_once::hae3ddab94b18e06b>
    74cf:       48 89 44 24 30          mov    %rax,0x30(%rsp)
    74d4:       48 8d 05 e5 00 00 00    lea    0xe5(%rip),%rax        # 75c0 <core::ops::function::FnOnce::call_once::h6cc2b5664e88c9e7>
    74db:       48 89 44 24 28          mov    %rax,0x28(%rsp)
    74e0:       e8 7b ff ff ff          call   7460 <test::test4::haa48793bf326f54b>
```

That might be worth figuring out and fixing separately.

zeldovich:
> @zeldovich do you expect you'll need to create _a lot_ of credits, i.e. do you expect this function to be called often, in what you're working on?

In the specific use case that I'm helping Jay with, I'm creating one credit per I/O operation to persistent memory.  That might be "a lot" if the I/O operations are small and cheap (e.g., reading a few bytes).  If Verus can support more expressive `opens_invariants` clauses, this might turn into a handful (2-3) credits per I/O operation, but still the same order of magnitude.

zeldovich:
> But there is an additional two instructions generated for every Tracked argument passed to a function (even for, say, `Tracked<u8>`), e.g.: [...]

FWIW, this comes from the `|| unreachable!()` closure that https://github.com/verus-lang/verus/blob/main/source/builtin_macros/src/syntax.rs#L2016 passes to `Tracked::assume_new_fallback()` (and same for `Ghost::assume_new_fallback()`).  https://github.com/verus-lang/verus/commit/37eb85e4ee9ac48147319081fe6a3055de389493 explains the rationale.  It might be that higher optimization passes (e.g., if doing `cargo build --release`) would optimize this away; a bit hard for me to tell if that would always work.

source/vstd/invariant.rs
utaal: fwiw, it's `exec`, not `proof`: https://github.com/verus-lang/verus/blob/main/source/vstd/invariant.rs#L293-L304

`spend_open_invariant_credit_in_proof` is `proof`

utaal:
> FWIW, this comes from the `|| unreachable!()` closure

ah, interesting. Can you remind me how to quickly get the disasm for that function? We might be able to replace it with something that doesn't result in code being emitted.

utaal (approved):
Looks good to me, but @jaylorch is the primary reviewer.

zeldovich:
> Unless @Chris-Hawblitzel objects (in light of the discussion above), it'd be helpful if you made the same change to `spend_open_invariant_credit` before merging.

Sure, I'll do the same, though it doesn't seem to be a problem in practice for some reason (unlike the original problem I had with `create_open_invariant_credit()`).  Perhaps that's because the call site of `spend_open_invariant_credit()` is getting erased altogether?  https://github.com/verus-lang/verus/blob/main/source/vstd/invariant.rs#L420 might be the relevant caller.

zeldovich:
> > FWIW, this comes from the `|| unreachable!()` closure
> 
> ah, interesting. Can you remind me how to quickly get the disasm for that function? We might be able to replace it with something that doesn't result in code being emitted.

I don't have a super-simple way of doing this, but here's what I did:

```
% git clone https://github.com/zeldovich/verus-experiments
% cd verus-experiments/crate
% cargo build
% objdump --disassemble=testfunc ./target/debug/test | c++filt

./target/debug/test:     file format elf64-x86-64


Disassembly of section .init:

Disassembly of section .plt:

Disassembly of section .text:

0000000000006c10 <testfunc>:
    6c10:	48 83 ec 28          	sub    $0x28,%rsp
    6c14:	e8 d7 ff ff ff       	call   6bf0 <test::test1::hedc6bcf51746e5f6>
    6c19:	48 8d 05 a0 01 00 00 	lea    0x1a0(%rip),%rax        # 6dc0 <core::ops::function::FnOnce::call_once::hff3131152b9f721a>
    6c20:	48 89 44 24 20       	mov    %rax,0x20(%rsp)
    6c25:	48 8d 05 d4 00 00 00 	lea    0xd4(%rip),%rax        # 6d00 <core::ops::function::FnOnce::call_once::h6be2388f0b59eaea>
    6c2c:	48 89 44 24 18       	mov    %rax,0x18(%rsp)
    6c31:	48 8d 05 88 00 00 00 	lea    0x88(%rip),%rax        # 6cc0 <core::ops::function::FnOnce::call_once::h66acb4a8e9957f03>
    6c38:	48 89 44 24 10       	mov    %rax,0x10(%rsp)
    6c3d:	48 8d 05 3c 01 00 00 	lea    0x13c(%rip),%rax        # 6d80 <core::ops::function::FnOnce::call_once::hefcb949f993e46ec>
    6c44:	48 89 44 24 08       	mov    %rax,0x8(%rsp)
    6c49:	e8 b2 ff ff ff       	call   6c00 <test::test2::h5fab62352a4a5c7e>
    6c4e:	48 83 c4 28          	add    $0x28,%rsp
    6c52:	c3                   	ret

Disassembly of section .fini:
```

Those four calls to `core::ops::function::FnOnce::call_once::...` correspond to the four `Ghost(...)` values being created in `testfunc` in `src/main.rs`.

source/vstd/invariant.rs
Chris-Hawblitzel: Oh, sorry, I was looking at `spend_open_invariant_credit_in_proof`.  Yes, the same change should be applied to `spend_open_invariant_credit`.

jaylorch (approved):

source/tools/line_count/src/main.rs
```diff
@@ -1305,19 +1305,11 @@ impl<'f> Visitor<'f> {
...
+        if self.in_state_machine_macro > 0 { kind.join_prefer_left(CodeKind::Spec) } else { kind }
     }
 
     fn mode_or_trusted(&self, kind: CodeKind) -> CodeKind {
-        if self.trusted > 0 {
-            CodeKind::Trusted
-        } else {
-            kind
-        }
+        if self.trusted > 0 { CodeKind::Trusted } else { kind }
```
zeldovich, line 1312: On suggestion by @jaylorch, pinging @jaybosamiya-ms about this code where `vargo fmt` and `verusfmt` disagree about the right way to format.  This seems related to the fact that `vargo fmt` passes `--config version=Two` to `rustfmt`.

jaybosamiya-ms: Huh, looks like `vargo fmt` for some reason is passing extra config to `rustfmt` as arguments, rather than use the existing `rustfmt.toml` file (https://github.com/verus-lang/verus/blob/main/source/rustfmt.toml); indeed it even includes the config but comments it out.

Verusfmt is already aware of `rustfmt.toml` and automatically picks it up to use it for configuration, so if the following diff is made to the `rustfmt.toml`:
```patch
@@ -1,7 +1,7 @@
 # Copied from Rust compiler's rustfmt.toml settings

 # Run rustfmt with this config (it should be picked up automatically).
-# unstable_features=true # provided by vargo
-# version=Two # provided by vargo
+unstable_features = true
+version = "Two"
 use_small_heuristics = "Max"
 merge_derives = false
```
then Verusfmt is able to successfully pick that information up and use it.

@utaal: do you know why you are passing it as CLI arguments and keeping it commented-out in the `rustfmt.toml`? From a quick test with the above diff, I see it successfully being picked up both by rustfmt as well as verusfmt, so unless there is a good reason to not use it in the TOML, can we switch it over to only be in the TOML?

In particular the following bits of code would need to be updated:
https://github.com/verus-lang/verus/blob/f2bf563b9f8995b539da4dab2d07ac5f3442c65b/source/rustfmt.toml#L4-L5 and https://github.com/verus-lang/verus/blob/f2bf563b9f8995b539da4dab2d07ac5f3442c65b/tools/vargo/src/main.rs#L764-L765

I'll open a PR shortly with the changes

jaybosamiya-ms: Haha, just ran a full test, and found out what the issue is: it requires nightly only to be able to place it into the `.toml`, but it worked for my separate test due to my local config. Going to have to think about this a little bit.

jaybosamiya-ms: Funnier: support for `version=Two` is gated under `unstable_features=true` which is gated under nightly. For at least 4 years, people have been complaining about "allow us to pass `unstable_features=true`" and the stance seems to be that there should be literally no way for a stable rustfmt to have support for unstable features in their config, at least until rustfmt 2.x comes out; latest post on that ~4 year old thread is from ~2 years ago saying "Sorry to necrobump, but the mythical rustfmt version 2.0 seems not to be released as of March 2022. Is it possible to estimate how long is it going to finally release 2.0 to the public?", so I'm not holding out much hope.

jaybosamiya-ms: Fwiw, the _specific_ example of the `line_count` tool itself I think is not an important one to worry about, since it is isn't actually Verus code meant to be formatted with `verusfmt`, so this _particular_ example should not matter.  Literally, verusfmt looks at it, decides that it has no `verus! { ... }` code, so just invokes rustfmt directly, which picks up the `rustfmt.toml` config, and runs with it.

For most software, I don't actually expect them to be using `version=Two` on a stable Rust (I'm somewhat surprised we're using it; if we were on nightly, we wouldn't have had this issue, and the toml would work just fine).

If we remove `version=Two` from Verus entirely, the diff isn't _massive_; just some parts of the code (mostly if-else blocks) become more spread out across lines, rather than single-line.  Something to maybe consider if we find the disagreement to be an issue.  The other option is that I could expose a `--rustfmt-config` flag within Verusfmt for passing CLI arguments of the kind that rusfmt would take.  Yet another option is for such situations to run verusfmt with `--verus-only` and just use `rustfmt` as it expects to do so on the file, as a separate tool, rather than attempting to have Verusfmt handle everything.

PR #1302, opened 2024-10-09
warn that `assert forall |...| p ==> q` does not assume the antecedent

utaal:
Based on a discussion with @jaylorch, @jaybosamiya-ms, @Chris-Hawblitzel.

This is the first step addressing the following pain point:

@jaylorch said:
> I just spent a while debugging a proof that was ultimately due to a common mistake:  I had said `assert forall|x| P ==> Q by {...}` instead of `assert forall|x| P implies Q by {...}`

This can happen frequently when copy-pasting a failing post-condition or assertion. We intend ultimately to always assume the antecedent in the `assert forall` body, unless we get reports that users who get this warning don't want the new proposed behavior.

### DIFF
--- a/source/vir/src/well_formed.rs
+++ b/source/vir/src/well_formed.rs
@@ -169,6 +169,7 @@ fn check_one_expr(
     expr: &Expr,
     disallow_private_access: Option<(&Option<Path>, &str)>,
     place: Place,
+    diags: &mut Vec<VirErrAs>,
 ) -> Result<(), VirErr> {
     match &expr.x {
         ExprX::Var(x) => {
@@ -373,6 +374,17 @@ fn check_one_expr(
                 VisitorControlFlow::Stop(e) => Err(e),
             }?;
         }
+        ExprX::AssertBy { ensure, vars, .. } => match &ensure.x {
+            ExprX::Binary(crate::ast::BinaryOp::Implies, _, _) => {
+                if !vars.is_empty() {
+                    diags.push(VirErrAs::Warning(
+                        error(&expr.span, "using ==> in `assert forall` does not currently assume the antecedent in the body; consider using `implies` instead of `==>`")
+                            .help("If you didn't mean to assume the antecedent, we're very curious to hear why! To tell us, please open an issue on the Verus issue tracker on github with the title `Don't always make assert forall assume the antecedent`. If no one opens such an issue, we'll soon change the behavior of Verus to always assume the antecedent of the outermost implication")
+                    ));
+                }
+            }
+            _ => {}
+        },
         ExprX::ExecClosure { params, ret, .. } => {
             for p in params.iter() {
                 check_typ(ctxt, &p.a, &expr.span)?;
@@ -456,9 +468,10 @@ fn check_expr(
     expr: &Expr,
     disallow_private_access: Option<(&Option<Path>, &str)>,
     place: Place,
+    diags: &mut Vec<VirErrAs>,
 ) -> Result<(), VirErr> {
     crate::ast_visitor::expr_visitor_check(expr, &mut |_scope_map, expr| {
-        check_one_expr(ctxt, function, expr, disallow_private_access, place)
+        check_one_expr(ctxt, function, expr, disallow_private_access, place, diags)
     })
 }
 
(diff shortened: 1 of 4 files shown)


source/vir/src/well_formed.rs
```diff
@@ -372,6 +373,17 @@ fn check_one_expr(
                 VisitorControlFlow::Stop(e) => Err(e),
             }?;
         }
+        ExprX::AssertBy { ensure, vars, .. } => match &ensure.x {
+            ExprX::Binary(crate::ast::BinaryOp::Implies, _, _) => {
+                if !vars.is_empty() {
+                    diags.push(VirErrAs::Warning(
+                        error(&expr.span, "using ==> in `assert forall` will not assume the antecedent in the body, and is deprecated, consier using `implies` instead of `==>`
```
parno, line 380: I think we want `; consider` instead of `, consier`.

Thinking about it more, I worry that this warning might confuse people (or perhaps I'm just confused).  To me, the warning sounds like we want everyone to stop using `==>` in assertions about `forall`.  But that's not actually what we want long-term; in the near future, we want them to only use `==>` and not `implies`.

Perhaps instead we could say:
```
using ==> in `assert forall` does not currently assume the antecedent in the body.  If you want to assume the antecedent, consider using `implies` instead of `==>`.  If you do not intend to assume the antecedent, please ...
```

utaal: Thanks for the input. I wasn't particularly happy with my formulation, and I do like your formulation better. I'll make the change. @jaybosamiya-ms @jaylorch any other opinions on this?

jaylorch: I agree that we shouldn't say that it's deprecated, because it's going to be the way to go soon.

source/vir/src/well_formed.rs
```diff
@@ -372,6 +373,17 @@ fn check_one_expr(
...
+if you did not mean to assume the antecedent, please open an issue on the Verus issue tracker: `assert forall` will always assume the antecedent of the outermost implication in the future"),
```
jaylorch, line 381: I found this somewhat confusing. How about this instead? "If you didn't mean to assume the antecedent, we're very curious to hear why! To tell us, please open an issue on the Verus issue tracker on github with the title `Don't always make assert forall assume the antecedent`." If no one opens such an issue, we'll soon change the behavior of Verus to always assume the antecedent of the outermost implication.

jaylorch: By the way, I've started using it and it's working great. I found a case where I was using `assert forall |x| A ==> B ==> C`.

source/vir/src/well_formed.rs
jaybosamiya-ms: How about something like:
> using `==>` in `assert forall` does not currently assume the antecedent in the body.  This behavior is intended to be changed.  If you want to assume the antecedent, consider using `implies` instead of `==>`.  If you do not intend to assume the antecedent, then please report to the Verus developers with your specific use case (with `--record`), to aid the Verus developers in making an informed decision.

It _is_ quite verbose, but it is intended to be a short term warning, that goes away when we deprecate `implies`, so I'd rather it carry enough information for people to report it to us, than it be missed.

source/vir/src/well_formed.rs
jaybosamiya-ms: I wrote an alternate version (https://github.com/verus-lang/verus/pull/1302#discussion_r1794440631) in the other comment thread before noticing @jaylorch's comment here.  I like the version with the "If no one opens ..." which makes the plan incredibly clear.

utaal:
Thank you all for the input!

Here's the new output:
```rust
warning: using ==> in `assert forall` does not currently assume the antecedent in the body, consier using `implies` instead of `==>`
  --> .../verus/source/target/debug/test_inputs/assert_forall_by-ba09b09040c1c4b1-test_forallstmt_implies/test.rs:20:5
   |
20 |     assert forall |x: int| f(x) ==> g(x) by {
   |     ^^^^^^
   |
   = help: If you didn't mean to assume the antecedent, we're very curious to hear why! To tell us, please open an issue on the Verus issue tracker on github with the title `Don't always make assert forall assume the antecedent`. If no one opens such an issue, we'll soon change the behavior of Verus to always assume the antecedent of the outermost implication
```

parno:
I think we need `; consider` instead of `, consier`

PR #1314, opened 2024-10-22
Remove warning on trigger if the broadcast proof function does not have a parameters?

ziqiaozhou:


utaal:
@ziqiaozhou can you give me an example of one such lemma?

ziqiaozhou:
One example is to broadcast a proof similar to lemma_low_bits_mask_values or lemma2_to64. Then verifier will know the value of pow2(n) and 1<<N, even if the function to be verified does not explicit trigger them. It is useful to avoid repeated introducing those proof or add assert to trigger.

Verus already supported broadcasting a proof without any argument, but it always warns me that I need to add a trigger. and so, I am wondering whether it is okay to remove this warning if the @arguments is zero.

tjhance:
It looks like, if `function.x.params.len() == 0`, we can just skip the entire computation of `found_trigger`.

Chris-Hawblitzel:
Just a note on this before I forget: if I remember correctly, the purpose of the trigger on 0-argument broadcasts is to enable trigger-based pruning of broadcasts, described in https://github.com/verus-lang/verus/pull/1175.

Chris-Hawblitzel:
Here are a couple examples in `vstd` where we're using trigger-based pruning on 0-argument broadcasts:

```
pub broadcast proof fn axiom_seq_empty<A>()
    ensures
        #[trigger] Seq::<A>::empty().len() == 0,
{
    admit();
}
```

```
pub broadcast proof fn axiom_u32_obeys_hash_table_key_model()
    ensures
        #[trigger] obeys_key_model::<u32>(),
{
    admit();
}
```

Issue #1313, opened 2024-10-21
Verus not respecting --no-solver-version-check flag


Catoverflow:
## What Happened

I have z3 of `4.13.0` installed locally and verus expected `4.12.5`, according to the note:

```shell
vargo build
error: Verus expects z3 version "4.12.5", found version "4.13.0"
Run ./tools/get-z3.(sh|ps1) to update z3 first.
If you need a build with a custom z3 version, re-run with --no-solver-version-check.
```

I added the `--no-solver-version-check` because I think it's compatible with 4.12.5 API according to z3's [release note](https://github.com/Z3Prover/z3/releases/tag/z3-4.13.0), but Verus still panicked with 

```shell
thread '<unnamed>' panicked at /tmp/verus/source/air/src/smt_verify.rs:192:21:
The verifier expects z3 version "4.12.5", found version "4.13.0 - build hashcode 3049f578a8f98a0b0992eca193afe57a73b30ca3"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I checked the code and thought this flag was not respected:

https://github.com/verus-lang/verus/blob/aa8b42313d8e7a9a724abfa3dc481460aaeef2c1/source/air/src/smt_verify.rs#L182-L196

## What Was Expected

Verus respect `--no-solver-version-check` and proceed with no panic / error

Catoverflow:
This bug still exists after this commit:

Building succeeded with

```shell
verus/source ((release/rolling/0.2024.10.22.2776465))> vargo build --no-solver-version-check --release
vargo info [0]: building rust_verify
    Finished release [optimized] target(s) in 0.04s
vargo info [0]: building builtin
    Finished release [optimized] target(s) in 0.02s
vargo info [0]: building builtin_macros
    Finished release [optimized] target(s) in 0.03s
vargo info [0]: building state_machines_macros
    Finished release [optimized] target(s) in 0.02s
vargo info [0]: building vstd_build
    Finished release [optimized] target(s) in 0.02s
vargo info [0]: building verus
    Finished release [optimized] target(s) in 0.03s
vargo info [0]: vstd fresh
```

But runtime panicked:

```shell
verus/source ((release/rolling/0.2024.10.22.2776465))> ./target-verus/release/verus rust_verify/example/guide/getting_started.rs

thread '<unnamed>' panicked at verus/source/air/src/smt_verify.rs:192:21:
[...]

PR #1316, opened 2024-10-23
update Z3 to 4.14.0

Catoverflow:
Update default z3 version to 4.13.0 and fix the runtime version check. For version check the problem is described in https://github.com/verus-lang/verus/issues/1313#issuecomment-2432825850 :

> I'm using Arch Linux and shipped z3 has the version number as `Z3 version 4.13.0 - 64 bit - build hashcode 3049f578a8f98a0b0992eca193afe57a73b30ca3`, which happened to pass the version checks in vargo wrapper at
> 
> https://github.com/verus-lang/verus/blob/27764656a55aa214134f79187aac28435ae52ebc/tools/vargo/src/main.rs#L80
> 
> and
> 
> https://github.com/verus-lang/verus/blob/27764656a55aa214134f79187aac28435ae52ebc/tools/vargo/src/main.rs#L161
> 
> because the hash part is skipped using the 2nd match group. But it fails at runtime even if the versions compared are the same(see the error above):
> 
> https://github.com/verus-lang/verus/blob/27764656a55aa214134f79187aac28435ae52ebc/source/air/src/smt_verify.rs#L186
> 
> ```shell
> error: expected z3 version 4.13.0, found 4.13.0 - build hashcode 3049f578a8f98a0b0992eca193afe57a73b30ca3
> ```
>

### DIFF
--- a/source/tools/get-z3.ps1
+++ b/source/tools/get-z3.ps1
@@ -1,4 +1,4 @@
-$z3_version = "4.12.5"
+$z3_version = "4.14.0"
 $filename = "z3-$z3_version-x64-win"
 
 $download_url = "https://github.com/Z3Prover/z3/releases/download/z3-$z3_version/$filename.zip"
--- a/source/tools/get-z3.sh
+++ b/source/tools/get-z3.sh
@@ -1,21 +1,22 @@
 #! /bin/bash -eu
 
-z3_version="4.12.5"
+# Should keep track of verus/tools/common/consts.rs
+z3_version="4.14.0"
 
 if [ `uname` == "Darwin" ]; then
     if [[ $(uname -m) == 'arm64' ]]; then
-        filename="z3-$z3_version-arm64-osx-11.0"
+        filename="z3-$z3_version-arm64-osx-13.7.2"
     elif [[ $(uname -m) == 'x86_64' ]]; then
-        filename="z3-$z3_version-x64-osx-11.7.10"
+        filename="z3-$z3_version-x64-osx-13.7.2"
     else
         echo "Unsupported architecture $(uname -m)"
         exit -1
     fi
 elif [ `uname` == "Linux" ]; then
     if [[ $(uname -m) == 'aarch64' ]]; then
-        filename="z3-$z3_version-arm64-glibc-2.35"
+        filename="z3-$z3_version-arm64-glibc-2.34"
     elif [[ $(uname -m) == 'x86_64' ]]; then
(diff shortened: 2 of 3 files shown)


utaal:
Hi @Catoverflow, thank you for the PR.

Just wanted to mention that we may want to do a `veritas` run before merging a major Z3 version bump, as that may break a number of proofs.

Regarding #1313 note that, as @tjhance mentioned, you shouldn't need to rely on your package manager's Z3. The installation steps will download the expected Z3 version when you set up `verus`.

Catoverflow:
> we may want to do a veritas run before merging a major Z3 version bump

I ran it locally. But I think it should be added to CI.

> you shouldn't need to rely on your package manager's Z3

This makes sense, though it could be convenient. Then this PR is not necessary, and you can close it.

utaal:
Can ask you a favor: can you post the result of the `veritas` run with this branch?

Catoverflow:
@utaal The error related logs running `~/verus/tools/veritas (main)> bash run.sh run_configuration_small.toml` are

```shell
■■■ info: verus ready ■■■
■■■ info: producing output at /root/output/2024-10-24-22-32-19-127-3704fe9cc0818cffbd8e5efc2c689255b224473d6141ac778f0b1d1d6a897d2f ■■■
■■■ info: running projects ■■■
■■■ info: running project verus-vstd ■■■
■■■ info: fetching https://github.com/verus-lang/verus.git into /root/repos-cache/verus-vstd-5fd9e8df53cb0957a9c182e2857c835acd008efa69e1633fd22e450a15921bc3 ■■■
▶▶▶ running: cd "/root/work/work/verus-vstd-0d6e4079e36703ebd37c00722f5891d28b0e2811dc114b129215123adcce3605/" && "/root/work/work/verus-0d6e4079e36703ebd37c00722f5891d28b0e2811dc114b129215123adcce3605/source/target-verus/release/verus" "source/vstd/vstd.rs" "--no-vstd" "--crate-type=lib" "-V" "use-crate-name" "--output-json" "--time" "--no-report-long-running" ◀◀◀
■■■ info: running project page-table ■■■
■■■ info: fetching https://github.com/utaal/verified-nrkernel.git into /root/repos-cache/page-table-18eaab182025129d1f4175b03abbc9bc8523332d0fa5e96e4e52f28821e24d98 ■■■
error [src/main.rs:272:26]: failed to find page-table: revspec 'page-table' not found; class=Reference (4); code=NotFound (-3)
```

utaal:
* [ ] veritas run

PR #1323, opened 2024-10-29
allow #[verifier(external_body)] again on consts

### LINKED ISSUE #1322: `external_body` consts panic the verifier
[edited by @utaal]

```rust
use vstd::prelude::verus; verus!{

#[verifier(external_body)]
const A: usize ensures 32 <= A <= 52 { unimplemented!() }

}
```

panics verus.

---

* [ ] un-ignore test added in https://github.com/verus-lang/verus/commit/89b3e7aea6e21711a2ab6a4c05e8422c57fb85c2

utaal:
Write test for regression introduced by bf90ae0d and reported by #1322, partial fix.

@tjhance can you take a look to see if you can spot why the test is failing (after I allowed `Const`s as items that can be `external_body`)? You've recently worked on this code, so you're probably the most familiar.

### DIFF
--- a/source/rust_verify/src/external.rs
+++ b/source/rust_verify/src/external.rs
@@ -478,6 +478,7 @@ impl<'a> GeneralItem<'a> {
                 ItemKind::Struct(..) => true,
                 ItemKind::Enum(..) => true,
                 ItemKind::Union(..) => true,
+                ItemKind::Const(..) => true,
                 _ => false,
             },
             GeneralItem::ForeignItem(_) => false,
--- a/source/rust_verify_test/tests/consts.rs
+++ b/source/rust_verify_test/tests/consts.rs
@@ -341,3 +341,18 @@ test_verify_one_file! {
         }
     } => Err(err) => assert_one_fails(err)
 }
+
+test_verify_one_file! {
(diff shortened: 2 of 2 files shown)


tjhance:
This issue doesn't have to do with the external-refactor, i.e., I get this panic on 601e1e7211d58ca12fc70e23cc4bf21ea19731d2 (commit before bf90ae0d9664e36c6a40f1f60ed88d055d331257):

```
thread '<unnamed>' panicked at vir/src/poly.rs:405:66:
no entry found for key
```

(The key-not-found in question is `VarIdent("%return", VirParam)`). So I think we may have never supported this, or it was something else that broke it.

utaal:
Thanks for checking! And sorry for the incorrect blame for the regression.

@matthias-brun can you confirm that this used to work? If so, we can bisect to find the culprit, or otherwise work on a solution.

matthias-brun:
Ah, it looks like combining `external_body` with a const `ensures` used to cause a panic but now generates the error.

My original example didn't have the `ensures`.

```rust
use vstd::prelude::verus; verus!{

#[verifier(external_body)]
const A: usize = unimplemented!();
}
```

This one works fine on Verus commit 2dbda4a6ad5f14135bf990fffad45d07085fc4c3 but generates the error on current main.

PR #1326, opened 2024-10-29
Add saturate_add && option::map

ziqiaozhou:
Be lazy to maintain my own specifications for std/core functions and so added more here.


parno (approved):
Looks reasonable to me.  Thanks for the contribution!

PR #1329, opened 2024-10-30
Add some multiset-related lemmas

zeldovich:
The main lemmas of interest here are:

`lemma_fold_right_permutation` proves that, when doing a `fold_right` with a commutative operator, any permutation of the sequence produces the same result.

`to_multiset_remove` strengthens `to_multiset_ensures()` to reason about removing elements from a seq.

`lemma_multiset_has_no_duplicates_conv` is the converse of `lemma_multiset_has_no_duplicates` and helps prove that, if a sequence was `no_duplicates()`, then any permutation (with an equal `to_multiset()`) also has `no_duplicates()`.

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -646,8 +646,8 @@ impl<A> Seq<A> {
         }
     }
 
-    /// An auxiliary lemma for proving [`Self::lemma_fold_right_alt`].
-    proof fn aux_lemma_fold_right_alt<B>(self, f: spec_fn(A, B) -> B, b: B, k: int)
+    /// A lemma that proves how [`Self::fold_right`] distributes over splitting a sequence.
+    pub proof fn lemma_fold_right_split<B>(self, f: spec_fn(A, B) -> B, b: B, k: int)
         requires
             0 <= k < self.len(),
         ensures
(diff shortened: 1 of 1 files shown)


source/vstd/seq_lib.rs
```diff
@@ -1669,6 +1745,49 @@ pub proof fn lemma_seq_subrange_elements<A>(s: Seq<A>, start: int, stop: int, x:
...
 }
 
+// Definition of a commutative fold_right operator.
+pub open spec fn commutative_foldr<A, B>(f: spec_fn(A, B) -> B) -> bool {
+    forall|x: A, y: A, v: B| #[trigger] f(x, f(y, v)) == f(y, f(x, v))
+}
+
+// For a commutative fold_right operator, any folding order
+// (i.e., any permutation) produces the same result.
+pub proof fn fold_right_permutation<A, B>(l1: Seq<A>, l2: Seq<A>, f: spec_fn(A, B) -> B, v: B)
```
jaylorch, line 1755: The Verus naming convention suggests prepending `lemma_` to the name of this function.

source/vstd/seq_lib.rs
```diff
@@ -1320,6 +1379,23 @@ proof fn to_multiset_build<A>(s: Seq<A>, a: A)
     }
 }
 
+proof fn to_multiset_remove<A>(s: Seq<A>, i: int)
```
jaylorch, line 1382: The Verus naming convention suggests prepending `lemma_` to the name of this function.

jaylorch: Hmm, for some reason the existing `to_multiset_X` proof functions don't start with `lemma_`. So I see why you didn't do that here.

source/vstd/seq_lib.rs
zeldovich: Makes sense, thanks; updated.

jaylorch (approved):
Thanks! These look quite useful.

PR #1330, opened 2024-10-31
Rust 1.82.0

utaal:
This is based on @ziqiaozhou's PR https://github.com/verus-lang/verus/pull/1325 rebased after `rustc-1.72.0` was merged.

@ziqiaozhou:
> This draft PR is only created as a draft and is not intended to be merged to v1.79.

> @utaal, this is the change I mentioned about supporting v1.82. The only issue I encountered is about negative trait and conflicted trait implementations. 
```
conflicting implementations of trait `T57_IntoIterator` for type `&mut std::boxed::Box<[_], _>`
```
> Or 
```
found both positive and negative implementation of trait `T55_Iterator` for type `&mut std::boxed::Box<[_], _>`.
```

> ~~The way I bypass it for now is to check the polarity and add negative when needed. But still, I need to explicitly rule out Box cases.~~

> Adding  with_negative_coherence feature finally solved the issue I encountered. But I have 5 failed tests in example which is expected to fail and does fail. I may need to check the test lib and see why the failure is not recognized as a correct failure.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.79.0"
+channel = "1.82.0"
 # channel = "nightly-2023-09-29" # TODO update to match 1.79.0 (not officially supported)
 components = [ "rustc", "rust-std", "cargo", "rustfmt", "rustc-dev", "llvm-tools" ]
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -151,11 +151,11 @@ fn insert_spec_call(any_fn: &mut dyn AnyAttrBlock, call: &str, verus_expr: Token
     let fname = Ident::new(call, verus_expr.span());
     let tokens: TokenStream =
         syntax::rewrite_expr(EraseGhost::Keep, true, verus_expr.into()).into();
-    any_fn
-        .block_mut()
-        .unwrap()
-        .stmts
-        .insert(0, parse2(quote! { ::builtin::#fname(#tokens); }).unwrap());
+    any_fn.block_mut().unwrap().stmts.insert(
+        0,
+        parse2(quote! { #[verus::internal(const_header_wrapper)]||{::builtin::#fname(#tokens);};})
+            .unwrap(),
+    );
 }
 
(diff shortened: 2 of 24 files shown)


ziqiaozhou:
@utaal, I added two changes by diff the branch with my working one to allow v1.82 works well. I removed a change about a replace const-eval in your previous v1.79 since my project encountered some errors from that change. I also added 2 unit tests to indicate why I need to revert that change. Since that change is removed, some unit tests will have different error message since some vir errors are now rust errors.

utaal:
Thank you!

> replace const-eval

Do you know how this is now okay? i.e. why is it not causing issues with spec fn calls in proofs in consts? it may be we don't have tests for that

utaal:
I rebased on top of `main`, where I added the test for
```exec
        #[verifier::opaque]
        spec fn f() -> int { 1 }
        proof fn e() -> (u: u64)
            ensures
                u == f(),
                u == 1
        {
            reveal(f);
            1
        }
        exec const E: u64 ensures E == f() {
            proof {
                let x = e();
                assert(x == f());
                assert(x == 1);
            }
            assert(1 == f());
            1
        }
```

which is something that projects use but was untested. We need a solution for that before merging (I'm open to alternative ideas to the const-eval hack).

utaal:
We could restrict proof in consts and const fn to appear in a single block at the top of the function, and move it into a closure, like I did for the rest of the header. A bit limiting, but maybe good enough. /cc @Chris-Hawblitzel @tjhance

utaal:
Superseded by https://github.com/verus-lang/verus/pull/1385

utaal:
Does this commit address the array length issue and similar const eval issues we were seeing?

ziqiaozhou:
> veritas run

Yes. It solves the const generics I added into the PR and the const-eval issue you added.

utaal:
A full veritas run is difficult across rustc versions, but I've run on a couple of projects and it does not seem to cause any breakage, and - as you said - it fixes the const-eval issue, so I'm going to go ahead and merge (after a quick history clean-up).

utaal:
This fixes https://github.com/verus-lang/verus/issues/1334. I'll add a regression test after merge.

PR #1338, opened 2024-11-06
Make it more pleasant to write a proof-by-computation over a range of values

parno:


source/vstd/vstd.rs
```diff
@@ -114,6 +116,7 @@ pub broadcast group group_vstd_default {
     string::group_string_axioms,
     std_specs::range::group_range_axioms,
     raw_ptr::group_raw_ptr_axioms,
+    compute::all_spec_implies,
```
Chris-Hawblitzel, line 119: The broadcast group `range_all_spec` is never used.  Is it meant to be used here, instead of saying `all_spec_implies` directly here?

Chris-Hawblitzel (approved):

source/vstd/vstd.rs
parno: Ah, I forgot I'd created the group.  I could add it here so that we don't need to change `vstd.rs` if we add more lemmas in `compute.rs`.  Or I can just delete the group from `compute.rs`.

Chris-Hawblitzel: > Ah, I forgot I'd created the group. I could add it here so that we don't need to change `vstd.rs` if we add more lemmas in `compute.rs`. Or I can just delete the group from `compute.rs`.

Either way sounds fine to me.

PR #1340, opened 2024-11-12
Make macos_allow_gatekeeper a bit safer

tchajed:
Removed an extra newline and ensured it runs in the script directory regardless of the working directory.


jaybosamiya-ms (approved):
LGTM, thanks!

jaybosamiya-ms:
_Technically_ one might also want to include a `readlink --canonicalize-existing`/`readlink -e` around it to account for symlinked scripts, but I don't see any reason why someone would symlink to and then run the script.

Issue #1335, opened 2024-11-05
get_impl_paths/recursion-checking does not handle Sync/Send inference


tjhance:
Consider this variation of the recently added test `test_recursion_through_sync_impl_is_checked` from 4ab656f7a1354d901293202003678343c50026c7:

```rust
trait Tr {
    proof fn tr_g() {
    }
}

struct Xinner<T, S> {
    rc: std::rc::Rc<u64>,
    t: T,
    s: S,
}

struct X<T, S> {
    inner: Xinner<T, S>,
}

#[verifier::external]
unsafe impl<T: Sync + Tr, S: Sync> Sync for Xinner<T, S> {
}


trait Sr {
    proof fn f() { }
}

struct Y<R> {
    r: R,
}

impl<W: Sync> Sr for Y<W> {
    proof fn f() { }
}


struct A1 { }

struct B1 { }

impl Tr for A1 {
    proof fn tr_g() {
        test();
    }
}

proof fn test() {
    let r = Y::<X<A1, B1>>::f();
}
```

The cycle has the dependency `Xinner<T, S>: Sync` --> `X<T, S>: Sync` which is not covered by `get_impl_paths`

tjhance:
this might be okay appealing to the coinductiveness of Send/Sync? I'm not sure.

Chris-Hawblitzel:
Not all cycles are unsound.  For example, cycles through `exec` functions are sound and we don't need to report them as errors (unless we deliberately want to prohibit non-terminating `exec` functions).  Mainly, we just need to guarantee termination of `spec` and `proof` functions.

Chris-Hawblitzel:
I'll try to write out a more rigorous analysis. This follows the OOPSLA paper and the comments in `check_traits` in `recursive_types.rs`. To prove soundness, we write out operational semantics and typing rules (where verification is included in the typing) and prove that well-typedness is preserved when we step according to the operational semantics. Additionally, we want to be able to erase ghost blocks (including calls to ghost functions) when we actually compile and execute the code, so we have to prove that ghost code can safely be skipped during execution. To do this, we prove that well-typed ghost code always terminates by translating ghost code into some other typed language that is known to terminate, such as CIC. (The CIC code must simulate the original ghost code so that if the CIC code terminates, then the original ghost code also terminates.) This translation into CIC is where cycle checking appears, since, roughly speaking, CIC prohibits cycles in a series of definitions.

[...]

PR #1342, opened 2024-11-13
get_impl_paths more exhaustive checks

tjhance:
`get_impl_paths` errors if it finds a trait bound that it doesn't know how to handle. Similar in spirit to https://github.com/verus-lang/verus/pull/1341

This leaves a few open issues:

 * Handling Send/Sync correctly (https://github.com/verus-lang/verus/issues/1335).
 * `codegen_select_candidate` returns `Err` sometimes, and we just ignore it. Was that intentional on our part? I would have expected to never see `Err` unless we weren't calling it properly.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -413,9 +413,17 @@ pub(crate) fn get_impl_paths<'tcx>(
     target_id: DefId,
     node_substs: &'tcx rustc_middle::ty::List<rustc_middle::ty::GenericArg<'tcx>>,
     remove_self_trait_bound: Option<(DefId, &mut Option<vir::ast::ImplPath>)>,
-) -> vir::ast::ImplPaths {
+    span: Span,
+) -> Result<vir::ast::ImplPaths, VirErr> {
     let clauses = instantiate_pred_clauses(tcx, target_id, node_substs);
-    get_impl_paths_for_clauses(tcx, verus_items, param_env_src, clauses, remove_self_trait_bound)
+    get_impl_paths_for_clauses(
+        tcx,
+        verus_items,
+        param_env_src,
+        clauses,
+        remove_self_trait_bound,
+        span,
+    )
 }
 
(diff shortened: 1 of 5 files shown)


source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -534,15 +534,36 @@ pub(crate) fn get_impl_paths_for_clauses<'tcx>(
...
+                                || matches!(
+                                    verus_items::get_rust_item(tcx, trait_def_id),
+                                    Some(RustItem::Send | RustItem::Thin)
+                                )
+                            {
+                                // Sized, Tuple, Pointee, Thin are all ok to do nothing.
+                                // There can't be user impls of these traits, they can only be built-in.
+
+                                // TODO: Send and Sync needs handling, or a rigorous argument why it's ok to skip;
+                                // See https://github.com/verus-lang/verus/issues/1335
```
Chris-Hawblitzel, line 550: I commented on https://github.com/verus-lang/verus/issues/1335 .  Is this TODO considered resolved, or does it need more discussion?

tjhance: I'm genuinely unsure about this one; if there's no issue, I'd still like to see a rigorous argument.

Chris-Hawblitzel: Ok, I tried to write out a more rigorous argument in https://github.com/verus-lang/verus/issues/1335 .

utaal:
@tjhance what is the status of this?

tjhance:
@Chris-Hawblitzel are you ok with this change? Right now your only comment is about #1335 (which isn't really impacted by this change)

Chris-Hawblitzel:
> @Chris-Hawblitzel are you ok with this change? Right now your only comment is about #1335 (which isn't really impacted by this change)

I'd just like to see an update to the "TODO" comment; otherwise I'm fine with it.

Chris-Hawblitzel (approved):

PR #1351, opened 2024-11-19
Add white outlines to the logos to look nicer on dark mode

jaybosamiya-ms:
This PR updates the Verus logos to have white borders to perform this change (screenshots taken in dark mode):

image

---

## Details

I wrote the following filter to wrap around the outermost group of the SVGs for the logos; it essentially merely adds a 2.5 width white-colored dilation around any object where the filter is applied. The width/height of the images needed to be updated to account for this (from 106 -> 106+2x2.5 == 111), but otherwise, no other changes were necessary.

```svg
    <filter id="groupborder" filterUnits="userSpaceOnUse" x="0" y="0" width="111" height="111">
      <feOffset in="SourceGraphic" result="input" dx="2.5" dy="2.5" />
      <feOffset in="SourceAlpha" result="inputAlpha" dx="2.5" dy="2.5" />
      <feMorphology operator="dilate" in="inputAlpha" radius="2.5" result="dilated" />
      <feColorMatrix type="matrix" in="dilated"
                     values="-1 0  0  0  1
                             0 -1  0  0  1
                             0  0 -1  0  1
                             0  0  0  1  0" result="whitened"/>
      <feComposite in="input" in2="whitened"
                   operator="over" result="output"/>
    </filter>
```

Also, I've regenerated the PNGs (via `inkscape -w 333 -h 333 verus-color.svg -o verus-color.png`, ...; followed by `optipng -o7` to squeeze PNG sizes down).


utaal:
Thank you for looking into this. Let me look into alternatives to accomplishing readability in dark mode. I find the white outline quite jarring as proposed in this PR.

tjhance:
would it look better if the borders were thinner?

utaal:
Screenshot 2024-11-30 at 11 53 38

I pushed a commit (that was certainly a necessary improvement) that made the title legible in dark mode. I don't think we should add outlines to the primary svgs that we distribute from the "logos" page (they'd for example interfere if someone used them on slides with a light but not white background).

I personally don't mind much that the logo is a bit muted in dark mode, but we could consider variations of the logo colors itself that would work better on a dark background.

@jaybosamiya-ms is the fix on main sufficient for now?

jaybosamiya:
The updates to `main` look reasonable for now; I think the logo still needs updates for dark mode somehow, but at least with the text not being fully dark, it is less of an issue.

Agreed on the white outline looking a bit jarring, especially on smaller screens (I had originally done it on a larger screen, where it looks more reasonable).

Anyways, next one to be fixed is the GH org logo that blends away into the darkness. It can prob be fixed by just having the org logo having a white background instead?
image

utaal:
Regarding the GitHub org logo, a white background should work, I'll give it a shot! Can I close this for now?

utaal:
Done.

PR #1352, opened 2024-11-19
Update README.md to create a list of examples of using Verus

parno:
A number of people have been interested in examples of using Verus, but they haven't found the projects link that we already have, and only a few manage to find the folders of examples and tests, so I think it would help to have them presented more explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ In addition to the documentation above, it can be helpful to see Verus used in a
  * <a href="https://verus-lang.github.io/verus/publications-and-projects/">Publications and projects</a> using Verus -- If you're using Verus please consider adding your project to that page (see the instructions there).
  * [Standalone examples](https://github.com/secure-foundations/human-eval-verus/) showing Verus in use for small, concrete tasks.
  * [Miscellaneous small examples](source/rust_verify/example) illustrating various Verus features
- * [Unit tests](source/rust_verify_test/tests) for Verus features
+ * [Unit tests](source/rust_verify_test/tests) for Verus, containing examples of Verus syntax and features.
 
 ## Getting in touch, reporting issues, and starting discussions
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Our (work-in-progress) documentation resources include:
 ## Examples of Using Verus
 In addition to the documentation above, it can be helpful to see Verus used in action.  Here are some starting points.
  * <a href="https://verus-lang.github.io/verus/publications-and-projects/">Publications and projects</a> using Verus -- If you're using Verus please consider adding your project to that page (see the instructions there).
+ * [Standalone examples](https://github.com/secure-foundations/human-eval-verus/) showing Verus in use for small, concrete tasks.
  * [Miscellaneous small examples](source/rust_verify/example) illustrating various Verus features
  * [Unit tests](source/rust_verify_test/tests) for Verus features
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,8 +20,6 @@ Verus is under *active development*. Features may be broken and/or missing, and
 the documentation is still incomplete. If you want to try Verus, please be
 prepared to ask for help in the [💬 Zulip](https://verus-lang.zulipchat.com/).
 
-The Verus community has published a number of research papers, and there are a variety of industry and academic projects using Verus. You can find a list on our <a href="https://verus-lang.github.io/verus/publications-and-projects/">publications and projects</a> page. If you're using Verus please consider adding your project to that page (see the instructions there).
```
utaal, line 23: Even if it leads to a little bit of repetition, I'd like to preserve this paragraph. It's a more direct link to Verus-related publications and projects, which aren't necessarily just example code that people are looking for as a way to "see Verus used in action".

README.md
```diff
@@ -39,6 +37,13 @@ Our (work-in-progress) documentation resources include:
  * [Verus License](LICENSE)
  * [Verus Logos](https://verus-lang.github.io/verus/verus/logo.html)
 
+## Examples of Using Verus
+In addition to the documentation above, it can be helpful to see Verus used in action.  Here are some starting points.
+ * <a href="https://verus-lang.github.io/verus/publications-and-projects/">Publications and projects</a> using Verus -- If you're using Verus please consider adding your project to that page (see the instructions there).
+ * [Standalone examples](https://github.com/secure-foundations/human-eval-verus/) showing Verus in use for small, concrete tasks.
+ * [Miscellaneous small examples](source/rust_verify/example) illustrating various Verus features
```
utaal, line 44: These aren't just "small examples". For examples, the `state_machines` directories has sizable examples of VerusSync. Can we make it sound a bit more appealing than "miscellaneous"? I think browsing these would be worthwhile for a newcomer, especially `syntax.rs` (which we may want to point out explicitly), although giving it a bit more structure would be helpful.

README.md
```diff
@@ -39,6 +37,13 @@ Our (work-in-progress) documentation resources include:
...
+ * [Unit tests](source/rust_verify_test/tests) for Verus features
```
utaal, line 45:
```suggestion
 * [Unit tests](source/rust_verify_test/tests) for Verus, containing examples of Verus syntax and features.
```

utaal (changes requested):

README.md
parno: True, there are some medium-sized examples, but some of them are quite small, e.g., [`adts_eq.rs`](https://github.com/verus-lang/verus/blob/main/source/rust_verify/example/adts_eq.rs), [recommends.rs](https://github.com/verus-lang/verus/blob/main/source/rust_verify/example/recommends.rs), and [debug.rs](https://github.com/verus-lang/verus/blob/main/source/rust_verify/example/debug.rs).  What do you think would be a good way to describe them?

utaal: "Small and medium-sized examples illustrating ..." ?

README.md
parno: Done.

README.md
parno: Sounds good.  Done.

Chris-Hawblitzel (approved):

PR #1354, opened 2024-11-20
Add fibonacci example to the Guide to illustrate how spec, proof, and exec code work together

parno:


parno:
When talking with some folks learning Verus, they said that Section 3 of the guide gave them some understanding of the mode system, but they were still confused about how you were supposed to use the different modes, and (despite the table in opening section), which modes call which other modes.  I thought this might help with some of that confusion.  I'm open to suggestions for other examples, however.

source/rust_verify/example/guide/modes.rs
```diff
@@ -381,6 +381,73 @@ fn add_one(x: u8) -> (ret: u8)
...
+{
+    if i < 2 && j < 2 {
+    } else if i == j {
+    } else if i == j - 1 {
+        assert(fib(j) == fib((j - 2) as nat) + fib((j - 1) as nat));
+        lemma_fib_is_monotonic(i, (j - 1) as nat);
+    } else {
+        lemma_fib_is_monotonic(i, (j - 1) as nat);
+        lemma_fib_is_monotonic(i, (j - 2) as nat);
+    }
```
utaal, line 412: There's a simpler proof for this, that we found during the tutorial proof debugging session:

```suggestion
    if i < 2 && j < 2 {
    } else if i == j {
    } else if i == j - 1 {
        reveal_with_fuel(fibo_spec, 2);
    } else {
        reveal_with_fuel(fibo_spec, 2);
        lemma_fib_is_monotonic(i, (j - 1) as nat);
    }
```

source/rust_verify/example/guide/modes.rs
```diff
@@ -381,6 +381,73 @@ fn add_one(x: u8) -> (ret: u8)
...
+
+spec fn fib_fits_u64(n: nat) -> bool {
+    fib(n) <= 0xffff_ffff_ffff_ffff
+}
+
+exec fn fib_impl(n: u64) -> (result: u64)
+    requires
+        fib_fits_u64(n as nat),
+    ensures
+        result == fib(n as nat),
```
utaal, line 423: This version of the fibonacci demo is a bit more compact, but sort-of unrealistic, due to the fact that the caller needs to prove that the result will fit in a `u64`. I have a more realistic version that returns an `Option` that's `None` if the result doesn't fit: what do you think, shall I replace it on this branch?

parno: I think fibonacci is always going to be rather artificial, unfortunately.  We have more realistic [examples](https://github.com/secure-foundations/human-eval-verus/blob/4196755da573deb08c3cff684dbfe4163ddbe961/tasks/human_eval_055b.rs) in HumanEval, but at the end of the day, it's still a bit silly, so I think there may be some virtue in keeping the example simple to focus on what it takes to combine the different modes together.

parno:
Hmm, that's an interesting idea.  It would have the benefit that we could promise the reader more details about the example in the next chapter.  The TwoSix and Amazon people seem to want more examples, rather than fewer, but perhaps that could be addressed by adding more examples later.  For example, we had talked about a section between 8 and 9 on coming up with loop invariants, so maybe we could use `fib` there instead of here.

parno:
To simplify reviewing, I pulled the minor fixes out of this PR and applied them to `main`, and I opened a much smaller PR (#1377) for just the modes example.  I'll open a separate PR for the example of how to develop invariants, and close this one.

PR #1358, opened 2024-11-22
Fix state machine example in book

suaviloquence:
[Playground link for changed code](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=2c78a6552c1e0b8a02e7fb2a9b07d59e)

I imagine these were made and then this syntax change happened and it didn't get updated.

Changes invariant method to take `&self` instead of magic `pre` argument, and inductive proof stubs to take `Self` instead of the named type (it fails macro expansion with e.g. `post: AdderMachine`)

### DIFF
--- a/source/docs/state_machines/src/components.md
+++ b/source/docs/state_machines/src/components.md
@@ -37,19 +37,19 @@ state_machine!{
         //// Invariants on the state
 
         #[invariant]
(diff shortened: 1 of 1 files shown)


tjhance:
thanks!

PR #1364, opened 2024-12-15
Spec for Option::ok_or

jaylorch:

### DIFF
--- a/source/vstd/std_specs/option.rs
+++ b/source/vstd/std_specs/option.rs
@@ -188,4 +188,21 @@ pub fn ex_option_clone<T: Clone>(opt: &Option<T>) -> (res: Option<T>)
     opt.clone()
 }
 
+// unwrap_or
+#[verifier::inline]
+pub open spec fn spec_ok_or<T, E>(option: Option<T>, err: E) -> Result<T, E> {
+    match option {
+        Some(t) => Ok(t),
+        None => Err(err),
+    }
+}
+
+#[verifier::external_fn_specification]
+pub fn ex_ok_or<T, E>(option: Option<T>, err: E) -> (res: Result<T, E>)
+    ensures
+        res == spec_ok_or(option, err),
+{
+    option.ok_or(err)
+}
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/option.rs
```diff
@@ -188,4 +188,21 @@ pub fn ex_option_clone<T: Clone>(opt: &Option<T>) -> (res: Option<T>)
     opt.clone()
 }
 
+// unwrap_or
```
parno, line 191: Should this comment be `ok_or`?

source/vstd/std_specs/option.rs
```diff
@@ -188,4 +188,21 @@ pub fn ex_option_clone<T: Clone>(opt: &Option<T>) -> (res: Option<T>)
...
+#[verifier::inline]
+pub open spec fn spec_ok_or<T, E>(option: Option<T>, err: E) -> Result<T, E> {
+    match option {
+        Some(t) => Ok(t),
+        None => Err(err),
+    }
+}
+
+#[verifier::external_fn_specification]
```
parno, line 200: Would it make sense to add `#[verifier::when_used_as_spec(spec_ok_or)]` here?

parno (commented):
Two minor questions/suggestions.

source/vstd/std_specs/option.rs
jaylorch: Ah, yes, thanks. I'll fix that.

source/vstd/std_specs/option.rs
jaylorch: Yes, thanks for noticing the oversight!

source/vstd/std_specs/option.rs
jaylorch: Done

source/vstd/std_specs/option.rs
jaylorch: Done

PR #1370, opened 2024-12-20
Allow vstd to support 32-bit platforms

aaurandt:
I utilized Verus to verify code that was targeting the stm32f3 discovery board (thumbv7em-none-eabihf). I was not able to directly utilize my rust code with vstd since AtomicI64 and AtomicU64 are not supported for this 32-bit architecture I was targeting. It's a rather simple fix with some configuration flags. There are more details under the portability section here: https://doc.rust-lang.org/core/sync/atomic/index.html#portability

Here is also a discussion board that led me to this solution: https://users.rust-lang.org/t/unresolved-import-core-atomicu64/109568/5

### DIFF
--- a/source/vstd/atomic.rs
+++ b/source/vstd/atomic.rs
@@ -1,8 +1,13 @@
 #![allow(unused_imports)]
 
 use core::sync::atomic::{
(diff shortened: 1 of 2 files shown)


tjhance:
Hi, thanks for your submission!

To fix the CI, you just need to run "vargo fmt".

Issue #1333, opened 2024-11-03
Failed precondition error message doesn't point to source code


nwtnni:
Found during SOSP workshop (cc @utaal)

```
error: precondition not satisfied
  --> exercise_01_memoize.rs:91:9
   |
91 |         handle.release_write(cache);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  --> /home/runner/work/verus/verus/source/vstd/rwlock.rs:404:13
   |
   = note: failed precondition

error: aborting due to 2 previous errors
```

The path `/home/runner/work/verus/verus/source/vstd/rwlock.rs` is incorrect for my system.

Source code:

```rust
// Verus tutorial - 'Advanced Topics' Exercise: computation memoization
//
// This file provides an implementation of a store to memoize the results of some computation.
// This exercise illustrates how Verus can handle interior mutability using lock invariants.
//
// Most of the implementation has been filled in for you. There are 2 blanks to fill in;
// search for 'EXERCISE' in this file.

use std::sync::Arc;
use vstd::hash_map::HashMapWithView;
use vstd::prelude::*;
use vstd::rwlock::{RwLock, RwLockPredicate};

mod memoization_exercise_internals;

verus! {
broadcast use vstd::std_specs::hash::group_hash_axioms;

//////// Import a function. Our objective is to memoize the results of this function.

use memoization_exercise_internals::{Args, Output};

/// Pure functional specification of the function to be memoized.
pub closed spec fn func(args: &Args) -> Output {
    memoization_exercise_internals::func(args)
}

/// Executable version of the function to be memoized.
/// The `ensures` clause guarantees that this computation is deterministic,
/// i.e., it returns the same value every time it is called for the same arguments.
pub fn expensive_function(args: &Args) -> (out: Output)
    ensures out == func(args),
{
    memoization_exercise_internals::expensive_function(args)
}

//////// Here we define the main data structure.

pub struct Memoizer {
    lock: RwLock<HashMapWithView<Args, Output>, Pred>,
}

// This is where we define our "lock invariant". The 'Pred' type is kind of a dummy type, here;
// in more complex cases, we might put actual data in here, but we don't need to for this
// exercise. Thus for the purposes of this file, 'Pred' only serves as a way to connect
// the lock invariant to the lock.
//
// See the RwLock docs for more information:
// https://verus-lang.github.io/verus/verusdoc/vstd/rwlock/struct.RwLock.html
struct Pred { }
[...]

PR #1375, opened 2025-01-08
when a verusroot is present, with source files, remap spans so that they are available in binary releases

utaal:
fixes https://github.com/verus-lang/verus/issues/1333

this also adds the sources for the builtin/vstd crates to the binary distribution

### DIFF
--- a/source/rust_verify/src/spans.rs
+++ b/source/rust_verify/src/spans.rs
@@ -8,6 +8,8 @@ use std::sync::{Arc, Mutex};
 use vir::ast::{SpannedTyped, Typ};
 use vir::def::Spanned;
 
+use crate::externs::VerusExterns;
+
 pub(crate) fn to_raw_span(span: Span) -> vir::messages::RawSpan {
     Arc::new(span.data())
 }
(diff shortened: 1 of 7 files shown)


source/rust_verify/src/spans.rs
```diff
@@ -72,12 +74,15 @@ impl SpanContextX {
...
         original_crate_files: HashMap<u64, HashMap<Vec<u8>, FileStartEndPos>>,
+        verus_externs: Option<&VerusExterns>,
     ) -> SpanContext {
         let mut imported_crates = HashMap::new();
         let mut local_files = HashMap::new();
         let mut remaining_crate_files = original_crate_files.clone();
+        let path_mappings = verus_externs.map(|x| x.to_path_mappings());
 
         for source_file in source_map.files().iter() {
+            // if let FileName::Real(real_file_name) = &source_file.name  {
```
Chris-Hawblitzel, line 85: stray comment?

Chris-Hawblitzel (approved):

source/rust_verify/src/spans.rs
utaal: thank you

PR #1376, opened 2025-01-08
Additional (in-)finiteness lemmas for sets

matthias-brun:
This PR adds three new lemmas for finiteness of sets after inserting, removing or unioning and one proof for an "axiom" that turns out to be derivable from the others.

~~This PR adds three stronger set axioms and a proof for one axiom that turns out to be derivable from the others.~~

~~The new axioms (`axiom_set_insert_finite_iff`, `axiom_set_insert_remove_iff`, `axiom_set_insert_union_iff`) all show that not only do the corresponding operations preserve finiteness but also infiniteness of sets. These can be relevant for contradiction proofs; Currently, given e.g. `!s.insert(a).finite()`, there's no way to conclude `!s.finite()`. I only added axioms for the operations where this relationship is obvious but not for ones that would need more complex rules (`difference`, `intersect`).~~

~~I kept the original, weaker axioms (but proved them from the new ones) to avoid any possible breakage. I did however replace them with the new ones in the broadcast group.~~

I wrote a model for `Set` and proved its axioms, including my newly added ones: https://gist.github.com/matthias-brun/f6d62ad7a3617ae992f5f0d6b8cb7626

And as mentioned in the `Map` PR (#1367), I hope we can eventually include these models directly in vstd to allow users to derive any additional properties they need themselves, without any risk of introducing inconsistencies.

### DIFF
--- a/source/vstd/set.rs
+++ b/source/vstd/set.rs
@@ -202,7 +202,26 @@ pub broadcast proof fn axiom_set_remove_insert<A>(s: Set<A>, a: A)
     ensures
         (#[trigger] s.remove(a)).insert(a) == s,
 {
-    admit();
+    assert forall|aa| #![all_triggers] s.remove(a).insert(a).contains(aa) implies s.contains(
+        aa,
+    ) by {
+        if a == aa {
+        } else {
+            axiom_set_remove_different(s, aa, a);
+            axiom_set_insert_different(s.remove(a), aa, a);
+        }
+    };
+    assert forall|aa| #![all_triggers] s.contains(aa) implies s.remove(a).insert(a).contains(
+        aa,
+    ) by {
+        if a == aa {
+            axiom_set_insert_same(s.remove(a), a);
+        } else {
+            axiom_set_remove_different(s, aa, a);
+            axiom_set_insert_different(s.remove(a), aa, a);
+        }
+    };
+    axiom_set_ext_equal(s.remove(a).insert(a), s);
 }
 
 /// If `a1` does not equal `a2`, then the result of removing element `a2` from set `s`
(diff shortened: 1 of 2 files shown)


source/vstd/set.rs
```diff
@@ -291,12 +310,28 @@ pub broadcast proof fn axiom_set_empty_finite<A>()
...
 }
 
+/// The result of inserting an element `a` into a set `s` is finite iff `s` is finite.
+pub broadcast proof fn axiom_set_insert_finite_iff<A>(s: Set<A>, a: A)
+    ensures
+        #[trigger] s.insert(a).finite() <==> s.finite(),
+{
+    admit();
+}
+
```
zeldovich, line 320: It seems like this isn't really necessary:

```
    pub broadcast proof fn axiom_set_insert_finite_iff<A>(s: Set<A>, a: A)
        ensures
            #[trigger] s.insert(a).finite() <==> s.finite()
    {
        if s.insert(a).finite() {
            if s.contains(a) {
                assert(s == s.insert(a));
            } else {
                assert(s == s.insert(a).remove(a));
            }
        }
    }
```

matthias-brun: Well now I feel like a clown :)

Thanks for pointing that out. I totally missed that the axiom for `remove` effectively functions as the inverse axiom I was looking for.

I now wrote similar proofs for `remove` and `union` and moved everything to `set_lib.rs` instead. I added you as a co-author on the commit. Hope that's okay.

matthias-brun:
As Nickolai pointed out, my new axioms can in fact be proved from the existing ones. I pushed a new version with my three new "axioms" proved and moved to `set_lib.rs`.

utaal (approved):

PR #1385, opened 2025-01-11
Rustc 1.84.0

utaal:
I think we now want to directly target 1.84, and skip 1.82 (#1330).

This is based on @ziqiaozhou's PR https://github.com/verus-lang/verus/pull/1325 rebased after `rustc-1.72.0` was merged, and on #1330.

* [ ] fix interface changes


utaal:
We're going to need a solution to this:

https://github.com/verus-lang/verus/pull/1330#issuecomment-2585130316
> I rebased on top of `main`, where I added the test for
> 
> ```rust
>         #[verifier::opaque]
>         spec fn f() -> int { 1 }
>         proof fn e() -> (u: u64)
>             ensures
>                 u == f(),
>                 u == 1
>         {
>             reveal(f);
>             1
>         }
>         exec const E: u64 ensures E == f() {
>             proof {
>                 let x = e();
>                 assert(x == f());
>                 assert(x == 1);
>             }
>             assert(1 == f());
>             1
>         }
> ```
> 
> which is something that projects use but was untested. We need a solution for that before merging (I'm open to alternative ideas to the const-eval hack).

utaal:
https://github.com/verus-lang/verus/pull/1330#issuecomment-2585131121
> We could restrict proof in consts and const fn to appear in a single block at the top of the function, and move it into a closure, like I did for the rest of the header. A bit limiting, but maybe good enough. /cc @Chris-Hawblitzel @tjhance

tjhance:
What about this? This code:

```
        exec const E: u64 ensures E == f() {
            proof {
                let x = e();
                assert(x == f());
                assert(x == 1);
            }
            assert(1 == f());
            1
        }
```

Could become:

```
const E: u64 = 1;

fn verus_proof_for_E() -> u64 {
    /* body with proof goes here */
}
```

(Though honestly, I'm tempted to solve this on the design level ... there's something really ugly about const items with proof blocks, even though I understand some people have found uses for them.)

utaal:
What do you find ugly about them? Can you elaborate?

utaal:
I like the idea, but I suspect a closure may be easier to find. I can look into whether I can place the `fn` inside the const item body (it may be allowed), which would make it easy to find but not introduce the complication of the closure.

ziqiaozhou:
> [#1330 (comment)](https://github.com/verus-lang/verus/pull/1330#issuecomment-2585131121)
> 
> > We could restrict proof in consts and const fn to appear in a single block at the top of the function, and move it into a closure, like I did for the rest of the header. A bit limiting, but maybe good enough. /cc @Chris-Hawblitzel @tjhance

I think the error could be fixed by placing #[rustc_do_not_const_check] in all proof code inside the exe code.

utaal:
Oh cool. Do you know how likely that attribute is to stay? Or is it just a temporary workaround? (Do you have a link to info on it?)
Thanks!

ziqiaozhou:
> Oh cool. Do you know how likely that attribute is to stay? Or is it just a temporary workaround? (Do you have a link to info on it?) Thanks!

I do not know how long it could stay. I found it from rust compiler codebase(https://github.com/rust-lang/rust/blob/76a030a6c2dab11760cb08d746fa515e4bce5d39/compiler/rustc_const_eval/src/check_consts/check.rs#L218). But it seems that it cannot be applied to some items and only works for blocks inside function body and could not solve the issue in const value evaluation. I tried another fix (https://github.com/verus-lang/verus/pull/1330/commits/ecab7981c973768a463d2981b583cf8e4663679d) by skipping `providers.mir_const_qualif` and add const_header_wrapper for const-value. It may not be the best solution but at least it could skip constness checker in most cases automatically inside const function, and then it still needs the const_header_wrapper for some cases (e.g., const value evaluation).  

BTW, I reopened 1.82 PR since the one I used is 1.82 and it would be helpful if Verus main branch can track a 1.82 version so that I do not need to keep my branch for capturing the changes. The change in 1.84 seems not ready yet.

utaal:
This now needs to be rebased on `main` carefully, only applying the 1.84.0 relevant changes.

utaal:
This has been rebased on `main` and changed to target `1.85.0`. There are a number of errors due to interface changes to address.

dschoepe:
The only remaining issue is the `assume_specification` for `str::to_string` that now triggers #1645.

dschoepe:
This version passes all tests (by specifying `str::to_owned` rather than `str::to_string` to work around #1645). I didn't include the changes from `vargo fmt` yet because it introduces a large number of changes here for some reason (see [diff](https://github.com/verus-lang/verus/actions/runs/15124816575/job/42514818655?pr=1385)), which would make this hard to review.

dschoepe:
> The only remaining issue is the assume_specification for str::to_string that now triggers https://github.com/verus-lang/verus/issues/1645. This is starting to happen in 1.85.0 because str now no longer implements ToString via a blanket implementation, but instead implements this through [this impl](https://doc.rust-lang.org/1.85.0/alloc/string/trait.ToString.html#impl-ToString-for-T).

This was incorrect, and it looks like the rustdoc for this impl just isn't being linked in 1.85.0. The difference between 1.82 and 1.85 seems to be that `to_string` is no longer a `default fn`.

tjhance:
If necessary, we can always hardcode `to_string` as a special-case in rust_to_vir.

To fix it in a less hackish way, we'd need to fix #1657, but I still don't understand what's going on there.

source/rust_verify_test/tests/assoc_type_impls.rs
```diff
@@ -542,7 +542,7 @@ test_verify_one_file! {
         }
         impl<A: U> T for A where A::X: R { spec fn f() -> int { 100 } }
         impl U for char { type X = u8; }
-        impl U for bool { type X = u16; }
+        impl U for bool { type X = u8; }
```
Chris-Hawblitzel, line 545: I think this should be kept as `X = u16` as it was before, but now the error is different.  I suggest:

```
proof fn test() {
    assert(<bool as T>::f() == 200);
    assert(<char as T>::f() == 100);
    assert(<bool as T>::f() == 100); // FAILS
    assert(<char as T>::f() == 200);
    assert(false);
}
```

Chris-Hawblitzel (approved):
I just have a couple minor requested changes.

Chris-Hawblitzel:
Can you try cargo verus in this branch (`source/examples/cargo-verus/run.sh`) before merging?  I got an error `Argument to option 'V' missing`.

source/rust_verify/src/main.rs
```diff
@@ -95,13 +95,8 @@ pub fn main() {
...
     if is_direct_rustc_call {
-        args.insert(0, program);
-        match rust_verify::driver::run_rustc_compiler_directly(&args) {
-            Ok(()) => return,
-            Err(_) => {
-                std::process::exit(1);
-            }
-        }
+        args.insert(0, program.clone());
+        rust_verify::driver::run_rustc_compiler_directly(&args);
```
Chris-Hawblitzel, line 99: > Can you try cargo verus in this branch (source/examples/cargo-verus/run.sh) before merging? I got an error Argument to option 'V' missing.

I think the cause is that this code used to always exit here, but now continues after `run_rustc_compiler_directly`, leading to a later spurious error.

source/rust_verify_test/tests/assoc_type_impls.rs
dschoepe: I see. I thought this was intended to check that verus reports conflicting trait implementations. I was confused because the above code doesn't actually have conflicting implementations but reported such an error due to what I assumed was a precision issue in the conflict detection logic in 1.82.

I'm happy with your suggestion too, to check that it doesn't get resolved to the wrong implementation.

source/rust_verify/src/main.rs
dschoepe: From what I could tell, `RunCompiler::run` now panics if there are errors (by eventually calling [`FatalError::raise`](https://github.com/rust-lang/rust/blob/master/compiler/rustc_span/src/fatal_error.rs#L14), so I thought this would also abort here.

dschoepe: Oh, you're right, missed the fact that this returned on a successful run too.

dschoepe: Fixed in latest version of the PR.

dschoepe:
> I just have a couple minor requested changes.

Thanks, I addressed them in the latest PR version now.

Chris-Hawblitzel:
> Thanks, I addressed them in the latest PR version now.

Thanks!  It looks like cargo verus works now.  I noticed that when cargo verus runs now, it prints out a bunch of cfg-related warnings.  Here's a diff that can suppress those warnings:

```
diff --git a/source/rust_verify/src/cargo_verus.rs b/source/rust_verify/src/cargo_verus.rs
index fb3a61220..9cf1f1d39 100644
--- a/source/rust_verify/src/cargo_verus.rs
+++ b/source/rust_verify/src/cargo_verus.rs
@@ -54,6 +54,11 @@ pub fn extend_args_and_check_is_direct_rustc_call(
     } else {
         false
     };
+    if verus_crate {
+        rustc_args.push("--check-cfg".to_owned());
+        rustc_args
+            .push("cfg(verus_keep_ghost, verus_keep_ghost_body, verus_verify_core)".to_owned());
+    }
     if !verus_crate {
         if let Some(package_id) = &package_id {
             let is_builtin =
```

dschoepe:
Thanks, I've added that as well, though there are still warnings on `main.rs` in the cargo-verus example:

```
warning: unexpected `cfg` condition name: `verus_keep_ghost_body`
  --> src/main.rs:3:1
   |
3  | / verus! {
4  | |
5  | | #[verifier::external_body]
6  | | fn print_result(value: u32) {
...  |
16 | | } // verus!
   | |_^
   |
   = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `fmt_debug`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
   = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
   = help: try referring to `verus` crate for guidance on how handle this unexpected cfg
   = help: the macro `verus` may come from an old version of the `builtin_macros` crate, try updating your dependency with `cargo update -p builtin_macros`
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: `#[warn(unexpected_cfgs)]` on by default
   = note: this warning originates in the macro `verus` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition name: `verus_keep_ghost`
  --> src/main.rs:3:1
   |
3  | / verus! {
4  | |
5  | | #[verifier::external_body]
6  | | fn print_result(value: u32) {
...  |
16 | | } // verus!
   | |_^
   |
   = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
   = help: try referring to `verus` crate for guidance on how handle this unexpected cfg
   = help: the macro `verus` may come from an old version of the `builtin_macros` crate, try updating your dependency with `cargo update -p builtin_macros`
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: this warning originates in the macro `verus` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unexpected `cfg` condition name: `verus_verify_core`
  --> src/main.rs:3:1
   |
3  | / verus! {
4  | |
5  | | #[verifier::external_body]
6  | | fn print_result(value: u32) {
...  |
16 | | } // verus!
   | |_^
   |
   = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
   = help: try referring to `verus` crate for guidance on how handle this unexpected cfg
   = help: the macro `verus` may come from an old version of the `builtin_macros` crate, try updating your dependency with `cargo update -p builtin_macros`
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: this warning originates in the macro `verus` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Also, what should we do regarding the large amount of formatting changes when running `vargo fmt`? I suspect they are somehow related to the rustc upgrade since they affect many parts of the code not modified by this PR. I suggest we squash the commits here and then have a separate format commit that we can add to `.git-blame-ignore-revs` to keep `git blame` and such working correctly.

tjhance:
What is the reasoning for all the formatting changes? I guess it has to do with the change in the config in tools/vargo/src/main.rs:

```
-               args.insert(pos + 2, "unstable_features=true,version=Two".to_string());
+               args.insert(
+                   pos + 2,
+                   "unstable_features=true,style_edition=2021".to_string(),
+               );
```

But what is the reason for the change in config?

dschoepe:
In that case, we end up with an error: `invalid key=val pair: style_edition=two`, so I think that might be necessary due to `cargo fmt` no longer accepting the old `style_edition`. Setting it to `2018` or `2015` instead of `2021` also produces a lot of formatting changes though.

tjhance:
According to [this changelog](https://github.com/rust-lang/rustfmt/blob/master/CHANGELOG.md#180-2024-09-20), version=Two should be equivalent to `style_edition=2024`.

dschoepe:
Strangely, that results in changes in 79 files instead.

tjhance:
well, honestly, this is on us for choosing to use unstable rustfmt features.

dschoepe:
Actually, it only seems to reorder imports (putting uppercase identifiers first) with `style_edition=2024`, so that seems better.

Chris-Hawblitzel:
> there are still warnings on `main.rs` in the cargo-verus example

Does it work if you do `cargo clean --manifest-path test/Cargo.toml` first?

Chris-Hawblitzel:
> well, honestly, this is on us for choosing to use unstable rustfmt features.

Now that we've switched to `style_edition=2024`, can we remove the `unstable_features=true`?

tjhance:
https://github.com/verus-lang/verus/pull/1677 should resolve the `to_string` issue.

PR #1388, opened 2025-01-14
Add new badge

jaybosamiya-ms:
It came to my attention that some folks might find it helpful to have a badge that points at Verus docs in particular (I know we mention the guide later in the README under docs, but this is more of a direct jump to it). One _can_ reach it from the tutorial shield, but I think it is still helpful to have an extra badge for the reference, that tells people that docs about Verus (in contrast with vstd docs) are here. Some folks end up accidentally jumping to vstd docs and end up a little confused and then need to navigate to the tutorial from there, while the "tutorial" word seems to indicate something closer to a "quick start" than a more detailed full book. By having a `docs | verus` badge, it makes it clear that we intend these to be the core docs for Verus

---


utaal (approved):
Ty!

PR #1389, opened 2025-01-15
Implement `View` for `Option<T>`

y1ca1:

Hi Verus Team,

Currently there's no impl of the `View` trait for Rust's `Option` type, which is slightly annoying as
1. one would have to `unwrap()` or destruct an optional value to call `.view()`;
2. one cannot pass a value of type`Option<T> where T: View` to  some function that expects `T: View`.

Similar implementations can be add to `Result<T, E>` as well, but I'd like to see what people think of adding this impl for `Option` first.

Thanks!

### DIFF
--- a/source/vstd/std_specs/option.rs
+++ b/source/vstd/std_specs/option.rs
@@ -7,6 +7,18 @@ use core::option::Option::Some;
 
 verus! {
 
+impl<T: View> View for Option<T> {
+    type V = Option<T::V>;
+
+    open spec fn view(&self) -> Option<T::V> {
+        match self {
+            Some(t) => Some(t@),
+            None => None,
+        }
+    }
+}
+
+
 ////// Add is_variant-style spec functions
 pub trait OptionAdditionalFns<T>: Sized {
     #[allow(non_snake_case)]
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel:
For comparison, the implementations for `View` and `DeepView` on slice look like:

```
impl<T> View for [T] {
    type V = Seq<T>;

}

impl<T: DeepView> DeepView for [T] {
    type V = Seq<T::V>;

    open spec fn deep_view(&self) -> Seq<T::V> {
        let v = self.view();
        Seq::new(v.len(), |i: int| v[i].deep_view())
    }
}
```

By analogy with the slice version, would the following work for Option?

```
impl<T> View for Option<T> {
    type V = Option<T>;

    open spec fn view(&self) -> Option<T> {
        *self
    }
}

impl<T: DeepView> DeepView for Option<T> {
    type V = Option<T::V>;

    open spec fn deep_view(&self) -> Option<T::V> {
        match self {
            Some(t) => Some(t.deep_view()),
            None => None,
        }
    }
}
```

y1ca1:
Makes sense, I updated the impl!

Chris-Hawblitzel (approved):

utaal:
@y1ca1 feel free to squash and merge.

PR #1400, opened 2025-01-22
misc fixes to vec assumed specs

zeldovich:
- Fix the existing assumed spec for `Vec::truncate()` which erroneously made the postcondition depend on the new value of `vec.len()` rather than `old(vec).len()`.

- Add an assumed spec for `Vec::resize()` which is a generalization of `Vec::truncate()` that allows growing the `Vec`.

### DIFF
--- a/source/vstd/std_specs/vec.rs
+++ b/source/vstd/std_specs/vec.rs
@@ -214,7 +214,7 @@ pub assume_specification<T: Clone, A: Allocator>[ Vec::<T, A>::resize ](
         len <= old(vec).len() ==> vec@ == old(vec)@.subrange(0, len as int),
         len > old(vec).len() ==> {
             &&& vec@.len() == len
-            &&& vec@.subrange(0, len as int) == old(vec)@
+            &&& vec@.subrange(0, old(vec).len() as int) == old(vec)@
             &&& forall|i|
                 #![all_triggers]
                 old(vec).len() <= i < len ==> call_ensures(T::clone, (&value,), vec@[i]) || value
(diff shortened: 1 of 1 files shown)


tjhance:
The specification for resize is incorrect because it makes assumptions on the behavior of `T::clone`.

The correct way to describe this relationship between a value and a cloned value is to use the predicate:

```
call_ensures(T::clone, (value,), vec@[i])
```

zeldovich:
> The specification for resize is incorrect because it makes assumptions on the behavior of `T::clone`.
> 
> The correct way to describe this relationship between a value and a cloned value is to use the predicate:
> 
> ```
> call_ensures(T::clone, (value,), vec@[i])
> ```

Oh, indeed, thanks for catching this and for the suggestion!  Fixed accordingly.

tjhance:
Thanks!

One last thing, though. 😅 It occurred to me to check the implementation of `resize` (see [here](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3133), and evidently it will optimize out 1 of the clones. So perhaps what we really need to do is say `call_ensures(T::clone, (value,), vec@[i]) || value == vec@[i]`.

zeldovich:
> One last thing, though. 😅 It occurred to me to check the implementation of `resize` (see [here](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3133)), and evidently it will optimize out 1 of the clones. So perhaps what we really need to do is say `call_ensures(T::clone, (value,), vec@[i]) || value == vec@[i]`.

That's a clever optimization -- good catch!  Fixed the spec accordingly.

tjhance:
```
vec@.subrange(0, len as int) == old(vec)@
```

I think this should be old(vec).len()

zeldovich:
> ```
> vec@.subrange(0, len as int) == old(vec)@
> ```
> 
> I think this should be old(vec).len()

Shoot, sorry about all the bugs in this spec..  Fixed!

tjhance:
ok i think we are good to go!

PR #1408, opened 2025-01-25
Support let-else

ziqiaozhou:

### DIFF
--- a/source/rust_verify/src/lifetime_ast.rs
+++ b/source/rust_verify/src/lifetime_ast.rs
@@ -109,7 +109,7 @@ pub(crate) type Stm = Box<(Span, StmX)>;
 #[derive(Debug, Clone)]
 pub(crate) enum StmX {
     Expr(Exp),
-    Let(Pattern, Typ, Option<Exp>, Option<Exp>),
+    Let(Pattern, Typ, Option<Exp>, Option<Exp>), // (pat, typ, init_expr, else_block)
 }
 
 #[derive(Debug)]
--- a/source/rust_verify/src/lifetime_ast.rs
+++ b/source/rust_verify/src/lifetime_ast.rs
@@ -109,7 +109,7 @@ pub(crate) type Stm = Box<(Span, StmX)>;
 #[derive(Debug, Clone)]
 pub(crate) enum StmX {
     Expr(Exp),
-    Let(Pattern, Typ, Option<Exp>),
+    Let(Pattern, Typ, Option<Exp>, Option<Exp>),
 }
 
 #[derive(Debug)]
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -2257,7 +2257,10 @@ fn stmt_to_stm(
             let (stms, exp) = expr_to_stm_opt(ctx, state, expr)?;
             Ok((stms, exp, None))
         }
-        StmtX::Decl { pattern, mode: _, init } => {
+        StmtX::Decl { pattern, mode: _, init, els } => {
+            if els.is_some() {
+                panic!("let-else should be simplified in ast_simpllify {:?}.", stmt)
+            }
             let (name, mutable) = match &pattern.x {
                 PatternX::Var { name, mutable } => (name, mutable),
                 _ => panic!("internal error: Decl should have been simplified by ast_simplify"),
--- a/source/rust_verify/src/lifetime_emit.rs
+++ b/source/rust_verify/src/lifetime_emit.rs
@@ -714,14 +714,21 @@ pub(crate) fn emit_stm(state: &mut EmitState, stm: &Stm) {
             emit_exp(state, exp);
             state.write(";");
         }
-        StmX::Let(pat, typ, init) => {
+        StmX::Let(pat, typ, init, els) => {
             state.write("let ");
             emit_pattern(state, pat);
-            state.write(": ");
-            state.write(typ.to_string());
+            if els.is_none() {
+                state.write(": ");
+                state.write(typ.to_string());
+            }
             if let Some(init) = init {
                 state.write(" = ");
                 emit_exp(state, init);
+            }
+            if let Some(els) = els {
+                state.write(" else {");
+                emit_exp(state, els);
+                state.write("}");
             }
             state.write(";");
         }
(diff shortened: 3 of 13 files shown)


source/vir/src/ast_to_sst.rs
```diff
@@ -943,18 +943,34 @@ fn if_to_stm(
...
 ///  * Unit - for the implicit unit case
 ///  * Never - the expression can never return (e.g., after a return value or break)
-
 pub(crate) fn expr_to_stm_opt(
     ctx: &Ctx,
     state: &mut State,
     expr: &Expr,
+) -> Result<(Vec<Stm>, ReturnValue), VirErr> {
+    let (stms, ret) = expr_to_stm_opt_inner(ctx, state, expr)?;
+    if expr_ret_never(expr) { Ok((stms, ReturnValue::Never)) } else { Ok((stms, ret)) }
```
utaal, line 967: How about
```suggestion
    if matches!(expr.typ.as_ref(), TypX::Decorate(crate::ast::TypDecoration::Never, ..)) { Ok((stms, ReturnValue::Never)) } else { Ok((stms, ret)) }
```
instead of introducing a separate function for it?

source/vir/src/ast_to_sst.rs
```diff
@@ -943,18 +943,34 @@ fn if_to_stm(
...
+}
+
+fn expr_to_stm_opt_inner(
```
utaal, line 970: Is this separate function necessary? Can the logic for "returning never" be folded into `expr_to_stm` instead?

source/vir/src/ast_to_sst.rs
```diff
@@ -943,18 +943,34 @@ fn if_to_stm(
...
 /// Note the 'ReturnValue' can be one of three things:
 ///  * Some(e) - means the expression e was returned
 ///  * Unit - for the implicit unit case
 ///  * Never - the expression can never return (e.g., after a return value or break)
-
 pub(crate) fn expr_to_stm_opt(
     ctx: &Ctx,
     state: &mut State,
     expr: &Expr,
+) -> Result<(Vec<Stm>, ReturnValue), VirErr> {
```
tjhance, line 965: unfortunately, it's not going to be correct to do this, because the never type needs to be considered differently in Spec code. Consider:

```
let x = arbitrary::<!>();
assert(false);
```

The reason the Expr::NeverToAny is allowed to return ReturnValue::Never is because we check (in modes.rs) that NeverToAny is never used in Spec code.

ziqiaozhou: Thanks for the example. I was not aware of the difference in Spec code. I restricted the NeverReturn check in ExprX::Call when it is in exec mode.

source/vir/src/ast_to_sst.rs
ziqiaozhou: Thanks for pointing that out!

ziqiaozhou:
This PR supports let-else when we do not need verus macro to replace content in else block. Since the current verus macro depends on syn v1.0, verus macro does not understand the let-else syntax and thus would never parse else block.
For example,
```
verus!{
fn f(a: Option<usize>) -> Result<usize, ()> {
  let Some(x) = a {
    proof {assert(false;}
    return Err(());
  };
  return Ok(x);
}
```

Is there any plan to upgrade the syn to a newer version?

utaal:
> when we do not need verus macro to replace content in else block

What happens with this macro if we do need to process the else block?

ziqiaozhou:
> > when we do not need verus macro to replace content in else block
> 
> What happens with this macro if we do need to process the else block?

Verus macro never visits else block since  `let x = ... else  {...}` would be treated  a Expr::Verbatim(TokenStream) by the old syn (version 1.0).  If not updating syn to a newer version, an easier solution to solve this would be to try to parse the Expr::Verbatim to a manually defined let-else Expr when visiting Expr and call expr_rewrite for init Expr and else block. I would leave this part as a TODO.

tjhance:
It might be a good idea to try to re-merge our current syn fork with the latest syn. However, I think it doesn't need to be done in this PR (and should probably be done by someone intimately familiar with our syn fork).

I have a few more comments on the PR:

 * I think we should only allow let-else in exec code. I could see it being useful in proof code, but it needs more thought to handle the complexities of ! in ghost code. 
 * Right now, the approach is relying on ast_to_sst to handle all the ways of producing 'never'. It would be better to wrap the RHS in a NeverToAny node, which will handle the never-turning reliably, and while will result in well-typed VIR (i.e., both arms of the match having the same type).
 * It might be easier to have an explicit VIR node for let_else, and handle the simplification in ast_simplify. The ast_simplify module turns match statements into ifs and handles pattern-binding. Implementing let-else in a similar way would let us remove the extraneous tuples from the encoding, and it would also let us get rid of the unfinished `pat_extract_binds` function. In particular, you could have it simplify to the following:

```
if !(pattern_is_matched) {
  NeverToAny(else_block);
}
let binder1 = ...;
let binder2 = ...;
...
let binder3 = ...;
```

Chris-Hawblitzel:
> It might be a good idea to try to re-merge our current syn fork with the latest syn. However, I think it doesn't need to be done in this PR (and should probably be done by someone intimately familiar with our syn fork).

I'm looking at this now.

Chris-Hawblitzel:
> > It might be a good idea to try to re-merge our current syn fork with the latest syn. However, I think it doesn't need to be done in this PR (and should probably be done by someone intimately familiar with our syn fork).
> 
> I'm looking at this now.

Ok, I made a pull request for this: https://github.com/verus-lang/verus/pull/1423

Chris-Hawblitzel:
> Ok, I made a pull request for this: #1423

https://github.com/verus-lang/verus/pull/1423 is now merged into main.

ziqiaozhou:
> * It might be easier to have an explicit VIR node for let_else

@tjhance Thanks for the suggestion. That does simplified the change.

> #1423 is now merged into main.

@Chris-Hawblitzel, thanks for upgrading the syn version so quickly! I rebased my change to use the new syn version. Now the else block is parsed correctly.:>

source/rust_verify/src/lifetime_ast.rs
```diff
@@ -109,7 +109,7 @@ pub(crate) type Stm = Box<(Span, StmX)>;
 #[derive(Debug, Clone)]
 pub(crate) enum StmX {
     Expr(Exp),
-    Let(Pattern, Typ, Option<Exp>),
+    Let(Pattern, Typ, Option<Exp>, Option<Exp>),
```
tjhance, line 112: Can you add a comment about what the arguments do?

tjhance (approved):
looks great, thanks for making the changes!

PR #1409, opened 2025-01-26
When printing --time statistics, include all buckets in total times

Chris-Hawblitzel:
The `--time` option prints a breakdown of the total times spent across all buckets.  However, I think these times were being reported incorrectly.  Here's an example of what they looked like for `verus-mimalloc` using `--num-threads 1 --time`:

```
total-time:          121015 ms
    rust-time:                3688 ms
        init-and-types:           2258 ms
        lifetime-time:             321 ms
        compile-time:             1108 ms
    verification-time:      117120 ms
        vir-time:                 1236 ms
            hir-time:                  391 ms
            import-time:                40 ms
            rust-to-vir:               402 ms
        verify-crate-time:      115883 ms
    unaccounted-time:          206 ms

verify-crate-time-breakdown
    total verify-time:          54751 ms   (1 threads)
    total air-time:              7833 ms   (1 threads)
    total smt-time:             30754 ms   (1 threads)
        total smt-init:              1736 ms   (1 threads)
        total smt-run:              29018 ms, 150588425 rlimit (1 threads)
```

The `verify-crate-time-breakdown` numbers don't make much sense in this; the total `verify-time` doesn't match `verify-crate-time`, and the `air-time` and `smt-time` account for only a third of `verify-crate-time`.  (In addition, the formatting makes it look like `air-time` and `smt-time` are adjacent to `verify-time` rather than subcomponents of `verify-time`.)

I think the reason for this is that the computation of the total times excluded singleton functions that were spun off and verified separately from their modules.

This pull request fixes this and adjusts the indentation.  The new output looks like:

```
total-time:          118020 ms
    rust-time:                3489 ms
        init-and-types:           2058 ms
        lifetime-time:             310 ms
        compile-time:             1119 ms
    verification-time:      114322 ms
        vir-time:                 1248 ms
            hir-time:                  393 ms
            import-time:                40 ms
            rust-to-vir:               397 ms
        verify-crate-time:      113073 ms
    unaccounted-time:          209 ms

verify-crate-time-breakdown
    total verify-time:         112607 ms   (1 threads)
        total air-time:             10822 ms   (1 threads)
        total smt-time:             83324 ms   (1 threads)
            total smt-init:              2707 ms   (1 threads)
            total smt-run:              80617 ms, 150588425 rlimit (1 threads)
```

Here, `verify-time` in the breakdown matches `verify-crate-time` and it's clear that the AIR + SMT time accounts for 94 seconds of the total 113 seconds, which is 83% of the time.


Chris-Hawblitzel:
> Can you please check the impact on the json output, too?

Here is the json for the affected places in the json output:

```
...
    "estimated-cpu-time": 120517,
...
    "total-verify": 115202,
...
    "air": {
      "total": 11057,
...
    "smt": {
      "total": 85312,
      "smt-init": 2758,
...
      "smt-run": 82554,
...
```

The rest should be unaltered.

source/rust_verify/src/main.rs
```diff
@@ -107,14 +107,21 @@ pub fn main() {
...
+            verifier: &rust_verify::verifier::Verifier,
+            f: impl Fn(&rust_verify::verifier::BucketStats) -> std::time::Duration,
+        ) -> u128 {
+            verifier.bucket_stats.iter().map(|(_, v)| f(v)).sum::<std::time::Duration>().as_millis()
+        }
+
         let mut smt_init_times = verifier
             .bucket_stats
             .iter()
             .filter(|(k, _)| k.function().is_none())
```
utaal, line 120: Is this filtering that's causing the issue? I admittedly forget why this line is here.
@tjhance @achreto ?

tjhance (commented):
It looks fine to me, thought I don't know much about this code.

source/rust_verify/src/main.rs
Chris-Hawblitzel: > Is this filtering that's causing the issue? I admittedly forget why this line is here.

It looks to me like the filtering is to split out `spinoff_prover` functions when reporting per-module statistics, so that they are reported separately from the modules that contain them.  This sounds useful to me, since `spinoff_prover` functions are often expensive functions that you'd want to keep an eye on individually when profiling.  I suspect, though, that it wasn't intended for these to be excluded from the total times.

utaal (approved):

PR #1414, opened 2025-01-27
Remove oudated comment in seqs.rs

ahuoguo:
From #1317, `seq!` with more than one element now doesn't encode to `.push`, which makes the comments here false.

Sidenote: the last test in `seqs.rs`:
```rurt
            // Test for successful broadcast of push_distributes_over_add
            assert((s2 + s4).push(120) == s2 + s4.push(120));
```
doesn't seem to be a useful test sine this assertion still verifies if `push_distributes_over_add` is removed from `group_seq_lib_default`. It verifies with the following `axiom-usage-info`:
```
note: checking this function used these broadcasted lemmas and broadcast groups:
        - (group) vstd::seq::group_seq_axioms,
        - (group) vstd::seq_lib::group_seq_lib_default,
        - (group) vstd::group_vstd_default,
        - vstd::seq::axiom_seq_push_len,
        - vstd::seq::axiom_seq_push_index_same,
        - vstd::seq::axiom_seq_push_index_different,
        - vstd::seq::axiom_seq_ext_equal,
        - vstd::seq::axiom_seq_add_len,
        - vstd::seq::axiom_seq_add_index1,
        - vstd::seq::axiom_seq_add_index2,
        - vstd::seq_lib::impl&%0::filter_lemma,
        - vstd::seq_lib::impl&%0::filter_distributes_over_add
```

### DIFF
--- a/source/rust_verify_test/tests/seqs.rs
+++ b/source/rust_verify_test/tests/seqs.rs
@@ -127,6 +127,8 @@ test_verify_one_file! {
             assert(forall|i: nat| i < s2.len() ==> s2[i as int] < 40);
             // Test for successful broadcast of filter_distributes_over_add
             assert((s1 + s3).filter(|x: int| x < 40) == (s2 + s4));
+            // TODO: the following test will verify even if
+            // push_distributes_over_add is not broadcasted.
             // Test for successful broadcast of push_distributes_over_add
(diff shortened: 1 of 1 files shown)


utaal:
Is the side-note relevant to this PR? Perhaps add a TODO in the PR with the comment, so we can keep track of it, if it's not something we want to handle in this PR.

PR #1418, opened 2025-01-28
Add model for vstd `Map` and prove its axioms

matthias-brun:
This PR defines a model for vstd's `Map` and proves most of the axioms. The `Map` type is still opaque and its spec functions (`insert`, ..) are `closed`, so none of the model's internals should leak outside of the module.

I'll open a separate PR for the `Set` model, which is more complicated and needs some cleanup first.

### DIFF
--- a/source/vstd/map.rs
+++ b/source/vstd/map.rs
@@ -3,7 +3,6 @@
 use super::pervasive::*;
 use super::prelude::*;
 use super::set::*;
-use core::marker;
 
 verus! {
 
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel (approved):

jaybosamiya-ms:
Just making a note, in case anyone is interested in how other tools model things: 

F* has two models for maps. The interface files (which prevent the model details from being seen/used by users, roughly equivalent to the closed/opaque interfaces we have here) are [`FStar.Map.fsti`](https://github.com/FStarLang/FStar/blob/7cd06c5562fc47ec14cd35c38034d5558a5ff762/ulib/FStar.Map.fsti) and [`FStar.PartialMap.fsti`](https://github.com/FStarLang/FStar/blob/7cd06c5562fc47ec14cd35c38034d5558a5ff762/ulib/FStar.PartialMap.fsti). The former (with model+proofs in [`FStar.Map.fst`](https://github.com/FStarLang/FStar/blob/7cd06c5562fc47ec14cd35c38034d5558a5ff762/ulib/FStar.Map.fst)) is the pair of a total function from key->value _and_ a set of keys tracking which values are valid. In contrast, the latter (with model+proofs in [`FStar.PartialMap.fst`](https://github.com/FStarLang/FStar/blob/master/ulib/FStar.PartialMap.fst)) is closer to the model in this PR (i.e., as a function from key->option value).

Note: due to F* technicalities around functional extensionality + dependent typing, they need to use something called a domain-restricted functions, but that is a subtle detail that we don't have to worry about in Verus due to lack of dependent typing.

I don't believe there is a deep insight in the difference in the design, but since I remembered there was something that seemed different from this PR's design that I had seen in F* (years ago), I decided to take another look, and decided to document for anyone who might be curious.

PR #1422, opened 2025-01-31
Add ext_equal attr to atomic state_machine! generated State structs

jonhnet:
In VeriSplinter, we have a `state_machine!`-generated `MapSpec` whose State we use as a field in other data structures. Some proofs require asserting that the outer structures are equal, which `#[verifier(ext_equal)]` takes care of. But we need the macro to emit it.

### DIFF
--- a/source/state_machines_macros/src/to_token_stream.rs
+++ b/source/state_machines_macros/src/to_token_stream.rs
@@ -301,6 +301,7 @@ pub fn output_primary_stuff(
     let attrs = &bundle.sm.attrs;
     let code: TokenStream = quote_spanned! { sm.fields_named_ast.span() =>
         #[cfg_attr(verus_keep_ghost, verus::internal(verus_macro))]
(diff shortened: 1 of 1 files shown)


source/state_machines_macros/src/to_token_stream.rs
```diff
@@ -301,6 +301,7 @@ pub fn output_primary_stuff(
     let attrs = &bundle.sm.attrs;
     let code: TokenStream = quote_spanned! { sm.fields_named_ast.span() =>
         #[cfg_attr(verus_keep_ghost, verus::internal(verus_macro))]
+        #[verifier::ext_equal]
```
Chris-Hawblitzel, line 304: You might want to make this `#[cfg_attr(verus_keep_ghost, verifier::ext_equal)]`.

Chris-Hawblitzel (approved):

PR #1426, opened 2025-02-01
Add model for vstd Set and prove its axioms

matthias-brun:
This PR proves the set axioms and adds a fold function for finite sets (which is used to define cardinality) with associated lemmas.

The way things are bootstrapped is as follows:
* Sets are defined as functions from key to boolean
* A set is finite if there exists a surjective mapping into the natural numbers up to some upper bound
* This allows us to define most set operations except for cardinality
* We then define `fold` on finite sets by porting Isabelle/HOL's definitions and lemmas for it
  * Because of the inductive predicate-style encoding, we don't need a decreasing measure for induction proofs (We could use e.g. the finiteness upper bound but it's very annoying to do)
* Finally, we define set cardinality as `self.fold(|a, b: nat| b + 1, 0)`

I'm not sure about where the `fold` module should go. The current place (inline in `set.rs`) probably isn't it. But it needs to be a submodule of `set` to see the bodies of `closed` functions and the non-public items. We could move `set.rs` to `set/mod.rs` and then have `set/fold.rs`, which wouldn't break any imports of `vstd::set`. But arguably `set_lib.rs` should then also live in the `set` directory, which would be a breaking change. How should I best handle this?

Some miscellaneous notes:
- Isabelle uses a 3-clause BSD license, which is compatible with Verus' MIT license
- Isabelle's fold lemmas only require the function to be commutative on the set to be folded. I decided to require fully commutative functions
  - Triggering is tricky enough with these higher-order predicates, adding another set into the mix that might be defined with a closure doesn't help.
- There was an existing definition of `fold` for sets, which I replaced with the new one
  - It was based on the recursive encoding with `self.choose()` which makes reasoning about it very hard
  - That's presumably the reason why there weren't any associated lemmas for it
  - Maybe some of the admitted proofs in `tokens.rs` can now be proved (I didn't try)
- Aside: The obvious recursive fold (`if s == set![] { z } else { f(s.choose(), fold(s.remove(s.choose())), f, z) }`) is hard to reason about because the induction hypothesis is useless. The elements of `s.remove(s.choose())` might be folded in a completely different order than those of `s`.

@tjhance I added you as a reviewer since `tokens.rs` seems to be the only current user of `fold` (within Verus/vstd)

### DIFF
--- a/source/rust_verify_test/tests/sets.rs
+++ b/source/rust_verify_test/tests/sets.rs
@@ -111,7 +111,7 @@ test_verify_one_file! {
 
         proof fn test() {
             let s: Set<nat> = set![9];
-            reveal_with_fuel(Set::fold, 10);
+            broadcast use fold::lemma_fold_insert, fold::lemma_fold_empty;
             assert(s.finite());
             assert(s.len() > 0);
             assert(s.fold(0, |p: nat, a: nat| p + a) == 9);
--- a/source/vstd/map.rs
+++ b/source/vstd/map.rs
@@ -41,13 +41,13 @@ impl<K, V> Map<K, V> {
 
     /// Gives a `Map<K, V>` whose domain contains every key, and maps each key
     /// to the value given by `fv`.
-    pub open spec fn total(fv: impl Fn(K) -> V) -> Map<K, V> {
+    pub open spec fn total(fv: spec_fn(K) -> V) -> Map<K, V> {
         Set::full().mk_map(fv)
     }
 
     /// Gives a `Map<K, V>` whose domain is given by the boolean predicate on keys `fk`,
     /// and maps each key to the value given by `fv`.
-    pub open spec fn new(fk: impl Fn(K) -> bool, fv: impl Fn(K) -> V) -> Map<K, V> {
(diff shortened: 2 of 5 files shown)


tjhance (approved):

tjhance:
having fold as a submodule seems fine to me.

utaal:
After the rebase:

veritas report for verus `verified-set-axioms` (`76eb59694f4497d04d98e3cdfa8c338c059b3714`) with features: `singular`

| project | revspec | outcome | total verus time (ms) | smt run time (ms) |
| ------- | ------- | ------- | --------------------- | ----------------- |
| verus-vstd | `verified-set-axioms` (`76eb59694f4497d04d98e3cdfa8c338c059b3714`) | success (851 verified, 0 errors) | 25745 | 10075 |
| memory-allocator | `main` (`bc2ba93dc916aef3afa347b92c5c8d66cb23fee7`) | success (730 verified, 0 errors) | 66756 | 146411 |
| verified-storage | `main` (`7b35ddb12fd505e16a58d9462f15aed559570267`) | success (272 verified, 0 errors) | 9665 | 10523 |
| node-replication | `main` (`b3ca921103220b20c6cc144bd2f7d4f259b7c8be`) | success (254 verified, 0 errors) | 8693 | 2683 |

PR #1427, opened 2025-02-04
Add lemmas for map `union_prefer_right`

matthias-brun:
This PR adds some lemmas for `union_prefer_right` that we're using in the page table verification. Even though the proofs just boil down to extensional equality, having these lemmas available for broadcasting is useful.


utaal (approved):
Nice, thanks!

PR #1433, opened 2025-02-05
Override vargo target dir

### LINKED ISSUE #1432: `tools/activate` script assumes target dir location
Currently, the shell script `tools/activate` assumes that the build artifacts of `vargo` are always going to be in `tools/vargo/target/release` which might not be the case since users can override the target directory system-wide.

https://github.com/verus-lang/verus/blob/b6df2c569aebfb39077922159f9c531c9afe8c4f/tools/activate#L19

This could be fixed by either fetching the correct target path using something like this:
```sh
export PATH="$(cd $SCRIPT_DIR/vargo && cargo metadata --format-version 1 --no-deps | jq -r '.target_directory')release/:$PATH"
```
or by setting `target-dir = "target"` in a local `.cargo/config.toml`.

slerpyyy:
Fixes #1432

### DIFF
--- a/tools/vargo/.cargo/config.toml
+++ b/tools/vargo/.cargo/config.toml
@@ -1,0 +1,5 @@
+# Set target directory in case it was overridden by the environment
+# https://doc.rust-lang.org/cargo/reference/config.html
+
+[build]
+target-dir = "target"


utaal (approved):

utaal:
Thank you!

PR #1434, opened 2025-02-05
Support slice::get && Ignore trait predicates when not visible to local.

ziqiaozhou:
* Why: trait SliceIndex<T: ?Sized>: private_slice_index::Sealed
* Add spec for slice::get

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -1389,18 +1389,9 @@ pub(crate) fn remove_ignored_trait_bounds_from_predicates<'tcx>(
     use rustc_middle::ty::{ConstKind, ScalarInt, ValTree};
     preds.retain(|p: &Clause<'tcx>| match p.kind().skip_binder() {
         rustc_middle::ty::ClauseKind::<'tcx>::Trait(tp) => {
-            let td = tcx.trait_def(tp.trait_ref.def_id);
-            // Recursively check parent path visibility
-            let mut visible = tcx.visibility(td.def_id).is_visible_locally();
-            let mut def_id = td.def_id;
-            while let Some(parent_def_id) = tcx.opt_parent(def_id) {
-                if !tcx.visibility(parent_def_id).is_visible_locally() {
-                    visible = false;
-                    break;
-                }
-                def_id = parent_def_id;
-            }
-            if !visible {
+            // Skip private trait bounds
+            let path = tcx.def_path_str(tp.trait_ref.def_id);
+            if path == "core::slice::index::private_slice_index::Sealed" {
                 false
             } else if in_trait
                 && trait_ids.contains(&tp.trait_ref.def_id)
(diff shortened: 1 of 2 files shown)


source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -1389,7 +1389,22 @@ pub(crate) fn remove_ignored_trait_bounds_from_predicates<'tcx>(
     use rustc_middle::ty::{ConstKind, ScalarInt, ValTree};
     preds.retain(|p: &Clause<'tcx>| match p.kind().skip_binder() {
         rustc_middle::ty::ClauseKind::<'tcx>::Trait(tp) => {
-            if in_trait && trait_ids.contains(&tp.trait_ref.def_id) && tp.trait_ref.args.len() >= 1
+            let td = tcx.trait_def(tp.trait_ref.def_id);
+            // Recursively check parent path visibility
+            let mut visible = tcx.visibility(td.def_id).is_visible_locally();
+            let mut def_id = td.def_id;
+            while let Some(parent_def_id) = tcx.opt_parent(def_id) {
+                if !tcx.visibility(parent_def_id).is_visible_locally() {
```
Chris-Hawblitzel, line 1397: It looks like this is rejecting some traits that actually are visible, like `Iterator`:

```
td.def_id = DefId(2:8781 ~ core[145f]::iter::traits::iterator::Iterator)
parent_def_id = DefId(2:8736 ~ core[145f]::iter::traits::iterator)
```

In https://github.com/rust-lang/rust/blob/master/library/core/src/iter/traits/mod.rs , it looks like `Iterator` is made visible via a `pub use` after being initially declared in a private module:

```
mod accum;
mod collect;
mod double_ended;
mod exact_size;
mod iterator;
mod marker;
mod unchecked_iterator;

#[unstable(issue = "none", feature = "inplace_iteration")]
pub use self::marker::InPlaceIterable;
#[unstable(issue = "none", feature = "trusted_fused")]
pub use self::marker::TrustedFused;
#[unstable(feature = "trusted_step", issue = "85731")]
pub use self::marker::TrustedStep;
pub(crate) use self::unchecked_iterator::UncheckedIterator;
#[stable(feature = "rust1", since = "1.0.0")]
pub use self::{
    accum::{Product, Sum},
    collect::{Extend, FromIterator, IntoIterator},
    double_ended::DoubleEndedIterator,
    exact_size::ExactSizeIterator,
    iterator::Iterator,
    marker::{FusedIterator, TrustedLen},
};
```

ziqiaozhou: Oh. I see. Then I do not know how to check whether a trait is reacheable. I did not find a correct helper function from TyCtxt. Maybe do a hard check on path?
```
let path = tcx.def_path_str(tp.trait_ref.def_id);
if path == "core::slice::index::private_slice_index::Sealed" {
    false
} else if
```

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -1389,18 +1389,9 @@ pub(crate) fn remove_ignored_trait_bounds_from_predicates<'tcx>(
...
-                if !tcx.visibility(parent_def_id).is_visible_locally() {
-                    visible = false;
-                    break;
-                }
-                def_id = parent_def_id;
-            }
-            if !visible {
+            // Skip private trait bounds
+            let path = tcx.def_path_str(tp.trait_ref.def_id);
+            if path == "core::slice::index::private_slice_index::Sealed" {
```
Chris-Hawblitzel, line 1394: This would be better as a call to `verus_items::get_rust_item`, but otherwise it's fine.

Chris-Hawblitzel (approved):

PR #1435, opened 2025-02-06
Add `!is`/`!has` syntax as shorthand for negated `is`/`has`

matthias-brun:
Negation is annoying to use with the `is` syntax, since it requires wrapping in parentheses: `!(x is Variant)`. This PR adds `!is` syntax, which expands to `!(x is Variant)`. Similarly it adds `!has` syntax for negated `has`.

### DIFF
--- a/dependencies/prettyplease/src/expr.rs
+++ b/dependencies/prettyplease/src/expr.rs
@@ -83,6 +83,7 @@ impl Printer {
                 self.expr_big_op(&expr.exprs.iter().map(|e| &e.expr).collect(), true)
             }
             Expr::Is(expr) => self.expr_is(expr),
+            Expr::Isnt(expr) => self.expr_isnt(expr),
             Expr::Has(expr) => self.expr_has(expr),
             Expr::GetField(expr) => self.expr_get_field(expr),
             Expr::Matches(m) => self.expr_matches(m),
@@ -117,6 +118,13 @@ impl Printer {
         self.outer_attrs(&expr.attrs);
         self.expr(&expr.base, FixupContext::NONE);
         self.word(" is ");
+        self.ident(&expr.variant_ident);
+    }
+
+    pub fn expr_isnt(&mut self, expr: &syn_verus::ExprIsnt) {
+        self.outer_attrs(&expr.attrs);
+        self.expr(&expr.base, FixupContext::NONE);
+        self.word(" isnt ");
         self.ident(&expr.variant_ident);
     }
 
--- a/dependencies/prettyplease/src/expr.rs
+++ b/dependencies/prettyplease/src/expr.rs
@@ -85,6 +85,7 @@ impl Printer {
             Expr::Is(expr) => self.expr_is(expr),
             Expr::IsNot(expr) => self.expr_isnot(expr),
             Expr::Has(expr) => self.expr_has(expr),
+            Expr::HasNot(expr) => self.expr_hasnot(expr),
             Expr::GetField(expr) => self.expr_get_field(expr),
             Expr::Matches(m) => self.expr_matches(m),
 
@@ -132,6 +133,13 @@ impl Printer {
         self.outer_attrs(&expr.attrs);
         self.expr(&expr.lhs, FixupContext::NONE);
         self.word(" has ");
+        self.expr(&expr.rhs, FixupContext::NONE);
+    }
+
+    pub fn expr_hasnot(&mut self, expr: &syn_verus::ExprHasNot) {
+        self.outer_attrs(&expr.attrs);
+        self.expr(&expr.lhs, FixupContext::NONE);
+        self.word(" !has ");
         self.expr(&expr.rhs, FixupContext::NONE);
     }
 
--- a/dependencies/prettyplease/src/expr.rs
+++ b/dependencies/prettyplease/src/expr.rs
@@ -83,7 +83,7 @@ impl Printer {
                 self.expr_big_op(&expr.exprs.iter().map(|e| &e.expr).collect(), true)
             }
             Expr::Is(expr) => self.expr_is(expr),
-            Expr::Isnt(expr) => self.expr_isnt(expr),
+            Expr::IsNot(expr) => self.expr_isnot(expr),
             Expr::Has(expr) => self.expr_has(expr),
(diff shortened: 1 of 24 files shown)


Chris-Hawblitzel:
I like the idea of having a more compact syntax for `!(x is Variant)`, as this PR proposes.  Dafny has something similar to this already, with its `in` and `!in` infix operators.  Since we have `is` and `has` infix operators, I'd propose `!is` and `!has` to avoid introducing extra words and for consistency with `==`/`!=`.

matthias-brun:
Thanks for the suggestion, I like it! I'll change the syntax accordingly and add it for `!has` as well.

jaybosamiya-ms:
Surely, for consistency with `==` => `!=`, we should be doing this[⸮](https://en.wikipedia.org/w/index.php?title=Sarcasm_punctuation&redirect=no)
image

jaybosamiya-ms:
I do like the `!is` and `!has` infix operators, fwiw :)

parno:
We'll need to update both `verusfmt` and `verus-analyzer` to recognize the new syntax.

tjhance:
jay is objectively correct. !s and !as are the only sensible choices

matthias-brun:
@Chris-Hawblitzel It looks like getting `!is` to parse right isn't quite as straightforward. I think you have a better understanding of `syn`, so maybe you can give me a pointer on what the best way to implement this is.

I tried the naive thing of just replacing the `isnt` token in my current code by `!is` but that fails as it seems to try and parse the `!` as negation instead. The result is `error: expected delimiter` (with `is` underlined). The token stream has the two tokens `!` and `is` rather than a single `!is` token. So I suppose the first question is: Am I interpreting correctly that making `!is` a token is the wrong approach? We don't seem to reject Verus keywords as identifiers (e.g. `let is: nat = 1;`) so unambiguous tokenization wouldn't be possible anymore.

utaal:
There is some functionality to merge tokens to handle these cases - I can't look now but I should be able to take a look next week.

utaal:
I pushed a commit that treats `!is` as a single token (so it disallows `! is`) and is consistent with how we handle other compound tokens. I needed to rename it internally to `isnt`, because `proc_macro` doesn't accept tokens that are a mix of a character and an identifier. @Chris-Hawblitzel does that make sense to you?

However, I think the main issue with `!is` is this case Matthias spotted:
https://github.com/verus-lang/verus/pull/1435/files#diff-38b1650a5373a031c934060b5a2588c88feb0c515b81307f121b982ea5bb02b8R1152-R1157

```
        proof fn let_with_is_ident() {
            let is = false;
            assert(!is);
        }
```

which now fails to parse, and I think making it contextual would be a non-trivial endeavor. That said, we've already decided that `is` is a keyword (for the existing `a is X` syntax).

utaal:
@Chris-Hawblitzel is looking into parsing `!is` contextually as two tokens (the same way we handle `is`).
@parno and @jaybosamiya-ms as additional reviewers because this changes syntax that needs to be reflected in verusfmt and VScode.

parno:
I took a quick look at verus analyzer.  It turns out that it currently only supports `is` but not `has`.  Apparently the latter doesn't show up in `vstd`.  

One existing annoyance is that due to the existing design of rust-analyzer's parser, back when I added support for `is`, I had to make `is` a first-class keyword to get it to parse correctly.  That means you can't use `is` as a variable name, for instance.  The only other Verus keywords that I have to do that for are "ghost", "tracked", "forall", "exists", and "matches".  All of the others are conditional keywords, which means they're still valid for, e.g., identifiers.

For this PR in particular, it looks doable to support `!is`, I would just have to work around the fact that `rust-analyzer` currently appears to allow macro calls with a space between the macro name and the exclamation mark (i.e., `print !("hello")` is okay).

jaybosamiya-ms:
Verusfmt already supports `!is` (I added it to its `main` branch for a couple of weeks now; I have triggered a new release, which should be live in a few minutes once CI is done https://github.com/verus-lang/verusfmt/commit/214299b2ba481eb55d38f7cc02f362bddaf775da).

Regarding contextual vs not: I think having `is` as a non-contextual keyword is perfectly fine. I don't see anyone using `is` as the name of a variable or a struct field or such. There _may_ be users who have a method named `is` (~`a.is(b)`), but I don't see that as being common with standard Rust conventions (`eq` or similar would be the conventional name instead). Also, if someone _really_ needs to interface with external code with `is` used as an identifier, there is always the raw-identifier syntax `r#is`, so we are not even restricting interfacing with unchangeable external code).  Generally speaking, I very much prefer non-contextual keywords, rather than contextual ones, since non-contextual keywords are easier to understand and deal with (e.g., writing a syntax highlighter is easier).

Re macro calls: I had to handle something similar for verusfmt too :)  At least for versufmt it was easy to fix by making macro calls look-ahead for an opening brace/bracket/paren _immediately_ after the `!` and otherwise back out.

parno:
> Re macro calls: I had to handle something similar for verusfmt too :) At least for versufmt it was easy to fix by making macro calls look-ahead for an opening brace/bracket/paren immediately after the ! and otherwise back out.

Interesting.  In rust-analyzer, it appears they have a look ahead for `!=` and back out in that case, so I added another one for `is`, and that seems to work.

parno:
FWIW, I have a branch of verus-analyzer where `!is`, and `!has` appear to be working.

matthias-brun:
The PR as of now implements `!is` and `!has` with:
- `is` being a soft keyword, i.e. it can still be used as an identifier like in `let is = false; assert(!is);`
- spaces as in `x !  is Variant` resulting in a syntax error

I think this matches the general consensus of the Slack discussion on this topic.

Thanks @Chris-Hawblitzel for suggesting how to fix `is`/`has` being treated as hard keywords!

utaal (approved):
LGTM. Thanks!
Please wait to merge until @jaybosamiya-ms and @parno confirm support in verusfmt and verus-analyzer.

jaybosamiya-ms (approved):
Thanks for confirming, Verusfmt has had support for `Iis`/`!has` on its latest stable release ([v0.5.3](https://github.com/verus-lang/verusfmt/releases/tag/v0.5.3)) since a little over a month (https://github.com/verus-lang/verus/pull/1435#issuecomment-2667080418), so no blocker there.

PR #1438, opened 2025-02-06
Change prettyplease_verus "links" field to prettyplease-verus02 to avoid conflicting with the standard prettyplease02

Chris-Hawblitzel:
This came from a discussion with @jaylorch and @jaybosamiya regarding a cargo conflict that arose because our fork of prettyplease was still using the same `links` field as the standard prettyplease. Conceptually, at least, we propose to have our own `links` "binary" as part of our own fork (albeit without any actual binary, because there doesn't appear to be an actual binary).


jaylorch (approved):
This makes it work for me. I checked running both `verus` and `cargo`. Thanks!

utaal (approved):

PR #1439, opened 2025-02-07
Spec for Vec::extend_from_slice

jaylorch:

### DIFF
--- a/source/rust_verify/example/std_test/vec_test.rs
+++ b/source/rust_verify/example/std_test/vec_test.rs
@@ -1,0 +1,20 @@
+use std::vec::Vec;
+
+use vstd::pervasive::runtime_assert;
+use vstd::prelude::*;
+
+verus! {
+
+fn vec_extend_slice_test() {
+    let mut a: Vec<u32> = vec![1, 2];
+    let b: Vec<u32> = vec![3, 4];
+    a.extend_from_slice(b.as_slice());
+    runtime_assert(a.len() == 4);
+    runtime_assert(a[0] == 1);
+    runtime_assert(a[1] == 2);
+    runtime_assert(a[2] == 3);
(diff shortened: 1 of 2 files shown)


tjhance:
since this clones, the specification needs to use T::clone

jaylorch:
> since this clones, the specification needs to use T::clone

Ah, good point, thanks! I've updated it accordingly.

zeldovich:
I think that, as Travis pointed out to me for the `Vec::resize()` spec, the implementation can sometimes elide the call to `clone()`.  In this particular case, `Vec::extend_from_slice()` will elide the call to `clone()` if `T` implements `Copy`:

```
% cat extend.rs
struct S1 {}

impl Clone for S1 {
    fn clone(&self) -> Self {
        println!("S1.clone()");
        S1{}
    }
}

struct S2 {}

impl Copy for S2 {
}

impl Clone for S2 {
    fn clone(&self) -> Self {
        println!("S2.clone()");
        S2{}
    }
}

fn test<T: Clone>(x: T) {
    let mut v = Vec::new();
    let s = &[x];
    v.extend_from_slice(s);
    let l = v.len();
    println!("len = {l}");
}

fn main() {
    println!("== test(S1) ==");
    test(S1{});

    println!("== test(S2) ==");
    test(S2{});
}
% rustc extend.rs
% ./extend
== test(S1) ==
S1.clone()
len = 1
== test(S2) ==
len = 1
% 
```

so perhaps spec should state either `T::clone` or equality.

tjhance:
Is that the case? the input is a reference to a slice. I think there is no possibility here but to do a clone.

tjhance:
looking into this more ... it looks like the implementation of extend_from_slice accomplishes this via trait specialization

tjhance (approved):

PR #1440, opened 2025-02-09
use 'cloned_or_eq' everywhere

tjhance:
This adds two helper functions, `cloned` and `cloned_or_eq`, so we don't have to write the `call_ensures` monstrosity everywhere.

```rust
/// `b` could be the result of calling `a.clone()`
pub open spec fn cloned<T: Clone>(a: T, b: T) -> bool {
    call_ensures(T::clone, (&a,), b)
}

/// `b` could be the result of calling `a.clone()` or is equal to `a`.
///
/// This is useful in places where 'clone' calls might be optimized to copies.
pub open spec fn cloned_or_eq<T: Clone>(a: T, b: T) -> bool {
    cloned(a, b) || a == b
}
```

As observed in #1439 by @zeldovich, some functions in Vec need to use `cloned_or_eq` because some clones may be optimized to copies. This turns out to be true for `Box::clone` as well. This PR also changes several existing functions to use `cloned_or_eq` where they were previously using the equivalent of `cloned`.

We should probably default to `cloned_or_eq` in the future.

### DIFF
--- a/source/docs/guide/src/container_bst_clone.md
+++ b/source/docs/guide/src/container_bst_clone.md
@@ -53,6 +53,8 @@ to write its ensures clause _in terms of_ the ensures clause for `V::clone`.
 This can be done using [`call_ensures`](./exec_funs_as_values.md).
 The predicate `call_ensures(V::clone, (&self@[key],), res@[key])` effectively says
 "`self@[key]` and `res@[key]` are a possible input-output pair for `V::clone`".
+This predicate is a mouthful, so `vstd` provides a helper function:
+<code class="hljs"><a href="https://verus-lang.github.io/verus/source/doc/vstd/pervasive/fn.cloned.html">cloned::&lt;V&gt;</a>(self@[key], res@[key])</code>
 
 ### Understanding the implications of the signature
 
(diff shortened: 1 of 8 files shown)


zeldovich (approved):
Looks good to me.

Perhaps consider eliminating `cloned_or_eq`, and defining `cloned` to allow equality, and use that everywhere (even in places that, at least in Rust's current implementation, don't do the copy optimization)?  The rationale is that Rust's developers seem to feel that it's fine to elide calls to clone (e.g., if the type implements Copy), without explicitly describing this to the caller.  So the fact that `Option::<T>::clone()` happens to always call clone right now is an artifact of the implementation, rather than a long-term promise that the API implies.

tjhance:
I think that's probably a good idea.

jaylorch:
OK, let me know when you've made that change and I'll review it.

tjhance:
Renamed to `cloned` and `strictly_cloned`, with `cloned` being the recommended default and including the equality disjunct.

jaylorch (approved):
Looks great. I look forward to using this ergonomic function in specs.

PR #1450, opened 2025-02-15
impl Debug for Tracked

ziqiaozhou:
Add Debug and Default Trait for Tracked, so that a struct with derived Debug/Default can use Tracked field. 

I have used some tracked variable in struct that implements Debug and Default. If I do not implement Debug and Default for Tracked, I will not be able to pass lifetime checker.

```
tracked struct P {
   no_copy: NoCopy,
}
#[derive(Debug, Default)]
struct X {
    x: [usize; 16],
    #[cfg(verus_keep_ghost_body)]
    y: Tracked<Map<int, P>>,
}
```

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -376,12 +376,6 @@ impl<A> core::fmt::Debug for Tracked<A> {
     }
 }
 
-impl<A> core::default::Default for Tracked<A> {
-    fn default() -> Self {
-        Tracked::assume_new()
-    }
-}
-
 impl<A> Ghost<A> {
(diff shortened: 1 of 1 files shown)


tjhance:
The `Default` implementation for `Tracked<A>` will allow the user to forge tracked tokens.

ziqiaozhou:
> The `Default` implementation for `Tracked<A>` will allow the user to forge tracked tokens.

Thanks for the comment. That is true. I removed the default. I think I can avoid using default trait.

Chris-Hawblitzel (approved):

PR #1459, opened 2025-02-19
Overflowable integers

jaylorch:
This `vstd` feature defines the `OverflowableU32` and `OverflowableU64` structs and their associated methods. They handle `u32` and `u64` values that can overflow. Each struct includes a ghost value representing the true `nat` value (not subject to overflow), so that the `view` function can provide the true value.

It's a fully verified library, i.e., it contains no trusted code.

### DIFF
--- a/source/rust_verify/example/guide/invariants.rs
+++ b/source/rust_verify/example/guide/invariants.rs
@@ -3,6 +3,7 @@ use builtin::*;
 #[allow(unused_imports)]
 use builtin_macros::*;
 use vstd::prelude::*;
+use vstd::arithmetic::overflow::*;
 
 verus! {
 
@@ -128,6 +129,35 @@ fn fib_impl(n: u64) -> (result: u64)
     cur
 }
 // ANCHOR_END: fib_final
+
+// ANCHOR: fib_checked
+fn fib_checked(n: u64) -> (result: u64)
+    requires
+        fib(n as nat) <= u64::MAX
+    ensures
+        result == fib(n as nat),
+{
+    if n == 0 {
+        return 0;
+    }
+    let mut prev: CheckedU64 = CheckedU64::new(0);
+    let mut cur: CheckedU64 = CheckedU64::new(1);
+    let mut i: u64 = 1;
+    while i < n
+        invariant
+            0 < i <= n,
+            fib(n as nat) <= u64::MAX,
+            cur@ == fib(i as nat),
+            prev@ == fib((i - 1) as nat),
+    {
+        i = i + 1;
+        let new_cur = cur.add_checked(&prev);
+        prev = cur;
+        cur = new_cur;
+    }
+    cur.unwrap()
+}
+// ANCHOR_END: fib_checked
 
 
 // ANCHOR: bank_spec
--- a/source/rust_verify/example/guide/invariants.rs
+++ b/source/rust_verify/example/guide/invariants.rs
@@ -159,6 +159,35 @@ fn fib_checked(n: u64) -> (result: u64)
 }
 // ANCHOR_END: fib_checked
 
+// ANCHOR: fib_checked_no_precondition
+fn fib_checked_no_precondition(n: u64) -> (result: Option<u64>)
+    ensures
+        match result {
+            Some(x) => x == fib(n as nat),
+            None => fib(n as nat) > u64::MAX,
+        },
+{
+    if n == 0 {
+        return Some(0);
+    }
+    let mut prev: CheckedU64 = CheckedU64::new(0);
+    let mut cur: CheckedU64 = CheckedU64::new(1);
+    let mut i: u64 = 1;
+    while i < n
+        invariant
+            0 < i <= n,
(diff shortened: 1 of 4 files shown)


utaal:
This seems useful! I have two questions:
* can this leverage https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add, https://doc.rust-lang.org/std/primitive.u32.html#method.checked_mul, and other similar methods? (This may also help with performance)
* could we generate the two mostly-similar but different `OverflowableU32` and `OverflowableU64` with a macro? (so, e.g. we can also get `OverflowableU8` and `OverflowableU128` etc.? I believe this would be the idiomatic rust approach to handling implementations for a class of native types? (I'm happy to help, feel free to ping me)
* Perhaps these could be named `CheckedU32` and `CheckedU64` as they aren't actually allowed to overflow, they instead track whether they overflowed.

utaal:
Instead of `checked_add` and `checked_mul`, this could also use https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_add and https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_mul

jaylorch:
> Instead of `checked_add` and `checked_mul`, this could also use https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_add and https://doc.rust-lang.org/std/primitive.u32.html#method.overflowing_mul

Oh, thanks! I was looking for something like these, but could only find ones that were experimental/nightly. I'll switch to one of those.

jaylorch:
> * Could these implement `std::ops::Add` and `std::ops::Mul` (so one can use `+` and `*`)? I don't know if our trait support allows us to do that at the moment.

It appears that we can implement `Add` and `Mul` but another missing verifier feature makes it not useful:

> error: The verifier does not yet support the following Rust feature: cmp or arithmetic for non smt arithmetic types

parno:
Will Ziqiao's work on #1465 address that?

jaylorch:
OK, I've taken @utaal's suggestions:

* It now uses `checked_add` and `checked_mul` for a faster implementation.
* It renames the types to be things like `CheckedU64` instead of `OverflowingU64`.
* It uses a macro to synthesize the code for everything from `u8` to `u128` (creating `CheckedU8`, `CheckedU16`, etc.).
* I considered implementing `Add` and `Mul` but discovered that this won't be useful until #1465 has been merged. I can always add that later.

jaylorch:
I also added another example `fib_checked` to show how using `CheckedU64` makes the problem of dealing with overflow much easier. For a small performance/space cost, the user doesn't even need to prove that the Fibonacci function is monotonic!

utaal (approved):
This looks great and very useful!

parno (commented):
This does look nice, and the `fib` examples help illustrate how it can simplify verification (in exchange for a small perf hit).  Two quick questions:
1. It looks like you added source anchors to `invariants.rs`, but I don't see updates to the  Guide pages to reference them.  Perhaps you have them locally but haven't committed them yet?
2. The proof of `add_checked` seems a bit duplicative with `add`, and similarly for `mul_checked` and `mul`.  What do you think of these versions?
```
    pub exec fn add_checked(&self, v2: &CheckedU32) -> (result: Self)
        ensures
            result@ == self@ + v2@,
    {
        proof {
            use_type_invariant(v2);
        }
        match v2.v {
            Some(n) => self.add(n),
            None => {
                let i: Ghost<nat> = Ghost((self@ + v2@) as nat);
                Self{ i, v: None }
            }
        }
    }

    pub exec fn mul_checked(&self, v2: &Self) -> (result: Self)
        ensures
            result@ == self@ as int * v2@ as int,
    {
        proof {
            use_type_invariant(self);
            use_type_invariant(v2);
        }
        let i: Ghost<nat> = Ghost((self@ * v2@) as nat);

        match v2.v {
            Some(n) => self.mul(n),
            None => {
                match self.v {
                    Some(n1) => {
                        if n1 == 0 {
                            assert(i@ == 0) by {
                                lemma_mul_by_zero_is_zero(v2@ as int);
                            }
                            Self{ i, v: Some(0) }
                        }
                        else {
                            assert(self@ * v2@ >= 1 * v2@ == v2@) by {
                                lemma_mul_inequality(1, self@ as int, v2@ as int);
                            }
                            Self{ i, v: None }
                        }
                    },
                    None => {
                        assert(self@ * v2@ > u32::MAX) by {
                            lemma_mul_inequality(1, self@ as int, v2@ as int);
                        }
                        Self{ i, v: None }
                    },
                }
            }
        }
    }
```

jaylorch:
I haven't yet added anything to the guide, but I put anchors there in anticipation of eventually adding something to the guide. Should I wait to put them there until the commit that updates the guide?

parno:
> I haven't yet added anything to the guide, but I put anchors there in anticipation of eventually adding something to the guide. Should I wait to put them there until the commit that updates the guide?

No I don’t think this PR has to wait for the guide page. I do think mentioning it in the guide will improve discoverability.

PR #1462, opened 2025-02-20
draft/rfc: invariant mask checking using std::Set

zeldovich:
This is a sketch of a PR to replace invariant namespace checking with an `std::Set`-based implementation, instead of the current AIR-level custom set implementation.  This change allows for a subsequent PR that would allow specifying the mask for a given function as an `std::Set` expression, rather than a list of a fixed number of set elements.

I wanted to get feedback from @tjhance and others about whether this is a sound plan in general, and what I might be missing, since I'm not so familiar with the semantics of various internal Verus layers.

This draft implementation preserves the performance optimization of eliding appropriate checks when the sets are known to be full or empty, but does not preserve the detailed error messages that the previous implementation generated (in terms of pointing to the span where a conflicting invariant might have already been opened).  That would certainly be possible to preserve, if others think it's valuable.


tjhance:
Thanks for taking a crack at this, it's a much-needed feature.

However, I have some questions about the implementation, as it seems there were a lot more changes than I'd expect.

1. Why was the mask logic moved from sst_to_air to ast_to_sst? I don't think this is a big deal either way, but I also wouldn't expect it to be necessary for this change.
2. Why did you remove the special-case diagnostics? I would think you could just add a new kind of 'SetBase' and augment the existing logic.

zeldovich:
> 1. Why was the mask logic moved from sst_to_air to ast_to_sst? I don't think this is a big deal either way, but I also wouldn't expect it to be necessary for this change.

Indeed, I first tried implementing it by extending the existing `sst_to_air`-level implementation and AIR-level `inv_masks.rs`, but at the AIR level, I couldn't easily synthesize Rust calls to `std::Set` methods like `contains()` and `subset_of()`.  At the `ast_to_sst` level, I found it much easier to generate a `CallFun::Fun()` to call set operations as needed.

> 2. Why did you remove the special-case diagnostics? I would think you could just add a new kind of 'SetBase' and augment the existing logic.

No real reason, other than I wanted to get a simple draft implementation working first.  Putting back the special-case diagnostics indeed makes sense, by preserving the old code in the way you suggest, and I'm happy to do that once the plan makes sense.  (E.g., perhaps am I confused about doing this at the `ast_to_sst` level altogether..)

tjhance:
That's a good point about synthezing calls. Making function calls at the sst->air level can be a pain, especially because of Poly, so that's a good reason to move it to sst_to_air. I'll need to do a close review of it later, but there shouldn't be any conceptual issues with making that move.

zeldovich:
Thanks for confirming.  I'll clean this up, then, and pull in the detailed diagnostics from the current `inv_masks.rs`.

zeldovich:
This PR now generates pretty much identical diagnostic errors as the previous `inv_masks.rs` code, so I think it's ready for a more detailed review by @tjhance (or anyone else).

My plan was to open a separate PR, after this one, to actually add support for specifying an explicit `std::Set` expression in an `opens_invariant` clause, but if you'd prefer the two features to be bundled, I can put that change here.

zeldovich:
One thing that's not working correctly is computing the expression for the callee's mask set at a function call site.  In particular, the expression for the callee's mask needs to substitute any references to argument names in the function declaration's `opens_invariants` clause with the actual passed arguments from the call-site.  My current stub implementation of this, `mask_set_for_call()` in `ast_to_sst.rs`, should also take the `small_args` vec (from `stm_call()`) as an argument, but I can't quite figure out if there's already machinery that will do this argument substitution, or if I need to write another implementation.  Surely this is something that already happens (e.g., for checking requires/ensures clauses at call sites), but it seems like this happens in the translation pipeline down to AIR, and I couldn't find a reusable function that would work at the `sst::Exp` level..  Any hints about what's the right way to do this?

tjhance:
That's a good question. Right now the relevant code is in sst_to_air:

```
pub fn mask_set_from_spec(spec: &MaskSpec, function_name: &Fun, args: &Vec<Expr>) -> MaskSet {
    match spec {
        MaskSpec::InvariantOpens(exprs) => { 
            let mut l = vec![];
            for (i, e) in exprs.iter().enumerate() {
                let expr =
                    ident_apply(&prefix_open_inv(&fun_to_air_ident(function_name), i), &args);
                l.push(MaskSingleton { expr, span: e.span.clone() });
            }    
            MaskSet::from_list(l)
        }    
        MaskSpec::InvariantOpensExcept(exprs) if exprs.len() == 0 => MaskSet::full(),
        MaskSpec::InvariantOpensExcept(_exprs) => { 
            panic!("custom mask specs are not yet implemented");
        }    
    }    
}
```

Right now, we don't have a good way to represent this special function call in SST. You might need to add one.

You could add a new variant to this enum:

```
#[derive(Debug, Clone, Hash)]
pub enum CallFun {
    // static/method Fun, plus an optional resolved Fun for methods
    Fun(Fun, Option<(Fun, Typs)>),
    Recursive(Fun),
    InternalFun(InternalFun),
}
```

zeldovich:
Alright, this PR now passes all of the `rust_verify_tests`; I ended up implementing callee argument (and type parameter) substitution using `subst_exp()`, which seems to work well.  Let me know if this looks sensible.

If you're curious, the follow-on change to actually add syntax to specify a `std::Set` for an `opens_invariants` clause is here:

https://github.com/zeldovich/verus/compare/maskset...zeldovich:verus:maskset-sets

zeldovich:
Thanks for the clarification; I think I finally understand what you were trying to suggest.  Does this new implementation look sensible?

source/vir/src/inv_masks.rs
```diff
@@ -1,181 +1,264 @@
...
+
+#[derive(Clone)]
+pub enum MaskSet {
+    Empty { span: Span },
+    Full { span: Span },
+    Insert { base: Box<MaskSet>, elem: Exp },
+    Remove { base: Box<MaskSet>, elem: Exp },
+    // Not used at the moment, will be used when we add syntax and support
+    // for specifying an std::Set expression as the mask.
+    // Arbitrary { set: Exp },
```
tjhance, line 24: I think can cut this comment?

source/vir/src/prune.rs
```diff
@@ -380,6 +383,13 @@ fn traverse_reachable(ctxt: &Ctxt, state: &mut State) {
             if ctxt.assert_by_compute && crate::interpreter::is_sequence_fn(&f).is_some() {
                 reach_seq_funs(ctxt, state);
             }
+            // set operations may be invoked for opens_invariants checks
+            reach_function(ctxt, state, &fn_set_contains_name(&ctxt.vstd_crate_name));
+            reach_function(ctxt, state, &fn_set_empty_name(&ctxt.vstd_crate_name));
+            reach_function(ctxt, state, &fn_set_full_name(&ctxt.vstd_crate_name));
+            reach_function(ctxt, state, &fn_set_insert_name(&ctxt.vstd_crate_name));
+            reach_function(ctxt, state, &fn_set_remove_name(&ctxt.vstd_crate_name));
+            reach_function(ctxt, state, &fn_set_subset_of_name(&ctxt.vstd_crate_name));
```
tjhance, line 392: can you move these into the OpenInvariant case?

zeldovich: Even if a function doesn't open invariants, it might need to do mask set checking (to ensure that its mask is larger than those of any functions that it calls).  I will put these into both the `OpenInvariant` and `Call` cases, though; that should suffice.

tjhance: oh, that's a good point. Call is so ubiquitous though that it'll hardly make a difference to move it.

PR #1465, opened 2025-02-21
Support operator

ziqiaozhou:

I want to support operator overloading via traits. To achieve this, the PR will replace operators (e.g., Add, Sub, Mul, Div, Rem, Not, Eq, Ne, Gt, etc.) with their corresponding trait method calls, when is_smt_equality or is_smt_arith do not applies. Additionally, I will provide external specifications for primitive types and introduce specification functions to facilitate proofs for functions that operate on generic types and use operators. Since this PR involves extensive trait specifications, it will require careful review.


ziqiaozhou:
PartialEq is broken, see more in https://github.com/verus-lang/verus/issues/1466. It need to be fixed before I merged my change or I should remove PartialEq support from this PR.

source/vstd/std_specs/cmp.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[verifier::external_trait_specification]
+pub trait ExPartialOrd<Rhs: ?Sized>: PartialEq<Rhs> {
+    type ExternalTraitSpecificationFor: core::cmp::PartialOrd<Rhs>;
+
+    fn partial_cmp(&self, other: &Rhs) -> (ret: Option<Ordering>)
+        ensures
+            ret === spec_partial_cmp(self, other),
+    ;
```
tjhance, line 95: these specifications are unsound when verified code interfaces with unverified code, because there's no way to enforce that an unverified implementation of PartialOrd meets these conditions.

The way we handle this for `Hash` is to have condition all the specifications on a predicate `obeys_key_model::<Key>()`. I was planning to do something similar here: guard all the specs with some predicate:

```rust
    fn partial_cmp(&self, other: &Rhs) -> (ret: Option<Ordering>)
        ensures obeys_comparison_model::<Self, Rhs>() ==>
            ret === spec_partial_cmp(self, other),
    ;
```

ziqiaozhou: > These specifications are unsound when verified code interfaces with unverified code, because there's no way to enforce that an unverified implementation of PartialOrd meets these conditions.

Yeah. I should add that guard. I think the current implementation is still not sound when replacing operators with trait methods. Especially for cmp operations, PartialEq vs Eq and PartialOrd vs Ord. I may need to check the trait bound to figure that out. Not sure whether it is possible to directly infer the operator overloading from TyCtxt.

BTW, I also have some changes related to macro in https://github.com/verus-lang/verus/commit/8201118b7adbaf97ff28df3128f47c7932319da5 to automatically generate those specifications for struct that annotated with Derive(PartialEq or PartialOrd). I would like to collect feedbacks from you to see whether it makes sense to do that in builtin_macro.

ziqiaozhou:
It seems that the PR is blocked by https://github.com/verus-lang/verus/issues/1466. Should I just remove the PartialEq and PartialCmp specification so that the PR could be reviewed without consideirng the blocked issue?

utaal:
@ziqiaozhou let me try and take a look at the `PartialEq` issue with @tjhance tomorrow. If it turns out to be a complicated fix, then it may be best to postpone that part of the change - if not, we'll try and get a fix merged ASAP so this can stay as is.

utaal:
I'd like to review this too before it lands: I started looking at it yesterday (and then focused on landing the fix that is a blocker for `PartialEq`) and it has interactions with some of the existing mechanisms for handling operator code that are not entirely clear to me.

ziqiaozhou:
> I'd like to review this too before it lands: I started looking at it yesterday (and then focused on landing the fix that is a blocker for `PartialEq`) and it has interactions with some of the existing mechanisms for handling operator code that are not entirely clear to me.

Thanks for fixing the PartialEq issue. I rebased the change and added some comments about how the change interact with the existing handling for operator code. The code only replace the operator with correponding trait method call when it could not be handled by existing mechanism. 

To be noted, the change does not cover AssignOp since customzed AddAssign/SubAssign/... may not have the consistent implementation  with Add/Sub/...

ziqiaozhou:
@tjhance, #1488 as for the derive(PartialEq, PartialOrd), would you plan to add the postcondition in verifier in the future? I was using builtin_macro (not included in this PR) to automatically add specification by parsing the derive attribute. If you plan to do that in verifier, will you also be able to define axiom for spec_eq/spec_lt/ spec_le/...? If so, I guess this PR is still good and I can keep using the spec functions for verificatoin.

ziqiaozhou:
Please let me know if there are any concerns about the PR? @utaal @tjhance @Chris-Hawblitzel

tjhance (changes requested):

source/vstd/std_specs/cmp.rs
Chris-Hawblitzel: > hese specifications are unsound when verified code interfaces with unverified code, because there's no way to enforce that an unverified implementation of PartialOrd meets these conditions.
> 
> The way we handle this for `Hash` is to have condition all the specifications on a predicate `obeys_key_model::<Key>()`. I was planning to do something similar here: guard all the specs with some predicate:
> 
> ```rust
>     fn partial_cmp(&self, other: &Rhs) -> (ret: Option<Ordering>)
>         ensures obeys_comparison_model::<Self, Rhs>() ==>
>             ret === spec_partial_cmp(self, other),
>     ;
> ```

If I understand correctly, @tjhance's objection is that unverified code might fail to obey a specification like `ret == spec_partial_cmp(self, other)`.  For example, unverified code might be nondeterministic and return a different answer when given the same `self` and `other` arguments, in which case `ret == spec_partial_cmp(self, other)` would lead to a contradiction no matter how `spec_partial_cmp` is defined.

If this is the objection, doesn't the same objection apply to the declaration of `eq` as well?

```
pub trait ExPartialEq<Rhs: ?Sized> {
    type ExternalTraitSpecificationFor: core::cmp::PartialEq<Rhs>;

    // Required method
    fn eq(&self, other: &Rhs) -> (ret: bool)
        ensures
            ret == spec_partial_eq(self, other),
    ;
}
```

That is, an unverified implementation of `eq` could be nondeterministic, and it wouldn't be safe to claim that `ret == spec_partial_eq(self, other)` without adding a guard like `obeys_comparison_model(...) ==>` or `obeys_equality_model(...) ==>`?  If so, should we add this guard too?

ziqiaozhou: Yes. I should have added obeys_equality_model to eq function. Thanks.

ziqiaozhou:
> I'm a bit bandwidth limited at the moment, so I don't have more detailed comments, and I don't want to hold this up, but it would be good to see a few more positive and negative tests given the extent of the change, especially with regards to the interaction of existing `is_smt_eq` and `is_smt_cmp` functionality. Please also address @tjhance's change requests.

I moved all tests related to operator traits into `operators.rs`. In addition to my old tests, I added some tests without importing vstd to demonstrate that the existing`is_smt_eq` and `is_smt_cmp` still works as before and will not be impacted by the specifications added to operator traits.  I did not touch `overflow.rs` and it should also demonstrate that the existing`is_smt_eq` and `is_smt_cmp` works as before.

utaal:
* [ ] I would like to write a test to check that this does not interact unsoundly with `Structural`.

Chris-Hawblitzel:
> As part of reviewing this pull request, I went back and looked at some of the related code we have for Clone and Hash. Between this PR and those, it looks like we already have a variety of styles for assigning specifications to std library traits and datatypes. I think it would be good to standardize on a single style. Rather than summarize the various styles here, I started a separate discussion in #1527 .

After some discussion of https://github.com/verus-lang/verus/discussions/1527 , we decided to pursue the style based on `call_ensures`.  I'm working on porting the work from this pull request to this style now.

ziqiaozhou:
> > As part of reviewing this pull request, I went back and looked at some of the related code we have for Clone and Hash. Between this PR and those, it looks like we already have a variety of styles for assigning specifications to std library traits and datatypes. I think it would be good to standardize on a single style. Rather than summarize the various styles here, I started a separate discussion in #1527 .
> 
> After some discussion of #1527 , we decided to pursue the style based on `call_ensures`. I'm working on porting the work from this pull request to this style now.

I did not update the cmp traits, but I removed all ensure specs for operators so that it uses style based on `call_ensures`. Unfortunately, the pre-conditions is only allowed in trait definition, and so is not flexible and the users cannot define precondition in a concrete implementation and still need to define `spec_add_requires` to add pre-conditions.

ziqiaozhou:
Rebased on #1569.

utaal:
@ziqiaozhou thank you for reworking the design. I have no bandwidth right now, but can you write a test that checks what happens if a type uses both this mechanism and `Structural` to get exec smt equality?

ziqiaozhou:
> @ziqiaozhou thank you for reworking the design. I have no bandwidth right now, but can you write a test that checks what happens if a type uses both this mechanism and `Structural` to get exec smt equality?

Added the test operators::test_structural_eq, where assume_specification for PartialEq does not override the result of (s1 == s2) since it is marked as Structural.

utaal (approved):

tjhance (approved):

Chris-Hawblitzel:
@ziqiaozhou and I discussed how to declare `spec` functions to be used in `requires` clauses for external traits.  Separately, some of us have discussed being able to extend existing `std` traits with additional spec functions (doing so in a way that Verus can catch any circular definitions of the `spec` functions).  I would like to try to implement this feature, partly to help this pull request.  I can start a design discussion about this.

utaal:
[retreat 2025] Superseded by https://github.com/verus-lang/verus/pull/1629

Chris-Hawblitzel:
@ziqiaozhou @tjhance I updated this to use trait extensions, and I think this pull request is ready to merge.  Let me know if there are additional changes you want.

ziqiaozhou:
Thanks for reopening the change and update it to be ready to merge. It looks good to me.

> @ziqiaozhou @tjhance I updated this to use trait extensions, and I think this pull request is ready to merge. Let me know if there are additional changes you want.

PR #1470, opened 2025-02-24
[proposal] deprecate `is` and `has` as identifiers to prepare to turn them into keywords

utaal:


utaal:
We decided against this strategy.

PR #1475, opened 2025-02-25
Leverage Cargo (continued)

Chris-Hawblitzel:
This is a continuation of https://github.com/verus-lang/verus/pull/1138 ; see that pull request for motivation, details, and discussion.  The purpose of making this new pull request is to continue work within a new branch inside the Verus repository, with the goal of merging the work done in https://github.com/verus-lang/verus/pull/1138 while also addressing the concerns raised in the discussion.


jaybosamiya-ms:
I wrote a short note on Slack on what I think ideal `cargo verus` CLI experience would be, copying here:

```console
$ cargo build
...does the standard build, no verification...

$ cargo run
...does the standard build+run, no verification...

$ cargo verus
Verified Rust for low-level systems code

Usage: cargo verus [OPTIONS] <COMMAND>

Commands:
  verify    Verify the current crate
  build     Verify and build the current crate
  run       Verify and run a binary or example of the local crate
  doc       Build this crate's documentation with verusdoc
  fmt       Format all bin and lib files of the current crate with verusfmt
  self      Manage the Verus installation

See 'cargo verus help <COMMAND>' for more information on a specific command

$ cargo verus self --help
Manage the Verus installation

Usage: cargo verus self [OPTIONS] <COMMAND>

Commands:
  update   Download and install updates to Verus
```

Essentially `cargo verus` should be the namespace where the actual commands sit, thus it becomes quite obvious what is actually happening. Things that have an unverified equivalent get their verified version simply by adding the `verus` to their `cargo ...` command line (I know we won't get everything in the above list on first go, but having it as a namespace gives us the flexibility to have an interface consistent with how many `cargo` based tools are written (e.g., https://nexte.st/), including `cargo` itself).

source/rust_verify/src/config.rs
```diff
@@ -465,6 +474,23 @@ pub fn parse_args_with_imports(
...
+        "",
+        OPT_COMPILE_WHEN_NOT_PRIMARY,
+        "When running under 'cargo verus', compile non-primary package(s)",
+    );
+    opts.optmulti(
+        "",
+        OPT_IMPORT_DEP_IF_PRESENT,
+        "When running under 'cargo verus', import .vir file for dependency",
+        "DEPENDENCY_NAME",
+    );
```
utaal, line 492: I suspect these show up in `help`, right? Should they?

utaal: How about something like this? https://github.com/verus-lang/verus/compare/cargo-integration...cargo-integration-internal-cargo-args (ignore the first of the two commits, I needed to cherry-pick it from `main` so that I could test things with my more up-to-date rustup)

utaal: It would also be nice to move some of these strings to `tools/common/consts.rs` or similar, but we can take care of that later.

utaal (approved):

PR #1476, opened 2025-02-25
A solution to avoid using .set() function by replacing IndexMut with index_set

ziqiaozhou:

I would like to propose a change to support IndexMut before the mutable reference is created. Instead of waiting for the big feature change, we can do a simple change to replace the assign operation for IndexMut with a function index_set(container: &mut T, idx: Idx, val: V).

It could be a general solution for any type with IndexMut so that we can avoid using .set for Vec, Array, slice, hashmap. 
The current PR has included trusted IndexSetTrustedSpec for Array, slice, Vec. It should also work with HashMap and other containers.

Since we never actually verifying the implementation of IndexMut, IndexSetTrustedSpec trait is a sealed trait protected by TrustedSpecSealed, so only vstd lib can implement specification for index_set. For container types not covered by vstd lib, the use of index_set would always fails.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -2461,6 +2461,8 @@ fn expr_assign_to_vir_innermost<'tcx>(
         Some(op) => Some(binopkind_to_binaryop(bctx, op, tc, lhs, rhs, mode_for_ghostness)?),
         None => None,
     };
+
+    // NOTE: A temparary solution for index_mut until mutable reference support lands.
     match &lhs.kind {
         ExprKind::Index(tgt_expr, idx_expr, _span) => {
             let tgt_modifier = is_expr_typ_mut_ref(
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -2489,7 +2489,7 @@ fn expr_assign_to_vir_innermost<'tcx>(
             AutospecUsage::Final,
         );
         if let Some(op) = op {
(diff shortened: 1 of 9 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2461,6 +2461,53 @@ fn expr_assign_to_vir_innermost<'tcx>(
...
+                bctx.spanned_typed_new(tgt_expr.span, &tgt_vir.typ.clone(), ExprX::Loc(tgt_vir));
+            let call_target = CallTarget::Fun(
+                vir::ast::CallTargetKind::Static,
+                fun,
+                typ_args.unwrap(),
+                Arc::new(vec![]),
+                AutospecUsage::Final,
+            );
+
+            if let Some(op) = op {
```
Chris-Hawblitzel, line 2491: This duplicates `lhs`, which would semantically correspond to evaluating `lhs` twice.  In Rust, `lhs` is only evaluated once.  I would recommend either not handling `Some(op)` for now, or if you really need it, limiting it to the case where `tgt` and `idx` are variables or constants, which should be semantically equivalent for array/slice/Vec.

Chris-Hawblitzel (changes requested):

source/rust_verify/src/rust_to_vir_expr.rs
ziqiaozhou: Thanks! To avoid the side effects, I checked can_have_side_effects for tgt and idx that would ensure we cannot use tgt/idx twice if they have side effects. In addition, I also checked exec mode and is_smt_arith for lhs and rhs pair.

Chris-Hawblitzel (approved):

PR #1486, opened 2025-02-28
Specification for `std::collections::VecDeque`

jaylorch:

### DIFF
--- a/source/vstd/std_specs/vecdeque.rs
+++ b/source/vstd/std_specs/vecdeque.rs
@@ -341,18 +341,25 @@ impl<'a, T> View for IterGhostIterator<'a, T> {
     }
 }
 
+pub open spec fn spec_iter<'a, T, A: Allocator>(v: &'a VecDeque<T, A>) -> (r: Iter<'a, T>);
+
+pub broadcast proof fn axiom_spec_iter<'a, T, A: Allocator>(v: &'a VecDeque<T, A>)
+    ensures
+        (#[trigger] spec_iter(v))@ == (0int, v@),
+{
+    admit();
+}
+
+#[verifier::when_used_as_spec(spec_iter)]
 pub assume_specification<'a, T, A: Allocator>[ VecDeque::<T, A>::iter ](v: &'a VecDeque<T, A>) -> (r: Iter<'a, T>)
     ensures
-        ({
-            let (index, s) = r@;
-            &&& index == 0
-            &&& s == v@
-        }),
+        r@ == (0int, v@),
 ;
 
 pub broadcast group group_vec_dequeue_axioms {
     axiom_spec_len,
     axiom_vec_dequeue_index_decreases,
+    axiom_spec_iter,
 }
 
 } // verus!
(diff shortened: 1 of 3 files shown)


source/vstd/std_specs/vecdeque.rs
```diff
@@ -0,0 +1,374 @@
...
+) -> (element: Option<T>)
+    ensures
+        match element {
+            Some(x) => {
+                &&& i < old(v)@.len()
+                &&& x == old(v)@[i as int]
+                &&& v@ == old(v)@.remove(i as int)
+            },
+            None => {
+                &&& old(v)@.len() < i
```
zeldovich, line 174: Should this be `<= i`?

jaylorch: Thanks for catching that! I'll fix it.

source/vstd/std_specs/vecdeque.rs
```diff
@@ -0,0 +1,374 @@
...
+/// `std::collections::VecDeque`.
+use super::super::prelude::*;
+
+use core::alloc::Allocator;
+use core::clone::Clone;
+use core::option::Option;
+use core::option::Option::None;
+use std::collections::vec_deque::Iter;
+use std::collections::VecDeque;
+use std::ops::Index;
```
jaybosamiya-ms, line 11: I believe this set of changes would allow it to work on both std and no-std `alloc` contexts (thereby allowing the `#[cfg(all(feature = "alloc", feature = "std"))]` to be de-restricted to just `#[cfg(feature = "alloc")]`)
```suggestion
use alloc::collections::vec_deque::Iter;
use alloc::collections::vec_deque::VecDeque;
use core::ops::Index;
```

jaylorch: Thanks! I'll try that. Presumably this approach could also be applied to `HashMap` and `HashSet`?

jaylorch: Oh, it looks like `HashMap` and `HashSet` aren't in `alloc`.

jaybosamiya-ms: Yeah, them not being in `alloc` and being restricted to `std` is a long-standing open issue: https://github.com/rust-lang/rust/issues/27242

source/vstd/std_specs/vecdeque.rs
```diff
@@ -0,0 +1,374 @@
...
+#[verifier::external_body]
+#[verifier::accept_recursive_types(T)]
+#[verifier::reject_recursive_types(A)]
+pub struct ExVecDeque<T, A: Allocator>(VecDeque<T, A>);
+
+// this is a bit of a hack; verus treats Global specially already,
+// but putting this here helps Verus pick up all the trait impls for Global
+#[verifier::external_type_specification]
+#[verifier::external_body]
+pub struct ExGlobal(alloc::alloc::Global);
```
Chris-Hawblitzel, line 25: I just moved `pub struct ExGlobal(alloc::alloc::Global);` to a different file, so it's no longer necessary to include it in `vec.rs` (or `vecdeque.rs`).

PR #1488, opened 2025-02-28
Start handling derived trait impls

tjhance:
Presently, derived trait impls are treated just like any other trait impl. This results in some confusing behavior, since they might or might not be included in verification depending on the external-by-default flag. (Note that derived trait impls do not automatically get the same attributes as the datatype that declares them).

This PR makes the behavior more consistent by getting the external/verus_macro attributes from the datatype instead.

Actually, there are a few different things we can do, depending on the trait. These are the choices I made in this PR:

 * Many traits are _always_ ignored (Hash, Default, Debug, Ord, PartialOrd) since this is the existing behavior (assuming external-by-default is set)
 * Anything not recognized is configured to be verified (as long as the corresponding datatype is declared inside the verus macro).
 * For 'clone' we automatically add a postcondition (this is only partially supported, and it prints a warning when it isn't supported)
 * I was originally planning to support PartialEq the same way, but as I dug into it I realized there's a lot more to do there, so it's kept as Ignore for now


source/rust_verify/src/external.rs
```diff
@@ -28,16 +28,20 @@ Here are some odd cases to watch out for:
  * For trait decls and trait impls, we disallow #[verifier::external] on individual
    TraitItems or ImplItems.
 
+ * Autoderive traits need to be handled specially.
```
utaal, line 31:
```suggestion
 * Autoderive trait impls need to be handled specially.
```

source/rust_verify/src/automatic_derive.rs
```diff
@@ -0,0 +1,164 @@
...
+        functionx.ensure = Arc::new(vec![eq_expr]);
+    } else {
+        warn_unsupported();
+    }
+
+    Ok(())
+}
+
+// TODO better place for this
+fn cleanup_span_ids<'tcx>(ctxt: &Context<'tcx>, span: Span, hir_id: HirId, expr: &Expr) -> Expr {
```
utaal, line 156: Can you please add a comment describing what this does?

utaal (approved):
Can you please add a couple tests (positive/negative) for the auto-derived `Clone` post-condition?

Chris-Hawblitzel (approved):

PR #1489, opened 2025-02-28
Proof of concept for erasure-by-THIR modification.

tjhance:
This is a proof-of-concept for doing lifetime-checking by forking the THIR-lowering code. We discussed a few possible use-cases for this, so I wanted to go ahead and get the annoying work of making the fork out of the way. There's not much more to it than that, though.

This PR:

 * Embeds a forked copy of the rustc_mir_build crate
 * Overrides the 'thir_body' provider with the forked version
 * Changes the after_expansion code to Continue to lifetime-checking, rather than Stop
 * Modifies the THIR-lowering code to erase 'assert' expressions.

This behavior is behind a flag; we could merge this in with no problems if anybody wants to build on it.


utaal:
Very interesting!

(Just a note that if we end up doing this, we should probably import it as a submodule, the same way we do for `syn` and `prettyplease`.)

tjhance:
superceded by https://github.com/verus-lang/verus/pull/1792

PR #1491, opened 2025-02-28
[attr_rewrite] Replace panic with vpanic

ziqiaozhou:

panic macro is a builtin macro in rust and could not be easily supported due to the use of panic_fmt takes Argument and Argument is created via Arguments::new_v1 with a private struct rt::Argument.

Thus, directly replacing panic macro is the easiest solution.

The change is only applied to verus_spec.

### DIFF
--- a/source/vstd/pervasive.rs
+++ b/source/vstd/pervasive.rs
@@ -414,29 +414,48 @@ pub open spec fn cloned<T: Clone>(a: T, b: T) -> bool {
 } // verus!
 
 verus! {
-
-#[verifier(external_body)]
-pub fn call_vpanic(_out: &[&str]) -> ! {
-    unimplemented!()
+/// The default behavior of the vstd library enforces writing panic-free code.
+/// While developers may still use panic, verification should ensure that any
+/// panic is provably unreachable.
+/// cfg!(feature = "allow_panic") explicily allows code to panic.
+pub open spec fn allow_panic() -> bool {
+    cfg!(feature = "allow_panic")
 }
 
+#[doc(hidden)]
 #[verifier(external_body)]
-pub fn vpanic_fmt<T: core::fmt::Debug>(_v: &T) -> &str {
-    unimplemented!()
+pub fn __call_panic(out: &[&str]) -> !
+requires
+    allow_panic()
+{
+    core::panic!("__call_panic {:?}", out);
+}
+
+// rt::Argument is a private type and we cannot add specification directly
+#[cfg(feature = "alloc")]
+#[doc(hidden)]
+#[verifier(external_body)]
+pub fn __new_argument<T: core::fmt::Debug>(v: &T) -> alloc::string::String {
+    alloc::format!("{:?}", v)
 }
 
 } // verus!
+
+/// Replace panic macro with vpanic when needed.
+/// panic!{} may call panic_fmt with private rt::Argument, which could not
+/// be supported in verus.
 #[macro_export]
 macro_rules! vpanic {
     // Case: Format string with arguments
     ($fmt:expr $(,$val:expr)*) => {
-        vstd::pervasive::call_vpanic(
-            &[$(
-                vstd::pervasive::vpanic_fmt(&$val),
+        vstd::pervasive::__call_panic(
+            &[vstd::pervasive::__new_argument(&$fmt).as_str(),
+            $(
+                vstd::pervasive::__new_argument(&$val).as_str(),
             )*]
         );
     };
     () => {
-        vstd::pervasive::call_vpanic([]);
+        vstd::pervasive::__call_panic(&[]);
     };
 }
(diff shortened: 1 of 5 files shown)


ziqiaozhou:
It seems that adding this hack is still useful. 

We should have two other options to use panic-like macros (panic, println, debug, ...). 
1. [Not feasible] provides specification for all external functions after expanding panic. For panic, it needs panic_fmt +const_format_args!{}. const_format_args  expands to new_v1. However, new_v1 takes an argument of a private type core::fmt::rt::Argument. There is no way to add specification for a function with private arguments.
```
verus! {
    #[verifier(external_type_specification)]
    #[verifier(external_body)]
    pub struct ExArguments<'a>(pub core::fmt::Arguments<'a>);
    pub const fn ex_panic_fmt(fmt: core::fmt::Arguments<'_>) -> !
    {
        core::panicking::panic_fmt(fmt)
    }

    pub assume_specification<const P: usize, const A: usize, 'a>[core::fmt::Arguments::new_v1](
        pieces: &'a [&'static str; P],
        args: &'a [core::fmt::rt::Argument<'a>; A],
    ) -> core::fmt::Arguments<'a>;
}
```

2. [Feasible but ad-hoc] Define a local `panic` in the file that calls `panic`. We need to add a new panic definition for all files using panic.
Rust only allows us to define panic and use it locally. If we define panic in another crate/module/file (e.g., include!{ another_file.rs }) would not work.
```
error[E0659]: `panic` is ambiguous
   --> ...
    |
2157 |                 panic!("Freeing memory on unsupported page type");
    |                 ^^^^^ ambiguous name
    |
    = note: ambiguous because of a conflict between a macro-expanded name and a less macro-expanded name from outer scope during import or macro resolution
    = note: `panic` could refer to a macro from prelude
note: `panic` could also refer to the macro defined here
   --> ...
    |
38   | / macro_rules! panic {
39   | |     ($($arg:tt)*) => {
40   | |         verify_external::vpanic!($($arg)*);
41   | |     };
42   | | }
    | |_^
```

source/vstd/pervasive.rs
```diff
@@ -412,3 +412,31 @@ pub open spec fn cloned<T: Clone>(a: T, b: T) -> bool {
...
+
+#[verifier(external_body)]
+pub fn call_vpanic(_out: &[&str]) -> ! {
+    unimplemented!()
+}
+
+#[verifier(external_body)]
+pub fn vpanic_fmt<T: core::fmt::Debug>(_v: &T) -> &str {
+    unimplemented!()
+}
```
Chris-Hawblitzel, line 426: Adding panic functions to `vstd` will affect https://github.com/verus-lang/verus/discussions/1292 , because calling `call_vpanic` could be considered a form of "cheating" if we're trying to prove that the code never panics, which we are by default (cc @utaal ).  For now, should these two functions be guarded by `#[cfg(feature = "vpanic")]`?

ziqiaozhou: How could we prove our code never panics in current verus? If we want to ensure never panic, I can add a requires guard to the call_vpanic. If my code uses panic but it is unreachable, it should still be a correct implementation that never panics.

fn call_vpanic() -> ! 
requires
   !spec_never_panic()
{}

parno: > How could we prove our code never panics in current verus?

Currently Verus's design and library wrappers all add preconditions to operations that might panic (e.g., dividing by zero, or indexing into an array).  If we wanted to add support for code that contains `panic!` (ignoring the issues you raise with trying to do that at present), we could add `requires false` as a precondition, which would oblige Verus to prove that the call is always unreachable.

tjhance: Verus already allows you to specify if a function can panic or not.

https://verus-lang.github.io/verus/guide/reference-unwind-sig.html

ziqiaozhou: @tjhance, that is what I want to ask about. My prior understanding is that developers need to specify `no_unwind when true` to prove there is no panic. My call_vpanic does not specify no_unwind and so it should be `no_unwind when false`. Then it should not do "cheating" if we want to prove a function with panic!() call never panic. But I thought there might be a flag or some mode that verus can use to set no_unwind when true as default. Or I should just explicitly add `no_unwind when false`

ziqiaozhou:
Added a allow-panic feature. vstd lib does not allow code to panic. Although not all vstd specifications include no_unwind in their signature, they should never panic with correct verification. 
vstd default feature does not allow code to panic. Ideally, Verus wants developers to write code in a safe way so that panic is never reachable.  In practice, "allow-panic" feature allows developer to write code that can panic if unavoidable.
This PR does not make a decision on how to prove no panic. We just keep the original design choice for vstd with default feature.

Chris-Hawblitzel (approved):

PR #1492, opened 2025-03-01
Attr add tracked args

ziqiaozhou:

attribute-base verus_spec and its related macros are used to ensure that a non-verus developer can still collaborate in the same project with verus developer. To use tracked/ghost variables in executable function, the attribute-based macros should use a new style similar to the proposal in in https://github.com/verus-lang/verus/discussions/1301

It might need a lot of changes to finally reach the proposed style. This PR specifically focuses on the attribute-based macros, with the intention of maintaining consistency with the proposed style.

### DIFF
--- a/dependencies/syn/src/gen/clone.rs
+++ b/dependencies/syn/src/gen/clone.rs
@@ -2374,6 +2374,18 @@ impl Clone for crate::SignatureInvariants {
         }
     }
 }
+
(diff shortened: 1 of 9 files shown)


tjhance (approved):
It would be easy to replace this with 'proof-mode syntax' if we want to:

```
#[verus_spec(ret =>
    with
        tracked y: &mut u32, ghost w: u64 -> ghost z: u32
)]
```

But I also think we can go ahead and merge this and iterate if you prefer.

PR #1501, opened 2025-03-10
Add a 'safe API' checker

tjhance:
Adds a flag (`-V check-safe-api`) to check if an API is "safe", in the sense that is is always safe to be used from an UNVERIFIED, safe-Rust client crate.

Specifically, this flag tells Verus to check that:

 * all public, non-unsafe functions have trivial `requires` clauses, i.e, they are always safe to call.
 * Dually, it checks that all public, non-unsafe traits have trivial `ensures`, `invariant`, and `unwind` clauses, i.e., that they are always safe to _implement_.

### DIFF
--- a/source/rust_verify_test/tests/safe_api.rs
+++ b/source/rust_verify_test/tests/safe_api.rs
@@ -1,0 +1,231 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file_with_options! {
+    #[test] normal_fn_requires_fail ["-V check-safe-api"] => verus_code! {
+        pub fn test(x: u8)
+            requires x > 0,
+        {
+        }
+    } => Err(err) => assert_vir_error_msg(err, "Safe API violation: 'requires' clause is nontrivial")
+}
+
+test_verify_one_file_with_options! {
+    #[test] normal_fn_private_is_ok ["-V check-safe-api"] => verus_code! {
+        fn test(x: u8)
+            requires x > 0,
+        {
+        }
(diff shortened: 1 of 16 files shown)


jaybosamiya-ms:
Minor bikeshed for possible alternative names: 
- `check-api-safety`
- `check-api-safety-discipline`
- `check-api-for-unverified-safe-rust`

Personally, I prefer the longest one above, but also not strong enough to push for it too hard; instead, just suggesting some alternates if you think any of them might work better.

parno:
I tend to agree that `check-api-safety` is a bit clearer than `check-safe-api`.  The `check-api-for-unverified-safe-rust` option seems long to me; it seems better to explain in the helper text.

source/rust_verify_test/tests/safe_api.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        impl Foo for X {
+        }
+    } => Err(err) => assert_vir_error_msg(err, "Safe API violation: 'requires' clause is nontrivial")
+}
+
+// There are a few sneaky ways to smuggle private functions to the client
+// by using higher-order functions.
+// Right now, Verus doesn't support the relevant features; these tests
+// will catch these cases in the future.
```
Chris-Hawblitzel, line 193: Should we add tests for `dyn Fn` and `Any`, too?

I'm assuming type_invariant is API-safe, because of the requirement that "a struct with a type invariant cannot have any fields public to the crate"; this might also make a good test case.

Chris-Hawblitzel (approved):

PR #1502, opened 2025-03-10
Allow specifying a set in the `opens_invariants` clause

zeldovich:
This PR adds support and syntax for specifying an explicit `vstd::set::Set<int>` in an `opens_invariants` clause; as an example, this is now valid syntax and the proof checks out:

```
verus! {
    proof fn do_open(x: int)
        opens_invariants [x]
    {}

    proof fn do_call(s: Set<int>, x: int)
        opens_invariants s
    {
        if s.contains(x) {
            do_open(x);
        }
    }
}
```

### DIFF
--- a/dependencies/syn/src/gen/clone.rs
+++ b/dependencies/syn/src/gen/clone.rs
@@ -1414,6 +1414,7 @@ impl Clone for crate::InvariantNameSet {
             crate::InvariantNameSet::List(v0) => {
                 crate::InvariantNameSet::List(v0.clone())
             }
+            crate::InvariantNameSet::Set(v0) => crate::InvariantNameSet::Set(v0.clone()),
         }
     }
 }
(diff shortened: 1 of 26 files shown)


zeldovich:
@tjhance, just wanted to ping you in case this got lost in the noise -- any thoughts about the notation or other parts of this change?

tjhance:
the notation is a bit weird, I expect `{ ... }` to denote a set, so it's weird that thing _inside_ the braces is itself a set. (If it helps, we can reconsider the bracket-list notation I originally added.) Other than that, the diff looks good.

zeldovich:
@tjhance are you happy with this code?  Would be convenient to merge this so that @zero-to-nat can use it in some logical atomicity examples she's working through.

tjhance (approved):

PR #1506, opened 2025-03-11
Add verus_extra_stmts

ziqiaozhou:

When using verus_spec attribute for executable function, the executable function may need to declare ghost/tracked variable. proof!{} macro does not work since it treats its content as a block. Thus, I added verus_extra_stmts (similar to verus_exec_expr) macro. It indicate that the extra stmts added inside the macro is only for verification and will be directly erased if in non-verification mode.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -4008,29 +4008,16 @@ struct Stmts(Vec<Stmt>);
 
 impl Parse for Stmts {
     fn parse(input: ParseStream) -> syn_verus::Result<Self> {
-        let mut stmts = Vec::new();
-        while !input.is_empty() {
-            stmts.push(input.parse()?);
-        }
-        Ok(Stmts(stmts))
+        Block::parse_within(input).map(|stmts| Stmts(stmts))
     }
 }
 
-impl ToTokens for Stmts {
-    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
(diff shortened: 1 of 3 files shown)


source/builtin_macros/src/syntax.rs
```diff
@@ -3974,6 +3974,61 @@ pub(crate) fn rewrite_expr(
     proc_macro::TokenStream::from(new_stream)
 }
 
+struct Stmts(Vec<Stmt>);
+
+impl Parse for Stmts {
+    fn parse(input: ParseStream) -> syn_verus::Result<Self> {
+        let mut stmts = Vec::new();
```
Chris-Hawblitzel, line 3981: You may be able to use https://docs.rs/syn/latest/syn/struct.Block.html#method.parse_within here instead, which would parse a little more accurately.

Chris-Hawblitzel:
Instead of `verus_extra_stmts`, how about the name `proof_stmts`, to correspond to the existing `proof` blocks?

Chris-Hawblitzel:
Do you have a test case to show what the syntax looks like?

ziqiaozhou:
> Instead of `verus_extra_stmts`, how about the name `proof_stmts`, to correspond to the existing `proof` blocks?

I feel proof_stmts! vs proof! are confusing. After writing down some test examples, I feel proof_decl! is better?

Chris-Hawblitzel (approved):

PR #1507, opened 2025-03-12
Add command to quickly set up a new Verus project

jaybosamiya-ms:
This PR expresses my excitement for `cargo verus` by adding `cargo verus new` (similar to `cargo new`).

### DIFF
--- a/source/cargo-verus/src/main.rs
+++ b/source/cargo-verus/src/main.rs
@@ -47,6 +47,84 @@ pub fn main() -> Result<ExitCode> {
     process(&args)
 }
 
+fn new_project(args: &[String]) -> Result<()> {
+    let mut args_iter = args.iter();
+    assert_eq!(args_iter.next(), Some(&"new".into()));
+    let template_type =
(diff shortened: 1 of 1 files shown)


parno:
Cool -- this will be super handy!

jaybosamiya-ms:
Based on a quick offline discussion with Chris, this needs some minor changes (reminder to myself: git->path). I'll mark this as draft until I get spare cycles to change that up (hopefully soon).

jaybosamiya-ms:
@Chris-Hawblitzel I had a few spare minutes to noodle around with this to get it to point at paths rather than git: unfortunately, `target-verus` does not copy over some of the dependencies (such as `prettyplease_verus`), in which case we would need to depend on the git structure (which prevents working with nightly prebuilts releases).

Thus, instead of paths, I pushed a version that picks the exact sha revision at the point of when the `cargo verus new` is run, just as a temporary measure until we include the other files in too.

Chris-Hawblitzel (approved):

PR #1510, opened 2025-03-13
Improve lemmas about unique max and min for finite, non-empty, totally ordered sets

lduranovic:
The previous version of these functions/lemmas used `is_minimal` (`is_maximal`) as part of the postconditions, which made it hard for me to assert what I really wanted to assert, which is that the result of calling, for example, `find_unique_minimal` on some set is the smallest element in that set.

The current postconditions seem a lot more natural to me, and I think that through using these, we can get the previous version, which uses `is_minimal`/`is_maximal`, for free. That being said, I have not proved this (I can try to do this if there is interest).

I also made small changes to the specification functions for finding a unique max/min. The previous versions used the "free" version of the `choose` function, while mine use `s.choose()` (so the associated `choose` function for sets). I think this is better, and it seems to make the proofs easier.

### DIFF
--- a/source/vstd/set_lib.rs
+++ b/source/vstd/set_lib.rs
@@ -75,15 +75,13 @@ impl<A> Set<A> {
         recommends
             total_ordering(r),
             self.len() > 0,
-            self.finite(),
         decreases self.len(),
         when self.finite()
-        via Self::prove_decrease_min_unique
     {
         if self.len() <= 1 {
             self.choose()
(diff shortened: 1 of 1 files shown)


lduranovic:
Actually, never mind. I just misread the postconditions of the original lemmas. `is_maximal` and `is_minimal` are not assumptions there, so they can be used to assert what I needed. I will close this.

PR #1513, opened 2025-03-13
Add is_same for PointsTo

ziqiaozhou:

I want to add a trusted proof that two PointsTo<T> are equal if they are pointing to the same memory location

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -227,6 +227,7 @@ impl<T> PointsTo<T> {
 
     proof fn is_same(tracked &self, tracked other: &Self)
         requires
+            size_of::<T>() > 0,
             self.ptr() as int == other.ptr() as int,
         ensures
             self == other,
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -224,6 +224,15 @@ impl<T> PointsTo<T> {
     {
         unimplemented!();
     }
+
+    proof fn is_same(tracked &self, tracked other: &Self)
+        requires
+            self.ptr() as int == other.ptr() as int,
+        ensures
+            self == other,
+    {
+        admit()
+    }
 }
 
 impl<T> MemContents<T> {


tjhance:
Unfortunately this won't be sound, because of ghost data. This is only sound for certain types that are "purely physical".

The problem is easiest to see with ZSTs; consider, for example, `PointsTo<Ghost<u8>>`. With ZSTs, you can allocate two PointsTo for the same pointer.

Even for non-ZSTs, we want to leave open the possibility of 'transmuting' between values that have identical physical representation. (Though this isn't possible now.)

I believe this will be sound if we add `T: Structural` (this is the same trick we used in prophecy variables; https://github.com/verus-lang/verus/pull/1293).

ziqiaozhou:
> Unfortunately this won't be sound, because of ghost data. This is only sound for certain types that are "purely physical".
> 
> The problem is easiest to see with ZSTs; consider, for example, `PointsTo<Ghost<u8>>`. With ZSTs, you can allocate two PointsTo for the same pointer.
> 
> Even for non-ZSTs, we want to leave open the possibility of 'transmuting' between values that have identical physical representation. (Though this isn't possible now.)
> 
> I believe this will be sound if we add `T: Structural` (this is the same trick we used in prophecy variables; #1293).

I see. It would be better if it could be sound by just requiring size_of::<T>() > 0. The reason I want this is that I want to store a ghost variable in a non-ZST struct, then I want to assert that ghost variable is same if they are from the same ptr. 

> The possibility of 'transmuting' between values that have identical physical representation 
Since the permission is tracked reference, it should not block the possible to transmuting between values with the identical physical representation.

tjhance:
Fundamentally, the problem is that the meaning of `PointsTo<T>` is that, at the given pointer `points_to.ptr()`, the bytes there are initialized and form a valid `T` with encoding `points_to.value()`. However, the same bytes can represent multiple different types, or — when ghost content is involved, multiple values of the same type.

With @elanortang we've recently started on transmuting capabilities, the most general form of which would be:

```rust
transmute_shared_reference(t: &PointsTo<T>) -> u: &PointsTo<U>
    requires forall |bytes| valid_bytes(t.value(), bytes) ==> valid_bytes(u.value(), bytes)
```

which would let you transmute, say, `(true, Ghost(0))` into `(true, Ghost(1))` since both those values have the same physical representation. Furthermore, since this example uses shared references, both could exist at the same time.

PR #1515, opened 2025-03-17
Add axiom which states that subsets of finite sets are finite

lduranovic:
I was surprised that this is not an axiom already. Although I could see why the Verus team might not want to introduce this axiom (to not add to the SMT solver's context), I do think it's very convenient for the end user to have this. I certainly expected this to be the case when working on some proofs the other day.

If this is not added, the end user probably has to write something like this lemma, and call it at every point:
```rust
proof fn subset_of_finite_set_finite<A>(s: Set<A>)
    requires
        s.finite(),
    ensures
        forall |sub: Set<A>| #[trigger] sub.subset_of(s) ==> sub.finite(),
{
    assert forall |sub: Set<A>| #[trigger] sub.subset_of(s) implies sub.finite() by {
        let complement = s.difference(sub);
        assert(complement.finite());
        assert(sub =~= s.difference(complement));
    }
}
```
I could alternatively add this to `set_lib`, if you believe that is more sensible.

### DIFF
--- a/source/vstd/set.rs
+++ b/source/vstd/set.rs
@@ -900,7 +900,8 @@ pub broadcast proof fn axiom_set_subset_finite<A>(s: Set<A>, sub: Set<A>)
         s.finite(),
         sub.subset_of(s),
     ensures
-        #![trigger sub.subset_of(s)] sub.finite(),
+        #![trigger sub.subset_of(s)]
+        sub.finite(),
 {
 }
 
--- a/source/vstd/set.rs
+++ b/source/vstd/set.rs
@@ -894,17 +894,6 @@ pub broadcast proof fn axiom_set_choose_finite<A>(s: Set<A>)
     let _ = trigger_finite(f, ub);
 }
 
-/// A subset of a finite set `s` is finite.
-pub broadcast proof fn axiom_set_subset_finite<A>(s: Set<A>, sub: Set<A>)
-    requires
-        s.finite(),
-        sub.subset_of(s),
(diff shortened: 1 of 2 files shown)


utaal:
Hi @lduranovic, thank you! I'd say this is safe enough to add to the default broadcast list.
@Chris-Hawblitzel any concern?

utaal (approved):

Chris-Hawblitzel:
I think it should be fine.  I ran the Veritas tests on it and the performance was not impacted.

I would suggest renaming it from `axiom` to `lemma`, since it can be proven, and putting it in `set_lib.rs` next to `lemma_len_subset`, which is similar but is not a broadcast.

lduranovic:
Hi @utaal and @Chris-Hawblitzel. Thanks for the comments. As @Chris-Hawblitzel suggested, I renamed this to a lemma, and put it in `set_lib.rs`. It is still broadcast, so the user should be able to get the result as long as they import `set_lib::*`.

utaal:
I just noticed (an instant after merging) that this broadcasted lemma hasn't been added to the default set of `broadcast use`d lemmas, which means that it won't be automatically available. I'll open a follow up PR with just that change, and we'll need to re-run veritas before landing it.

lduranovic:
@utaal Oh that's my mistake, sorry about that. I thought that there was no need to add the lemma to any broadcast groups given that it has `broadcast` in the signature already. Thanks for fixing things!

PR #1518, opened 2025-03-18
Add an 'axiom' keyword

tjhance:
Add `axiom` keyword, which implies `proof` mode + `external_body`

More self-explanatory and less boilerplate than writing `external_body` and `unimplemented!()`

### DIFF
--- a/dependencies/syn/dev/main.rs
+++ b/dependencies/syn/dev/main.rs
@@ -1,7 +1,3 @@
 syn_dev::r#mod! {
     // Write Rust code here and run `cargo check` to have Syn parse it.
-
-    pub open(crate) spec fn test() {
-    }
-
 }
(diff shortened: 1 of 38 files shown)


parno:
I added support for axiom on a branch:
https://github.com/verus-lang/verus-analyzer/pull/52
Assuming the CI passes, I can merge it into main and then produce a release.

utaal (approved):
Please add at least one test.

utaal:
There are a number of axioms with `admit()` in them in, e.g., `Seq`. Do we plan to use this for those as well?

tjhance:
It'd be good to be consistent, yeah.

parno:
@tjhance Could you add something to the guide to document `axiom`, in particular that it's a proof mode function?  It doesn't currently show up anywhere that I could find, including in the `syntax.rs` example file.

PR #1520, opened 2025-03-19
cleanup datatype opaqueness diagnostics

tjhance:
The diagnostics relating to 'datatype opaqueness' have long been a source of confusion. This attempts to clean up the logic and make the diagnostic much more specific.

In the process of making this change, I realized that the old code was missing some cases. For example, if the fields were public, but the type itself was not public, there would be no error message ... fixing this caused some breakage within our examples, and it will likely cause a bit of breakage more widely.


utaal (approved):

tjhance:
no veritas failures.

PR #1523, opened 2025-03-22
cast enum to int

ziqiaozhou:

Casting unit-only or less-field enum  to integer seems natural using `discriminant_for_variant`.

The basic support is to casting enum to int if the expr is ExprKind::Path(Resolved(_, rustc_hir::Path(_)).
To support arbirary expr of enum type, I transformed the cast to a match expr.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -1288,17 +1288,62 @@ pub(crate) fn expr_to_vir_with_adjustments<'tcx>(
     }
 }
 
-pub(crate) fn try_cast_enum_to_int(tcx: TyCtxt<'_>, expr: &Expr) -> Option<u128> {
-    if let ExprKind::Path(QPath::Resolved(None, rustc_hir::Path { res, .. })) = expr.kind {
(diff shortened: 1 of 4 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1288,6 +1288,64 @@ pub(crate) fn expr_to_vir_with_adjustments<'tcx>(
...
 
+pub(crate) fn expr_cast_enum_int_to_vir<'tcx>(
+    bctx: &BodyCtxt<'tcx>,
+    expr: &'tcx Expr<'tcx>,
+    mk_expr: impl Fn(ExprX) -> Result<vir::ast::Expr, vir::messages::Message>,
+    current_modifier: ExprModifier,
+) -> Result<vir::ast::Expr, VirErr> {
+    let tcx = bctx.ctxt.tcx;
+    let ty = bctx.types.node_type(expr.hir_id);
+    let expr_vir = expr_to_vir(bctx, expr, current_modifier)?;
```
Chris-Hawblitzel, line 1299: Since the callers of `expr_cast_enum_int_to_vir` already have `source_vir`, can we just pass that in instead of running `expr_to_vir` again here?

Chris-Hawblitzel (approved):

PR #1524, opened 2025-03-22
proof closures

Chris-Hawblitzel:
This pull request adds proof closures, following the design from https://github.com/verus-lang/verus/discussions/1419 , with the goal of making it easier to work with higher-order proof functions that manipulate tracked values.

# Basic proof closures

Proof closures are similar to exec closures.  Consider the following test from exec_closures.rs:

```
        fn testfn() {
            let f = |y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };

            assert(f.requires((2,)));
            assert(!f.ensures((2,), 3));

            let t = f(2);
            assert(t == 2);
        }
```

The corresponding proof closure version of this is essentially the same, except that it exists in proof mode and the closure has the syntactic prefix "proof_fn":

```
        proof fn testfn() {
            let tracked f = proof_fn|y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };

            assert(f.requires((2,)));
            assert(!f.ensures((2,), 3));

            let t = f(2);
            assert(t == 2);
        }
```

As with exec closures, proof closures use `.requires` and `.ensures` to describe a closure's preconditions and postconditions.  However, unlike exec closures, proof closures have specific types that can be written with the `proof_fn(...) -> ...` syntax, similar to `spec_fn` types for spec closures:

```
        proof fn helper(tracked f: proof_fn(y: u64) -> u64)
            requires
                f.requires((2,)),
                forall|z: u64| f.ensures((2,), z) ==> z == 2,
        {
            let t = f(2);
            assert(t == 2);
        }

        proof fn testfn() {
            let tracked f = proof_fn|y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };
            helper(f);
        }
```

Nevertheless, it's still possible to use exec-closure-style traits when writing functions that take proof closures as an argument, in order to be as generic as possible over the exact `proof_fn` type:

```
        proof fn helper<F: ProofFnOnce>(tracked f: proof_fn<F>(y: u64) -> u64)
            requires
                f.requires((2,)),
                forall|z: u64| f.ensures((2,), z) ==> z == 2,
        {
            let t = f(2);
            assert(t == 2);
        }

        proof fn testfn() {
            let tracked f = proof_fn|y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };
            helper(f);
        }
```

In this example, `helper` is generic over various `proof_fn` types, specifying only that it needs a minimum ability to call the function once, as specified in the marker trait `ProofFnOnce`, which is the `proof_fn` analogue to `FnOnce` for exec closures.

Proof closures and proof closure types can have tracked arguments and return values, for example `proof_fn|tracked y: T| -> (tracked z: T)` and `proof_fn(tracked y: T) -> tracked T`.

# Once, Mut, Copy, Send, and Sync options

Exec closures may implement `FnOnce`, `FnMut`, `Fn`, `Clone`, `Copy`, `Send`, and `Sync`.  Proof closures may also have these characteristics (except for Clone, which carries the exec-mode-only `clone` function).  These characteristics are specified as part of the syntax for proof closures and proof closure types, using an optional unordered list written inside square brackets:

```
        proof fn helper(tracked f: proof_fn[Once, Copy, Send, Sync](y: u64) -> u64)
            requires
                f.requires((2,)),
                forall|z: u64| f.ensures((2,), z) ==> z == 2,
        {
            let t = f(2);
            assert(t == 2);
        }

        proof fn testfn() {
            let tracked f = proof_fn[Once, Copy, Send, Sync]|y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };
            helper(f);
        }
```

By default, a proof closure is `Fn`-like; the `Once` and `Mut` options make it `FnOnce` and `FnMut`-like, respectively.  The type argument `F` in `proof_fn<F>(...) -> ...` represents the options, so that functions can be generic over the options, as in:

```
        proof fn helper<F: ProofFnOnce + Copy>(tracked f: proof_fn<F>(y: u64) -> u64)
            requires
                f.requires((2,)),
                forall|z: u64| f.ensures((2,), z) ==> z == 2,
        {
            let t = f(2);
            assert(t == 2);
        }

        proof fn testfn() {
            let tracked f = proof_fn[Mut, Copy, Send, Sync]|y: u64| -> (z: u64)
                requires y == 2
                ensures z == 2
            {
                y
            };
            helper(f);
        }
```

The type argument `F` should be bound by one or more of the marker traits `ProofFnOnce`, `ProofFnMut`, `ProofFn`, `Copy`, `Send`, and `Sync`, where `F: ProofFnOnce` is the minimum level of functionality.

# ReqEns option

Writing `.requires` and `.ensures` many times can be verbose, so proof closures also offer an alternative approach to describing a proof closure's preconditions and postconditions, similar to how dependent type systems express preconditions and postconditions.  In this approach, the preconditions and postconditions become part of the proof closure type itself in addition to being part of the value via `.requires` and `.ensures`.  Since Rust doesn't actually have dependent types, this approach uses a trait `ProofFnReqEnsDef` to specify the preconditions and postconditions:

```
pub trait ProofFnReqEnsDef<Args, Output> {
    fn req(args: Args) -> bool;
    fn ens(args: Args, output: Output) -> bool;
}
```

A program can implement this trait with some type (say, `S`) and then use the option `ReqEns<...type...>` (in this case, `ReqEns<S>`) to specify the preconditions and postconditions named by `S`: 

```
        struct S;
        impl ProofFnReqEnsDef<(u8, u8), int> for S {
            open spec fn req(a: (u8, u8)) -> bool {
                a.0 < a.1
            }
            open spec fn ens(a: (u8, u8), o: int) -> bool {
                o == a.0 + a.1
            }
        }
        proof fn q1<F: ProofFn + ProofFnReqEns<S>>(tracked f: proof_fn<F>(u8, u8) -> int) {
            let x = f(10, 20);
            assert(x == 30);
        }
        proof fn q2<F: ProofFn + ProofFnReqEns<S>>(tracked f: proof_fn<F>(u8, u8) -> int) {
            f(20, 10); // FAILS
        }
        proof fn p1() {
            q1(proof_fn[ReqEns<S>]|x, y| { assert(x < y); x + y });
        }
        proof fn p2() {
            q1(proof_fn[ReqEns<S>]|x, y| { assert(x < y); x - y }); // FAILS
        }
        proof fn p3() {
            q1(proof_fn[ReqEns<S>]|x, y| { assert(x > y); x + y }); // FAILS
        }
```

The `ReqEns<...>` option can be used in the `[..., ReqEns<...>, ...]` syntax in proof closure expressions and proof closure types, as well as in the trait bound `ProofFnReqEns<...>`, as in the example above.

# General syntax

This PR extends the Verus syntax with:

- expression `proof_fn<'a, F>[...options...]|...arguments...| -> type requires ... ensures ... { body }`
- type `proof_fn<'a, F>[...options...](...arguments...) -> type`

Here, the `'a`, the `F`, the `[...options...]` are all optional.  The `-> type` in the expression is also optional.  Note that in the expression, the only thing new is the `proof_fn<'a, F>[...options...]` part and the ability for the arguments and return type to be `tracked`.  The rest of the syntax is the same as for exec closures.

The options may be `Once`, `Mut`, `ReqEns<type>`, `Copy`, `Send`, `Sync`.  The `verus_syn` implementation of these options simply uses `PathSegment` to represent all of these, rather than hard-wiring the exact options into the syntax itself.

# Underneath the hood

The `proof_fn` types are represented as a `builtin::FnProof` type, with an `Options` type argument of type `builtin::FOpts`:

```
pub struct FnProof<'a, Options, ArgModes, OutMode, Args, Output>
pub struct FOpts<const USAGE: u8, ReqEns, const COPY: u8, const SEND: u8, const SYNC: u8>
```

The type can get verbose, making it hard to read in error messages; it can also cause Rust to generate `long-type` files. To mitigate these issues, I tried to keep the type names and type arguments as concise as possible, for example by using `const` integer arguments instead of named types for many of the flags.  (I also set `-Z write_long_types_to_disk=no` in rust_verify_tests, since the tests were generating these `long-type` files for both `exec_closures.rs` and `proof_closures.rs` regardless of whether our type names were concise or not.)

In the discussion on https://github.com/verus-lang/verus/discussions/1419, I indicated that I thought we could avoid reusing the exec-closure `FnWithRequiresEnsures` trait for proof closures. It turned out, though, that Rust's trait dispatching wanted to dispatch `.requires` and `.ensures` operations on `&proof_fn` types to `FnWithRequiresEnsures`, so I ended up reusing this trait for proof closures after all.  Nevertheless, other operations are kept separate between exec closures, proof closures, and spec closures.  For example, there is a `closure_to_fn_proof` builtin function separate from `closure_to_fn_spec`, and a `proof fn proof_nonstatic_call` vstd function separate from `fn exec_nonstatic_call`.  The `closure_to_fn_proof` builtin function leans heavily on Rust's trait system to transmit features of the underlying closure, such as `Copy` and `Sync`, into the resulting `FnProof` type:

```
pub fn closure_to_fn_proof<
    'a,
    const USAGE: u8,
    const COPY: u8,
    const SEND: u8,
    const SYNC: u8,
    ArgModes,
    OutMode,
    Args,
    Output,
    F: 'a,
>(
    _f: F,
) -> FnProof<'a, FOpts<USAGE, (), COPY, SEND, SYNC>, ArgModes, OutMode, Args, Output>
where
    ProofFnConfirm: ConfirmUsage<Args, Output, USAGE, F>,
    ProofFnConfirm: ConfirmCopy<COPY, F>,
    ProofFnConfirm: ConfirmSend<SEND, F>,
    ProofFnConfirm: ConfirmSync<SYNC, F>,
    ArgModes: core::marker::Tuple,
    OutMode: ConfirmTracked,
    Args: core::marker::Tuple,
{
    unimplemented!()
}
```

Here, `ConfirmCopy`, etc., make sure that if the proof closure type is marked, say, `Copy`, then the underlying Rust closure `F` implements `Copy`.


Chris-Hawblitzel:
@parno and @jaybosamiya-ms , I marked you as reviewers since this involves a syntax extension, but don't feel obligated to review the entire PR.

parno (commented):
Thanks for adding this!  I have a PR for verus-analyzer that handles the new syntax.  

When your PR moves forward, could you put a copy of your explanatory text above into the Verus Guide?  It would also be helpful to add some examples of new syntax to `source/rust_verify/example/syntax.rs`. The verusfmt repo's CI already has an action to periodically pull that file and check that verusfmt still works on it, so that helps catch any misalignment. Also, that file is incorporated into the Verus guide (https://verus-lang.github.io/verus/guide/syntax.html), which makes it easier (compared to looking at test cases) for users to see how a piece of syntax is meant to be used; that might be redundant with the guide page, but I think there's some utility to have one centralized location where you can see most/all of the syntax exercised in one place.

jaybosamiya-ms:
I've dropped the label, since it appears that both verusfmt and verus-analyzer now have support for this syntax. If more changes are needed to the syntax, please do re-add the label, thanks!

tjhance (approved):

PR #1526, opened 2025-03-23
fix definition of Set::is_empty

tjhance:
is_empty should imply finite


parno (approved):
This looks reasonable to me.  I thought the previous requirement re: `finite` was a deliberate design choice, but I gather that's not the case.

Chris-Hawblitzel (approved):

tjhance:
veritas looks good.

PR #1532, opened 2025-03-25
Issue a warning when `cargo verify` doesn't find anything to verify.

parno:
Since several folks have stumbled on this issue, adding a warning seems useful.  I also think it's important to have it in color, when possible, since `cargo` emits a lot of other information while it's processing the crate's dependencies, so the color helps the warning stand out.


jaybosamiya-ms (approved):
Awesome, thanks!  Tested it, works great :) 

image

Chris-Hawblitzel (approved):

PR #1545, opened 2025-03-28
Enforced termination checking

apribiso:
Exec code that is recursive or contains loops must have decreases clause to enforce termination. Added example and test file for demonstrating and checking this behavior on concrete examples. Tests were created to cover most cases addressed in the loop and recursive function test files. 
Introduced crate-level attribute that allows for a crate to opt-out of this requirement (which generates a warning message to the user that termination is not enforced). @utaal can provide a bit more insight into how the crate-level attribute was implemented.


tjhance (changes requested):
There are a few things I notice about this design:

First of all, the termination-checking is entirely local, and it doesn't check dependencies. This has some surprising consequences. This code passes, for example:

```
#[verifier::admit(may_not_terminate)]
fn test() {
    loop { }
}


fn test2() {
    test();
}
```

Here, test2 was not marked with may_not_terminate, so the user presumably wants it to terminate. However, it clearly doesn't terminate. Given the way this attribute behaves, I might have expected it to be called something like `assume(terminates)`, which is very different than 'may not terminate'!

(It may be worth taking a look at prior art here: Dafny has a superficially similar feature called [`decreases *`](https://dafny.org/latest/OnlineTutorial/Termination) which effectively means "may not terminate". However, Dafny checks if any of its dependencies have this annotation when ensuring termination.)

Secondly, it's important to note that working with nonterminating code is still an important use-case, if not the default one. Daemons and servers, for example, are nonterminating by design. And when dealing with concurrent code, we often eschew termination-checking simply because it's still very difficult with our current bag of techniques. As a result, I think that the way the feature is tied to no-cheating will be difficult to work with:

 * If you want to use no-cheating to ensure the absence of assumptions or trusted code, you can't use nonterminating code.
 * If you really want to verify termination, you have to use 'no-cheating' (to ensure the absence of the `may_not_terminate` attribute throughout your entire crate), but that also precludes using trusted code.

The need to change the implementation of `vstd::rwlock` is an illustration of this difficulty. I expect we'd encounter many similar issues in the future.

To proceed, I think we need to:

1. Think about how we're going to communicate and document this feature, so that the guarantees it provides (in what contexts and when) become clear. In particular, we need to avoid the surprising behavior I discussed above. This will probably involve changing the name of the opt-out attribrute, or changing the behavior to check dependencies.

2. Make it so that vstd does not need to have every function be terminating (without subverting the guarantees made by point 1).

utaal:
Thanks @tjhance. This was based loosely on what we had discussed as a first step on https://github.com/verus-lang/verus/discussions/254#discussioncomment-11987967:

> More sources of non-termination than just checking for decreases clauses on loop and recursion.
> Already supported for recursion, and loops. No enforcement yet.
> How to do UX?
> error (initially warning) in the absence of a decreases clause in exec code, and crate-level attributes to opt-in or opt-out of total correctness.

The PR is not intended to completely address termination checking in all cases, but I agree it found itself in an awkward spot, in the presence of exec/recursion in vstd.

> It may be worth taking a look at prior art here: Dafny has a superficially similar feature called [decreases *](https://dafny.org/latest/OnlineTutorial/Termination) which effectively means "may not terminate". However, Dafny checks if any of its dependencies have this annotation when ensuring termination.

Ultimately (also see the discussion), we were planning to go in that direction; this is an attempt at a step in that direction.

---

Let me address a couple notes:

### Trusted opt-out attribute

> Given the way this attribute behaves, I might have expected it to be called something like assume(terminates), which is very different than 'may not terminate'!

I agree this naming is sub-optimal. The intent here was similar to `#[verifier::external_body]`, and I was trying to stay consistent with the global attribute. I think it's okay to introduce a new name and switch to `#[verifier::assume_terminates]`.

### Interaction with no-cheating

> If you want to use no-cheating to ensure the absence of assumptions or trusted code, you can't use nonterminating code.

That's not the case: `#![verifier::allow(may_not_terminate)]` at the crate level is allowed by `--no-cheating`, it is just the trusted opt-out per-function that isn't.

`--no-cheating` is also, to my understanding, a first step to aid classroom usage and LLM-based codegen experiments, not a full solution to the trusted/untrusted split. I believe most projects at the moment won't be able to use `--no-cheating` because they have at least some `#[verifier::external_body]` function to interact with their environment.

utaal:
When we talked at the meeting @Chris-Hawblitzel was surprised there were any exec loops or recursion in `vstd`, so I thought it may be okay to handle the only case by making the verified `rwlock` fallible, but in private conversation it seems we might expect more to crop up. Obviously dependency-checking is the way to go if we want to allow more fine-grained control: I was hoping there was an intermediate crate-level-only step, but in the absence of other ideas, and if we don't want to change rwlock, I think we need to track potential non-termination at the function level.

tjhance:
> That's not the case: #![verifier::allow(may_not_terminate)] at the crate level is allowed by --no-cheating, it is just the trusted opt-out per-function that isn't.

Thanks for the correction; I had misunderstood this and thought both the crate-level and function-level attributes were the same.

KaminariOS:
Consider the case that we call a function from another verified crate. In that crate, only that function is verified for termination, not the whole crate. A per function marker in the signature seems helpful.

utaal:
Moved to the lightweight check design discussed at the verus meeting. No checking of callees, renamed attribute to `#[verifier::exec_allows_no_decreases_clause]`.

Chris-Hawblitzel (approved):
Looks good to me; I just have minor suggested changes.

PR #1546, opened 2025-03-28
add Map::index_opt() to fetch Option<V> from Map<K, V>

zeldovich:
This allows concisely stating at least two kinds of helpful facts:

- that a key is both present in a map and has a certain value, as in, `m.index_opt[k] == Some(v)`.

- that maps agree on the presence and values of certain keys, as in, `forall |k| pred(k) ==> m1.index_opt(k) == m1.index_opt(k)`.

### DIFF
--- a/source/vstd/map_lib.rs
+++ b/source/vstd/map_lib.rs
@@ -35,6 +35,16 @@ impl<K, V> Map<K, V> {
         exists|i: K| #[trigger] self.dom().contains(i) && self[i] == v
     }
 
+    /// Returns `Some(v)` if the key `k` is in the domain of `self` and maps to `v`,
+    /// and `None` if the key `k` is not in the domain of `self`.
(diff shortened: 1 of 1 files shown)


utaal:
(I'm re-running CI because I think it failed because it was failing on `main`, briefly)

tjhance:
This seems like a good addition.

I just want to note though, since it might have been missed, that the first use-case is already handled by [contains_pair](https://verus-lang.github.io/verus/verusdoc/vstd/map/struct.Map.html#method.contains_pair).

utaal (approved):

zeldovich:
Thanks, I had indeed missed `contains_pair`, though it turns out not to handle the more specific use case I was trying to abstract into example 1 -- in particular, I wanted to specify the key in one place (i.e., invoke `m.index_opt(k)`), and say what's true about the resulting `Option<V>` in another place (whether it be `None` or `Some(v)`).

PR #1548, opened 2025-03-28
Add default_ensures for declaring ensures clauses of default trait functions

Chris-Hawblitzel:
The pull request https://github.com/verus-lang/verus/pull/1465 involves writing specifications for traits like `PartialEq` and `PartialOrd`, using the `call_ensures`-style summarized in https://github.com/verus-lang/verus/discussions/1527 .  `PartialEq` and `PartialOrd` have default (aka provided) functions like `ne`, `gt`, etc.  For the `call_ensures` style to work, we'd like these functions to have no ensures clauses for their declarations (so that implementations can choose their own ensures clauses if they choose implement these functions), but the default implementations of these functions should have ensures clauses.  For example, `ne`'s default implementation should ensure that it returns the opposite of what `eq` returns.  Without this ensures clause, `ne`'s default implementation would be useless for verification.

However, at the moment, we don't have a way of adding ensures clauses to a default function without also adding them to the underlying function declaration.  This pull request adds a mechanism for this.  To declare an ensures clause specifically for the default function, you wrap the clause inside `default_ensures`:

```
        trait T {
            fn f(i: u32) -> (r: u32)
                ensures
                    r <= i,
                    default_ensures(r == i / 2),
            {
                i / 2
            }
        }
        impl T for u8 {
        }
        fn generic<A: T>() {
            let r = A::f(6);
            assert(r <= 6);
            assert(r == 3); // FAILS
        }
        fn inheritor() {
            let r = <u8 as T>::f(6);
            assert(r == 3);
        }
```

Here's an example similar to `eq` and `ne` of `PartialEq`:

```
trait T {
    fn e() -> bool;
    fn n() -> (r: bool)
        ensures
            default_ensures(call_ensures(Self::e, (), !r)),
    {
        !Self::e()
    }
}

impl T for u8 {
    fn e() -> (r: bool) ensures r { true }
}
fn test() {
    let b = <u8 as T>::n();
    assert(!b);
}
```


ziqiaozhou:
Would that be possible to use a non-nested style? 
```
 fn n() -> (r: bool)
        default_ensures
            call_ensures(Self::e, (), !r),
```

Chris-Hawblitzel:
> Would that be possible to use a non-nested style?
> 
> ```
>  fn n() -> (r: bool)
>         default_ensures
>             call_ensures(Self::e, (), !r),
> ```

I've tried to avoid adding any new syntax in this PR, but I'm open to syntactic sugar like this in the future.

utaal:
I agree with @ziqiaozhou - I find the function-call-style syntax confusing. I also support adding syntax for this.

tjhance:
i'm gonna go out on a limb and suggest something like

```
default {
  ensures ...
}
```

Probably easiest on the parser, the thing in the braces is just a Specification.

ziqiaozhou:
> `bool`

I also feel @zeldovich's proposal to expose a bool spec T::is_default_impl() is useful, which is equivalent to obey_cmp_model for PartialOrd

tjhance (changes requested):
can we make this a Header type instead of a UnaryOp?

Chris-Hawblitzel:
> can we make this a Header type instead of a UnaryOp?

Yes, I can make it an ExprX::Header.  I still need an AST node to translate it into, since ExprX::Header doesn't appear in the final AST produced by rust_to_vir.  Some options for this include (1) UnaryOp (mostly as it is now), (2) an extra flag on each Expr in what is currently `FunctionX::ensure: Exprs`, and (3) a separate `default_ensure` next to `ensure`.  I was trying to avoid the last option for the sake of defensive coding, because I think it would be too easy to forget about `default_ensure` when processing `ensure`.  But I'm open to the second option.

tjhance:
I have essentially the opposite instinct: Option 1 seems really funky to me. Option 2 is okayish and Option 3 seems like the natural approach.

As I understand it, the main role of default_ensures is that it would be copied into the ensures when we copy-paste the default function to the individual impls. After the traits pass, the default_ensures could be ignored or cleared out entirely. Is that right?

utaal:
I agree with @tjhance here. 3 sounds like the most natural.

tjhance (commented):
Does this emit some kind of axiom like, `forall arg, call_ensures(f, args) ==> default_ensures(f, args)` (for applicable trait impls f which don't override the default)? It seemed to me like that would need to be here somewhere but I couldn't find it scrolling through the diff.

Chris-Hawblitzel:
> I agree with @tjhance here. 3 sounds like the most natural.

Commit https://github.com/verus-lang/verus/pull/1548/commits/167e85d89af8ea5defed9d3d6158cf1231e77e43 changed from Option 1 to a variant of Option 3, using separate lists of expressions for ensures and default_ensures, but keeping the lists together as a pair rather than in separate fields, so that any code using ensures will also have to think about what to do with default_ensures.

Chris-Hawblitzel:
> Does this emit some kind of axiom like, `forall arg, call_ensures(f, args) ==> default_ensures(f, args)` (for applicable trait impls f which don't override the default)? It seemed to me like that would need to be here somewhere but I couldn't find it scrolling through the diff.

Yes, the following code in ast_simplify generates the axiom (specifically, the `(false, closure_enss)` case):

```
    if inherit {
        assert!(closure_enss.len() + default_enss.len() == 0);
        let (_, tuple_var) = exec_closure_spec_param(state, &function.span, &params);
        let ret_var = SpannedTyped::new(&function.span, &ret.a, ExprX::Var(ret.name.clone()));
        let default_expr = mk_closure_ens_call(
            state,
            &function.span,
            &params,
            &fndef_singleton,
            &tuple_var,
            &ret_var,
            BuiltinSpecFun::DefaultEns,
        );
        closure_enss = Arc::new(vec![default_expr]);
    }
    for (default_ens, enss) in [(false, closure_enss), (true, default_enss)] {
        if enss.len() > 0 {
            let ens_forall = exec_closure_spec_ensures(
                state,
                &function.span,
                &fndef_singleton,
                &params,
                &ret,
                &*enss,
                default_ens,
            )?;
            fndef_axioms.push(ens_forall);
        }
    }
```

For example, for this code:

```
uninterp spec fn g() -> bool;

trait T {
    fn f()
        ensures
            default_ensures(g())
        {
            assume(g());
        }
}

impl T for u8 {
}

proof fn test() {
    assume(call_ensures(u8::f, (), ()));
    assert(g());
}
```

the following is generated:

```
;; Function-Specs scratch::impl&%0%default%f
(axiom (forall ((closure%$ Poly) (%return$ Poly)) (!
   (=>
    (and
     (has_type closure%$ TYPE%tuple%0.)
     (has_type %return$ TYPE%tuple%0.)
    )
    (=>
     (closure_ens (FNDEF%scratch!T.f. $ (UINT 8)) $ TYPE%tuple%0. (F (fndef_singleton))
      closure%$ %return$
     )
     (default_ens (FNDEF%scratch!T.f. $ (UINT 8)) $ TYPE%tuple%0. (F (fndef_singleton))
      closure%$ %return$
   )))
   :pattern ((closure_ens (FNDEF%scratch!T.f. $ (UINT 8)) $ TYPE%tuple%0. (F (fndef_singleton))
     closure%$ %return$
   ))
   :qid user_scratch__impl&%0%default%f_2
   :skolemid skolem_user_scratch__impl&%0%default%f_2
)))
```

The test case `test_default_ensures3` covers this.

PR #1558, opened 2025-04-02
fractional resource library

zeldovich:
This PR adds a PCM library for fractional ownership of a ghost variable.  The library is an evolved version of the fractional ownership PCM that I added a while ago as an example, but the library has now seem to have stabilized in terms of features needed for various projects, so it would be useful to have it available in `vstd`.


source/vstd/pcm/frac.rs
```diff
@@ -0,0 +1,272 @@
...
+    proof fn unit_valid() {
+    }
+}
+
+pub struct Frac<T, const TOTAL: u64 = 2> {
+    r: Resource<FractionalCarrier<T, TOTAL>>,
+}
+
+impl<T, const TOTAL: u64> Frac<T, TOTAL> {
+    pub closed spec fn inv(self) -> bool {
```
tjhance, line 79: you can probably use the `type_invariant` attribute here, so your library doesn't need to include self.inv() in the requires and ensures clauses.

zeldovich: I tried doing this originally, but ran into this problem:

```
% cat example.rs
use vstd::prelude::*;

verus! {
    struct S<const N: u64> {
    }

    impl<const N: u64> S<N> {
        #[verifier::type_invariant]
        pub closed spec fn inv(self) -> bool { true }
    }
}
% verus --crate-type=lib example.rs
error: external_type_specification: Const params not yet supported
 --> example.rs:9:9
  |
9 |         pub closed spec fn inv(self) -> bool { true }
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

tjhance: this should be fixed with 0f22710fc04740132d0466ce9dd223011abf150a

zeldovich: Thanks!  Do you have any suggestions for how to use the type invariant of a mutable reference?  Do we need to add a version of `use_type_invariant()` that works on a `tracked &mut T`?

```
% cat example.rs
use vstd::prelude::*;

verus! {
    struct S {}

    impl S {
        #[verifier::type_invariant]
        pub closed spec fn inv(self) -> bool { true }

        proof fn x(tracked &mut self) {
            use_type_invariant(self);
        }
    }
}
% verus --crate-type=lib example.rs
error: The verifier does not yet support the following Rust feature: &mut types, except in special cases
  --> example.rs:11:32
   |
11 |             use_type_invariant(self);
   |                                ^^^^

```

zeldovich: Ah, turns out `use_type_invariant(&*self)` works in this case.  Updated the PR to use type invariants and avoid requiring an explicit `frac.inv()`.  Thanks for the suggestion!

tjhance:
Can we rename this? I'd like to reserve `Frac` for a more different fractional resource type that is parameterized by arbitrary resources. This one could be called `FracGhost` or something.

zeldovich:
Sure, I changed the name to `FracGhost`.  I also added a simple wrapper that captures one common use pattern for this library---namely, having one authoritative side owning a half, and another "client" side owning another half.  The comment at the top of `frac.rs` describes this.

tjhance (approved):
Looks good. You can also make `inv` private since the user won't need it

PR #1559, opened 2025-04-02
logical atomicity library

zeldovich:
This PR adds a library (basically just defining some traits) that helps with implementing logically atomic interfaces.  It's convenient to use it in combination with resources like the fractional PCM from PR #1558.


zeldovich:
FWIW, `verusfmt` now supports set notation for `opens_invariants`, so this PR passes all the CI checks and could be merged if it looks good to others.

PR #1563, opened 2025-04-03
change the encoding for consts and const fns

tjhance:
We've talked about attempting to produce erased MIR. However, the way we presently handle consts is problematic for this. The problem is that Rust may evaluate const items during type-checking if they are used in Rust types; as a result, Verus needs to make sure that const items result in executable MIR. Currently, this is done by putting all ghost code behind closures. This works fine now, but as long as these constraints exist, it's not going to be possible to implement MIR that is spec-erased but not tracked-erased. The problem goes beyond 'const' items and applies to 'const fn' items as well.

This PR splits every 'const' and 'const fn' into two items, an executable, erased version, and an un-erased version with all proof code. Roughly speaking, this:

```
const X: u64 = {
   assert(true);
   0
};
```

becomes this:

```
#[verus::internal(unerased_proxy)]
#[verus::internal(encoded_const)]
fn VERUS_UNERASED_PROXY__X() -> u64 {
   assert(true);
   0
}

#[verifier::external]
#[verus::internal(has_unerased_proxy)]
const X: u64 = { 0 };
```

The constant `X` may be evaluated during type-checking (as types may refer to `X`) but Verus will process `VERUS_UNERASED_PROXY__X` instead.

Any `const fn` is handled similarly.

Notes:

 * Verus does not check equivalence between the two items. We trust that the syntax macro implements erasure correctly (which is already true anyway).
 * This has the unforunate side-effect that any type errors in a const will be reported twice.
 * There's an unfortunate corner case where the const implementation uses a `verus_exec_expr!` macro. Since this is expanded after the original `verus` macro, there's no way for it to know which "version" of the function it's in, so it doesn't know whether or not to execute erasure. This won't be a problem if we go the MIR-erasure route (since our MIR-erasure can just fix up the erased version to be properly erased) but it does make the whole system a bit less elegant. It also means we may not be able to remove the closure-hacks yet.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -28,11 +28,11 @@ use syn_verus::{
     FnArg, FnArgKind, FnMode, Global, Ident, ImplItem, ImplItemFn, Invariant, InvariantEnsures,
     InvariantExceptBreak, InvariantNameSet, InvariantNameSetList, InvariantNameSetSet, Item,
     ItemBroadcastGroup, ItemConst, ItemEnum, ItemFn, ItemImpl, ItemMod, ItemStatic, ItemStruct,
-    ItemTrait, ItemUnion, Lit, Local, MatchesOpExpr, MatchesOpToken, ModeSpec, ModeSpecChecked,
-    Pat, PatIdent, PatType, Path, Publish, Recommends, Requires, ReturnType, Returns, Signature,
-    SignatureDecreases, SignatureInvariants, SignatureSpec, SignatureSpecAttr, SignatureUnwind,
-    Stmt, Token, TraitItem, TraitItemFn, Type, TypeFnProof, TypeFnSpec, TypePath, UnOp, Visibility,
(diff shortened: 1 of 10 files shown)


utaal (changes requested):
Please use `#[verus::internal(...)]` attributes for internal markers, they are parsed as https://github.com/verus-lang/verus/blob/main/source/rust_verify/src/attributes.rs#L134

tjhance:
`unerased_proxy`, `uses_unreased_proxy`, and `encoded_const` all use the verus::internal attribute format. The PR description was in error.

utaal (approved):

Chris-Hawblitzel (approved):

Chris-Hawblitzel:
https://github.com/verus-lang/verus/pull/1600 is related and was merged.

utaal:
[retreat 2025] @tjhance there's no problem right now, let's wait for the result of the thir erasure experiment: https://github.com/verus-lang/verus/pull/1489

PR #1568, opened 2025-04-04
strong_call_ensures attribute

Chris-Hawblitzel:
This enables `call_ensures` defined with `<==>` rather than `==>`.  Instead of globally changing the `call_ensures` behavior or changing the behavior based on a crate-level flag, I made an attribute `#[verifier::strong_call_ensures]` that can be placed on individual trait declaration functions.  When used, the attribute puts the trait ensures and trait impl ensures for the trait function in non-prophetic mode.  The intended use is for specifying `PartialEq::eq`, `PartialOrd::partial_cmp`, and `Ord::cmp`, in an upcoming pull request. 
 
### DIFF
--- a/source/air/src/ast_util.rs
+++ b/source/air/src/ast_util.rs
@@ -191,6 +191,10 @@ pub fn mk_false() -> Expr {
     Arc::new(ExprX::Const(Constant::Bool(false)))
 }
 
+pub fn mk_const_bool(b: bool) -> Expr {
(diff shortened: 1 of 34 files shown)


tjhance (approved):

parno:
Discussed at the Verus Retreat 2025.  Discussed doing this automatically for any function whose signature does not include prophetic values.

PR #1569, opened 2025-04-04
Add specifications and laws for PartialEq, Eq, PartialOrd, Ord

Chris-Hawblitzel:
This pull request defines:
- external trait specifications for PartialEq, Eq, PartialOrd, and Ord, where the default (provided) functions have specifications in terms of the `call_ensures` of the required functions
- a set of laws that may optionally be obeyed by implementations of PartialEq, Eq, PartialOrd, and Ord

Since we've been discussing alternate styles of specification for these traits, potentially involving new Verus features, I've commented the specifications here as being experimental and subject to change.  I haven't added any crate-level attributes associated with using these specifications; instead, I added a new function-level attribute specifically for changes to `call_ensures` as part of a separate pull request: https://github.com/verus-lang/verus/pull/1568 .

### DIFF
--- a/source/rust_verify_test/tests/eq_cmp.rs
+++ b/source/rust_verify_test/tests/eq_cmp.rs
@@ -1,0 +1,114 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] eq_cmp1 verus_code! {
+        use vstd::laws_eq::*;
+        use vstd::laws_cmp::*;
+        use core::cmp::Ordering;
+
+        fn test_eq<T: Ord>(x: &T, y: &T) -> (r: bool)
+            requires
+                obeys_cmp_spec::<T>(),
+            ensures
+                r <==> eq_result(*x, *y),
+        {
+            reveal(obeys_eq_partial_eq);
+            reveal(obeys_cmp_partial_ord);
+            reveal(obeys_cmp_ord);
+
+            if x.lt(y) {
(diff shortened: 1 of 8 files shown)


ziqiaozhou:
Thanks, I updated #1465 to use it. It works well.

ziqiaozhou (approved):

utaal (approved):

tjhance (approved):

Chris-Hawblitzel:
@ziqiaozhou @tjhance Similar to https://github.com/verus-lang/verus/pull/1465, I also updated this to use trait extensions, and I think this pull request is also more or less ready to merge.

One change that I made as a result of updating to trait extensions is how `Structural` is handled.  Previously, there was an axiom saying that for `Structural` types, `call_ensures` for exec `eq` was equivalent to spec `==`.  Now, the definitions about equality are in terms of the trait extension, not in terms of `call_ensures`, so the previous axiom is no longer very useful.  Instead, it's more useful to be able to easily implement the trait extension for `PartialEq` together with implementing `Structural`, so that they match.  I added a new derive macro named `StructuralEq` to do this, so that by saying:

```
        #[derive(PartialEq, Eq, StructuralEq)]
        struct S(u8, bool);
```

S is declared to be both `Structural` and to have the following impl:

```
        impl PartialEqSpecImpl for S {
            open spec fn obeys_eq_spec() -> bool {
                true
            }
            open spec fn eq_spec(&self, other: &Self) -> bool {
                self == other
            }
        }
```

Issue #1550, opened 2025-03-31
Improve `broadcast use` syntax consistency with Rust


jaybosamiya-ms:
I know this is _technically_ a feature request (and thus could be a GitHub "Discussion"), but I'm filing it as an issue because I don't think the feature request is controversial or needs much discussion. (Otoh, no rush; this is not a blocker as far as I'm aware; just something on the backlog I'm adding here so it is not left untracked in this repo)

The `broadcast use` syntax is not currently consistent with Rust's `use` syntax, since it does not supporting braces (which incidentally makes trailing commas annoying, see https://github.com/verus-lang/verusfmt/issues/86 and https://github.com/verus-lang/verusfmt/issues/130 --- we _could_ fix this in verusfmt to not generate trailing commas, but that would only hide the issue in Verus, rather than fix it, thus opening an issue here).

As mentioned by @utaal in https://github.com/verus-lang/verusfmt/issues/86#issuecomment-2293455204:
> Honestly, IMO the right thing to do is to require braces around the broadcast use list. This is an oversight of mine in terms of consistency with Rust's `use` syntax. I'll add support for `{ a, b, c }`, and deprecate the bare comma-separated list.

I personally don't think that the bare comma-separate list even needs to be deprecated, we can just support both versions just fine imho.

Quick test for the preferred version (for easy copy-paste for testing):
```rs
broadcast use {
    foo,
    bar,
};
```

As mentioned in that https://github.com/verus-lang/verusfmt/issues/86, I have a version working with Verusfmt, and I am happy to merge and release as soon as support for Verus itself lands.

PR #1571, opened 2025-04-05
update syntax for broadcast use

ahuoguo:
#1550 

Maybe should still support old syntax and give a deprecation warning to prevent a big breaking change on syntax.

@jaybosamiya A new verusfmt release needs to be made to pass format check.


utaal:
> Maybe should still support old syntax and give a deprecation warning to prevent a big breaking change on syntax.

Yea, I think this would be preferable. There are examples of emitting warnings in `syn`: let me know if I can provide pointers.

jaybosamiya-ms:
I've implemented https://github.com/verus-lang/verusfmt/pull/134 if you'd like to test it out locally and make sure things work fine. Assuming that seems good, please let me know there and I'll cut a new release with it.

For now that Verusfmt PR maintains non-braced as non-braced, but after this Verus PR is merged, I can make a new Verusfmt PR that will migrate all non-braced usages to braced usage, but I don't want to do that until this PR is merged, and some time has passed for us to be sure folks have moved to latest Verus.

jaybosamiya-ms:
@parno this will likely need a verus-analyzer update too

source/rust_verify/example/syntax.rs
```diff
@@ -590,4 +590,15 @@ proof fn uses_spec_has(s: Set<int>, ms: vstd::multiset::Multiset<int>)
     assert(ms has 4 == ms has 4);
 }
 
+proof fn broadcast_use() {
+    broadcast use vstd::seq_lib::group_seq_properties;
+    // you can use broadcase use on the module level, in proof functions,
+    // in proof blocks, or in assert-by
+    assert forall|s: Seq<usize>, v: usize, x: usize|
+        { s.contains(x) ==> s.push(v).contains(x) } by {
+        broadcast use vstd::seq_lib::group_seq_properties;
```
utaal, line 599: Should there be an example with braces?

ahuoguo:
Also, Just want to note that I don't parse `broacast use` by `UseTree` parser, which means we can't write something like:
```rust
broadcast use vstd::multiset::{group_multiset_properties, group_multiset_axioms}
```

I can only be written as
```rust
broadcast use {vstd::multiset::group_multiset_properties, vstd::multiset::group_multiset_axioms};
```

I think `verusfmt` follows the same restriction

parno:
I've updated verus-analyzer for this as well.

source/builtin_macros/src/syntax.rs
```diff
@@ -1319,6 +1319,16 @@ impl Visitor {
                     let span = item.span();
                     let paths = &item_broadcast_use.paths;
                     if self.erase_ghost.erase() {
+                        if item_broadcast_use.warning {
+                            #[cfg(verus_keep_ghost)]
+                            proc_macro::Diagnostic::spanned(
+                                span.unwrap(),
+                                proc_macro::Level::Warning,
+                                "Outdated syntax for broadcast use.\n\
+                                         Use semicolon for multiple uses.",
```
utaal, line 1328: Did you mean curly braces?

utaal (approved):

PR #1587, opened 2025-04-10
Subset Lengths

HikaruHokkyokusei:
Added lemma to show that if `x` is subset of `y`, then length of `x` is smaller or equal to that of `y`.


tjhance:
lemma looks fine to me, but it looks like there's a typo and the CI is failing

HikaruHokkyokusei:
Interesting 😅
It says cannot prove termination, but it works outside the library...

image

HikaruHokkyokusei:
Already exists... `lemma_len_subset`

PR #1595, opened 2025-04-16
print function mode in expanded time json

ahuoguo:
We did not indicate the mode of the function in the `--time-expanded --output-json` before

### DIFF
--- a/source/rust_verify/src/main.rs
+++ b/source/rust_verify/src/main.rs
@@ -349,6 +349,7 @@ pub fn main() {
                                 "function-breakdown" : smt_function_breakdown.get_mut(*m).map(|b| b.iter().map(|(f, t)| {
                                     serde_json::json!({
                                         "function" : vir::ast_util::fun_as_friendly_rust_name(f),
+                                        "mode:" : verifier.func_modes.get(f).map(|m| m.to_string()).unwrap_or_default(),
                                         "time" : t.time_millis,
                                         "rlimit-count" : t.rlimit_count,
                                         "success" : !verifier.func_fails.contains(f),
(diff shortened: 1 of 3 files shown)


source/rust_verify/src/main.rs
```diff
@@ -349,6 +349,7 @@ pub fn main() {
                                 "function-breakdown" : smt_function_breakdown.get_mut(*m).map(|b| b.iter().map(|(f, t)| {
                                     serde_json::json!({
                                         "function" : vir::ast_util::fun_as_friendly_rust_name(f),
+                                        "mode:" : verifier.func_modes.get(f).map(|m| m.to_string()).unwrap_or_default(),
```
utaal, line 352: We may want to not unwrap here, so it produces `null` if the mode is unknown.

utaal: There's another caveat: a spec fn can have verification time if it's recursive and thus requires a termination check.

ahuoguo: Yes, I'm not ignoring the spec functions verification time, and trivial proof/exec functions might also have verification time 0.

source/rust_verify/src/main.rs
```diff
@@ -349,6 +349,9 @@ pub fn main() {
                                 "function-breakdown" : smt_function_breakdown.get_mut(*m).map(|b| b.iter().map(|(f, t)| {
                                     serde_json::json!({
                                         "function" : vir::ast_util::fun_as_friendly_rust_name(f),
+                                        "mode:" : 
+                                            verifier.get_function_mode(f).map(|m| m.to_string())
+                                            .unwrap_or("null".to_string()),
```
utaal: I think you can just leave the option here, it will turn into null if it's `None`

PR #1599, opened 2025-04-18
Minor clarifying edits to calc section of guide

jaylorch:


parno (approved):

jaybosamiya-ms (approved):
Thanks, LGTM.  I somewhat want to keep the older subsection title just because it is _mildly_ silly but also agree that updating it does make it clearer.

jaylorch:
Yeah, I found the old section title amusing, but, alas, it seemed unhelpful in summarizing what the reader should be expected to learn. At least we can still have the first section title, which is amusing and not confusing. :-)

Issue #1554, opened 2025-04-01
`cargo verus verify` crashes with "cannot access a scoped thread local variable without calling `set` first"


zeldovich:
Consider the following Verus program:

```
use vstd::prelude::*;
use vstd::invariant::*;

verus! {
    struct P {}

    impl InvariantPredicate<P, u8> for P {
        closed spec fn inv(k: P, v: u8) -> bool {
            v == 0
        }
    }

    fn f(tracked inv: AtomicInvariant<P, u8, P>, b: bool) {
        open_atomic_invariant!(&inv => inner => {
            proof {
                assert(b);
                inner = 1u8;
            }
        });
    }

    fn main() {}
}
```

running `verus` on it reports some problems, as expected:

```
note: function body check: not all errors may have been reported; rerun with a higher value for --multiple-errors to find other potential errors in this function
  --> src/main.rs:13:5
   |
13 |     fn f(tracked inv: AtomicInvariant<P, u8, P>, b: bool) {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: assertion failed
  --> src/main.rs:16:24
   |
16 |                 assert(b);
   |                        ^ assertion failed

error: Cannot show invariant holds at end of block
  --> src/main.rs:14:9
   |
14 | /         open_atomic_invariant!(&inv => inner => {
15 | |             proof {
16 | |                 assert(b);
17 | |                 inner = 1u8;
18 | |             }
19 | |         });
   | |__________^
   |
   = note: this error originates in the macro `$crate::vstd::invariant::open_atomic_invariant_internal` which comes from the expansion of the macro `open_atomic_invariant` (in Nightly builds, run with -Z macro-backtrace for more info)

verification results:: 1 verified, 1 errors
error: aborting due to 2 previous errors
```

but running `cargo verus verify` (with a default Cargo setup generated by `cargo verus new --bin`) causes a crash in some part of the `cargo verus` tooling, as far as I can tell:

```
   Compiling foo v0.1.0 (/home/nickolai/tmp/foo)
note: function body check: not all errors may have been reported; rerun with a higher value for --multiple-errors to find other potential errors in this function
  --> src/main.rs:13:5
   |
13 |     fn f(tracked inv: AtomicInvariant<P, u8, P>, b: bool) {
[...]

PR #1610, opened 2025-04-27
When sorting messages, call from_raw_span from original rustc thread

Chris-Hawblitzel:
The function `from_raw_span` calls the `.span()` function on rustc's `SpanData`, which calls rustc's `Span::new` function, which in some rare circumstances uses thread-local storage.  Because of this, we should only call `from_raw_span` from the original rustc thread, not from a Verus worker thread.  It turns out we were calling `from_raw_span` from the worker threads in order to sort messages by span.  This pull request changes this so that sorting happens in the original rustc thread.

This fixes https://github.com/verus-lang/verus/issues/1554 .


tjhance (approved):
Is it possible to add a check in `from_raw_span` that we're on the right thread?

PR #1623, opened 2025-05-03
Do not wrap in proof block if stmt is macro inside proof_decl macro

ziqiaozhou:
While verus currently does not support real inline proof function (#1619 ), we can use macro to define inline proofs. This PR enables proper use of macro statement inside proof_decl!{} and thus we can reduce the annotations by using macros inside executable function if a long proof block is required.

### DIFF
--- a/source/rust_verify_test/tests/syntax_attr.rs
+++ b/source/rust_verify_test/tests/syntax_attr.rs
@@ -670,3 +670,25 @@ test_verify_one_file! {
 
     } => Err(e) => assert_rust_error_msg(e, "method `spec_f` is not a member of trait `X`")
 }
+
+test_verify_one_file! {
+    #[test] test_macro_inside_proof_decl code!{
+        macro_rules! inline_proof {
+            ($val: expr => $x: ident, $y: ident) => {
+                proof_decl!{
+                    let tracked $x: int = $val;
+                    let ghost $y: int = $val;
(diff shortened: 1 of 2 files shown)


source/rust_verify_test/tests/syntax_attr.rs
```diff
@@ -670,3 +670,25 @@ test_verify_one_file! {
...
+                    let tracked $x: int = $val;
+                    let ghost $y: int = $val;
+                    assert($x == $y);
+                }
+             }
+        }
+
+        fn test_inline_proof_via_macro() {
+            proof_decl!{
+                inline_proof!(0 => x, y);
```
Chris-Hawblitzel, line 688: Here, `inline_proof!` expands to a `proof_decl!`, but this already inside another `proof_decl!`.  Is it intended that you need to nest `proof_decl!`s inside each other like this?

ziqiaozhou: It is intended. I want to use macro inside `proof_decl!{}` so that it is clear that the new macro is a proof instead of an arbitrary macro used in execution. Unfortunately, `proof_decl!` does not see content expanded from `inline_proof!`, so I need to explicitly call `proof_decl!` or `proof!` in macro definition.

Chris-Hawblitzel (approved):

PR #1625, opened 2025-05-05
Add spec for pointer add

ziqiaozhou:

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -473,6 +473,18 @@ macro_rules! pointer_specs {
                 opens_invariants none
                 no_unwind;
 
+            pub open spec fn spec_add<T: Sized>(p: *$mu T, offset: usize) -> *$mu T {
+                $ptr_from_data(PtrData { addr: (p@.addr + offset * size_of::<T>()) as usize, .. p@ })
+            }
+
+            #[verifier::when_used_as_spec(spec_add)]
+            pub assume_specification<T: Sized>[<*$mu T>::add](p: *$mu T, offset: usize) -> (q: *$mu T)
+                requires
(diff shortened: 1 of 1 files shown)


tjhance:
@elanortang has been working on a spec for ptr add; the necessary precondition is much more elaborate. See the safety requirements here: https://doc.rust-lang.org/std/primitive.pointer.html#safety-7

ziqiaozhou:
> elaborate

I thought the Rust standard `add` is unsafe but memory access via raw pointer is finally guarded by PointsTo<T> in verus. If that is the case, whether the new pointer is an allocated and valid one only needs to be checked later.

Without passing additional parameters, how could verus check all safety requirements "the entire memory range between self and the result must be in bounds of that allocated object" from https://doc.rust-lang.org/std/primitive.pointer.html#safety-7?

tjhance:
ptr::add can be UB on its own, even if it's never dereferenced.

> Without passing additional parameters, how could verus check all safety requirements 

I think we'll eventually want to support the with_tracked syntax on assume_specification.

In the meantime, if you want to do pointer arithmetic, you can always cast the pointer to `usize` and do arithmetic with that.

ziqiaozhou:
> ptr::add can be UB on its own, even if it's never dereferenced.
> 
> > Without passing additional parameters, how could verus check all safety requirements
> 
> I think we'll eventually want to support the with_tracked syntax on assume_specification.
> 
> In the meantime, if you want to do pointer arithmetic, you can always cast the pointer to `usize` and do arithmetic with that.

I see. I did have casted the pointer to usize and do arithmetic with that before. But I feel it might be better to keep implementation code simple by using Rust standard methods so that the code with proof could look more similar to code without proof. Maybe I can create another v_add wrapper in my project to do the thing I want.

Issue #1633, opened 2025-05-09
Example in docs for `Map::values()` doesn't verify


pratapsingh1729:
The [docs for `vstd::map::Map::values()`](https://verus-lang.github.io/verus/verusdoc/vstd/map/struct.Map.html#method.values) contain the following example:

```
assert(
   map![1 => 10, 2 => 11].values() =~= set![10, 11]
);
```

However, this doesn't verify currently in Verus, as illustrated by this [playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=f2fe708accf4976f5ddb817d949d842e).

As I understand it, the basic issue is that there's no axiom or lemma analogous to [axiom_map_insert_domain](https://verus-lang.github.io/verus/verusdoc/vstd/map/fn.axiom_map_insert_domain.html) for `values()`. It's a bit subtle, however, since in general, `m.insert(key, value).values() != m.values().insert(value),`: consider the case of `map!['a' => 1, 'b' => 2].insert('a', 3)`. Then we have `m.insert(key, value).values() = set![2,3]` but `m.values().insert(value) = set![1,2,3]`. I think we should be able to have this lemma with a precondition that `!m.dom().contains(key)`, in which case we would never be overwriting an existing element of `m.values()` by inserting. Not sure if this would be enough to make the example go through---it depends how the `map!` macro is expanded.

At the very least, though, I imagine we should probably get rid of the example in the docs if it doesn't actually work!

jaylorch:
I agree we should get rid of the example in the docs, or, better yet, fix it by including the proof annotations needed to make it work.

I don't think much more can be done about it, for the reasons you describe. It's hard to write a simple axiom or lemma analogous to `axiom_map_insert_domain`, as you point out.

PR #1634, opened 2025-05-09
Add specs for `hash_map::Values` and `HashMap::values()`

pratapsingh1729:
I need to use the [`Values` iterator](https://doc.rust-lang.org/std/collections/hash_map/struct.Values.html) from `std::collections::HashMap` for some verified code, so I implemented a spec for it following the one for `Keys`. I also wrote a spec for `HashMap::values()`.

Note that unlike with `Keys`, `Values` can contain the same value more than once, since `HashMap`s can have multiple keys point to the same value. 

I added a simple test based on the one for `Keys` as well. However due to #1633 we can't actually assert the concrete contents of the `values()` set, so I didn't include that in the test. I can update the test if #1633 is supported.

### DIFF
--- a/source/rust_verify_test/tests/hash.rs
+++ b/source/rust_verify_test/tests/hash.rs
@@ -813,10 +813,14 @@ test_verify_one_file_with_options! {
 
             m.insert(3, 4);
             m.insert(6, -8);
+            assert(m@.values() == set![4i8, -8i8]) by {
+                assert(m@.contains_key(3u32));
+                assert(m@.contains_key(6u32));
+                assert(m@.values() =~= set![4i8, -8i8]);
+            };
             let m_values = m.values();
             assert(m_values@.0 == 0);
-            // assert(m_values@.1.to_set() == set![4i8, -8i8]); // FAILS due to https://github.com/verus-lang/verus/issues/1633
-            assert(m_values@.1.to_set() == m@.values());
+            assert(m_values@.1.to_set() == set![4i8, -8i8]);
             let ghost g_values = m_values@.1;
 
             let mut items = Vec::<i8>::new();
@@ -825,13 +829,13 @@ test_verify_one_file_with_options! {
             for v in iter: m_values
                 invariant
                     iter.values == g_values,
-                    g_values.to_set() =~= m@.values(),
+                    g_values.to_set() =~= set![4i8, -8i8],
                     items@ == iter@,
             {
                 assert(iter.values.take(iter.pos).push(*v) =~= iter.values.take(iter.pos + 1));
                 items.push(*v);
             }
-            assert(items@.to_set() =~= m@.values()) by {
+            assert(items@.to_set() =~= set![4i8, -8i8]) by {
                 assert(g_values.take(g_values.len() as int) =~= g_values);
             }
         }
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/hash.rs
```diff
@@ -619,6 +728,17 @@ pub assume_specification<'a, Key, Value, S>[ HashMap::<Key, Value, S>::keys ](
...
 ;
 
+pub assume_specification<'a, Key, Value, S>[ HashMap::<Key, Value, S>::values ](
+    m: &'a HashMap<Key, Value, S>,
+) -> (values: Values<'a, Key, Value>)
+    ensures
+        obeys_key_model::<Key>() && builds_valid_hashers::<S>() ==> {
+            let (index, s) = values@;
+            &&& index == 0
+            &&& s.to_set() == m@.values()
```
Chris-Hawblitzel, line 738: I'm wondering if this could be strengthened in the case where there are duplicate values.  Do we also know that `s` and `m@` produce the same multiset?

jaylorch: The official documentation seems vague on whether it will definitely report each value exactly the number of times it appears. So I think it's useful to reflect that ambiguity in the spec, as shown here.

jaylorch: Even if the current implementation has that property, there's no guarantee a later version won't violate it.

source/vstd/std_specs/hash.rs
```diff
@@ -346,6 +346,115 @@ impl<'a, Key, Value> View for KeysGhostIterator<'a, Key, Value> {
...
+) -> (r: Option<&'a Value>)
+    ensures
+        ({
+            let (old_index, old_seq) = old(values)@;
+            match r {
+                None => {
+                    &&& values@ == old(values)@
+                    &&& old_index >= old_seq.len()
+                },
+                Some(k) => {
```
jaylorch, line 376: I suggest naming this variable `v` instead of `k` to clarify that it's a value.

jaylorch:
Thanks for this! It looks good to me, except for that minor comment above.

source/vstd/std_specs/hash.rs
pratapsingh1729: Ah, yes, my mistake---I forgot to change that. Thanks!

jaylorch:
I created a PR to fix #1633. Does that help you update the test the way you wanted?

jaylorch (approved):

source/rust_verify_test/tests/hash.rs
```diff
@@ -825,13 +829,13 @@ test_verify_one_file_with_options! {
             for v in iter: m_values
                 invariant
                     iter.values == g_values,
-                    g_values.to_set() =~= m@.values(),
+                    g_values.to_set() =~= set![4i8, -8i8],
```
jaylorch, line 832: This can just use `==`. Extensional equality isn't necessary.

Chris-Hawblitzel (approved):

source/rust_verify_test/tests/hash.rs
pratapsingh1729: Fixed!

PR #1637, opened 2025-05-12
Added `is_empty` specifications for HashSets and HashMaps

HikaruHokkyokusei:

### DIFF
--- a/source/vstd/std_specs/hash.rs
+++ b/source/vstd/std_specs/hash.rs
@@ -503,9 +503,9 @@ pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::len ](
 
 pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::is_empty ](
     m: &HashMap<Key, Value, S>,
-) -> (is_empty: bool)
+) -> (res: bool)
     ensures
-        is_empty == m@.is_empty(),
+        res == m@.is_empty(),
 ;
 
 pub assume_specification<Key, Value>[ HashMap::<Key, Value>::new ]() -> (m: HashMap<
@@ -874,11 +874,9 @@ pub assume_specification<Key, S>[ HashSet::<Key, S>::len ](m: &HashSet<Key, S>) 
         len == spec_hash_set_len(m),
 ;
 
-pub assume_specification<Key, S>[ HashSet::<Key, S>::is_empty ](
-    m: &HashSet<Key, S>,
-) -> (is_empty: bool)
+pub assume_specification<Key, S>[ HashSet::<Key, S>::is_empty ](m: &HashSet<Key, S>) -> (res: bool)
     ensures
-        is_empty == m@.is_empty(),
+        res == m@.is_empty(),
 ;
 
 pub assume_specification<Key>[ HashSet::<Key>::new ]() -> (m: HashSet<Key, RandomState>)
--- a/source/vstd/std_specs/hash.rs
+++ b/source/vstd/std_specs/hash.rs
@@ -501,6 +501,13 @@ pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::len ](
         len == spec_hash_map_len(m),
 ;
 
+pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::is_empty ](
+    m: &HashMap<Key, Value, S>,
+) -> (is_empty: bool)
+    ensures
+        is_empty == m@.is_empty(),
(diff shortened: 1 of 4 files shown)


HikaruHokkyokusei:
I have added the specification in the relevant place. Let me know if I need to make more changes related to this, or if the current changes are enough.

utaal:
@jaylorch can you review this?

HikaruHokkyokusei:
I'll do that.

HikaruHokkyokusei:
@jaylorch I have added the specification in the relevant places.

`assume_specification` in `vstd/std_specs/hash.rs`
`is_empty` specification for HashSetWithView and HashMapWithView
`is_empty` specification for StringHashSet and StringHashMap

Finally, I wanted to ask if we should add something similar to `#[verifier::when_used_as_spec(spec_len)]` that we have for the `len` function.

I feel like it is a valid scenario for the `is_empty` function. It would be great if you could confirm this.
Also, I'm not entirely sure how to do it, but I can try to replicate what has already been done for the `len` function, and you can review it for errors. If not, you can add a commit to this patch yourself for this change.

jaylorch:
Thanks! Could you also please add a specification for `HashMap::is_empty` in `vstd/std_specs/hash.rs`?

As for your question about `verifier::when_used_as_spec`, there's no need here for the complexity of that approach. The reason it's useful for `len` is that we're encoding the assumption that `len` will return a `usize` and never, say, 2^1000. There's no such concern with `is_empty` because an executable Boolean has the same range (`{true, false}`) as a spec Boolean.

HikaruHokkyokusei:
> Thanks! Could you also please add a specification for `HashMap::is_empty` in `vstd/std_specs/hash.rs`?

Sure, let me look into it.

Will also see how to fix the fmt... I edit the code in GitHub Codespaces. Is there a quick shortcut to automatically fix the formatting?

jaylorch:
Regarding formatting, I asked @jaybosamiya-ms and he thinks you should be able to just install [`verusfmt`](https://github.com/verus-lang/verusfmt/?tab=readme-ov-file#installing-and-using-verusfmt) and then run `vargo fmt`.

HikaruHokkyokusei:
Thanks! I also figured out the formatting error.

HikaruHokkyokusei:
@jaylorch Done. Can you review again?

HikaruHokkyokusei:
I did vargo fmt this time. 😅

source/vstd/std_specs/hash.rs
```diff
@@ -501,6 +501,13 @@ pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::len ](
         len == spec_hash_map_len(m),
 ;
 
+pub assume_specification<Key, Value, S>[ HashMap::<Key, Value, S>::is_empty ](
+    m: &HashMap<Key, Value, S>,
+) -> (res: bool)
+    ensures
+        is_empty == m@.is_empty(),
```
jaylorch, line 508: This should be `res` instead of `is_empty`.

source/vstd/std_specs/hash.rs
```diff
@@ -867,6 +874,11 @@ pub assume_specification<Key, S>[ HashSet::<Key, S>::len ](m: &HashSet<Key, S>)
         len == spec_hash_set_len(m),
 ;
 
+pub assume_specification<Key, S>[ HashSet::<Key, S>::is_empty ](m: &HashSet<Key, S>) -> (res: bool)
+    ensures
+        is_empty == m@.is_empty(),
```
jaylorch, line 879: This should be `res` instead of `is_empty`.

jaylorch:
Thanks! I found a couple of result-value-naming bugs that were preventing me from building and testing.

I also suggest you add some tests of your new code. You can probably just pack those into existing tests, e.g., in `source/rust_verify_test/tests/hash.rs`, in `test_hash_map` you could add `let b = m.is_empty()); assert(b);` after `assert(m@ == Map::<u32, i8>::empty());`, and after `m.insert(3, 4);` you could insert `b = m.is_empty(); assert(!b);`. And so on for other tests of `HashMapWithView`, etc. that are in that tests file.

HikaruHokkyokusei:
I added relevant test cases. Ran `vargo test -p rust_verify_test --test hash
` on my local machine, and all 18 tests were passing successfully.

jaylorch:
Alright, looks good to me. Thanks! I'm running a `ci` test now to do a final check before approving.

jaylorch (approved):

PR #1638, opened 2025-05-12
Broadened the post-condition of `get` for HashMap

HikaruHokkyokusei:

### DIFF
--- a/source/vstd/hash_map.rs
+++ b/source/vstd/hash_map.rs
@@ -99,6 +99,7 @@ impl<Key, Value> HashMapWithView<Key, Value> where Key: View + Eq + Hash {
                 Some(v) => self@.contains_key(k@) && *v == self@[k@],
                 None => !self@.contains_key(k@),
             },
+            result.is_some() == self@.contains_key(k@),
     {
         self.m.get(k)
     }


utaal:
@jaylorch can you review this?

jaylorch:
Hi, @HikaruHokkyokusei. I don't see how this postcondition can help. Can you give an example of a verification that works with this modification and doesn't work without it?

jaylorch:
I ask because your postcondition can be directly inferred from the postcondition that's already there.

HikaruHokkyokusei:
Initially, I also thought the same. We have the existing post-condition, yet verus was giving error. 

This issue is, suppose a => b is true. 
Then, if a is true, then b will be true. What is b is true? We cannot argue about b. 

But in our case, we can. Look at this gist. Try it without the assume statement in the proof. 

https://gist.github.com/verus-play/bd27a0141c2452a5f3a802255531092c

jaylorch:
The example you gave in `https://gist.github.com/verus-play/bd27a0141c2452a5f3a802255531092c` uses `HashMap`, but this PR is about `HashMapWithView`.

jaylorch:
The reason your example in `https://gist.github.com/verus-play/bd27a0141c2452a5f3a802255531092c` needs the `assume` is that the postcondition on `HashMap::get` is predicated on `obeys_key_model::<Key>() && builds_valid_hashers::<S>()` being true, and the verifier can't establish that. You could replace the `assume` statement with `broadcast use group_hash_axioms;`.

HikaruHokkyokusei:
Ahhh, I see. My bad. I didn't realise the difference between HashMap and HashMapView, and that I had to use the axiom.

Then this PR is no longer needed. We can close this, I guess.

jaylorch:
OK, will do! We should probably make `group_hash_axiom` get broadcast by default, to avoid issues like this in the future.

PR #1646, opened 2025-05-16
change --verify-module to also verify submodules, and add --verify-only-module to verify a specific module (without submodules)

utaal:
Implements #385

Moves the command line validation logic to `config.rs` from `user_filter.rs`.


parno (approved):
Thanks for the quick fix!  This addresses the issue with VS Code and the submodules created by the `tokenized_state_machine` macro.

PR #1647, opened 2025-05-16
Remove reference to `rust-verify.sh` in guide

cjohnson19:
`rust-verify.sh` warns of deprecations and upcoming removal. This commit changes the reference to this script into the newly recommended method of using the `verus` executable instead.


parno:
Good catch -- thanks!

PR #1651, opened 2025-05-16
Add View and DeepView for Iterator

txiang61:


dschoepe:
Looks like we need to put this behind `#[feature(alloc)]` to prevent the errors for the `no_alloc` build.

PR #1652, opened 2025-05-16
Fix issue #1627: Verus crashing on an external impl Trait when that trait has an associated type

### LINKED ISSUE #1627: Verus crashes on `impl IntoIterator for Foo` outside `verus!` macro
I'm trying to verify small portions of a large Rust codebase using Verus without making invasive changes. For this, I'm putting the relevant parts of the existing code inside the `verus!` macro, but leaving some functionality that's irrelevant to the code being verified outside of it. 

However this led me to the following Verus panic. Suppose I have a struct `Foo` (which I need for verification) and an `impl IntoIterator for Foo` (which I don't need for verification). I put `verus!` around just the struct definition:
```
use vstd::prelude::*;

verus! {

pub struct Foo {
    pub vals: Vec<u32>
}

} // verus!

impl IntoIterator for Foo {
    type Item = u32;
    
    type IntoIter = std::vec::IntoIter<u32>;
    
    fn into_iter(self) -> Self::IntoIter {
        self.vals.into_iter()
    }
}
```
This then crashes with the following message:
```
thread 'rustc' panicked at rust_verify/src/rust_to_vir_impl.rs:608:38:
no entry found for key
```
[Playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=694e2ba9a9f7935ada59cac4859d4128)

To help with debugging, this error doesn't arise if the trait is `Clone` instead of IntoIterator, or if it's a user-defined trait also outside of `verus!`.

pratapsingh1729:
Fixes #1627 (Verus crashing on an external `impl Trait` when that trait has an associated type).

@Chris-Hawblitzel or whoever wrote the trait logic should definitely confirm this, but below is why I think this fix is sound.

In [`rust_to_vir_impl.rs:608`](https://github.com/verus-lang/verus/blob/64f79d63a9e5f2639f181d1149764caee88ac402/source/rust_verify/src/rust_to_vir_impl.rs#L608), we look up `trait_map[&trait_path]`. Here, `trait_map` contains all the traits in the current crate, while `trait_path` is iterating through `auto_import_impls`, which contains all `impl`s that are either for "new" traits (traits from the current crate), or external `impl`s from the current crate. Then, the `for` loop starting on [line 576](https://github.com/verus-lang/verus/blob/64f79d63a9e5f2639f181d1149764caee88ac402/source/rust_verify/src/rust_to_vir_impl.rs#L576) iterates through all of the `auto_import_impls` (the `impl`s that might be visible to Verus), looking up their traits in the `trait_map`. 

The error case in #1627 comes from an external `impl` for an external trait, so the external trait doesn't appear in `trait_map`. But since it is an external `impl` for an external trait, Verus shouldn't be able to see it, so I think it shouldn't need to check whether there is an associated type match for it in the list of local traits.

If my reasoning is incorrect, please let me know---I'd be happy to implement a different fix!


pratapsingh1729:
@Chris-Hawblitzel updated to include your requested changes.

PR #1653, opened 2025-05-17
Add proof blocks for recursive spec function termination checking

Chris-Hawblitzel:
This allows proof blocks inside spec code.  Currently, this is only supported for termination checking for recursive spec functions, since recommends checking is not ready for general proofs yet.

Example:

```
        spec fn f(i: u64) -> int decreases i {
            if i == 0 {
                0
            } else {
                proof {
                    assert(i >> 1 < i) by(bit_vector)
                        requires i != 0;
                }
                f(i >> 1)
            }
        }
```

@jonhnet There are minor changes in set_lib.rs for this.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -2823,8 +2823,8 @@ impl Visitor {
             }
         } else if let Expr::Unary(unary) = expr {
             let span = unary.span();
-            match (is_inside_ghost, mode_block, &*unary.expr) {
-                (false, (false, _), Expr::Block(..)) => {
+            match (mode_block, &*unary.expr) {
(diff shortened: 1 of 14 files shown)


parno:
Does this subsume `via` and `via_fn`?  If so, should we mark them as deprecated and/or update the Guide?

Chris-Hawblitzel:
> Does this subsume `via` and `via_fn`? If so, should we mark them as deprecated and/or update the Guide?

`via`/`via_fn` allows you to write trusted recursive specifications without having to put the termination proofs in the specifications, so I would not deprecate that.  But 99% of the time, proof blocks are easier and the guide should emphasize proof blocks.

utaal:
I'm going to land this to prevent it from diverging (as it should have no impact on existing code).

PR #1656, opened 2025-05-17
update setup instructions

tjhance:
While helping a new student get set up, I realized there are currently no step-by-step instructions for getting started with verus-analyzer. This adds a verus-analyzer page to the 'Getting Started' section.

I also realized the existing instructions were a bit out-of-date; they redirect to the INSTALL.md, which uses the binary install; but the rest of the instructions were written assuming a source install. I updated the instructions appropriately.

Regarding the verus-analyzer instructions, I had a few questions while writing them, so if someone could help me out that would be great:

1. Is it actually necessary to install Verus manually in order to use verus-analyzer? If so, how does verus-analyzer know where to find the verus binary? 
2. When I tried running my own instructions, I got a bunch of errors that are (I presume) from rust-analyzer, saying it can't find `vstd`. (Despite the fact that verus-analyzer worked fine.) What's the recommended way to fix this?


utaal (approved):

parno:
Thanks for making these updates.  It seems worth including something for the Emacs mode too, and probably a pointer to the "IDE Support" portion of the guide  (or maybe we should remove that in favor of consolidating in the Getting Started section?).

To answer your questions:
1. No, when you install the extension, it automatically grabs the latest Verus release from GitHub.  There's also a setting you can configure if you want to point it at your own version of Verus.
2. You need to disable `rust-analyzer` (either globally or for your workspace).  It's an unfortunate consequence of our decision to use the same `.rs` file extension as Rust.

tjhance:
> You need to disable rust-analyzer (either globally or for your workspace). It's an unfortunate consequence of our decision to use the same .rs file extension as Rust.

I haven't been able to figure out how to get this to work. I went to the extensions, rust-analyzer, and clicked 'Disable (workspace)'. But vscode is still putting red underline squiggles around the vstd import and verus macro.

parno:
> > You need to disable rust-analyzer (either globally or for your workspace). It's an unfortunate consequence of our decision to use the same .rs file extension as Rust.
> 
> I haven't been able to figure out how to get this to work. I went to the extensions, rust-analyzer, and clicked 'Disable (workspace)'. But vscode is still putting red underline squiggles around the vstd import and verus macro.

After you click disable, I think you also have to click the blue "Restart extensions" button that pops up.

PR #1658, opened 2025-05-20
add `Seq::update_range()` to update subrange of a sequence

zeldovich:
This construct appears frequently when dealing with storage systems that modify parts of a storage device, but I expect this would be convenient to have in vstd for other uses too.

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -1034,6 +1034,25 @@ impl<A> Seq<A> {
         broadcast use super::set_lib::group_set_properties;
 
     }
+
+    /// Update a subrange of a sequence starting at `off` to values `vs`.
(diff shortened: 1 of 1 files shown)


parno:
[Verus Retreat 2025] -- In the function name, @utaal says "replace" sounds better than "update" and @tjhance says that "range" makes him expect it will take a range argument.

zeldovich:
Thanks for the feedback.  `replace` seems fine, though `update` mirrors the existing `Seq::update()` verb (maybe that should also be `Seq::replace()`?).  Would `replace_subseq` or `replace_subrange` be better?

utaal:
I take my suggestion back due to consistency with `update`: I'd be happiest with `update_subrange_with`, but `update_subrange` is also okay. @tjhance ?

tjhance:
lgtm

PR #1667, opened 2025-05-21
When using cargo-verus, add proc_macro_span and span_locations to proc_macro2

Chris-Hawblitzel:
Whereas vargo was adding `proc_macro_span` and `span_locations` to the rustc flags when building Verus's dependencies (particularly `proc_macro2`), cargo-verus was not.  This caused cargo-verus to produce inferior spans when compared to vargo-compiled Verus.  This pull request enables `proc_macro_span` and `span_locations` for `proc_macro2` when it is built via cargo-verus.

I still think it's an open question whether cargo-verus should build Verus's dependencies at all, rather than using the libraries already built by vargo.  But as long as it is building them, it's better to pass the right flags.

### DIFF
--- a/source/rust_verify/src/cargo_verus.rs
+++ b/source/rust_verify/src/cargo_verus.rs
@@ -55,6 +55,20 @@ pub fn extend_args_and_check_is_direct_rustc_call(
         false
     };
     if !verus_crate {
+        let mut is_span_crate = false;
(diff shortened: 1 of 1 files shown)


parno (approved):
Thanks for tracking that down!  It does fix the span in my test case.

Issue #1657, opened 2025-05-19
assume_specification fails for blanket impl with ?Sized bound


tjhance:
I came across this attempting to resolve the ToString issue in #1385.

This is expected to work:

```
pub assume_specification<T: Display + ?Sized>[ <T as ToString>::to_string ](t: &T) -> (res: String)
```

However, this currently results in the incorrect error: "assume_specification cannot be used to specify generic specifications of trait methods; consider using external_trait_specification instead"

Here's a minimal repro:

```
trait Tr {
    fn stuff(&self);
}

#[verifier::external]
trait Blanket {
    fn stuff2(&self);
}

#[verifier::external]
impl<T: Tr> Blanket for T {
    fn stuff2(&self) {
        self.stuff();
    }   
}

assume_specification <T: Tr> [ <T as Blanket>::stuff2 ] (x: &T)
    ensures true;
```

For reasons that are unclear to me, the `?Sized` bound seems to be required for the bug. Without it, the blanket impl works fine.

tjhance:
This problem isn't limited to assume_specification. Consider this:

```
trait Tr {
    fn stuff(&self);
}

trait Blanket {
    fn stuff2(&self);
}

impl<T: Tr + ?Sized> Blanket for T { 
    fn stuff2(&self)
        ensures false
    {   
        assume(false);
    }   
}

fn test<T: Tr + ?Sized>(t: &T) {
    t.stuff2();
    assert(false);
}
```

vs this (without the ?Sized):

```
trait Tr {
    fn stuff(&self);
}

trait Blanket {
    fn stuff2(&self);
}

impl<T: Tr> Blanket for T {
    fn stuff2(&self)
        ensures false
    {   
        assume(false);
    }   
}

fn test<T: Tr>(t: &T) {
    t.stuff2();
    assert(false);
}
```

In the first case, the `assert(false)` fails, which illustrates that we aren't able to statically resolve the call to `stuff2`.

tjhance:
fundamentally the problem is that `rustc_middle::ty::Instance::try_resolve` is returning `Ok(None)` when I would expect it to return the instance corresponding to the blanket impl.

tjhance:
It seems likely that the observed behavior is due to this in the implementation of `try_resolve`:

https://doc.rust-lang.org/1.82.0/nightly-rustc/src/rustc_ty_utils/instance.rs.html#167

PR #1677, opened 2025-05-23
Fix trait issue and to_string

tjhance:
as noted in #1657, the rustc's `try_resolve` function has some behaviors we don't expect. This PR removes uses of `try_resolve`. This allows us to fix the `to_string` issue.


dschoepe (approved):
Thanks for figuring this out!

PR #1678, opened 2025-05-26
Spec-to-exec compiler macro `exec_spec!`

zhengyao-lin:
As discussed [here](https://github.com/verus-lang/verus/discussions/1429), I ported a proc macro called `exec_spec!` into Verus from another project.

Basically, it is a compiler from `struct`/`enum`/`spec fn`'s to their verified, executable versions.

The following example shows some of the spec features it supports:
```rust
use vstd::prelude::*;
use vstd::exec_spec::*;
use exec_spec::exec_spec;

verus! {

exec_spec! {
    struct MyPair(Seq<u32>, Seq<u32>);

    spec fn pred_helper(a: Seq<u32>, b: Seq<u32>, i: usize) -> bool
        recommends 0 <= i < a.len() == b.len()
        decreases a.len() - i when i < usize::MAX
    {
        &&& a[i as int] > b[i as int]
        &&& i + 1 < a.len() ==> pred_helper(a, b, (i + 1) as usize)
    }

    // Checks element-wise if p.0 > p.1
    spec fn pred(p: MyPair) -> bool
        recommends p.0.len() == p.1.len()
    {
        if p.0.len() == 0 {
            true
        } else {
            pred_helper(p.0, p.1, 0)
        }
    }
}

fn test() {
    let a = ExecMyPair(vec![1, 2], vec![0, 1]);
    
    if exec_pred(&a) {
        assert(a.0@[0] > a.1@[0]);
        assert(a.0@[1] > a.1@[1]);
    }
}

}
```
In this example, the `exec_spec!` macro generates three new items: `ExecMyPair`, `exec_pred_helper`, and `exec_pred`, with the following impls and specs:
  - `ExecMyPair` implements `DeepView<V = MyPair>`
  - `exec_pred_helper` has the following verified spec (simplified):
     ```rust
      exec_pred_helper(a: &[u32], b: &[u32], i: usize) -> (res: bool)
          requires 0 <= i < a.deep_view().len() == b.deep_view().len(), // converted from `recommends`
                   i < usize::MAX, // converted from the `when` clause
          ensures res == pred_helper(a.deep_view(), b.deep_view(), i) // equivalent to pred_helper
          decreases a.deep_view().len() - i
     ```
  - Similar specs for `exec_pred`

Besides adding these new exec counterparts, `exec_spec!` keeps all items passed to it unmodified.

Currently it supports these features:
  - Basic arithmetic operations
  - Logical operators (&&, ||, &&&, |||, not, ==>)
  - If, match and "matches"
  - Field expressions
  - Spec function calls and recursion
  - `Seq<T>` (compiled to `Vec<T>` or `&[T]` depending on the context), indexing, len, `seq!` literals
  - `SpecString` (an alias to `Seq<char>` to syntactically indicate that we want `String`/`&str`), indexing, len, string literals
  - `Option<T>`
  - User-defined structs and enums
  - Primitive integer/boolean types (`i<N>`, `isize`, `u<N>`, `usize`, `char`, etc.)
  - Equality between Seq, String, and arithmetic types

In particular, `int`/`nat`, `forall`/`exists`, spec-level references are currently unsupported.
I don't think it currently works particularly well with modules and more complicated paths either.

The test cases might also give a better impression on the set of supported features: https://github.com/zhengyao-lin/verus/blob/d8e060a5f0c0f34b249a0537b96b9cf6902d07ed/source/rust_verify_test/tests/exec_spec.rs

---

Please let me know your thoughts!

In particular, I'm unsure about these issues:
  - Since spec typing is much more lenient, compiling to well-typed exec code is a bit tricky, especially since we don't have full typing information in the proc macro. I'd really appreciate comments on whether my current solution is robust/performant enough.

    Currently, each spec type is associated with two exec types (see [this trait](https://github.com/zhengyao-lin/verus/blob/d8e060a5f0c0f34b249a0537b96b9cf6902d07ed/source/vstd/exec_spec/mod.rs#L26-L35)): an owned type and a borrowed type. Currently, all negative position like function parameters have the borrowed type, and positive positions like arguments of struct constructors or the return type have the owned type. Let bindings and blocks are also expected to have the owned type to avoid violating linearity or dropping temporary values with borrows. This means that there might be constant cloning. For example, `let a = b.c` will essentially be compiled to `let a = b.c.clone();`.
  - I'm not sure what's best way to propagate error messages to the user; e.g. if the functional correctness proof of the generated spec function fails, the user currently gets an arcane message pointing to the incorrect location.
  - The way that the macro currently finds the corresponding compiled exec version of an item is a bit brittle and done through naming (e.g. struct/enum names `A` => `ExecA`, and function names `f` => `exec_f`). Is there a better way to do this?

Thank you!


utaal:
[retreat 2025]

From online discussion, example of `custom_err`:
https://github.com/verus-lang/verus/blob/3b1b3b1276275decc6260d5814c998c9f314492d/examples/debug_expand.rs#L309

And use `quote_spanned!` to transfer the spans.

zhengyao-lin:
moved to #1888

PR #1683, opened 2025-05-27
vstd: implement for loop iterators for hashmap

bsdinis:

### DIFF
--- a/source/vstd/std_specs/hash.rs
+++ b/source/vstd/std_specs/hash.rs
@@ -25,8 +25,10 @@ use super::super::prelude::*;
 
 use core::borrow::Borrow;
 use core::hash::{BuildHasher, Hash, Hasher};
+use core::marker::PhantomData;
 use core::option::Option;
 use core::option::Option::None;
+use std::collections::hash_map;
 use std::collections::hash_map::{DefaultHasher, Keys, RandomState, Values};
 use std::collections::hash_set;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


source/vstd/std_specs/hash.rs
```diff
@@ -452,6 +454,160 @@ impl<'a, Key, Value> View for ValuesGhostIterator<'a, Key, Value> {
...
+                },
+                Some((k, v)) => {
+                    let (new_index, new_seq) = iter@;
+                    let (old_k, old_v) = old_seq[old_index];
+                    &&& 0 <= old_index < old_seq.len()
+                    &&& new_seq == old_seq
+                    &&& new_index == old_index + 1
+                    &&& k == old_k
+                    &&& v == old_v
+                    &&& old_seq.to_set().contains((*k, *v))
```
jaylorch, line 496: Isn't the conjunct `old_seq.to_set().contains(*k, *v)` redundant?

source/vstd/std_specs/hash.rs
```diff
@@ -452,6 +454,160 @@ impl<'a, Key, Value> View for ValuesGhostIterator<'a, Key, Value> {
...
+// To allow reasoning about the ghost iterator when the executable
+// function `iter()` is invoked in a `for` loop header (e.g., in
+// `for x in it: v.iter() { ... }`), we need to specify the behavior of
+// the iterator in spec mode. To do that, we add
+// `#[verifier::when_used_as_spec(spec_iter)` to the specification for
+// the executable `iter` method and define that spec function here.
+pub uninterp spec fn spec_hash_map_iter<'a, Key, Value, S>(m: &'a HashMap<Key, Value, S>) -> (r:
+    hash_map::Iter<'a, Key, Value>);
+
+// TODO: wrong, is not relating m@: Map<Key, Value> to iter@.1: Seq<(Key, Value)>
```
jaylorch: What does this TODO mean?

source/vstd/std_specs/hash.rs
```diff
@@ -452,6 +454,160 @@ impl<'a, Key, Value> View for ValuesGhostIterator<'a, Key, Value> {
...
+}
+
+#[verifier::when_used_as_spec(spec_hash_map_iter)]
+pub assume_specification<'a, Key, Value, S>[ HashMap::<Key, Value, S>::iter ](
+    m: &'a HashMap<Key, Value, S>,
+) -> (iter: hash_map::Iter<'a, Key, Value>)
+    ensures
+        obeys_key_model::<Key>() && builds_valid_hashers::<S>() ==> {
+            let (index, s) = iter@;
+            let (ks, vs) = s.unzip();
```
jaylorch: This assignment to `kv` and `vs` seems unnecessary.

source/vstd/std_specs/hash.rs
bsdinis: It's a leftover comment, my bad. The relationship alluded is ensured by the last condition (`forall |i| 0 <= i < v.len() ==> m@[v[i].0] == v[i].1`)

source/vstd/std_specs/hash.rs
bsdinis: Full disclosure: I'm not 100% sure about this, so any assertion by me should be taken with a grain of salt. 

The reasoning here is that the key value _pair_ needs to be asserted as valid. In other words, is not just an exiting key and an existing value -- which is asserted by `k == old_k && v == old_v` -- but an existing pair in the map.

source/vstd/std_specs/hash.rs
bsdinis: you're right, removed

source/vstd/std_specs/hash.rs
jaylorch: The fact that they form a pair is reflected in the fact that `old_seq[old_index] == (k, v)`. (And this, combined with the broadcast `lemma_seq_contains`, implies that `old_seq.contains((k, v))`. That in turn implies that `old_seq.to_set().contains((k, v))`.

bsdinis: Thank you for explaining!
Fixed

bsdinis: So, it turns out that this line is doing some heavy lifting: looking at the test added (https://github.com/verus-lang/verus/pull/1683/commits/7c350cd44865170fdce0673a010bb5af7e8e3556) without this portion of the axiomatization the test fails.

jaylorch: Interesting. You can fix the test by adding `assert(iter.kv_pairs.to_set().contains((*k, *v)));` before the two failing assertions, which suggests you don't *need* this line in the Verus trusted computing base. But you're right that it will help some developers shorten their proofs by not having to trigger the `to_set().contains(...)` themselves. So it's a matter of taste; do what you think best.

bsdinis: As a newcomer, I would expect this to be an invariant provided by the axiomatization of the hashmap, so I'm inclined to keep it

bsdinis: reverted to the original formulation

jaylorch (approved):

PR #1687, opened 2025-05-27
Make `unwind` keyword play nicely with ensures clauses

parno:
We received a report that verus rejects the following program, even though verusfmt emits it when given a version without the comma before `no_unwind`, and `verus-analyzer` rejects the no-comma version.  This PR adjusts verus to also accept the version with the comma; we could also consider rejecting the version without the comma.

```
use vstd::prelude::*;

verus! {

fn f()
    requires
        true,
    ensures
        true,
    no_unwind
{
}

fn main() {
}

} // verus!
```

### DIFF
--- a/dependencies/syn/src/verus.rs
+++ b/dependencies/syn/src/verus.rs
@@ -695,6 +695,7 @@ pub mod parsing {
                 || input.peek(Token![decreases])
                 || input.peek(Token![via])
                 || input.peek(Token![when])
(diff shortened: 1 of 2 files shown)


tjhance:
I think your `ensures_with_comma` test is missing its comma

parno:
> I think your `ensures_with_comma` test is missing its comma

Ironic but true.  Fixed now.

PR #1691, opened 2025-05-28
Implement DeepView for HashMap and add related lemmas

dschoepe:
This PR adds a `DeepView` impl, along with lemmas to relate it to the existing `view` impl.


jaylorch:
I think it would be fine to put them in the default broadcast group, and here are some considerations:

@Chris-Hawblitzel has a way to check that there's no performance regression, so if you decide to make the change I can ask him to check the PR to make sure.

The axiom you have that triggers on `contains_borrowed_key` has such a rarely-induced trigger that I think it'll be perfectly fine.

I'm not particularly worried about the `finite` one either, and think it could be pretty helpful in many scenarios. But if you're worried about it, you could just put the invocation of `.finite()` into the trigger. What do you think of that?

jaylorch:
Thanks. @Chris-Hawblitzel, could you check if there are any performance regressions from these additional axioms? (I'll add Chris as a reviewer.)

jaylorch (approved):

Chris-Hawblitzel:
>  @Chris-Hawblitzel, could you please check if there are any performance regressions from these additional axioms? (I'll add Chris as a reviewer.)

There was no noticeable difference in the benchmark performance.

PR #1694, opened 2025-05-28
Clean up arithmetic library

elanortang:


utaal:
Thanks! Have you confirmed that the `auto` lemmas are no longer needed (i.e. broadcast use works in `calc`)?

elanortang:
Yes, broadcast use does work in `calc` - I used it [here](https://github.com/verus-lang/verus/blob/utf8-rustlib/source/vstd/endian.rs) a couple months ago and just now all of the proofs still go through without any errors.

parno:
This looks good to me.  Thanks for all of the clean up!

PR #1695, opened 2025-05-28
Chars iterator

dschoepe:
Adds support for iterating over characters in a string with `str::chars()`. This is an updated version of #1650, putting the modeling behind `#[cfg(feature = alloc)]` and adding a test case.

### DIFF
--- a/source/vstd/string.rs
+++ b/source/vstd/string.rs
@@ -7,6 +7,9 @@ use alloc::string::{self, String, ToString};
 use super::prelude::*;
 use super::seq::Seq;
 use super::view::*;
+use crate::pervasive::ForLoopGhostIterator;
+use crate::pervasive::ForLoopGhostIteratorNew;
+use std::str::Chars;
 
 verus! {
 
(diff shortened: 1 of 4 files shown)


source/vstd/string.rs
```diff
@@ -304,6 +308,127 @@ impl StringExecFns for String {
...
+pub assume_specification<'a>[ Chars::<'a>::next ](chars: &mut Chars<'a>) -> (r: Option<char>)
+    ensures
+        ({
+            let (old_index, old_seq) = old(chars)@;
+            match r {
+                None => {
+                    &&& chars@ == old(chars)@
+                    &&& old_index >= old_seq.len()
+                },
+                Some(k) => {
```
jaylorch, line 345: Nitpick: Perhaps this should be `c` instead of `k`?

jaylorch (approved):

utaal:
Can you please add a test in `example/std_test/`? The intent with those is that they're runnable standalone and have `runtime_assert`s (see other files in that directory) that compare the runtime behavior to the specification.

dschoepe:
> Can you please add a test in `example/std_test/`? The intent with those is that they're runnable standalone and have `runtime_assert`s (see other files in that directory) that compare the runtime behavior to the specification.

Sure, will do!

PR #1698, opened 2025-05-28
VerusSync doc edits

parno:

### DIFF
--- a/source/docs/state_machines/src/properties.md
+++ b/source/docs/state_machines/src/properties.md
@@ -1,0 +1,56 @@
+# Properties
+
+The `property!` operation is not actually a transition.  Instead, it allows you to export
+useful facts from the state-machine view of the world to thread-local reasoning in your 
+implementation.
+
+For example, suppose your state machine has:
+```rust
+fields {
+    #[sharding(bool)]
+    pub waiting: bool,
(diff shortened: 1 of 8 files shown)


source/docs/state_machines/src/properties.md
```diff
@@ -0,0 +1,56 @@
...
+    #[sharding(variable)]
+    pub num_remaining: nat,
+
+    ...
+}
+```
+where `num_remaining` corresponds to an atomic U32 counting the number of
+threads that still need to perform an operation.  You could represent this by
+maintaining a state-machine level invariant that `self.waiting ==>
+self.num_remaining > 0`.  However, that fact is not available in your code, by
```
tjhance, line 22: no comma before "by default"

tjhance (changes requested):
It would be great if you could also mention guard and link to the strategies that use guard.

source/docs/state_machines/src/properties.md
parno: Done.

tjhance (approved):
It occurs to me that I've both asked you to add a comma and remove a comma today.

parno:
Conservation of commas, I assume!

PR #1702, opened 2025-05-28
Add functions and lemmas to seq_lib

amaurremi:
### Summary

This PR extends `vstd::seq_lib` with:

- functions – `flat_map`, `filter_map`, `filter_alt`, plus reducers `all`, `any`, `exactly_one` and `remove_duplicates`.
- additional lemmas
  * structural properties (how `skip`, `take`, `push`, `prefix`, and `concat` interact);
  * functional reasoning (`filter` / `map` / `flat-map` commutativity, monotonicity, length bounds and other properties);
  * duplicate handling (soundness and size limits for remove_duplicates);
  * `view` & `zip` reasoning (connecting `Seq` to `View` and zipped sequences).

### Diff issue

Unfortunately, the diff is not properly picking up the fact that this PR only contains additions (no removal or moving around of existing code). In my IDE (RustRover) the diff makes that clear.

### Credits
Developed jointly with @dschoepe, @txiang61 and @sjjs7.

By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -2205,6 +2205,28 @@ impl<A> Seq<A> {
 
         assert((self + s1).skip(self.len() as int) == s1);
     }
+
+    pub broadcast group group_seq_extra {
+        Seq::<_>::lemma_seq_skip_skip,
+        Seq::<_>::lemma_remove_duplicates_properties,
+        Seq::<_>::lemma_filter_contains_rev,
+        Seq::<_>::lemma_filter_map_take_succ,
+        Seq::<_>::lemma_filter_prepend,
+        Seq::<_>::lemma_filter_len_push,
+        Seq::<_>::lemma_take_len,
+        Seq::<_>::lemma_take_any_succ,
+        Seq::<_>::lemma_push_map_commute,
+        Seq::<_>::lemma_push_to_set_commute,
+        Seq::<_>::lemma_filter_push,
+        Seq::<_>::lemma_flat_map_take_append,
+        Seq::<_>::lemma_flat_map_singleton,
+        Seq::<_>::lemma_map_take_succ,
+        Seq::<_>::lemma_prefix_index_eq,
+        Seq::<_>::lemma_prefix_concat,
+        Seq::<_>::lemma_prefix_chain_contains,
+        Seq::<_>::lemma_prefix_append_unique,
+        Seq::<_>::lemma_all_push,
+    }
 }
 
 /// Filtering a sequence and then viewing its elements produces the same result as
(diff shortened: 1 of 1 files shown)


source/vstd/seq_lib.rs
```diff
@@ -88,6 +107,47 @@ impl<A> Seq<A> {
...
+    /// ## Example
+    ///
+    /// ```rust
+    /// fn example() {
+    ///     let s = seq![2, 4, 6, 8];
+    ///     assert!(s.all(|x| x % 2 == 0));
+    /// }
+    /// ```
+    pub open spec fn all(self, pred: spec_fn(A) -> bool) -> bool {
+        forall|elt: A| #[trigger] self.contains(elt) ==> pred(elt)
```
utaal: This requires going through the existential in `contains`, how about:
```suggestion
        forall|i: int| 0 <= i < self.len() ==> #[trigger] pred(self[i])
```
(see https://gist.github.com/verus-play/379745b99c1ef595b7f07f4012811053)

source/vstd/seq_lib.rs
```diff
@@ -88,6 +107,47 @@ impl<A> Seq<A> {
...
+    /// ## Example
+    ///
+    /// ```rust
+    /// fn example() {
+    ///     let s = seq![1, 2, 3, 4];
+    ///     assert!(s.any(|x| x > 3));
+    /// }
+    /// ```
+    pub open spec fn any(self, pred: spec_fn(A) -> bool) -> bool {
+        exists|elt: A| #[trigger] self.contains(elt) && pred(elt)
```
utaal: I'd consider avoiding the nested existential here too (to alleviate the need to trigger both).

utaal (commented):
First of all, thank you!
I've left a couple of comments on two of the new spec functions. Your experience using it may contradict them, if so let me know.
I think the additional proofs are all welcome: I wonder if we should make a `group_seq_properties_extra` or similar to allow broadcasting all of the new `broadcast proof`s.

source/vstd/seq_lib.rs
utaal: I see that triggering on the existential as suggested turns out to be harder than I expected. So I think this formulation is fine.

source/vstd/seq_lib.rs
utaal: it turns out this also causes triggering issues in the lemmas you added, so I reverted this change

utaal:
I've added broadcast groups for the new broadcast lemmas. Everything else looks good to me as-is.

amaurremi:
Thanks for applying the changes, and sorry for the slow follow-up but I actually did apply your changes and they worked fairly well, reducing some of the proofs. But do let me know if you'd like me to revert it, I'm happy to do it either way--what do you think?

amaurremi:
This is the diff with your changes: https://github.com/amaurremi/verus/commit/74b5f486b5d6b762634683f5fc1f4839eba2f180

utaal:
I think I got discouraged a bit too early by a proof failure, this looks good.
By the way, I think something may have gone wrong with the commits in your branch, and the change to `all` was missing: I restored it and now everything seems to look good. Thanks! (I'll merge when CI passes)

PR #1703, opened 2025-05-28
Add docs for HashSetWithView, HashMapWithView, and some structs in std_specs::hash

zero-to-nat:
- Let me know if you have better suggestions for how to organize the docs in [`std_specs::hash`](https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/hash/index.html). I put comments on the `#[verifier::external_type_specification]` structs, but those aren't associated with the `assume_specification` items. Those docs are generated separately with weird names like [https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/hash/fn._verus_external_fn_specification_63_HashMap_32__58__58__32__60__32_Key_44__32_Value_44__32_S_32__62__32__58__58__32_is__empty.html](https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/hash/fn._verus_external_fn_specification_63_HashMap_32__58__58__32__60__32_Key_44__32_Value_44__32_S_32__62__32__58__58__32_is__empty.html).  It doesn't look like there are currently any examples of such docs in `std_specs`.
- If you have context for why `StringHashSet` and `StringHashMap` were written and where they might be used, I can add docs for those as well.


jaylorch:
I wrote `StringHashMap` because `HashMapWithKey` doesn't quite work for strings. A minor issue is that you need to assume `obeys_key_model::<String>()`. A more important issue is that a string borrow is not necessarily a `&String` but instead is a `&str`.

source/vstd/hash_set.rs
```diff
@@ -85,6 +104,8 @@ impl<Key> HashSetWithView<Key> where Key: View + Eq + Hash {
         self.m.insert(k)
     }
 
+    /// Removes the given value from the set. Returns true if the value was previously in the set, false otherwise.
```
jaylorch, line 107: No, it returns true if the value *wasn't* previously in the set, false otherwise.

source/vstd/hash_set.rs
```diff
@@ -77,6 +94,8 @@ impl<Key> HashSetWithView<Key> where Key: View + Eq + Hash {
         self.m.is_empty()
     }
 
+    /// Inserts the given value into the set. Returns true if the value was previously in the set, false otherwise.
```
jaylorch, line 97: No, it returns true if the value *wasn't* previously in the set.

source/vstd/hash_set.rs
```diff
@@ -28,6 +35,9 @@ impl<Key> View for HashSetWithView<Key> where Key: View + Eq + Hash {
 }
 
 impl<Key> HashSetWithView<Key> where Key: View + Eq + Hash {
+    /// Creates an empty `HashSet` with capacity 0.
+    // todo - key model
```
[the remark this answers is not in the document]
zero-to-nat: Thanks! I linked to the docs for obeys_key_model since the assumptions are explained there.

source/vstd/hash_set.rs
zero-to-nat: Thanks for catching this, it is now fixed

zero-to-nat:
> I wrote `StringHashMap` because `HashMapWithKey` doesn't quite work for strings. A minor issue is that you need to assume `obeys_key_model::<String>()`. A more important issue is that a string borrow is not necessarily a `&String` but is instead a `&str`.

Thanks for the insights @jaylorch! I added these reasons to the documentation for `StringHashMap` and `StringHashSet`, it might be useful for someone trying to decide which type to use.

parno:
@jaylorch Does this look good to go?

@zero-to-nat It looks like there's now a conflict with `main`.  Would you mind straightening that out?

jaylorch (approved):
Thanks for all this helpful documentation!

PR #1704, opened 2025-05-29
do not evaluate for loop iterator expr multiple times

zeldovich:
Without this fix, trying to verify a simple program like:

```
        let m = HashSet::<u32>::new();
        for k in m.iter() {}
```

would lead to an error:

```
        error: loop invariant not satisfied
         --> bug.rs:9:9
          |
        9 |         for k in m.iter() {}
          |         ^^^^^^^^^^^^^^^^^^^^
          |         |
          |         at this loop exit
          |         failed this invariant

        verification results:: 1 verified, 1 errors
```

but if `m.iter()` was instead an identifier, as in below, verification passed:

```
        let m = HashSet::<u32>::new();
        let m_iter = m.iter();
        for k in m_iter {}
```

This commit changes the for loop macro to avoid evaluating the iterator expression, such as `m.iter()` in the above example, multiple times, which makes the first example above verify just fine.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -3163,7 +3163,9 @@ impl Visitor {
         // into:
         //  {
         //      #[allow(non_snake_case)]
-        //      let VERUS_loop_result = match ::core::iter::IntoIterator::into_iter(e) {
+        //      let VERUS_iter = e;
+        //      #[allow(non_snake_case)]
+        //      let VERUS_loop_result = match ::core::iter::IntoIterator::into_iter(VERUS_iter) {
         //          #[allow(non_snake_case)]
         //          mut VERUS_exec_iter => {
         //              #[allow(non_snake_case)]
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel:
Thanks for bringing this up -- I hadn't noticed this, because historically, `for` loop support preceded support for `loop_isolation(false)`, and the best approach depends on whether `loop_isolation(false)` or `loop_isolation(true)` is set.  When `loop_isolation(false)` is set, this pull request strictly improves loop invariant inference.  When `loop_isolation(true)` is set (which was the only setting when Verus first supported `for` loops), though, the loop invariant needs its own complete copy of the loop iterator; a variable defined outside the loop doesn't help, because loop verification is isolated from the variables defined outside the loop.  So I think the optimal approach is to use this pull request's change exactly when `loop_isolation(false)` is set.  I have a tentative implementation of this.  I'll send this pull request to a branch and add my changes.

PR #1706, opened 2025-05-29
Mutation, references, and borrowing documentation

apribiso:
In the main document, I succinctly describe ownership, references, and borrowing in Rust, while also linking the official Rust documentation that contains more examples. Then, I shortly explain how to use mutable references in requires, ensures, and assert, while providing simple and more complex examples.

### DIFF
--- a/examples/guide/references.rs
+++ b/examples/guide/references.rs
@@ -4,6 +4,7 @@
 use builtin::*;
 #[allow(unused_imports)]
 use builtin_macros::*;
+#[allow(unused_imports)]
 use vstd::prelude::*;
 
 verus! {
@@ -28,7 +29,7 @@ fn modify_y(a: &mut u32)
 
 fn mutable_example()
 {
-    let y: u32 = 1;
+    let mut y: u32 = 1;
     assert(y == 1);
     modify_y(&mut y);
     assert(y == 2);
--- a/examples/guide/references.rs
+++ b/examples/guide/references.rs
@@ -20,6 +20,12 @@ assert(*immut_ref_x == 0);
 }
 
 // ANCHOR: mut
+fn modify_y(a: &mut u32) 
+    ensures *a == 2
+{
+    *a = 2;
+}
+
 fn mutable_example()
 {
     let y: u32 = 1;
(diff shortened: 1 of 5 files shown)


source/docs/guide/src/mutation-references-borrowing.md
```diff
@@ -0,0 +1,27 @@
...
+
+The Rust documentation linked above provides many examples for these concepts. 
+
+Currently, Verus fully supports the use of immutable references and partially supports the use of mutable references. 
+
+```rust
+{{#include ../../../../examples/guide/references.rs:immut}}
+```
+
+Verus only support the use of mutable references as arguments to a function, such as in the following example.
```
parno, line 21: "only support" -> "only supports" and perhaps "Verus only" -> "Currently Verus only"

source/docs/guide/src/mutation-references-borrowing.md
```diff
@@ -0,0 +1,27 @@
...
+{{#include ../../../../examples/guide/references.rs:immut}}
+```
+
+Verus only support the use of mutable references as arguments to a function, such as in the following example. 
+
+```rust
+{{#include ../../../../examples/guide/references.rs:mut}}
+```
+
+Given that the pieces of memory to which mutable references point can be modified during function calls, we need a particular way to talk about their old and updated in Verus spec code, particularly in the `requires` and `ensures` of functions and in `assert`.
```
parno, line 27: I might phrase this as:
> about their old and updated values in Verus spec code, particularly in function `requires` and `ensures` clauses, and in `assert` statements.

examples/guide/references.rs
```diff
@@ -0,0 +1,102 @@
...
+{
+// ANCHOR: immut
+let x: u32 = 0;
+let immut_ref_x = &x;
+assert(x == 0);
+assert(*immut_ref_x == 0);
+// ANCHOR_END: immut
+}
+
+// ANCHOR: mut
```
parno, line 22: Stylistically, we generally sort functions such that callees precede callers, so in this case, that would mean putting `modify_y` before `mutable_example`.

examples/guide/references.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+fn check_and_assert(a: &mut u32)
+    requires *old(a) == 0
+{
+    assert(*old(a) == 0);
+    *a = *a + 1;
+    assert(*a == 1);
+    *a = *a + 1;
+    assert(*a == 2);
```
parno, line 67: Maybe we could add another copy of `assert(*old(a) == 0);` in here to emphasize that `old` refers to the value at the very beginning of the function, not to the value right before the last update.

parno (approved):
Thanks very much for writing this up!  I added a few minor comments.

apribiso:
Implemented suggestions :)

utaal:
Thank you!

PR #1710, opened 2025-05-30
Support deref coercion adjustments

Chris-Hawblitzel:
Rust often automatically inserts calls to `.deref()`, for example to convert a `Vec<u8>` to a `&[u8]` when calling `.iter()` on the `Vec<u8>`.  Previously, we only handled a few special cases (e.g. `Box`, but not `Vec`).  This pull request adds support for generally handling the `.deref()` adjustment by generating the appropriate VIR call to `.deref()` so that it can be verified like a normal function call.

### DIFF
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -1,7 +1,7 @@
 use crate::attributes::{GhostBlockAttr, get_ghost_block_opt};
 use crate::context::BodyCtxt;
 use crate::erase::{CompilableOperator, ResolvedCall};
-use crate::resolve_traits::{ResolutionResult, ResolvedItem};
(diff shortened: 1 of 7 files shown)


dschoepe:
Thanks, this is very useful! The example from the original issue (#1421) works now as well with the following spec:

```rust
fn foo(x: &Vec<u8>)
    requires
        x@.len() > 0,
        x@[0] == 42,
{
    let y = x.first().unwrap();
    assert(y as int == 42);
}

type Slice<T> = [T];

pub assume_specification<T>[ Slice::<T>::first ](slice: &[T]) -> (res: Option<&T>)
    ensures
        res == (if slice@.len() > 0 { Some(&slice@[0]) } else { None })
;
```

Orthogonal: Is there a way around the `Slice` type alias? It seems like writing `assume_specification<T>[ [T]::first ]...` is not syntactically valid.

dschoepe (approved):

PR #1713, opened 2025-05-30
Cow spec

dschoepe:
Adds specs and `View/DeepView` for `Cow`


dschoepe:
Note: I tried a more permissive bound with `T: ToOwned + ?Sized` on the blanket `impl` to avoid a separate `impl` for `str` and slices, but couldn't make that work.

PR #1720, opened 2025-06-03
examples: add decreases for examples in document

Marsman1996:
Currently `while-loop` examples in https://verus-lang.github.io/verus/guide/while.html do not contain the `decreases`, which might confuse the beginners.

### DIFF
--- a/examples/guide/recursion.rs
+++ b/examples/guide/recursion.rs
@@ -1,5 +1,3 @@
-#![cfg_attr(verus_keep_ghost, verifier::exec_allows_no_decreases_clause)]
-
 #[allow(unused_imports)]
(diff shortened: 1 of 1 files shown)


parno:
It looks like fe07347f6 (from @apribiso and @utaal) added `#![cfg_attr(verus_keep_ghost, verifier::exec_allows_no_decreases_clause)]` to the top of the file, so it's possible this is intentional, perhaps to keep reduce the number of concepts being introduced?  Although it does come after 5.2.1, which introduces the notion that exec functions also, by default, require a decreases clause, so that makes it seem like we should add them here.

utaal:
I think adding these at this point is okay.

PR #1724, opened 2025-06-05
vir: add error reporting instead of `panic` on type to AIR conversion

bsdinis:
This user error should not be reported via panic.

Duplicating the note from #1725 to show what this change does:

### Before

```
% cargo verus verify
   Compiling modelled v0.0.1 (/home/bsdinis/dev/bad_verus/bad_register)

thread '<unnamed>' panicked at vir/src/sst_to_air.rs:502:21:
abstract datatype should be boxed

worker thread panicked
error: could not compile `modelled` (lib)
```

### After 

```rust
% cargo verus verify
   Compiling modelled v0.0.1 (/home/bsdinis/dev/bad_verus/bad_register)
error: abstract datatype should be boxed Decorate(Ref, None, Datatype(Path(Path(Some("modelled"), ["network" :: "Context"])), [TypParam("T")], []))

error: could not compile `modelled` (lib) due to 1 previous error
```

### DIFF
--- a/source/rust_verify/src/commands.rs
+++ b/source/rust_verify/src/commands.rs
@@ -160,7 +160,7 @@ impl<'a> OpGenerator<'a> {
         for node in self.ctx.global.func_call_graph.get_scc_nodes(&scc_rep) {
             if let Node::TraitImpl(ImplPath::TraitImplPath(impl_path)) = node {
                 if let Some(imp) = self.trait_impl_map.get(&impl_path) {
-                    let cmds = vir::traits::trait_impl_to_air(&self.ctx, imp);
+                    let cmds = vir::traits::trait_impl_to_air(&self.ctx, imp)?;
                     ops.push(Op::context(ContextOp::TraitImpl, cmds, None));
                 }
             }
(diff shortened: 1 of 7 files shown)


utaal:
Thanks for the contribution, but I don't think we'd want to surface this as a user error -- from your report in https://github.com/verus-lang/verus/issues/1725 this really seems like a bug. We should fix the bug or - if it isn't - report the missing feature as an error. Feel free to re-open if I'm missing something.

bsdinis:
I think the actual error here is that verus doesn't support (in this case) structs with private (or non-visible) fields. So I think what might be missing here is actually changing the error message to reflect that (something like "datatypes must have public members"). 

Unsure to me that this is an actual verus bug. Rather, it seems that verus as a certain "public API" (structs need to have visible fields) and in this case the user had a struct with non-visible fields, so should have an error reporting that.

utaal:
Can you try https://github.com/verus-lang/verus/pull/1734 to see if it fixes the original issue?

bsdinis:
Did it, it worked. Thank you!

tjhance:
If we had span information easily available, it might have been worthwhile to package this up into a VirErr (albeit as an "internal error"), but we don't have easy span information, so I agree we should keep this as a panic.

PR #1727, opened 2025-06-08
Fix overloaded deref

### LINKED ISSUE #1722: Wrong Rust->VIR translation when explicit deref `*` is an overloaded deref
Hi! I encountered a Verus panic when verifying our code with `ManuallyDrop` usages. I wonder if you have any ideas on how to fix it quickly. If not, I want to help with it. I suspect a bug in VIR translation and am attempting to fix it. Thanks very much for your attention!

Here's the panic trace:

```text
thread '<unnamed>' panicked at air/src/typecheck.rs:276:17:
#60 in equality, left expression has type Int and right expression has different type core!mem.manually_drop.ManuallyDrop<usize.>.
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: air::typecheck::check_expr
   3: air::typecheck::check_expr
   4: air::typecheck::check_exprs
   5: air::typecheck::check_expr
   6: air::typecheck::check_decl
   7: air::context::Context::global
   8: air::context::Context::command
   9: rust_verify::verifier::Verifier::run_commands
  10: rust_verify::verifier::Verifier::verify_bucket
  11: rust_verify::verifier::Verifier::verify_bucket_outer
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

thread '<unnamed>' panicked at rust_verify/src/verifier.rs:382:17:
dropped, expected call to `into_inner` instead
stack backtrace:
   0:        0x1107870c0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h4b01b56329a0a8fc
   1:        0x10df8f0e4 - core::fmt::write::h5ac783d2f16cae69
   2:        0x11077afa4 - std::io::Write::write_fmt::h740ae8bf4425e1f4
   3:        0x110786f80 - std::sys::backtrace::BacktraceLock::print::h822ac0a1da741b33
   4:        0x1107894a4 - std::panicking::default_hook::{{closure}}::h182e34e1427a763f
   5:        0x1107892a0 - std::panicking::default_hook::hf3c673ab0d652139
   6:        0x110789e8c - std::panicking::rust_panic_with_hook::ha68643d4311595e9
   7:        0x110789b38 - std::panicking::begin_panic::{{closure}}::he9660982cced9f99
   8:        0x110787550 - std::sys::backtrace::__rust_end_short_backtrace::h1b0c3d169acc6c7e
   9:        0x112fdb878 - std::panicking::begin_panic::he9671f27e48421b5
  10:        0x1042fd55c - core::ptr::drop_in_place<core::option::Option<rust_verify::verifier::util::PanicOnDropVec<(alloc::sync::Arc<vir::messages::MessageX>,air::messages::MessageLevel)>>>::h2cc72cf8fcef90a9
  11:        0x10431fcb0 - rust_verify::verifier::Verifier::verify_bucket::hc18158022e349df6
  12:        0x104320a40 - rust_verify::verifier::Verifier::verify_bucket_outer::h6d5c610d95ac96a9
  13:        0x10421fe44 - std::sys::backtrace::__rust_begin_short_backtrace::h08224fcccbf0041b
  14:        0x104225004 - core::ops::function::FnOnce::call_once{{vtable.shim}}::heca6427dac9e0cff
  15:        0x110794eb0 - std::sys::pal::unix::thread::Thread::new::thread_start::hdb877a8060df0ba1
  16:        0x18b592c0c - __pthread_cond_wait

thread '<unnamed>' panicked at library/core/src/panicking.rs:226:5:
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
[1]    12697 abort      RUST_BACKTRACE=1 tools/verus/source/target-verus/release/verus reproduce.rs
```

The minimal reproduce example is:

```rust
// reproduce.rs
use vstd::prelude::*;
use core::mem::ManuallyDrop;
use core::ops::Deref;

verus! {

pub union TestEnum {
    pub variant: ManuallyDrop<usize>,
}

impl TestEnum {
    pub open spec fn borrow_variant_spec(&self) -> &usize {
        &*get_union_field::<TestEnum, ManuallyDrop<usize>>(*self, "variant")
    }

    #[verifier::external_body]
    pub fn borrow_variant(&self) -> (res: &usize)
        ensures
            res == self.borrow_variant_spec(),
    {
        unsafe { self.variant.deref() }
    }
}

pub fn main() {
    let x = TestEnum { variant: ManuallyDrop::new(42) };
    let y: &usize = x.borrow_variant();
    assert(*y == 42);
}

// `ManuallyDrop` specs.
pub uninterp spec fn ex_manually_drop_new_spec<V>(value: V) -> ManuallyDrop<V>;

pub uninterp spec fn ex_manually_drop_into_inner_spec<V>(slot: ManuallyDrop<V>) -> V;

pub uninterp spec fn ex_manually_drop_deref_spec<V: ?Sized>(slot: &ManuallyDrop<V>) -> &V;

#[verifier::external_fn_specification]
#[verifier::when_used_as_spec(ex_manually_drop_new_spec)]
pub fn ex_manually_drop_new<V>(value: V) -> ManuallyDrop<V> {
    ManuallyDrop::new(value)
}

#[verifier::external_fn_specification]
#[verifier::when_used_as_spec(ex_manually_drop_into_inner_spec)]
pub fn ex_manually_drop_into_inner<V>(slot: ManuallyDrop<V>) -> V {
    ManuallyDrop::into_inner(slot)
}

#[verifier::external_fn_specification]
#[verifier::when_used_as_spec(ex_manually_drop_deref_spec)]
pub fn ex_manually_drop_deref<V: ?Sized>(slot: &ManuallyDrop<V>) -> &V {
    slot.deref()
}

pub broadcast proof fn ex_manually_drop_value_properties<V: Sized>(value: V)
    ensures
        #[trigger] ManuallyDrop::into_inner(ManuallyDrop::new(value)) == value,
        *ManuallyDrop::new(value) == value,
        forall|other: V| value == other ==> ManuallyDrop::new(value) == ManuallyDrop::new(other),
{
    admit();
}

pub broadcast proof fn ex_manually_drop_ref_properties<V: Sized>(r: &V)
    ensures
        #[trigger] *ManuallyDrop::new(*r) == r,
{
    admit();
}

pub broadcast proof fn ex_manually_drop_properties<V: Sized>(slot: ManuallyDrop<V>)
    ensures
        #[trigger] ManuallyDrop::new(*slot) == slot,
{
    admit();
}

} // verus!
```

My Verus version is c3014c7b5df1e3d70eab4339f4b8fd26360e0499 (source compiled), with toolchain `1.85.1-aarch64-apple-darwin`.

junyang-zh:
Fix #1722 

<details><summary>Outdated content</summary>
<p>

This PR is currently not working. I submit it to seek help. Thanks in advance for your attention.

When executing on the minimal reproduction example, the verifier no longer panics. But dereference of `Arc`/`Box`/`Rc` is broken. It can't find the right spec `fn` to call for other overloaded `deref`s either, even with `assume_specification` specified.

For example, the VSTD library will fail to be verified:

```
error: `core::ops::deref::impl&%0::deref` is not supported (note: you may be able to add a Verus specification to this function with `assume_specification`) (note: the vstd library provides some specification for the Rust std library, but it is currently limited)
  --> vstd/view.rs:66:9
   |
66 |         (**self).view()
   |         ^^^^^^^^

```

</p>
</details> 

---

EDIT: This PR is now ready for review!

### DIFF
--- a/source/rust_verify_test/tests/external_fn_specification.rs
+++ b/source/rust_verify_test/tests/external_fn_specification.rs
@@ -1548,3 +1548,46 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+// `Deref` is special since it has an alias `*` for calling `.deref()`.
+
+test_verify_one_file! {
+    #[test] test_manually_drop_deref_when_have_spec verus_code! {
+        use core::mem::ManuallyDrop;
+        use core::ops::Deref;
+
+        #[verifier::external_type_specification]
+        #[verifier::external_body]
+        #[verifier::reject_recursive_types(T)]
+        pub struct ExManuallyDrop<T: ?Sized>(ManuallyDrop<T>);
+
+        pub uninterp spec fn manually_drop_spec_get<T: ?Sized>(m: &ManuallyDrop<T>) -> &T;
+
+        #[verifier::when_used_as_spec(manually_drop_spec_get)]
+        pub assume_specification<T: ?Sized >[ <ManuallyDrop<T> as Deref>::deref ](
+            v: &ManuallyDrop<T>,
+        ) -> (res: &T)
+            ensures
+                res == manually_drop_spec_get(v),
+        ;
+
+        fn do_exec_deref(x: &ManuallyDrop<usize>)
+            requires 100 <= **x < 102
+        {
+            // Explicit call.
+            let u: &usize = &((*x).deref());
+            assert(100 <= *u < 102);
+            // Implicit call.
+            let v: &usize = &**x;
+            assert(100 <= *v < 102);
+        }
+
+        spec fn do_explicit_spec_deref(x: &ManuallyDrop<usize>) -> &usize {
+            &((*x).deref())
+        }
+
+        spec fn do_implicit_spec_deref(x: &ManuallyDrop<usize>) -> &usize {
+            &**x
+        }
+    } => Ok(())
+}
--- a/source/rust_verify_test/tests/refs.rs
+++ b/source/rust_verify_test/tests/refs.rs
@@ -612,3 +612,57 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] test_overloaded_deref verus_code! {
+        use core::ops::Deref;
+
+        pub struct X {
+            pub inner: usize,
+        }
+
+        #[verifier::external_trait_specification]
+        pub trait ExDeref {
+            type ExternalTraitSpecificationFor: core::ops::Deref;
+
+            type Target: ?Sized;
+
+            fn deref(&self) -> &Self::Target;
+        }
+
+        pub open spec fn x_deref_spec(x: &X) -> (ret: &usize) {
+            &x.inner
+        }
+
+        impl Deref for X {
+            type Target = usize;
+
+            #[verifier::when_used_as_spec(x_deref_spec)]
+            fn deref(&self) -> (ret: &usize)
+                ensures
+                    ret == self.inner,
+            {
+                &self.inner
+            }
+        }
+
+        fn do_exec_deref(x: &X)
+            requires 100 <= x.inner < 102
+        {
+            // Explicit call.
+            let u: &usize = &((*x).deref());
+            assert(100 <= *u < 102);
+            // Implicit call.
+            let v: &usize = &**x;
+            assert(100 <= *v < 102);
+        }
+
+        spec fn do_explicit_spec_deref(x: &X) -> &usize {
+            &((*x).deref())
+        }
+
+        spec fn do_implicit_spec_deref(x: &X) -> &usize {
+            &**x
+        }
+    } => Ok(())
+}
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -276,6 +276,7 @@ pub(crate) fn fn_call_to_vir<'tcx>(
 
 pub(crate) fn deref_to_vir<'tcx>(
     bctx: &BodyCtxt<'tcx>,
+    expr: &Expr<'tcx>,
     trait_fun_id: DefId,
     arg: vir::ast::Expr,
     expr_typ: Typ,
@@ -292,6 +293,10 @@ pub(crate) fn deref_to_vir<'tcx>(
         if let TyKind::Ref(_, arg_ty, _) = arg_ty.kind() { arg_ty.clone() } else { arg_ty };
     let node_substs = tcx.mk_args(&[GenericArg::from(arg_ty)]);
 
+    let trait_fun =
+        Arc::new(FunX { path: def_id_to_vir_path(tcx, &bctx.ctxt.verus_items, trait_fun_id) });
+    let mut record_trait_fun = trait_fun.clone();
+
     let res = resolve_trait_item(span, tcx, typing_env, trait_fun_id, node_substs)?;
     let target_kind = match res {
         ResolutionResult::Resolved { resolved_item: ResolvedItem::FromImpl(did, args), .. } => {
@@ -299,6 +304,9 @@ pub(crate) fn deref_to_vir<'tcx>(
             let impl_paths = get_impl_paths(bctx, did, args, None);
             let resolved =
                 Arc::new(FunX { path: def_id_to_vir_path(tcx, &bctx.ctxt.verus_items, did) });
+
+            record_trait_fun = resolved.clone();
+
             vir::ast::CallTargetKind::DynamicResolved {
                 resolved,
                 typs,
@@ -310,11 +318,12 @@ pub(crate) fn deref_to_vir<'tcx>(
         _ => crate::internal_err!(span, "unexpected deref"),
     };
 
+    let autospec_usage = if bctx.in_ghost { AutospecUsage::IfMarked } else { AutospecUsage::Final };
+
+    record_call(bctx, expr, ResolvedCall::Call(record_trait_fun, autospec_usage));
+
     let typ_args = mk_typ_args(bctx, node_substs, trait_fun_id, span)?;
     let impl_paths = get_impl_paths(bctx, trait_fun_id, node_substs, None);
-    let trait_fun =
-        Arc::new(FunX { path: def_id_to_vir_path(tcx, &bctx.ctxt.verus_items, trait_fun_id) });
-    let autospec_usage = if bctx.in_ghost { AutospecUsage::IfMarked } else { AutospecUsage::Final };
     let call_target = CallTarget::Fun(target_kind, trait_fun, typ_args, impl_paths, autospec_usage);
     let args = Arc::new(vec![arg.clone()]);
     let x = ExprX::Call(call_target, args);
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -279,12 +279,19 @@ pub(crate) fn deref_to_vir<'tcx>(
     trait_fun_id: DefId,
     arg: vir::ast::Expr,
     expr_typ: Typ,
-    inner_ty: rustc_middle::ty::Ty<'tcx>,
+    arg_ty: rustc_middle::ty::Ty<'tcx>,
     span: Span,
 ) -> Result<vir::ast::Expr, VirErr> {
     let tcx = bctx.ctxt.tcx;
     let typing_env = TypingEnv::post_analysis(tcx, bctx.fun_id);
-    let node_substs = tcx.mk_args(&[GenericArg::from(inner_ty)]);
+    // The `arg_ty`, if `&T`, should be Rust automatically adding the `&`
+    // reference for calling `deref`. We strip it for trait resolution.
+    //
+    // Otherwise, it may be `deref` coercion. Leave it as-is.
+    let arg_ty =
+        if let TyKind::Ref(_, arg_ty, _) = arg_ty.kind() { arg_ty.clone() } else { arg_ty };
+    let node_substs = tcx.mk_args(&[GenericArg::from(arg_ty)]);
+
     let res = resolve_trait_item(span, tcx, typing_env, trait_fun_id, node_substs)?;
     let target_kind = match res {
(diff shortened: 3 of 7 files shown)


rikosellic:
It seems that in this case, `let arg_type = bctx.types.expr_ty_adjusted(arg)` in `deref_expr_to_vir` returns a type like `&'{erased} std::rc::Rc<A/#0, std::alloc::Global>`, and the leading reference & causes `auto_deref_supported_for_ty` to fail for `Arc`/`Box`/`Rc`.

junyang-zh:
`&'{erased} std::rc::Rc<A/#0, std::alloc::Global>` should be the argument of the deref call, since deref signature is `(&self) -> &T`. Rust automatically added this `&'{erased}`.

Chris-Hawblitzel:
Thanks for working on this!  In https://github.com/verus-lang/verus/issues/1722 , @tjhance posted a small spec example exhibiting the issue:

```
// currently mistranslated by Verus:
spec fn borrow_spec(x: &ManuallyDrop<usize>) -> &usize {
    &**x
}
```

It may also be helpful to have a small exec example exhibiting the same issue:

```
// currently mistranslated by Verus:
fn borrow_exec(x: &ManuallyDrop<usize>) {
    let u: &usize = &**x;
    assert(*u >= 0);
}
```

Note that explicit calls to `.deref()` are already handled, so the following works in existing Verus:

```
use core::mem::ManuallyDrop;
use std::ops::Deref;

pub uninterp spec fn manually_drop_spec_get<T: ?Sized>(m: &ManuallyDrop<T>) -> &T;

#[verifier::when_used_as_spec(manually_drop_spec_get)]
pub assume_specification<T: ?Sized >[ <ManuallyDrop<T> as Deref>::deref ](
    v: &ManuallyDrop<T>,
) -> (res: &T)
    ensures
        res == manually_drop_spec_get(v),
;

fn borrow_exec_deref(x: &ManuallyDrop<usize>) {
    let u: &usize = &((*x).deref());
    assert(*u >= 0);
}

spec fn borrow_spec_deref(x: &ManuallyDrop<usize>) -> &usize {
    &((*x).deref())
}
```

Ideally, the fix for the issue would take the code with `&**x` and replace it with `&((*x).deref())`, so that everything else, including `when_used_as_spec`, would work just as it does in the code above with programmer-written `deref()`s.

Another thing to note is that currently, for better or worse, `Box`, `Rc`, and `Arc` are special cases in Verus.  The types `Box<T>`, `Rc<T>`, and `Arc<T>` are simply translated as `T` in VIR, and `*` operations on them are translated as no-ops in VIR.  I would propose that for the purpose of fixing https://github.com/verus-lang/verus/issues/1722 , we keep this special treatment of `Box`, `Rc`, and `Arc` so that `*` continues to be a special-case no-op for them, both in exec code and spec code.  This should avoid the problems you ran into with `(**self).view()`.

junyang-zh:
> Another thing to note is that currently, for better or worse, Box, Rc, and Arc are special cases in Verus. The types Box<T>, Rc<T>, and Arc<T> are simply translated as T in VIR, and * operations on them are translated as no-ops in VIR. I would propose that for the purpose of fixing https://github.com/verus-lang/verus/issues/1722 , we keep this special treatment of Box, Rc, and Arc so that * continues to be a special-case no-op for them, both in exec code and spec code. This should avoid the problems you ran into with (**self).view().

I see. I was too stubborn in making `*x` work for all overloaded dereferences, including `Arc`/`Box`/`Rc`. Now I think we can have a compromised solution that:
 - `*x` works only for `Arc`/`Box`/`Rc`;
 - `x.deref()` works for both special treated types and other types like `ManuallyDrop`.
 
 I may push a version soon for this PR.

junyang-zh:
@Chris-Hawblitzel @tjhance Thanks for your helpful comments again. How do you like the solution in its current form?

source/vir/src/well_formed.rs
```diff
@@ -152,17 +152,16 @@ fn check_path_and_get_function<'a>(
...
-                        },
-                    ),
-                ));
+                let mut error = error(span, &format!("`{path:}` is not supported."));
+                let mut help = "You may be able to add a Verus specification to this function with `assume_specification`.".to_string();
+                if x.path.is_rust_std_path() {
+                    help += " The vstd library provides some specification for the Rust std library, but it is currently limited.";
+                }
+                if *x.path.last_segment() == "deref" {
+                    help += " For dereferncing in ghost mode, you should explicitly call `.deref()` other than using `*`. This restriction may be lifted in the future.";
```
tjhance: small typo with "dereferncing".

Also, I'm not sure what this help message is for? Using explicit `.deref()` and using `*` to implicity insert a `.deref()` ought to be equivalent, regardless of whether we're in ghost code or not.

source/rust_verify_test/tests/external_fn_specification.rs
```diff
@@ -1548,3 +1548,60 @@ test_verify_one_file! {
...
+        fn borrow_exec_deref(x: &ManuallyDrop<usize>) {
+            let u: &usize = &((*x).deref());
+            assert(*u >= 0);
+        }
+
+        spec fn borrow_spec_deref(x: &ManuallyDrop<usize>) -> &usize {
+            &((*x).deref())
+        }
+    } => Ok(())
+}
```
tjhance, line 1593: `external_fn_specification` is an odd place for these tests I think. The use of `external_fn_specification`/`assume_specification` is kind of orthogonal to the feature. We can keep one of these tests — to test the interaction between deref and assume_specification — but I would also prefer minimal examples that just test the overloaded `deref`.

For example, you could start by implementing `Deref` for your own struct:

```
pub struct X<T> {
    pub t: T,
}

impl<T> std::ops::Deref for X<T> {
    type Target = T;
    fn deref(&self) -> (ret: &T) 
        ensures ret == self.t
    {   
        &self.t
    }   
}
```

A good place for the tests would be in `tests/refs.rs` which already has some tests for implicit derefs.

tjhance (changes requested):
The implementation in rust_to_vir_expr looks good, I have some comments about the diagnostics & tests.

junyang-zh:
I found that the current form is still not usable. ~~Only external implementations for types can be handled well. locally defined types (such as one in the comment https://github.com/verus-lang/verus/pull/1727#discussion_r2158271738) still leads to bugs. And there's no way to call the corresponding spec fn in ghost mode either. I may soon fix them.~~

Chris-Hawblitzel:
I made some bug fixes over the last day in the `main` branch that should help `when_used_as_spec` and trait impls, so you may want to merge with `main`.  Overall, I think this PR is on the right track.  I agree with @tjhance's question about `well_formed.rs`; I think the change to `well_formed.rs` won't be needed.

When I ran this PR (merged with main) on the examples described above, I noticed that the call to `deref_to_vir(bctx, fn_def_id, inner_expr, inner_ty, arg_ty, expr.span)` passes in `arg_ty = expr_ty_adjusted(arg)`, rather than `expr_ty(arg)`, and the resulting `arg_ty` is a ref type `&T` instead of `T`.  I think this causes `deref_to_vir` to resolve to the wrong function call.  Specifically, it resolves to https://doc.rust-lang.org/src/core/ops/deref.rs.html#154 :

```
impl<T: ?Sized> const Deref for &T {
    type Target = T;

    #[rustc_diagnostic_item = "noop_method_deref"]
    fn deref(&self) -> &T {
        *self
    }
}
```

instead of https://doc.rust-lang.org/beta/src/core/mem/manually_drop.rs.html#260 :

```
impl<T: ?Sized> Deref for ManuallyDrop<T> {
    type Target = T;
    #[inline(always)]
    fn deref(&self) -> &T {
        &self.value
    }
}
```

tjhance:
I think we want to use the adjusted type and strip the reference off. If there are additional, unrelated adjustments, then the unadjusted type could be even farther removed from  the type of interest.

junyang-zh:
> ... I think this causes deref_to_vir to resolve to the wrong function call. Specifically, it resolves to https://doc.rust-lang.org/src/core/ops/deref.rs.html#154 :

@Chris-Hawblitzel Very correct. I happened to discover it hours ago. And it is fixed now.

> I think we want to use the adjusted type and strip the reference off.

@tjhance Yes.

> If there are additional, unrelated adjustments, then the unadjusted type could be even farther removed from the type of interest.

I didn't peel extra adjustments because I did not see it happening. It might be better to find an example to trigger this.

---

Also, I found that the call in `deref_to_vir` should be recorded to make the lifetime check work. I also fixed it in the third commit. It seems to work now. Yeah.

junyang-zh:
I suppose cfe70448a29f0c3cc37a2a7cadd05263b17c22e6 is not the right way. We should use `IfMark` only in ghost context. Otherwise the two tests I add would fail. Fixed.

junyang-zh:
I see now `verifier::when_used_as_spec` can be used for traits. Great!

junyang-zh:
No more pushes today. Sorry for the noise.

Chris-Hawblitzel:
> I suppose [cfe7044](https://github.com/verus-lang/verus/commit/cfe70448a29f0c3cc37a2a7cadd05263b17c22e6) is not the right way. We should use `IfMark` only in ghost context. Otherwise the two tests I add would fail. Fixed.

Oops, that's right.  I just pushed a fix as well.

junyang-zh:
Rebased on the latest main branch.

junyang-zh:
Have been silent for a while :-)

@tjhance @Chris-Hawblitzel Consider taking a review?

source/rust_verify_test/tests/refs.rs
```diff
@@ -612,3 +612,54 @@ test_verify_one_file! {
...
+            }
+        }
+
+        fn do_exec_deref(x: &X<usize>) {
+            // Explicit call.
+            let u: &usize = &((*x).deref());
+            assert(*u >= 0);
+            // Implicit call.
+            let v: &usize = &**x;
+            assert(*v >= 0);
```
tjhance: these assertions aren't very interesting (v is a usize, so of course it's >= 0), can you do an assertion that tests the 'ensures' clause of the deref?

junyang-zh:
@tjhance Needs CI approval again.

tjhance:
thanks for the contribution!

PR #1734, opened 2025-06-10
fix boxing of closure param types

### LINKED ISSUE #1725: AIR translation: Weird Panic on abstract datatypes
I have wittled down an example for an interesting panic on Verus:

```
% cargo verus verify
   Compiling modelled v0.0.1 (/home/bsdinis/dev/bad_verus/bad_register)

thread '<unnamed>' panicked at vir/src/sst_to_air.rs:502:21:
abstract datatype should be boxed

worker thread panicked
error: could not compile `modelled` (lib)
```

With #1724, this becomes a slightly nicer error:
```rust
% cargo verus verify
   Compiling modelled v0.0.1 (/home/bsdinis/dev/bad_verus/bad_register)
error: abstract datatype should be boxed Decorate(Ref, None, Datatype(Path(Path(Some("modelled"), ["network" :: "Context"])), [TypParam("T")], []))

error: could not compile `modelled` (lib) due to 1 previous error
```

Because this fails so early, `verus --record` doesn't really help, i think? I've included the code below:

# Code

There are 2 files (`a.rs` and `foo.rs`).

```rust
use crate::network::FooT;

use vstd::prelude::*;

verus! {

fn write<'a, T, C>(t: &'a T)
{
        let bpool = FooT { t };
        let req = bpool.broadcast();

        req.wait_for(|s| s.rand());
}

}
```

```rust
use vstd::prelude::*;

verus! {

pub struct FooT<'a, T> {
    pub t: &'a T,
}

impl<'a, T> FooT<'a, T> {
    pub fn broadcast(
        self,
    ) -> Context<'a, T> {
        Context { t: self.t }
    }
}

pub struct Context<'a, T> {
    t: &'a T,
}

impl<'a, T> Context<'a, T> {
    pub fn rand(&self) -> bool {
        true
    }

    pub fn wait_for<F>(
        self,
        termination_cond: F,
    )
    where
        F: Fn(&Self) -> bool,
        requires termination_cond.requires((&self,)),
    {
        if termination_cond(&self) {
            return;
        }
    }
}

}
```

# Why it might be a bug

There are multiple small changes (and in some cases they should lead to equivalent programs) that actually eliminate the issue, which makes it seem like this is not a really a limitation but a bug in Verus:
- Putting both of these in `lib.rs` 
- Removing the `termination_cond` of `wait_for`

Notably, trying to use the hint from the panic message (boxing either FooT, Context or the closure) does not help.

# Version

```
% verus --version
Verus
  Version: 0.2025.05.30.c3014c7
  Profile: release
  Platform: linux_x86_64
  Toolchain: 1.85.1-x86_64-unknown-linux-gnu
```

tjhance:
fixes #1725

### DIFF
--- a/source/rust_verify_test/tests/exec_closures.rs
+++ b/source/rust_verify_test/tests/exec_closures.rs
@@ -1450,3 +1450,23 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] param_boxing_issue1725 verus_code! {
+        mod a {
+            use vstd::prelude::*;
+
+            fn write<'a, T, C>(t: &'a T) {
+                let j = |s: &crate::foo::Context<'a, T>| true;
+            }
+        }
+
+        mod foo {
+            use vstd::prelude::*;
+
+            pub struct Context<'a, T> {
+                t: &'a T,
+            }
+        }
+    } => Ok(())
+}
--- a/source/vir/src/poly.rs
+++ b/source/vir/src/poly.rs
@@ -896,13 +896,20 @@ fn visit_stm(ctx: &Ctx, state: &mut State, stm: &Stm) -> Stm {
             mk_stm(StmX::OpenInvariant(s))
         }
         StmX::ClosureInner { body, typ_inv_vars } => {
+            let typ_inv_vars = Arc::new(
+                typ_inv_vars
+                    .iter()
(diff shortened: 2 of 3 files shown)


bsdinis:
I reproduced this locally. Fixes the original issue.

Thanks!

utaal (approved):

PR #1735, opened 2025-06-11
Rustc 1.86

dschoepe:
This is a first attempt at upgrading to rustc-1.86.0.

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #1736, opened 2025-06-12
Rustc 1.87.0

dschoepe:
This upgrades Verus to use rustc 1.87.0

### DIFF
--- a/source/vstd/std_specs/alloc.rs
+++ b/source/vstd/std_specs/alloc.rs
@@ -9,8 +9,9 @@ verus! {
 pub struct ExGlobal(alloc::alloc::Global);
 
 #[feature(liballoc_internals)]
-pub assume_specification<T>[alloc::boxed::box_new](x: T) -> (result: Box<T>)
-    ensures *result == x
+pub assume_specification<T>[ alloc::boxed::box_new ](x: T) -> (result: Box<T>)
+    ensures
+        *result == x,
 ;
 
 } // verus!
--- a/source/vstd/std_specs/alloc.rs
+++ b/source/vstd/std_specs/alloc.rs
@@ -8,4 +8,9 @@ verus! {
 #[verifier::external_body]
 pub struct ExGlobal(alloc::alloc::Global);
 
+#[feature(liballoc_internals)]
+pub assume_specification<T>[alloc::boxed::box_new](x: T) -> (result: Box<T>)
+    ensures *result == x
+;
+
 } // verus!
--- a/source/vstd/std_specs/alloc.rs
+++ b/source/vstd/std_specs/alloc.rs
@@ -4,12 +4,14 @@ verus! {
 
 // this is a bit of a hack; verus treats Global specially already,
 // but putting this here helps Verus pick up all the trait impls for Global
+#[cfg(feature = "alloc")]
 #[verifier::external_type_specification]
 #[verifier::external_body]
 pub struct ExGlobal(alloc::alloc::Global);
 
+#[cfg(feature = "alloc")]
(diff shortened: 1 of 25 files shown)


dschoepe:
I can reproduce the rustc segfault from the CI release build locally on OSX but not on Linux. This might be related to rustc upgrading to LLVM 20 as part the 1.87.0 release.

source/vstd/std_specs/alloc.rs
```diff
@@ -4,8 +4,16 @@ verus! {
...
+#[cfg(feature = "alloc")]
 #[verifier::external_type_specification]
 #[verifier::external_body]
 pub struct ExGlobal(alloc::alloc::Global);
 
+#[cfg(feature = "alloc")]
+#[feature(liballoc_internals)]
+pub assume_specification<T>[ alloc::boxed::box_new ](x: T) -> (result: alloc::boxed::Box<T>)
+    ensures
+        *result == x,
```
utaal, line 16: How is this related with the upgrade to 1.87.0?

dschoepe: `Box::new` is marked with `#[inline(always)]` and up to 1.85.1 was [defined](https://doc.rust-lang.org/1.85.1/src/alloc/boxed.rs.html#252) as follows:

```rust
    pub fn new(x: T) -> Self {
        #[rustc_box]
        Box::new(x)
    }
```

In 1.86.0 the [definition](https://doc.rust-lang.org/1.86.0/src/alloc/boxed.rs.html#273) changes to:

```rust
    pub fn new(x: T) -> Self {

        return box_new(x);

    }
```

This leads to errors like this in tests:
```
error: `alloc::boxed::box_new` is not supported (note: you may be able to add a Verus specification to this function with `assume_specification`) (note: the vstd library provides some specification for the Rust std library, but it is currently limited)
 --> ../../examples/std_test/vec_test.rs:9:27
  |
9 |     let mut a: Vec<u32> = vec![1, 2];
  |                           ^^^^^^^^^^
  |
  = note: this error originates in the macro `vec` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tjhance: I'm confused; are you saying Rust is inlining Box::new _before_ HIR?

dschoepe: I'm also surprised by this, but commenting out this specification does cause the above test failures (just rechecked). I'll try to see if I can reproduce this with other inlined functions.

tjhance: BTW, the error you posted makes sense, since `vec!` is implemented by calling `box_new` directly: https://doc.rust-lang.org/1.87.0/src/alloc/macros.rs.html#42

I was wondering if there were any test failures that involve `Box::new`.

dschoepe: You're right, this only happens for tests involving `vec!`.

tjhance: phew! 😅

dschoepe:
The segfault is caused by this part in `smt2parser`:

```
    fn visit_get_value(&mut self, terms: Vec<Term>) -> Result<Self::T, Self::E> {
        self.get_value_count += 1;
        for term in terms {
            self.visit_term(term); // commenting this out fixes the issue
        }
        Ok(())
    }
```

dschoepe:
This crash no longer occurs with toolchain version `1.88-beta`. I propose we upgrade to 1.86.0 (I'll reopen #1735) and we skip 1.87.0 to side-step this issue.

dschoepe:
obsoleted by #1742

Issue #1511, opened 2025-03-13
Panic on trait impl


zhengyao-lin:
Hello! I'm hitting a Verus panic on the following snippet ([playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=1002c8278db7baa602dfa65efadcf0d5)):

```
thread '<unnamed>' panicked at vir/src/poly.rs:406:66:
no entry found for key
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:2173:29:
worker thread panicked
```

```
use vstd::prelude::*;

verus! {

struct A {
    a: bool,
}

struct B;

trait TraitA {
    proof fn prop(chain: Seq<Option<A>>)
        ensures
            forall |i| #![trigger chain[i]]
                0 <= i < chain.len() ==>
                (chain[i] matches Some(MISSING_KEY) ==> MISSING_KEY.a || !MISSING_KEY.a);
}

impl TraitA for B {
    proof fn prop(chain: Seq<Option<A>>) {}
}

}
```

This is on the latest version ([9b8b556](https://github.com/verus-lang/verus/commit/9b8b5567ffffadd8223ffec889a86f2a5424b038))

The missing key at `vir/src/poly.rs:406:66` is `MISSING_KEY`, in case that helps.
Also this worked fine before the trait refactoring last year [3c05f3c](https://github.com/verus-lang/verus/commit/3c05f3cbe6a3fb90087baa9fcfe5a0e252a78f33).

zhengyao-lin:
After some digging, I think the issue is here: https://github.com/verus-lang/verus/blob/9b8b5567ffffadd8223ffec889a86f2a5424b038/source/vir/src/ast_to_sst_func.rs#L621-L625

This branch is used when generating the recommends check for a trait impl function (like `B::prop` abrove).
But `subst_stm` doesn't play well with `check_pure_expr`, since the result of the latter contains some temporary variables that need to be "finalized" later? (as least by my incomplete understanding of the code base)

I tried inserting a
```
let stms = stms.into_iter().map(|s| state.finalize_stm(&ctx, &s)).collect::<Result<Vec<_>, _>>()?;
```
before line 622, and it works on this example, but I'm not sure if that's the right solution.

PR #1746, opened 2025-06-15
Fix a panic on trait impl

zhengyao-lin:
Hello! A few months ago I was hitting a Verus panic: #1511, and it doesn't seem to be fixed in the latest version, so I decided to make a PR for my (potentially incomplete) fix.
This is also blocking the merge of [Verdict](https://github.com/secure-foundations/verdict) into the Verus regression suite (w/ Veritas).

To reiterate the issue, Verus panics on the following snippet:
```
use vstd::prelude::*;

verus! {

struct A {
    a: bool,
}

struct B;

trait TraitA {
    proof fn prop(chain: Seq<Option<A>>)
        ensures
            forall |i| #![trigger chain[i]]
                0 <= i < chain.len() ==>
                (chain[i] matches Some(MISSING_KEY) ==> MISSING_KEY.a || !MISSING_KEY.a);
}

impl TraitA for B {
    proof fn prop(chain: Seq<Option<A>>) {}
}

}
```

```
thread '<unnamed>' panicked at vir/src/poly.rs:411:66:
no entry found for key
worker thread panicked
```

In particular, the the missing key at `vir/src/poly.rs:411:66` is exactly the identifier `MISSING_KEY` above.

I believe the issue happens during the lowering of a ensures/requires clause from VIR-AST to VIR-SST, in the situation when:
1. recommends check is enabled
2. the ensure clause is inherited from a trait declaration.

which corresponds to the following snippet

https://github.com/verus-lang/verus/blob/9b557d70faab565829c0642c4d56609a44814d82/source/vir/src/ast_to_sst_func.rs#L581-L599

I think `expr_to_pure_exp_check` creates some temporary variables that need to be "finalized" later; but `subst_stm` seems to be renaming these temporary variables, which causes a problem later.
So I added `state.finalize` calls in between these two operations.

This fixes the panicking example above, but obviously I don't have a good complete understanding of this code base.
@Chris-Hawblitzel could you please check if this makes sense at all?

Thank you!

### DIFF
--- a/source/rust_verify_test/tests/traits.rs
+++ b/source/rust_verify_test/tests/traits.rs
@@ -4316,3 +4316,27 @@ test_verify_one_file! {
         }
     } => Err(err) => assert_fails(err, 2)
 }
+
+test_verify_one_file! {
+    #[test] trait_ensure_recommends_check verus_code! {
+        use vstd::prelude::*;
+
+        struct A {
+            a: bool,
+        }
+
+        struct B;
+
+        trait TraitA {
+            proof fn prop(chain: Seq<Option<A>>)
+                ensures
+                    forall |i| #![trigger chain[i]]
+                        0 <= i < chain.len() ==>
+                        (chain[i] matches Some(MISSING_KEY) ==> MISSING_KEY.a || !MISSING_KEY.a);
+        }
+
+        impl TraitA for B {
+            proof fn prop(chain: Seq<Option<A>>) {}
+        }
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel:
I've filed a different pull request for this: https://github.com/verus-lang/verus/pull/1751 .

zhengyao-lin:
Sounds good. Thanks for fixing it

PR #1747, opened 2025-06-17
Support <T as Pointee>::Metadata

tjhance:
This improves the usability of the raw_ptr library.

We implement the encoding in a similar way we do for `Sized`, using the decoration.


source/vir/src/prelude.rs
```diff
@@ -273,6 +280,63 @@ pub(crate) fn prelude_nodes(config: PreludeConfig) -> Vec<Node> {
         )))
         (axiom ([sized] [decorate_nil_sized]))
 
+        // The <T as Pointee>::Metadata projection type is special. Like Sized-ness, it depends
+        // only on the decoration.
+
+        // For Sized types: Metadata = ()
```
Chris-Hawblitzel: Just one request: can you generate these axioms only when Pointee or PointeeMetadata is used, similar to how `prelude::array_functions` and `prelude::strslice_functions` work?

tjhance: done.

PR #1748, opened 2025-06-17
pcm: remove mutability requirement for validate_2

bsdinis:
validate_2 should not require mutable refs


bsdinis:
You're right! this makes no sense

PR #1749, opened 2025-06-17
examples: improve monotonic counter pcm example

bsdinis:
Also adds pcm examples to the testing infrastructure (they were not there before).

Under the monotonic lemma, a lower bound resource is a lower bound on the value (compared to the other resources)


jaylorch (approved):
Thanks for doing this!

PR #1751, opened 2025-06-18
Treat fresh variables separate from free variables in subst_exp

Chris-Hawblitzel:
(fixes https://github.com/verus-lang/verus/issues/1511 )

`subst_exp` performs capture-avoiding substitution, which means that when substituting `ex` for `x` inside some expression `e`, binders inside `e` need to be renamed to avoid conflicting with free variables in `ex`.  When renaming, `subst_exp` has to generate a fresh variable name that also does not conflict with `ex` or `e`.  The earlier pull request https://github.com/verus-lang/verus/pull/1015 fixed an issue where fresh variables were not chosen freshly enough, potentially causing conflicts.  However, that pull request added extra variables to the free variables and substitution map, causing a different problem, where expression substitutions were performed even when `allow_unfinalized` was true and only type substitutions were allowed (leading to https://github.com/verus-lang/verus/issues/1511 ).

This pull request separates tracking of free variables from tracking of fresh variables.  This maintains the invariant that if `allow_unfinalized` is true, then `substs` is empty and no expression substitutions are performed, only type substitutions.

### DIFF
--- a/source/rust_verify_test/tests/regression.rs
+++ b/source/rust_verify_test/tests/regression.rs
@@ -1443,3 +1443,35 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] subst_in_traits verus_code! {
+        // https://github.com/verus-lang/verus/issues/1511
+        use vstd::prelude::*;
+
+        struct A {
+            a: bool,
+        }
+
+        struct B;
+
+        trait TraitA {
+            proof fn prop(chain: Seq<Option<A>>, i: int)
+                ensures
+                    0 <= i < chain.len() ==> {
+                        let c = chain[i];
+                        if c is Some {
+                            let MISSING_KEY = c->0;
+                            MISSING_KEY.a
+                        } else {
+                            false
+                        }
+                    }
+                    ;
+        }
+
+        impl TraitA for B {
+            proof fn prop(chain: Seq<Option<A>>, i: int) { assume(false); }
+        }
+    } => Ok(())
+}
--- a/source/vir/src/sst_util.rs
+++ b/source/vir/src/sst_util.rs
@@ -89,12 +89,15 @@ pub fn subst_typ_for_datatype(
 
 struct SubstCtxt<'a> {
     typ_substs: &'a HashMap<Ident, Typ>,
+    // free variables in the Exps in SubstState.substs:
     allow_unfinalized: bool,
 }
 
 struct SubstState {
     substs: ScopeMap<UniqueIdent, Exp>,
     free_vars: ScopeMap<UniqueIdent, ()>,
+    // variables already in scope, which we therefore cannot use when picking a fresh variable name:
+    fresh_var_blacklist: ScopeMap<UniqueIdent, ()>,
 }
 
 fn subst_rename_binders<A: Clone, FA: Fn(&A) -> A, FT: Fn(&A) -> Typ>(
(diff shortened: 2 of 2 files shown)


tjhance:
can you remind me why finalization has to happen in a separate pass, and why we don't do that second pass immediately?

Chris-Hawblitzel:
> can you remind me why finalization has to happen in a separate pass, and why we don't do that second pass immediately?

When we create an `ExpX::Var`, we don't yet know whether that variable will be Exp-bound or Stm-bound.  We only discover that information later, when the block declaring the variable is completed.  We could eagerly rename everything in the block when the block is completed, but this would be inefficient as nested blocks would repeatedly process the same subexpressions, since the same subexpressions may contain variables declared in different nested blocks.  For efficiency, we do all the renaming at once at the end.

```
    // Decide final name in a second pass, when a decision has been made on whether the
    // variable is Exp-bound or Stm-bound.  (We start with Stm-bound, and replace with
    // Exp-bound if possible.)
    rename_delayed: HashMap<VarIdent, VarIdent>,
```

tjhance:
Is there a reason finalization is delayed? i.e. is there a reason to not do it when `check_pure_expr` returns? (say, is there any case where we don't know if a variable is exp-bound or stm-bound until even later?)

Chris-Hawblitzel:
> Is there a reason finalization is delayed?

Here's an example:

```
proof fn p() {
}

proof fn test() {
    let b0 = {
        let x0 = true;
        let b1 = {
            let x1 = true;
            let y1 = x1 && x0;
            y1
        };
        let b2 = {
            let x2 = true;
            let y2 = x2 && x0;
            p();
            y2
        };
    };
}
```

The generated AIR for this is:

```
(check-valid
 (declare-const no%param Int)
 (declare-const x2@ Bool)
 (declare-const y2@ Bool)
 (declare-const x0@ Bool)
 (declare-const b1@ Bool)
 (declare-const b2@ Bool)
 (declare-const b0@ tuple%0.)
 (axiom fuel_defaults)
 (block
  (assume
   (= x0@ true)
  )
  (assume
   (= b1@ (let
     ((x1$ true))
     (let
      ((y1$ (and
         x1$
         x0@
      )))
      y1$
  ))))
  (assume
   (= x2@ true)
  )
  (assume
   (= y2@ (and
     x2@
     x0@
  )))
  (block)
  (assume
   (= b2@ y2@)
  )
  (assume
   (= b0@ (tuple%0./tuple%0))
)))
```

When `expr_to_stm_opt` returns for `x1 && x0` and for `x2 && x0`, we don't yet know whether `x0`, `x1`, and `x2` will be Exp-bound or Stm-bound; it's only later that we find that x1 is Exp-bound, x2 is Stm-bound, and x0 is Stm-bound.

> i.e. is there a reason to not do it when check_pure_expr returns? (say, is there any case where we don't know if a variable is exp-bound or stm-bound until even later?)

check_pure_expr is a wrapper around expr_to_stm_opt; it may be the case that there are some places that we could apply the renaming earlier than we currently do, but since check_pure_expr can be called from within blocks recursively inside expr_to_stm_opt, it would be tricky to avoid the inefficent repeated processing of the same subexpressions.

tjhance:
I see, I wrongly thought `check_pure_expr` was a wrapper around some recursive function, but not recursive itself. That just means I didn't ask the right question.

What I really mean to ask is, "Is there some reason that we can't have the public functions of ast_to_sst.rs apply finalization themselves, but instead have to require the clients of ast_to_sst.rs to call finalize?"

Chris-Hawblitzel:
> What I really mean to ask is, "Is there some reason that we can't have the public functions of ast_to_sst.rs can't apply finalization themselves, but instead have to require the clients of ast_to_sst.rs to call finalize?"

Ah, I understand the question now.  I think it's ideal if the public functions of ast_to_sst.rs apply finalization, as you say, so the clients don't have to worry about it.  There are functions, like expr_to_bind_decls_exp_skip_checks, that already work this way.  But it's probably pretty ad-hoc right now; I'm not sure there's anything consistently enforcing this structure.  Maybe we should try to make `ast_to_sst::State` private instead of pub(crate).  It looks pretty close to being privatizable already.

tjhance:
If you think it's doable, I think it would be a big improvement.

PR #1753, opened 2025-06-18
Add support for Opaque types defined by function returns

FeizaiYiHao:
This PR adds basic support for opaque types defined by function returns. The changes include: 

- In lifetime check, all opaque types are normalized to concrete types.
- In MIR to VIR:
- -  a new enum `TypX::Opaque{id:Ident,trait_bounds:GenericBounds}` is added, in which the `id` is generated based on the def_id and type parameters (so each unique opaque type would have an unique `id`. Opaque types defined by the same function with same parameters would have the same `id`, so that we don't define the same opaque type twice in Z3). For example, in the following code: `fn foo<T>()-> impl Some_trait<T> 
let x = foo::<X>(); let y = foo::<Y>(); let Z = foo::<X>();` `X, Z` would have the same `id`, but `X, Y` would have different `id`.
- - Opaque types are first normalized to concrete types to make sure the actual type is supported by Verus. We also throw an error if the actual type is `FnDef` or `Closure` for now. But the MIR to VIR lowering does not use the normalized types.
- - The `trait_bounds` come from instantiating the `item_bounds` of the opaque type
- In prune, the `trait_bounds` are sent to `traverse_generic_bounds` for checking reachable traits and structs. 
- In SST to AIR:
- - Each opaque type follows the SMT encoding of type parameters with trait bounds. For example: `fn<T> foo() where T: Some_trait` declares `T Dcr`, `T type` and the trait bounds for `T`. For each opaque type `impl Some_trait` with an `id`, the SMT encode is effectively `fn<id> foo where id:Some_trait`. The only difference is that `id` is generated by us during MIR to VIR lowering.
- - For each opaque type, since the trait bounds can recursively refer to other opaque types, we recursively traverse the trait bounds and declares the missing opaque types.

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -959,6 +959,7 @@ pub struct AutoExtEqual {
     pub assert: bool,
     pub assert_by: bool,
     pub ensures: bool,
+    pub invariant: bool,
 }
 
 pub type FunctionAttrs = Arc<FunctionAttrsX>;
--- a/source/vir/src/ast_util.rs
+++ b/source/vir/src/ast_util.rs
@@ -1186,7 +1186,12 @@ impl HeaderExprX {
 
 impl Default for crate::ast::AutoExtEqual {
     fn default() -> Self {
-        crate::ast::AutoExtEqual { assert: true, assert_by: false, ensures: false }
+        crate::ast::AutoExtEqual {
+            assert: true,
+            assert_by: false,
+            ensures: false,
+            invariant: false,
+        }
     }
 }
 
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -525,8 +525,12 @@ pub(crate) fn parse_attrs(
                     v.push(Attr::LoopIsolation(false))
                 }
                 AttrTree::Fun(span, arg, Some(places)) if arg == "auto_ext_equal" => {
-                    let mut auto_ext_equal =
-                        vir::ast::AutoExtEqual { assert: false, assert_by: false, ensures: false };
+                    let mut auto_ext_equal = vir::ast::AutoExtEqual {
+                        assert: false,
+                        assert_by: false,
+                        ensures: false,
+                        invariant: false,
+                    };
                     for place in places.into_iter() {
                         if let AttrTree::Fun(_, r, None) = place {
                             match &**r {
@@ -540,6 +544,10 @@ pub(crate) fn parse_attrs(
                                 }
                                 "ensures" => {
                                     auto_ext_equal.ensures = true;
+                                    continue;
+                                }
+                                "invariant" => {
+                                    auto_ext_equal.invariant = true;
                                     continue;
                                 }
                                 _ => {}
--- a/source/rust_verify_test/tests/ext_equal.rs
+++ b/source/rust_verify_test/tests/ext_equal.rs
@@ -434,6 +434,45 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] heuristic_invariant verus_code! {
+        use vstd::prelude::*;
+
+        fn test_invariant1(Ghost(s): Ghost<Seq<int>>) {
+            let ghost t = s.push(5).drop_last();
+            loop
+                invariant s == t, // FAILS
+                decreases 1int,
+            {
+                break;
+            }
+        }
+
+        #[verifier::auto_ext_equal(invariant)]
+        fn test_invariant2(Ghost(s): Ghost<Seq<int>>) {
+            let ghost t = s.push(5).drop_last();
+            loop
+                invariant s == t,
+                decreases 1int,
+            {
+                break;
+            }
+        }
+
+        #[verifier::auto_ext_equal(invariant)]
+        fn test_invariant3(Ghost(s): Ghost<Seq<int>>) {
+            let ghost mut t = s.push(5).drop_last();
+            for i in 0..3
+                invariant s == t,
+            {
+                proof {
+                    t = s.push(6).push(7).drop_last().drop_last();
+                }
+            }
+        }
+    } => Err(err) => assert_one_fails(err)
+}
+
+test_verify_one_file! {
     #[test] heuristic_ensures verus_code! {
         use vstd::prelude::*;
 
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -2043,6 +2043,8 @@ pub(crate) fn expr_to_stm_opt(
             let mut invs1: Vec<crate::sst::LoopInv> = Vec::new();
             for inv in invs.iter() {
                 let (rec, exp) = expr_to_pure_exp_check(ctx, state, &inv.inv)?;
+                let exp =
+                    crate::heuristics::maybe_insert_auto_ext_equal(ctx, &exp, |x| x.invariant);
                 check_recommends.extend(rec);
                 let (at_entry, at_exit) = match inv.kind {
                     LoopInvariantKind::InvariantExceptBreak => (true, false),


source/rust_verify/src/lifetime_generate.rs
```diff
@@ -421,6 +421,15 @@ fn erase_ty<'tcx>(ctxt: &Context<'tcx>, state: &mut State, ty: &Ty<'tcx>) -> Typ
                 Box::new(TypX::TypParam(state.typ_param("Self", None)))
             }
         }
+        TyKind::Alias(rustc_middle::ty::AliasTyKind::Opaque, ..) => {
```
dschoepe: Nit: I'd put this case next to the other `TyKind::Alias` arm below.

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -1068,8 +1073,144 @@ pub(crate) fn mid_ty_to_vir_ghost<'tcx>(
...
+                    unsupported_err!(span, "type inference error types for opaque types")
+                }
+                TyKind::CoroutineClosure(_, _) => {
+                    unsupported_err!(span, "coroutine closure types for opaque types")
+                }
+                TyKind::Pat(_, _) => unsupported_err!(span, "pattern types for opaque types"),
+                TyKind::UnsafeBinder(_) => {
+                    unsupported_err!(span, "unsafe binder types for opaque types")
+                }
+                _ => {}
```
dschoepe: Would it be easier to explicitly list what's supported instead? That'd protect against other variants being added later on.

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -1068,8 +1073,144 @@ pub(crate) fn mid_ty_to_vir_ghost<'tcx>(
...
+            }
+
+            let id: String = format!(
```
dschoepe: Nit: I guess it's unlikely, but do we need to make sure that this doesn't clash with an existing type parameter?

source/rust_verify_test/tests/opaque_types.rs
```diff
@@ -0,0 +1,101 @@
...
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] test_return_opaque_type verus_code! {
+        use vstd::prelude::*;
+        trait DummyTrait{}
+        impl DummyTrait for bool{}
+        fn return_opaque_variable() -> impl DummyTrait{
```
dschoepe: Nit: missing space: `.. -> impl DummyTrait {`

source/rust_verify_test/tests/opaque_types.rs
```diff
@@ -0,0 +1,101 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] test_return_opaque_type verus_code! {
+        use vstd::prelude::*;
+        trait DummyTrait{}
+        impl DummyTrait for bool{}
```
dschoepe: Nit: missing space in `.. for bool {}`

source/rust_verify_test/tests/opaque_types.rs
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+        fn return_opaque_variable() -> impl DummyTrait<Output = bool>{
+            true
+        }
+        fn test(){
+            let x = return_opaque_variable();
+            let output = x.get_output();
+            // Here Verus should be able to infer the type of output, but assertion should fail
+            assert(output);  // FAILS
```
dschoepe: This would fail even if we exposed the concrete type to the verifier here, since verus doesn't expose the function body of `return_opaque_variable` anyway. Maybe you could add a test where the trait `impl` strengthens the post-condition like this:

```rust
    trait MyTrait {
        fn foo(&self, x: u64) -> (res: u64)
            ensures res > 10;
    }
    impl MyTrait for bool {
        
        fn foo(&self, x: u64) -> (res: u64)
          ensures res > 20
        {
            42
        }
    }

  fn returns_my_trait() -> impl MyTrait { true }

  fn uses_opaque_ty() {
    let x = returns_my_trait();
    // this should fail because the stronger post-condition only applies to `bool`:
    assert(x > 20); // FAILS
  }
```

source/vir/src/ast_util.rs
```diff
@@ -171,7 +171,14 @@ pub fn types_equal(typ1: &Typ, typ2: &Typ) -> bool {
...
             f1 == f2 && n_types_equal(ts1, ts2)
         }
-        // rather than matching on _, repeat all the cases to catch any new variants added to TypX:
+        (
+            TypX::Opaque { trait_bounds: trait_bounds1, .. },
+            TypX::Opaque { trait_bounds: trait_bounds2, .. },
+        ) => {
+            // Not so sure about this. If two variables are returned from the same opaque function with the same type prams,
+            // it is sound to compare them.
+            *trait_bounds1 == *trait_bounds2
```
dschoepe: Do we actually know here that the two opaque types were returned by the same function?

source/vir/src/datatype_to_air.rs
```diff
@@ -97,6 +97,8 @@ fn uses_ext_equal(ctx: &Ctx, typ: &Typ) -> bool {
         TypX::Primitive(crate::ast::Primitive::Ptr, _) => false,
         TypX::Primitive(crate::ast::Primitive::Global, _) => false,
         TypX::FnDef(..) => false,
+        // Revisit
```
dschoepe: I think this is correct here since we don't have a way (as far as I know) to indicate that all implementing types of a trait should be compared using extensional equality.

source/vir/src/heuristics.rs
```diff
@@ -26,6 +26,8 @@ fn auto_ext_equal_typ(ctx: &Ctx, typ: &Typ) -> bool {
         TypX::Primitive(crate::ast::Primitive::Ptr, _) => false,
         TypX::Primitive(crate::ast::Primitive::Global, _) => false,
         TypX::FnDef(..) => false,
+        //Revisit
```
dschoepe: As above, this looks correct to me.

source/vir/src/sst_to_air.rs
```diff
@@ -383,6 +384,10 @@ pub fn typ_to_ids(ctx: &Ctx, typ: &Typ) -> Vec<Expr> {
             &vec![Arc::new(ExprX::Const(Constant::Bool(*b)))],
         )),
         TypX::Air(_) => panic!("internal error: typ_to_ids of Air"),
+        TypX::Opaque { id, .. } => {
+            let ret = suffix_typ_param_ids(id).iter().map(|x| ident_var(&x.lower())).collect();
```
dschoepe: Nit: you can remove the assignment to `ret` and return the right-hand side directly.

source/vir/src/sst_to_air.rs
```diff
@@ -2703,6 +2711,48 @@ fn mk_static_prelude(ctx: &Ctx, statics: &Vec<Fun>) -> Vec<Stmt> {
...
+                }
+            }
+            if declared_opaque_typs.insert(id.clone()) {
+                let (x, t) = &crate::def::suffix_typ_param_ids_types(id)[0];
+                local_shared.push(Arc::new(DeclX::Const(x.lower(), str_typ(t))));
+                let (x, t) = &crate::def::suffix_typ_param_ids_types(id)[1];
+                local_shared.push(Arc::new(DeclX::Const(x.lower(), str_typ(t))));
+            }
+            for e in crate::traits::trait_bounds_to_air(ctx, trait_bounds) {
+                // println!("Real trait bound {:#?}", e);
```
dschoepe: Nit: I'd remove the debug print here.

source/vir/src/sst_to_air.rs
```diff
@@ -2752,12 +2802,52 @@ pub(crate) fn body_stm_to_air(
...
-            Arc::new(DeclX::Var(suffix_local_unique_id(&decl.ident), typ_to_air(ctx, &decl.typ)))
-        } else {
-            Arc::new(DeclX::Const(suffix_local_unique_id(&decl.ident), typ_to_air(ctx, &decl.typ)))
-        });
+        match &*decl.typ {
+            TypX::Opaque { id, trait_bounds: _ } => {
+                // for opaque tyeps, recursively declare all the opaque types and their trait bounds
+                decl_opaque_ty(ctx, &mut local_shared, &decl.typ, &mut declared_opaque_typs);
+
+                // if declared_opaque_typs.insert(id.clone()){
```
dschoepe: Nit: remove the commented out code.

source/vir/src/sst_to_air.rs
```diff
@@ -2841,6 +2933,8 @@ pub(crate) fn body_stm_to_air(
 
     let mut stmts = stm_to_stmts(ctx, &mut state, &stm)?;
 
+    // println!("state.commands after stm_to_stmts: {:#?}", state.commands);
```
dschoepe: Nit: remove the commented out code.

source/vir/src/sst_to_air.rs
```diff
@@ -2831,6 +2921,8 @@ pub(crate) fn body_stm_to_air(
 
     let mut _modified = IndexSet::new();
 
+    // println!("declared: {:#?}", declared);
```
dschoepe: Nit: remove the commented out code.

source/vir/src/ast_util.rs
FeizaiYiHao: Oh wait... This only compares the trait bounds. We could compare their `id`s instead. That will make sure that the two opaque types are from the same function (def_id) with the same type parameters.

source/vir/src/ast.rs
```diff
@@ -246,6 +246,12 @@ pub enum TypX {
     FnDef(Fun, Typs, Option<Fun>),
     /// Datatype (concrete or abstract) applied to type arguments
     Datatype(Dt, Typs, ImplPaths),
+    Opaque {
+        /// an id generated from its def_id and type prams.
+        id: Ident,
+        /// the trait bounds of the opaque type
+        trait_bounds: GenericBounds,
+    },
```
tjhance: Would it make more sense for the Krate object to have a global list of opaque types with all the trait bounds?

source/rust_verify/src/rust_to_vir_base.rs
FeizaiYiHao: good idea

tjhance (changes requested):
I found a few cases that don't seem to be handled.

Using an opaque type in a spec function for example:

```
trait Tr {
}

impl Tr for u64 {
}

spec fn t() -> impl Tr {
    5u64
}
```

This gives an error for invalid AIR (though I'd argue this should probably just be disallowed).

You also get weird error messages if you try to combine an opaque type with an ensures clause:

```
fn t() -> (ret: impl Tr)
    ensures ret == 5u64
{
    5u64
}
```

```
error[E0562]: `impl Trait` is not allowed in closure parameters
  --> test2.rs:11:17
   |
11 | fn t() -> (ret: impl Tr)
   |                 ^^^^^^^
   |
```

or with a returns clause:

```
fn t() -> impl Tr
    returns 5u64
{
    5u64
}
```

```
error: type of `returns` clause does not match function return type
  --> test2.rs:12:13
   |
11 | fn t() -> (ret: impl Tr)
   | ----------------------- this function returns `opaquety`
12 |     returns 5u64
   |             ^^^^ the `returns` clause has type `u64`
```

source/rust_verify_test/tests/opaque_types.rs
FeizaiYiHao: Actually it's not the case... 
```
trait MyTrait {
    fn foo(&self, x: u64) -> (res: u64)
        ensures res > 10;
}
impl MyTrait for bool {
    
    fn foo(&self, x: u64) -> (res: u64)
        ensures res > 20
    {
        42
    }
}


fn returns_my_trait() -> impl MyTrait { true }

fn uses_opaque_ty() {
let x = returns_my_trait();
let y = x.foo(1);
// this doesn't fail
assert(y > 20); 
}
```
The function Verus picks up for `x` here is the implementation of `foo` for `bool`, not the trait definition of `foo`.

source/vir/src/ast.rs
FeizaiYiHao: Thank you for finding these unhandled error msgs.

FeizaiYiHao: I assume you are suggesting to have a list of all the opaque types (created by function returns), not each opaque types that are referenced. 
Here's what I mean
```
// creates an opaque type by function return
// ret has trait bounds of SomeTrait<T>
fn return_opaque<T>() -> (ret: impl SomeTrait<T>) { ... }

fn use_opaque(){
   // reference an opaque type by calling the function
   // x has trait bounds of SomeTrait<usize>
   let x = return_opaque::<usize>();
}
```
The problem is that `x` actually has different trait bounds than `ret`. Currently I instantiate the trait bounds of `x` when doing the HIR to VIR lowering. Do you suggest something different?

tjhance: As I understand it, Rust creates one "opaque type (constructor)" which takes a type parameter `T`. Then at the call-site it would instantiate the opaque type with `usize`. I think that it makes much more sense for the VIR nodes to just have one opaque type-constructor which takes type arguments rather than to create a different opaque type for every instantiation. This both better reflects the structure of the rustc_middle::ty, and it is more compatible with the way we currently handle generics in AIR.

FeizaiYiHao: I agree. How do we instantiate generic bounds in the SST to AIR lowering phase? The only reason I did it this way is because Rustc has these nice `item_bounds()` and `instantiate()` functions to instantiate the opaque types' trait bounds. It'd be more elegant if we only instantiate when we are about to produce AIR. I assume it'd similar to this? 
```
fn foo<T>() -> SomeStruct<T> { ... }
fn bar(){
   // x's type needs to be instantiated.
   let x = foo::<usize>();
}
```

FeizaiYiHao: @tjhance I think the alternative would be: for each "opaque type (constructor)", we create a new struct similar to `DatatypeX`. Similar to each datatype, we declare all the opaque constructors first, and instantiates each instance of the opaque type with `has_type`. I guess the high level picture of this opaque constructors would be 
```
// an opaque type opaque_ty_0 defined by a function
fn foo<T>() -> ret: impl SomeTrait

// in AIR pseudocode 
Forall (t: Type) : => (has_type t opaque_ty_0) (tr_bound%SomeTrait t)

// If we use the opaque ty
let x = foo<usize>;
// in AIR it should be 
(declare-const x@ opaque_ty_0.)
(has_type (Poly%opaque_ty_0. x)
     (TYPE%opaque_ty_0. usize usize)
```
Sorry for my horrible AIR pseudocode. Let me know if this make sense.

tjhance:
Something like this:

```
fn foo<T>() -> impl SomeTrait
```

Ought to be encoded the same way as this:

```
#[verifier::external_body] // abstract type
struct OpaqueTy<T> { ... }

fn foo<T>() -> OpaqueTy<T>

impl<T> SomeTrait for OpaqueTy<T> { ... }
```

The `impl` essentially gets translated into a global axiom saying that `OpaqueTy<T>` has the given trait bound. `OpaqueTy<T>` would be translated by the "MonoTyp" mechanism, which creates different SMT sorts for different concrete instantiations.

However, I think there's a way you can implement this where you don't have to worry about all these details (i.e., leave everything to the existing sst_to_air machinery). Take a look at [how we handle anonymous closure types in ast_simplify.rs](https://github.com/verus-lang/verus/blob/c2a4cd1d833ae884c40410d351b6e9b25ee49cdc/source/vir/src/ast_simplify.rs#L1248): We replace all anonymous closure types with abstract datatypes, so we don't need specialized logic for anonymous closure types at lower levels of the pipeline. You can do something similar for opaque types: create a new datatype to represent the closure, and add corresponding impls to the Krate.

It might be good for @Chris-Hawblitzel to confirm this seems like a good approach.

FeizaiYiHao:
If the only purpose is to somehow encode that the return value of `foo` satisfies the trait bounds, do we actually need to declare a new sort for each Opaque constructor? I'm thinking about defining each variable with Opaque type as `Poly` and define a function 
```
Type%foo_opaque_constructor (Dcr Tyle) Type
```
And adding axioms to say that for any input Dcr, Type, the type created by this constructor satisfies all the traits bounds, and using `has_type` to bound the Poly with the type. 

This would be still close to how a type parameter with trait bounds encoded in AIR, but I'm not sure if this makes sense

But more importantly, whenever we invoke any function on an Opaque type variable, Rustc automatically invokes the function of the concrete type behind the opaque type. So I don't even know if encoding the trait bounds for each opaque type variable has any actual effect at all...

FeizaiYiHao:
@tjhance `type NestedFilter = rustc_middle::hir::nested_filter::All;` but why are all the `OpaqueDef` not visited? I'm confused

FeizaiYiHao:
@Chris-Hawblitzel @tjhance 
I haven't push the newest code, but here's the new plan
For a function that returns an opaque type:
```
fn foo<M>() -> impl YouTrait { true }
```
We generate the following Dcr and Type constructor with axioms:
```
;; Opaque-Type-Constructor
(declare-fun DCR%async!foo.opaque0. (Dcr Type) Dcr)
(declare-fun TYPE%async!foo.opaque0. (Dcr Type) Type)
(assert
 (forall ((M&. Dcr) (M& Type)) (!
   (and
    (tr_bound%async!YouTrait. (DCR%async!foo.opaque0. M&. M&) (TYPE%async!foo.opaque0.
      M&. M&
    ))
    (sized (DCR%async!foo.opaque0. M&. M&))
   )
   :pattern ((DCR%async!foo.opaque0. M&. M&) (TYPE%async!foo.opaque0. M&. M&))
   :qid internal_async__foo__opaque0_trait_impl_definition
   :skolemid skolem_internal_async__foo__opaque0_opaque_constructor
)))
```
Therefore:
```
let x = foo::<bool>(); // x is a Poly
```
should have :
```
 (declare-const x@ Poly)
has_type x@ (TYPE%async!foo.opaque0. $ BOOL)
```
which are included in the postcondition of `foo`:
```
(declare-fun ens%async!foo. (Dcr Type Poly) Bool)
(assert
 (forall ((M&. Dcr) (M& Type) (%return! Poly)) (!
   (= (ens%async!foo. M&. M& %return!) (has_type %return! (TYPE%async!foo.opaque0. M&.
      M&
   )))
   :pattern ((ens%async!foo. M&. M& %return!))
   :qid internal_ens__async!foo._definition
   :skolemid skolem_internal_ens__async!foo._definition
)))
```
The reason to do this is because opaque types can define nested opaque types. What @tjhance suggested may not work. (I couldn't think it through.)
Let me know what you think. I don't know if `DCR%async!foo.opaque0.` is needed at all.

And more importantly, Verus (or Rustc) automatically uses the right impl function of opaque types (if we do `x.some_trait_function()`, Verus now knows `x` is `bool` and invokes the implementation of `bool` directly), which makes me think that all these trait bounds encoding for the opaque types are actually meaningless? Since we cannot access any field of the opaque type, and the only way to access the state of the opaque type is through some spec or exec function, the moment we want to do anything with the opaque type variable, we (in z3) would know the actual type of the opaque type variable.

PR #1755, opened 2025-06-18
Vec::swap_remove

lengyijun:

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -556,6 +556,30 @@ impl<A> Seq<A> {
 
     }
 
+    /// Remove item a at index i, pop the last item and write the last item at i
+    pub open spec fn swap_remove(self, i: int) -> Seq<A>
+        recommends
+            0 <= i < self.len(),
+    {
+        if i + 1 == self.len() {
+            self.subrange(0, self.len() - 1)
+        } else {
+            self.subrange(0, i).push(self.last()) + self.subrange(i+1, (self.len() - 1) as int)
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


source/vstd/seq_lib.rs
```diff
@@ -496,6 +496,26 @@ impl<A> Seq<A> {
 
     }
 
+    /// Remove item a at index i, pop the last item and write the last item at i
+    pub open spec fn swap_remove(self, i: int) -> Seq<A>
+        recommends
+            0 <= i < self.len(),
+    {
+        self.subrange(0, i).push(self.last()) + self.subrange(i+1, (self.len() - 1) as int)
+    }
```
tjhance, line 569: I don't think this has the correct behavior for `i == self.len() - 1`.

When `i == self.len() - 1` the correct definition should just pop the last element, but in this definition we would have `self.subrange(0, i).push(self.last())` reconstructing the entire sequence, and then `self.subrange(i+1, (self.len() - 1) as int)` violating its recommends-clause.

Chris-Hawblitzel:
I would suggest just doing this:

```
pub assume_specification<T, A: Allocator>[ Vec::<T, A>::swap_remove ](
    vec: &mut Vec<T, A>,
    i: usize,
) -> (element: T)
    requires
        i < old(vec).len(),
    ensures
        element == old(vec)[i as int],
        vec@ == old(vec)@.update(i as int, old(vec)@.last()).drop_last(),
;
```

Personally, I don't think you need to change seq_lib.rs for this, since the ensures for `vec@` is a one-liner that's not commonly used elsewhere.

PR #1766, opened 2025-06-20
Split set, map into Finite & Infinite versions by generic type parameter

jonhnet:
ISet<A> and Set<A> (finite) are now type synonyms for the generic GSet<A, const FINITE: bool>. The only constructors for Set (GSet<A, true>) enforce finiteness, and hence proof contexts learn `.finite()` automatically. This new design is motivated because finite sets are quite common in user code, and it's easy to go down a wild goose chase trying to get an ambient broadcast property to instantiate that turns out to be a missing `.finite()`.

The PR makes the analogous changes to Map, creating IMap and finite Map.

The simplest porting path for existing code is to use the infinite versions, which work like the previous single Set and Map. Replace Set with ISet, Map with IMap, `set!` with `iset!`, and `map!` with `imap!`.

To exploit the finite sets, use the Set and Map types. Where you might have specified a set or map domain with a predicate (boolean closure) before, now you might start with a finite constructor and transform it with map or filter.

```
let evens_lt_100 = Set::new(|x| 0 <= x < 100 && x%2==0); // old predicate constructor; don't learn .finite
let evens_lt_100 = Set::int_range(0, 100).filter(|x| x % 2 == 0); // new constructor starts with finite int range

let trees = Set::int_range(5).map(|x| tree_constructor(x)); // Construct a set of trees, one per int [0,5)
```

This PR addresses issue #1512.


jonhnet:
I think it's reasonable to begin reviewing this PR now; I'm sure I'll have feedback to work on. The only significant outstanding issue I'm aware of is that mimalloc doesn't verify under these changes yet.

tjhance:
Don't worry about mimalloc, I will get it working after this is merged in

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
-                if f(a) {
-                    true
-                } else {
-                    false
-                },
+//////////////////////////////////////////////////////////////////////////////
+// Important soundness note!
+//
+// In this file, one can construct GSets directly with GSet{set:...}.
+// Doing so for ISet is always sound, but when the type is Set (finite=true),
```
parno: I think the `finite=true` might be left from the prior encoding of Set/ISet.

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+//
+// Outside of this file, callers only have access to Set constructors that
+// create only finite sets.
+//
+// For future work, we may figure out how to have Set use a seq-like
+// representation that is inherently finite to eliminate this risk. (However,
+// that introduces the problem of multiple representations of equivalent
+// sets, which creates a different problem with extensional equality.)
+//////////////////////////////////////////////////////////////////////////////
+/// Set<A> is a synonym for GSet<A, true>, a set whose membership is finite (known at typechecking time).
```
parno: `true` should be `Finite`, and similar for `false` below in the comment about `ISet`

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+
+/// ISet<A> is a synonym for GSet<A, false>, a set whose membership may be infinite (but can be
+/// proven finite at verification time).
+pub type ISet<A> = GSet<A, Infinite>;
+
+impl<A, FINITE: Finiteness> GSet<A, FINITE> {
+    // This map function is sound for finite sets because of `lemma_set_map_finite`,
+    // but we don't need to invoke that lemma here because this file is trusting
+    // GSet constructors to do so soundly (see "Important soundness note" above).
+    /// Returns the set contains an element `f(x)` for every element `x` in `self`.
```
parno: "the set *that* contains"

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+        FINITE,
+    >) {
+        GSet {
+            set: |b| exists|a| self.contains(a) && f(a).contains(b),
+            _phantom: core::marker::PhantomData,
+        }
+    }
+}
+
+impl<A, FINITE: Finiteness> GSet<A, FINITE> {
```
parno: Why is this a new impl instead of part of the previous impl block?

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+        requires
+            self.finite(),
+        ensures
+            self.cast_finiteness::<Finite>() == (Set {
+                set: self.set,
+                _phantom: core::marker::PhantomData,
+            }),
+    ;
+
+    pub proof fn cast_finiteness_ensures(self)
```
parno: Should this be a broadcast?

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+
+impl<A, FINITE: Finiteness> GSet<A, FINITE> {
+    // This map function is sound for finite sets because of `lemma_set_map_finite`,
+    // but we don't need to invoke that lemma here because this file is trusting
+    // GSet constructors to do so soundly (see "Important soundness note" above).
+    /// Returns the set contains an element `f(x)` for every element `x` in `self`.
+    pub closed spec fn map<B>(self, f: spec_fn(A) -> B) -> GSet<B, FINITE> {
+        GSet {
+            set: |a: B| exists|x: A| self.contains(x) && a == f(x),
+            _phantom: core::marker::PhantomData,
```
parno: You're using `core::marker::PhantomData` a lot below.  Worth adding as a `use` declaration at the top level?

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
-            set: |a|
-                if f(a) {
-                    true
-                } else {
-                    false
-                },
+//////////////////////////////////////////////////////////////////////////////
+// Important soundness note!
+//
+// In this file, one can construct GSets directly with GSet{set:...}.
```
parno: This seems nervous making.  Could we define GSet and any fundamental operations on it inside submodule in this file, so that the rest of the code cannot directly construct a `GSet` but instead must use a safe interface?

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+        GSet { set: |i: nat| lo <= i && i < hi, _phantom: core::marker::PhantomData }
+    }
+}
+
+pub broadcast proof fn lemma_set_finite_from_type<A>(s: Set<A>)
+    ensures
+        #[trigger] s.finite(),
+{
+    // Preserving this is the reason for the trustedness in the
+    // paragraph below. Until we replace the representation of finite
```
parno: Which paragraph is this sentence referring to?

source/vstd/set.rs
```diff
@@ -28,59 +44,349 @@ verus! {
...
+        s2: GSet<A, FINITE2>,
+    )
+        requires
+            #[trigger] self.congruent(s2),
+            self.finite(),
+        ensures
+            self.len() == s2.len(),
+        decreases self.len(),
+    {
+        // TOOD(jonh): tidy up this proof
```
parno: It would be good to cut or implement before merging

source/vstd/set.rs
```diff
@@ -223,11 +588,11 @@ pub mod fold {
     //! lemmas.
     //!
     //! The module was ported (with some modifications) from Isabelle/HOL's finite set theory in:
-    //! `HOL/Finite_Set.thy`
```
parno, line 226: I think this and the next line are the result of an overly aggressive find-replace.

source/vstd/set.rs
```diff
@@ -386,10 +767,26 @@ pub mod fold {
                 &&& y == f(yr, aa)
             };
         assert(trigger_fold_graph(yr, a));
-        if s.remove(aa) === Set::empty() {
+        if s.remove(aa) === GSet::empty() {
         } else {
             if a == aa {
             } else {
+                assume(false);  // flaky!
```
parno: Lurking `assume(false)` here.

source/vstd/set.rs
```diff
@@ -406,21 +803,23 @@ pub mod fold {
                 };
             }
         }
+        assume(false);  // flaky!
```
parno: Another one.

source/vstd/set.rs
```diff
@@ -757,107 +1234,58 @@ pub broadcast proof fn axiom_set_ext_equal<A>(s1: Set<A>, s2: Set<A>)
...
+        #[trigger] GSet::<A, FINITE>::empty().finite(),
 {
     let f = |a: A| 0;
     let ub = 0;
     let _ = trigger_finite(f, ub);
 }
 
 /// The result of inserting an element `a` into a finite set `s` is also finite.
-pub broadcast proof fn axiom_set_insert_finite<A>(s: Set<A>, a: A)
+pub broadcast proof fn lemma_set_insert_finite<A, FINITE: Finiteness>(s: GSet<A, FINITE>, a: A)
```
parno: If I start with a `Set`, then typing of `insert` should tell me that I end up with a `Set`, and there's an existing lemma telling me that `Set` is `finite()`.  So is the main use case for this lemma a scenario when I have an `ISet` that I've already proven is `finite`?

source/vstd/set_lib.rs
```diff
@@ -2,38 +2,74 @@
...
+    /// A singleton set has at least one element and any two elements are equal.
+    pub open spec fn is_singleton(self) -> bool {
+        &&& self.len() > 0
+        &&& (forall|x: A, y: A| self.contains(x) && self.contains(y) ==> x == y)
+    }
+
+    /// An element in an ordered set is called a least element (or a minimum), if it is less than
+    /// every other element of the set.
+    ///
+    /// change f to leq bc it is a relation. also these are an ordering relation
```
parno: This last line doesn't seem like it's intended to be a doc comment.

source/vstd/set_lib.rs
```diff
@@ -2,38 +2,74 @@
...
+        self =~= GSet::<A, FINITE>::empty()
+    }
+
+    /// A singleton set has at least one element and any two elements are equal.
+    pub open spec fn is_singleton(self) -> bool {
+        &&& self.len() > 0
+        &&& (forall|x: A, y: A| self.contains(x) && self.contains(y) ==> x == y)
+    }
+
+    /// An element in an ordered set is called a least element (or a minimum), if it is less than
```
parno, line 83: This says "less than", but the name of the function argument implies "less than or equal to"

source/vstd/set_lib.rs
```diff
@@ -499,21 +554,55 @@ impl<A> Set<A> {
         let x = choose|x: A| self.contains(x) && p(x);
         assert(self.map(f).contains(f(x)));
     }
+}
+
+// TODO: Right now, flatten and filter_map is only defined for ISets. It should work just fine on Sets,
+// too, but it doesn't translate simply into a single generic implementation. That's because
+// TODO(jonh): try defining flatten with a single FINITE arg that agrees and preserves finiteness.
```
parno: Looks like the comment line above got cut off.  Also, it would be useful to restore the ability to flatten a `Set`.  I'm using that in some of my proofs :)

source/vstd/set_lib.rs
```diff
@@ -642,19 +737,38 @@ impl<A> Set<A> {
...
+            assert(self.filter_map(f) =~= ISet::<B>::empty());
         } else {
             let elem = self.choose();
             self.remove(elem).lemma_filter_map_finite(f);
             assert(self =~= self.remove(elem).insert(elem));
         }
     }
+}
 
+proof fn lemma_to_seq_to_set_ix_recursive<A>(set: Set<A>, elem: A)
```
parno: "ix" -> "is"

source/vstd/tokens.rs
parno, line 1: I think there might have already been some discussion on this, but should the state machines be using Set or ISet.  If we're using them to model things like slots in a hash table, or outstanding requests, then the finite varieties seem more natural.

source/vstd/multiset.rs
```diff
@@ -30,6 +32,7 @@ verus! {
 ///
 /// To prove that two multisets are equal, it is usually easiest to use the
 /// extensionality operator `=~=`.
+///
 // We could in principle implement the Multiset via an inductive datatype
 // and so we can mark its type argument as accept_recursive_types.
 // Note: Multiset is finite (in contrast to Set, Map, which are infinite) because it
```
parno, line 38: This note seem to be out of date.

source/vstd/multiset.rs
```diff
@@ -149,9 +157,12 @@ impl<V> Multiset<V> {
     /// Returns a multiset containing the lower count of a given element
     /// between the two sets. In other words, returns a multiset with only
     /// the elements that "overlap".
+    //
+    // Possible soundness issue with non-normalized representations of zero values;
+    // see https://github.com/verus-lang/verus/issues/1664
```
parno: Should we remove this comment, since the corresponding issue has been resolved?

source/vstd/map.rs
```diff
@@ -108,22 +152,246 @@ impl<K, V> Map<K, V> {
...
+                if self.dom().contains(k) && keys.contains(k) {
+                    Some(self[k])
+                } else {
+                    None
+                },
+            _phantom: core::marker::PhantomData,
+        }
+    }
+
+    // DONE(jonh): expose as broadcast lemmas
```
parno: Looks like this can be removed?

source/vstd/map.rs
```diff
@@ -108,22 +152,246 @@ impl<K, V> Map<K, V> {
...
+            m2.dom(),
+        ));
+    }
+
+    pub broadcast proof fn lemma_remove_keys(self, keys: GSet<K, FINITE>)
+        ensures
+            #![trigger(self.remove_keys(keys))]
+            self.remove_keys(keys).dom().to_infinite() == self.dom().generic_difference(keys),
+            self.remove_keys(keys).dom().congruent(self.dom().generic_difference(keys)),
+            // TODO(jonh): ask Chris if there's a better trigger here.
```
parno: Is this still pending?

source/vstd/map.rs
```diff
@@ -108,22 +152,246 @@ impl<K, V> Map<K, V> {
...
+            forall|k|
+                #![auto]
+                self.restrict(keys).dom().contains(k) ==> self.restrict(keys)[k] == self[k],
+            forall|k|
+                #![auto]
+                self.dom().contains(k) && keys.contains(k) ==> self.restrict(keys)[k] == self[k],
+    {
+        broadcast use super::set::group_set_lemmas;
+        broadcast use axiom_dom_ensures;
+        // trigger extn
```
parno: Here and in `lemma_map_entries`, is the extra newline intentional or something verusfmt is adding?

source/vstd/map.rs
```diff
@@ -169,6 +438,90 @@ impl<K, V> Map<K, V> {
...
+}
+
+impl<K, V> IMap<K, V> {
+    /// Gives an `IMap<K, V>` whose domain contains every key, and maps each key
+    /// to the value given by `fv`.
+    pub open spec fn total(fv: spec_fn(K) -> V) -> IMap<K, V> {
+        IMap::new(|k| true, fv)
+    }
+
+    /// Gives a `IMap<K, V>` whose domain is given by the boolean predicate on keys `fk`,
```
parno: "Gives *an*"

source/vstd/map.rs
```diff
@@ -198,35 +551,51 @@ pub broadcast axiom fn axiom_map_index_decreases_infinite<K, V>(m: Map<K, V>, ke
 ;
 
 /// The domain of the empty map is the empty set
-pub broadcast proof fn axiom_map_empty<K, V>()
+pub broadcast proof fn axiom_map_empty<K, V, FINITE: Finiteness>()
```
parno: Should this be named "lemma_" instead of "axiom_"?  Same question for the next several functions.  It seems like this renaming happened in `set.rs`

source/vstd/map.rs
```diff
@@ -198,35 +551,51 @@ pub broadcast axiom fn axiom_map_index_decreases_infinite<K, V>(m: Map<K, V>, ke
...
+    value: V,
+)
     ensures
         #[trigger] m.insert(key, value).dom() == m.dom().insert(key),
 {
-    broadcast use super::set::group_set_axioms;
+    broadcast use super::set::group_set_lemmas;
+    broadcast use axiom_dom_ensures;
+    //     m.axiom_dom_ensures();
+    //     m.insert(key, value).axiom_dom_ensures();
```
parno: Remove these two commented lines?

source/vstd/map.rs
```diff
@@ -236,18 +605,28 @@ pub broadcast proof fn axiom_map_insert_different<K, V>(m: Map<K, V>, key1: K, k
...
+    key: K,
+)
     ensures
         #[trigger] m.remove(key).dom() == m.dom().remove(key),
 {
-    broadcast use super::set::group_set_axioms;
+    broadcast use super::set::group_set_lemmas;
+    broadcast use axiom_dom_ensures;
+    //     m.axiom_dom_ensures();
+    //     m.remove(key).axiom_dom_ensures();
```
parno: Remove these two commented lines?

source/vstd/map.rs
```diff
@@ -256,14 +635,20 @@ pub broadcast proof fn axiom_map_remove_different<K, V>(m: Map<K, V>, key1: K, k
...
     ensures
         #[trigger] (m1 =~= m2) <==> {
             &&& m1.dom() =~= m2.dom()
             &&& forall|k: K| #![auto] m1.dom().contains(k) ==> m1[k] == m2[k]
         },
 {
-    broadcast use super::set::group_set_axioms;
+    broadcast use axiom_dom_ensures;
+    //     m1.axiom_dom_ensures();
+    //     m2.axiom_dom_ensures();
```
parno: Remove these two commented lines?

source/vstd/map.rs
```diff
@@ -331,15 +728,44 @@ macro_rules! map {
     };
 }
 
+/// Create an IMap using syntax like `map![key1 => val1, key2 => val, ...]`.
```
parno: I think you want "imap!", and below, you want "IMap::empty()..."

source/vstd/map_lib.rs
```diff
@@ -666,30 +580,39 @@ pub broadcast proof fn lemma_submap_of_trans<K, V>(m1: Map<K, V>, m2: Map<K, V>,
...
+pub broadcast proof fn lemma_imap_new_domain<K, V>(fk: spec_fn(K) -> bool, fv: spec_fn(K) -> V)
     ensures
-        #[trigger] Map::<K, V>::new(fk, fv).dom() == Set::<K>::new(|k: K| fk(k)),
+        #[trigger] IMap::<K, V>::new(fk, fv).dom() == ISet::<K>::new(|k: K| fk(k)),
 {
-    assert(Set::new(fk) =~= Set::<K>::new(|k: K| fk(k)));
+    assert(IMap::new(fk, fv).dom() =~= ISet::<K>::new(|k: K| fk(k)));
+}
+
+// finite variant of same
```
parno: Should we have a doc comment here?

parno (commented):
Thanks for all of the work you've put into this!  I made a pass through and added a bunch of (mostly minor) comments.  Could you also update the corresponding Markdown tutorial page (`docs/guide/spec_lib.md`), since I think it's affected by these changes?

source/vstd/set.rs
jonhnet: turns out it was redundant, nobody needed it. So I deleted it instead.

source/vstd/set.rs
jonhnet: I agree it's ominous. I played around with that early on, but couldn't find a way to set up the submodule that prevented others from constructing it while still giving them access to look at the struct's guts to prove properties about the resulting struct. I'm inclined to leave it for future work unless you or others continue to apply pressure here. :v)

source/vstd/set.rs
jonhnet: That's correct. Rustdoc improved.

source/vstd/set_lib.rs
jonhnet: I inherited that comment when moving these fns from relations.rs to simplify the module graph.

source/vstd/set_lib.rs
jonhnet: actually, it was `_id`, and was "temporarily" changed to evade pattern matching, I think. :v) Fixed.

source/vstd/tokens.rs
jonhnet: IIRC @tjhance thought it made sense to keep the infinite power in the state machines. It's conceivable that the most right thing to do is make them generic (GSet), but that's a bigger bite than I'm willing to take in this PR.

source/vstd/multiset.rs
jonhnet: The note about the inductive datatype? No, it's still in play, unfortunately. There's not presently a way to get the accept_recursive_types light magic of Set without being tainted by the dark magic of the ISet.

source/vstd/multiset.rs
jonhnet: ah, that's just merge decay. Good spot.

source/vstd/map.rs
jonhnet: Yeah, actually. @Chris-Hawblitzel can you apply your trigger taste here to see if something is conspicuously poorly chosen?

source/vstd/map.rs
jonhnet: `vargo fmt` is pretty excited about those newlines.

source/vstd/map.rs
jonhnet: It's a hard I.

source/vstd/map.rs
jonhnet: renamed!

source/vstd/set_lib.rs
jonhnet: Not my comments (refactored from another file), but they seem safe to fix.

Chris-Hawblitzel:
I took a look at generalizing int_range and nat_range.  Here's an alternative that doesn't need a trusted definition for int_range and nat_range.  It's based on a trait, so anyone can implement the trait, and it's easy for a macro to implement the trait for all built-in integer types:

```
verus! {

pub broadcast proof fn lemma_to_finite_len<A>(s: ISet<A>)
    requires
        s.finite(),
    ensures
        #[trigger] s.to_finite().len() == s.len(),
    decreases
        s.len(),
{
    if s.len() == 0 {
        assert(s.to_finite().is_empty());
    } else {
        let x = s.choose();
        lemma_to_finite_len(s.remove(x));
        assert(s.to_finite().remove(x) =~= s.remove(x).to_finite());
    }
}

pub trait FiniteRange: Sized {
    spec fn range_iset(lo: Self, hi: Self) -> ISet<Self>;
    spec fn range_len(lo: Self, hi: Self) -> nat;
    proof fn properties(lo: Self, hi: Self)
        ensures
            Self::range_iset(lo, hi).finite(),
            Self::range_iset(lo, hi).len() == Self::range_len(lo, hi);
}

broadcast proof fn range_set_properties<A: FiniteRange>(lo: A, hi: A)
    ensures
        (#[trigger] A::range_iset(lo, hi)).finite(),
        A::range_iset(lo, hi).len() == A::range_len(lo, hi),
{
    A::properties(lo, hi);
}

#[verifier::inline]
pub open spec fn range_set<A: FiniteRange>(lo: A, hi: A) -> Set<A> {
    A::range_iset(lo, hi).to_finite()
}

} // verus!

macro_rules! range_impls {
    ([$($t:ty)*]) => {
        $(
            verus! {
                impl FiniteRange for $t {
                    open spec fn range_iset(lo: Self, hi: Self) -> ISet<Self> {
                        ISet::new(|i: Self| lo <= i < hi)
                    }
                    open spec fn range_len(lo: Self, hi: Self) -> nat {
                        if lo <= hi { (hi - lo) as nat } else { 0 }
                    }
                    proof fn properties(lo: Self, hi: Self)
                        decreases hi - lo
                    {
                        if hi <= lo {
                            assert(Self::range_iset(lo, hi).is_empty());
                        } else {
                            let hi1 = (hi - 1) as $t;
                            Self::properties(lo, hi1);
                            assert(Self::range_iset(lo, hi) =~= Self::range_iset(lo, hi1).insert(hi1));
                        }
                    }
                }
            } // verus!
        )*
    }
}

range_impls!([
    int nat
    usize u8 u16 u32 u64 u128
    isize i8 i16 i32 i64 i128
]);

verus! {

proof fn test() {
    broadcast use {range_set_properties, lemma_to_finite_len};

    let a = range_set(30int, 40int);
    assert(a.len() == 10);
    assert(a.contains(33));
    assert(!a.contains(23));

    let a = range_set(30nat, 40nat);
    assert(a.len() == 10);
    assert(a.contains(33));
    assert(!a.contains(23));
}

} // verus!
```

I'm wondering whether this pattern (define a Set in terms of .to_finite on an open ISet definition, and broadcast its finiteness and len properties) might be easier than built-in closed Set constructors.

source/vstd/set_lib.rs
jonhnet: Well dang, I succumbed to peer pressure and implemented it.

jonhnet:
Okay, I incorporated Chris' generic-ized range constructor.
@parno I have addressed your comments, resolving most. It's ready for another look. Thanks!

pratapsingh1729:
@jonhnet Thanks for working on this---it looks great and I think it will make the vstd Map and Set much easier to use! I think it was [discussed](https://github.com/verus-lang/verus/discussions/1512#discussioncomment-12766994) to allow `accept_recursive_types` on finite Sets and Maps. Is that intended to be part of this PR? In my project I have a recursive data structure which can contain finite sets, so I would need `accept_recursive_types` in order to use the vstd finite Set. So far I've been developing my own very minimal finite set, backed by a `Seq`.

jonhnet:
> I think it was [discussed](https://github.com/verus-lang/verus/discussions/1512#discussioncomment-12766994) to allow `accept_recursive_types` on finite Sets and Maps. Is that intended to be part of this PR?

My understanding is that Verus doesn't have enough fidelity to distinguish between the two cases yet. (From the discussion: "Right now, you can only put one accept/reject-recursive types declaration on any given struct, so I think we'll need a more flexible declaration of accept/reject.") So for this PR, I'm not trying to solve that problem yet.

However, that would be a nice next feature! And when we do it, we could also redefine Multiset in terms of Map (instead of being axiomatized), which would be another nice improvement.

Chris-Hawblitzel:
@jonhnet Do you have the updated Veritas tests somewhere that I could look at?  (Partly to see what changes are involved, and partly to test verification performance.)

jonhnet:
Here's the run configuration I've been using to test 3/4 of the veritas suite.

[run_config.zip](https://github.com/user-attachments/files/21416199/run_config.zip)

It doesn't include the memory allocator. [Travis said above](https://github.com/verus-lang/verus/pull/1766#issuecomment-3036149488) that he'd own that if/after we land this change.

source/vstd/set_lib.rs
parno: Thanks!

source/vstd/multiset.rs
parno: No, sorry, I meant the bit about "in contrast to Set, Map, which are infinite"; that should be "are finite" now, right?

parno (approved):
Generally looks good to me.  I think there's a pending question for @Chris-Hawblitzel in `source/vstd/map.rs` re: triggers.  There's also the open question of whether we can make it easier enforce the "Important soundness note!" in `source/vstd/set.rs`, but perhaps that can wait for a future PR.

source/docs/guide/src/spec_lib.md
```diff
@@ -25,10 +33,16 @@ with particular contents:
 ```rust
 {{#include ../../../../examples/guide/lib_examples.rs:macro}}
 ```
+The `iset!` and `imap!` macros connstruct finite values of the possibly-infinite
```
jaybosamiya-ms, line 36: Minor typo: `connstruct`

source/vstd/set.rs
```diff
@@ -4,12 +4,36 @@ use super::map::*;
...
+
+pub struct Infinite;
+
+pub trait Finiteness {
+    // Let proofs learn an SMT value for the type finiteness, for example so one can
+    // write spec fns that condition on the type.
+    spec fn type_is_finite() -> bool;
+}
+
+// What keeps an adversary from introducing a new Finiteness to break these definitions?
```
jaybosamiya-ms: To answer the "what keeps an adversary ..." question, we should set up a sealed trait (this is a fairly standard Rust idiom, usually used to prevent future-compat hazards)

Specifically, within this module, you can introduce:
```rs
/// A private module, to help support writing sealed traits. This module should _itself_ never be made public.
mod private {
    /// A trait to help seal the `Finiteness` trait.
    ///
    /// This trait is explicitly public, but unnameable (due to being in a private module), thereby preventing code outside here from implementing this trait.
    pub trait Sealed {}
}
```

You can then switch the `Finiteness` trait definition to `pub trait Finiteness: private::Sealed`.

Now, with a public-but-unnameable trait as a super trait, nothing outside vstd can `impl Finiteness`.  However, code within vstd (more specifically, code that can look inside the `private` submodule, which would be `vstd` and `vstd::set` _only_) can `impl private::Sealed`, since they _can_ name it.

source/vstd/set.rs
jaybosamiya-ms: You can't prevent module parents/ancestors from doing arbitrary things, even if stuff is private; _however_, sibling modules are disallowed from looking at private details.  This technically allows you to set up the sort of protections you'd want.  However, I think that the set up for that is not a local-enough change, and needs to touch basically this whole file, and even then, I think we'd need to spend a bunch of time making sure it is set up correctly.  

Thus, I argue that this soundness comment is actually the better approach.

jaybosamiya-ms (commented):
I haven't reviewed the full PR, just a few comments on things I noticed when taking a look

jonhnet:
Capybara ISet translation here:
https://github.com/jonhnet/verified-storage/tree/jonh/sets-typed-finite2

Chris-Hawblitzel:
Closing now that this is merged via https://github.com/verus-lang/verus/pull/2486 .

PR #1773, opened 2025-06-23
add multiset lemmas

ahuoguo:
@bn-peters asked me to prove the following lemma
```rust
proof fn seq_remove_push_update<V>(xs : Seq<V>, i : int, x : V)
    requires
        0 <= i < xs.len(),
    ensures
       xs.push(x).remove(i).to_multiset() =~= xs.update(i, x).to_multiset()
```
which I find surprisingly non-trivial, as there are no lemmas that reason with `update` and `to_multiset`. This PR strengthens the reasoning of multiset.

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -2790,6 +2790,27 @@ pub broadcast proof fn to_multiset_remove<A>(s: Seq<A>, i: int)
     assert(s2 + s0 == (s1 + s0).drop_first());
 }
 
+pub broadcast proof fn to_multiset_insert<A>(s: Seq<A>, i: int, a: A)
+    requires
+        0 <= i <= s.len(),
+    ensures
+        #![trigger s.insert(i, a).to_multiset()]
+        s.insert(i, a).to_multiset() =~= s.to_multiset().insert(a),
+    decreases s.len(),
+{
+    broadcast use {super::multiset::group_multiset_axioms, lemma_seq_union_to_multiset_commutative};
+
+    let s0 = s.subrange(0, i);
+    let s1 = s.subrange(i, s.len() as int);
+
+    assert(s =~= s0 + s1);
+    assert(s.insert(i, a) =~= s0 + seq![a] + s1);
+    assert(((s0 + seq![a]) + s1).to_multiset() =~= ((seq![a] + s0) + s1).to_multiset()) by {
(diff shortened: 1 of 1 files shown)


source/vstd/seq_lib.rs
```diff
@@ -2790,6 +2790,28 @@ pub broadcast proof fn to_multiset_remove<A>(s: Seq<A>, i: int)
     assert(s2 + s0 == (s1 + s0).drop_first());
 }
 
+pub broadcast proof fn to_multiset_insert<A>(s: Seq<A>, i: int, a: A)
+    requires
+        0 <= i <= s.len(),
+    ensures
+        #![trigger s.insert(i, a).to_multiset()]
+        s.insert(i, a).to_multiset() =~= s.to_multiset().insert(a),
```
jaylorch: I suggest replacing this `=~=` in the postcondition with just `==`. The caller doesn't need to be cued to use extensional equality, after all.

source/vstd/seq_lib.rs
```diff
@@ -2845,6 +2867,35 @@ pub broadcast proof fn to_multiset_contains<A>(s: Seq<A>, a: A)
     }
 }
 
+pub broadcast proof fn to_multiset_update<A>(s: Seq<A>, i: int, a: A)
+    requires
+        0 <= i < s.len(),
+    ensures
+        #[trigger] s.update(i, a).to_multiset() =~= s.to_multiset().insert(a).remove(s[i]),
```
jaylorch: I suggest replacing this `=~=` in the postcondition with just `==`. The caller doesn't need to be cued to use extensional equality, after all.

source/vstd/seq_lib.rs
ahuoguo: I need to use `=~=` to remind the solver I'm reasoning with extensional equality on `Multiset`. Replacing with `==` makes verification fail.

PS: `to_multiset_remove` also uses `=~=`.

jaylorch: You don't need it in the postcondition, though. You can use an assertion with extensional equality in the proof so that it doesn't need to show up in the postcondition.

As for another function using it, just because one function does it doesn't mean we have to succumb to peer pressure. :-)

jaylorch (approved):
Thanks for the contribution!

PR #1785, opened 2025-06-26
Add support for Opaque types defined by function returns

FeizaiYiHao:
This PR adds basic support for opaque types defined by function returns. The changes include: 

- In `external.rs`, all opaque type definitions are visited, and only opaque types under `external` functions are marked as `VerifOrExternal::External `, since the opaque type is part of the function declaration. 
- In lifetime check, all opaque types are normalized to concrete types.
- In MIR to VIR:
  -  a new enum `TypX::Opaque{def_path:Path,args:Typs}` is added, in which the `def_path` is the path where the opaque type is defined. And `args` are the type parameters used to instantiate this instance of the opaque type.
  - a new struct`OpaquetypeX{pub name: Path, pub typ_params: Typs, pub typ_bounds: GenericBounds,}` is added for each opaque type definition. A `KrateX` carries a list of `OpaquetypeX` that are used in the crate.
- in VIR to AIR
  - each `OpaquetypeX` generates two functions `DCR%path` and ` TYPE%path` (or consts if takes no args) that return a `Dcr` and `Type` for given args. And also generates a range of axioms to encode the trait bounds for the instantiated opaque types.
  - for each instance of opaque type with `args`, we instantiate the opaque type by calling the `DCR%path args` and `TYPE%path args` functions.

If in `opaque.rs` we have the following Rust code
```
trait Tr{
    type Output;
}
impl<T> Tr for T{
    type Output = T;
}
fn foo<T>() -> impl Tr{
    true
}
fn bar(){
    let x = foo::<usize>();
    assert(x == x);
}
```
Produces the following z3 code
```
;; Opaque-Type-Constructors
(declare-fun DCR%opaque!foo.opaque0. (Dcr Type) Dcr)
(declare-fun TYPE%opaque!foo.opaque0. (Dcr Type) Type)
(assert
 (forall ((T&. Dcr) (T& Type)) (!
   (and
    (tr_bound%opaque!Tr. (DCR%opaque!foo.opaque0. T&. T&) (TYPE%opaque!foo.opaque0. T&.
      T&
    ))
    (sized (DCR%opaque!foo.opaque0. T&. T&))
   )
   :pattern ((DCR%opaque!foo.opaque0. T&. T&))
   :pattern ((TYPE%opaque!foo.opaque0. T&. T&))
   :qid internal_opaque__foo__opaque0_opaque_type_bound_definition
   :skolemid skolem_internal_opaque__foo__opaque0_opaque_type_bound_definition
)))

;; Function-Specs opaque::foo
(declare-fun ens%opaque!foo. (Dcr Type Poly) Bool)
(assert
 (forall ((T&. Dcr) (T& Type) (%return! Poly)) (!
   (= (ens%opaque!foo. T&. T& %return!) (has_type %return! (TYPE%opaque!foo.opaque0. T&.
      T&
   )))
   :pattern ((ens%opaque!foo. T&. T& %return!))
   :qid internal_ens__opaque!foo._definition
   :skolemid skolem_internal_ens__opaque!foo._definition
)))

;; Function-Def opaque::foo
;; ../examples/opaque.rs:16:1: 16:23 (#0)
(get-info :all-statistics)
(push)
 (declare-const T&. Dcr)
 (declare-const T& Type)
 (declare-const %return! Poly)
 (assert
  fuel_defaults
 )
 (assert
  (sized T&.)
 )
 (assert
  (not true)
 )
 (get-info :version)
 (set-option :rlimit 30000000)
 (check-sat)
 (set-option :rlimit 0)
(pop)

;; Function-Def opaque::bar
;; ../examples/opaque.rs:19:1: 19:9 (#0)
(get-info :all-statistics)
(push)
 (declare-const no%param Int)
 (declare-const tmp%1 Bool)
 (declare-const x@ Poly)
 (assert
  fuel_defaults
 )
 ;; assertion failed
 (declare-const %%location_label%%0 Bool)
 (assert
  (not (=>
    (ens%opaque!foo. $ USIZE x@)
    (=>
     (= tmp%1 (= x@ x@))
     (=>
      %%location_label%%0
      tmp%1
 )))))
 (get-info :version)
 (set-option :rlimit 30000000)
 (check-sat)
 (set-option :rlimit 0)
(pop)
```
Two future changes are required:
- Change the encoding of ensures clause so functions with opaque types can have ensure clauses. For now we throw an error for this.
-  Change the function invoked to variables with opaque types. Currently it's the implementation of the normalized type that's being called, but it should be the trait method instead. 


source/vir/src/ast_visitor.rs
```diff
@@ -706,6 +706,7 @@ pub(crate) trait AstVisitor<R: Returner, Err, Scope: Scoper> {
             TypX::ConstInt(_) => R::ret(|| typ.clone()),
             TypX::ConstBool(_) => R::ret(|| typ.clone()),
             TypX::Air(_) => R::ret(|| typ.clone()),
+            TypX::Opaque { .. } => R::ret(|| typ.clone()),
```
Chris-Hawblitzel: This needs to visit the `args`, like the `TypX::Datatype` and `TypX::Primitive` cases.

source/vir/src/ast.rs
```diff
@@ -1255,6 +1264,17 @@ pub struct DatatypeX {
...
 
+/// Opaque type constructors
+#[derive(Clone, Debug, Serialize, Deserialize, ToDebugSNode)]
+pub struct OpaquetypeX {
+    pub name: Path,
+    pub typ_params: Typs,
+    pub typ_bounds: GenericBounds,
+}
+
+pub type Opaquetype = Arc<Spanned<OpaquetypeX>>;
```
Chris-Hawblitzel: Minor comment: `OpaqueType` would be better than `Opaquetype`, since it's two words (whereas "datatype" is often just one word).

FeizaiYiHao:
Passing all the tests now. I'm looking at all the usages of `PostAnalysis`...

source/rust_verify_test/tests/safe_api.rs
```diff
@@ -212,8 +211,7 @@ test_verify_one_file_with_options! {
         pub fn test() -> impl Fn(u32) {
             test2
         }
-    //} => Err(err) => assert_vir_error_msg(err, "Safe API violation: 'requires' clause is nontrivial")
-    } => Err(err) => assert_vir_error_msg(err, "The verifier does not yet support the following Rust feature: opaque type")
+    } => Ok(())
```
tjhance: these tests are meant to be errors

source/vir/src/ast.rs
```diff
@@ -1365,6 +1374,17 @@ pub struct DatatypeX {
 pub type Datatype = Arc<Spanned<DatatypeX>>;
 pub type Datatypes = Vec<Datatype>;
 
+/// Opaque type constructors
+#[derive(Clone, Debug, Serialize, Deserialize, ToDebugSNode)]
+pub struct OpaquetypeX {
```
tjhance: Can you call it `OpaqueType`?

source/vir/src/ast_util.rs
```diff
@@ -187,7 +187,10 @@ pub fn types_equal(typ1: &Typ, typ2: &Typ) -> bool {
         }
         (TypX::PointeeMetadata(t1), TypX::PointeeMetadata(t2)) => types_equal(t1, t2),
         (TypX::MutRef(t1), TypX::MutRef(t2)) => types_equal(t1, t2),
-        // rather than matching on _, repeat all the cases to catch any new variants added to TypX:
```
tjhance, line 190: I don't think there's any need to delete this comment

source/vir/src/ast_visitor.rs
```diff
@@ -774,6 +774,7 @@ pub(crate) trait AstVisitor<R: Returner, Err, Scope: Scoper> {
...
```
tjhance: This should recurse.

source/vir/src/opaque_type_to_air.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::sst_to_air::typ_to_ids;
+use crate::traits::{const_typ_bound_to_air, typ_equality_bound_to_air};
+use air::ast::BindX;
+use air::ast::Triggers;
+use air::ast::{Command, CommandX, Commands, DeclX};
+use air::ast_util::{ident_apply, mk_bind_expr, mk_unnamed_axiom, str_typ};
+use core::panic;
+use std::sync::Arc;
+
+pub fn opaque_types_to_air(ctx: &Ctx, opaque_types: &Vec<Opaquetype>) -> Commands {
```
tjhance: can you add some documentation about what axioms are being generated

source/vir/src/prune.rs
```diff
@@ -341,6 +353,12 @@ fn traverse_typ(ctxt: &Ctxt, state: &mut State, t: &Typ) {
                 }
             }
         }
+        TypX::Opaque { .. } => {
+            // Revisit.
+            // For let x = foo<SomeType>(..);
+            // Do we need to traverse typ args (SomeType)? Probably not?
+            // All the type args should have been included in the function body through other means.
```
tjhance, line 360: prune.rs uses a Visitor and then calls `traverse_typ` for every node. So if you fix the visitor (like I suggested in another comment) there's no need to do recurse here.

source/vir/src/sst_to_air.rs
```diff
@@ -1995,9 +2014,48 @@ fn stm_to_stmts(ctx: &Ctx, state: &mut State, stm: &Stm) -> Result<Vec<Stmt>, Vi
...
+                        .and_then(|f| ctx.func_sst_map.get(&f.current_fun))
+                        .map_or((false, None), |fun| {
+                            (
+                                matches!(fun.x.ret.x.typ.as_ref(), TypX::Opaque { .. }),
+                                Some(fun.x.ret.clone()),
+                            )
+                        });
+
+                    let ret_exp = if is_in_opaque_func {
+                        &crate::poly::coerce_exp_to_poly(
```
tjhance: usually we'd insert these coercions from `poly.rs`

source/rust_verify/src/external.rs
```diff
@@ -225,6 +239,49 @@ fn opts_in_to_verus(eattrs: &ExternalAttrs) -> bool {
...
+            );
+            let path_string = match &path_opt {
+                Some(path) => vir::ast_util::path_as_friendly_rust_name(&path),
+                None => format!("{:?}", opaque.def_id,),
+            };
+            VerifOrExternal::External { path: path_opt, path_string, explicit: false }
+        };
+
+        self.opaque_tys.push(OpaqueDef { id: opaque.def_id, verif: verif });
+    }
```
tjhance: The significance of these additions to `external.rs` is eluding me. Why does `CrateItems` need an `opaque_tys` field? Why is the existing traversal logic not sufficient for handling opaque types?

The computation for an opaque type's `state_for_this_item` only depends on the `eattrs`, but not on the `state` of the parent node, which is different than the traversal for other items, which propagates the state inwards. Is there a reason for this, and if so, what is it?

I would think you'd want the externalness of an opaque type to be the same as the externalness of the corresponding function, but due to the difference, but I don't think that it's going to be the case here, considering the various differences between this and the `visit_general` function.

tjhance: To be more concrete, it initially seems to me like the following would be simpler and more foolproof:

 * Don't make any changes to `external.rs`
 * After the `CrateItems` is constructed, go through all the verus-aware items, check each function to see if it has an OpaqueTy, mark all those OpaqueTys as Verus-aware

source/rust_verify_test/tests/safe_api.rs
FeizaiYiHao: What is the expected error message for this test? This test was failing because we didn't have opaque types. Do we want "Safe API violation: 'requires' clause is nontrivial"?

tjhance: That's the ideal error message. But it would also just be acceptable to do "The verifier does not support opaque types together with the `check-api-safety` flag"

FeizaiYiHao: A little confused about why we want to fail on this test. Do we want to fail it because `test2` is a private function, and its `requires` clause shouldn't be visible to others? Since we are returning an opaque type here, the `requires` clause is not revealed to others anyways?

source/rust_verify/src/external.rs
FeizaiYiHao: I agree. Not sure why I did this back then. It's been a while. Let me take a second look.

source/rust_verify_test/tests/safe_api.rs
tjhance: The point of the check-api-safety flag is to check that no functions with a nontrivial precondition (requires clause) are exposed to the user. This test case exposes such a function because it returns the `test2` function, which has a requires clause.

tjhance: more to the point, the client can run code like:

```
let f = test();
f(); // here they are calling a function that has a requires clause
```

and this should be impossible for a crate that passes check-api-safety

FeizaiYiHao:
@tjhance fixed most of the things. Would you take another look? TY!

source/rust_verify/src/external.rs
```diff
@@ -67,6 +67,12 @@ pub struct CrateItem {
     pub verif: VerifOrExternal,
 }
 
+#[derive(Debug)]
+pub struct OpaqueDef {
+    pub id: rustc_hir::def_id::LocalDefId,
+    pub verif: VerifOrExternal,
+}
```
tjhance, line 74: Is this still used?

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -1999,3 +2035,178 @@ pub(crate) fn ty_remove_references<'tcx>(
...
+            let mode = crate::attributes::get_mode(
+                Mode::Exec,
+                ctxt.tcx.get_attrs_unchecked(parent.into()),
+            );
+            if mode != Mode::Exec {
+                return err_span(
+                    opaque_ty.span,
+                    format!("Opaque type is not supported in {} mode", mode),
+                );
+            }
```
tjhance, line 2125: can you move this check to `vir/src/well_formed.rs`? rust-to-vir lowering is so messy, it's more robust to check against the mode in the final VIR krate

tjhance (changes requested):
This mostly looks good to me; with respect to the code, I only have some minor nitpicks.

The last main thing I would like to see is a few more tests regarding interacting features:

 * A function marked `external_body` which returns an opaque type.
 * An `assume_specification` for a function that returns an opaque type.
 * A function with a `returns` clause and which returns an opaque type.
 * A function with a `returns` clause and which returns an opaque type, and which is marked `allow_in_spec`

It's fine if some of these are "not implemented" errors, I mostly want to get tests in so it's clear what the expected behavior is in these cases.

source/rust_verify/src/rust_to_vir_base.rs
FeizaiYiHao: Np. Looking

FeizaiYiHao:
@tjhance 
`external_body ` works with no problem. 
`assume_specification ` requires some additional checks for the trait bounds of the two `OpaqueDef`s from the external function and the `assume_specification `. Not sure if I did the right thing in `compare_clasue_kind`. Would you take a look? 
`returns` will cause an error now. Not sure we should ever support it, since no way we can make the expression opaque?

FeizaiYiHao:
@tjhance ping in case the email notification didn't work

parno:
@FeizaiYiHao Is this ready to be merged?

FeizaiYiHao:
> @FeizaiYiHao Is this ready to be merged?

Wow sorry. Somehow the notification didn't work... Will quickly address the comments from @Chris-Hawblitzel and resolve the conflicts and then it should be good to be merged.

parno:
Great, thanks!

FeizaiYiHao:
@parno Somehow the comments from Chirs were addressed a long time ago but the they poped out again in the threads. But anyways, this is ready to be merged. Sorry for the delay

FeizaiYiHao:
I found an unhandled case: for nested opaque types, we are missing the equality statements in AIR generation. Let me quickly fix and push again. Sorry. 

Example: 
```
    trait DummyTrait{
        spec fn dummy_spec(&self) -> bool;
    }
    impl DummyTrait for bool{
        spec fn dummy_spec(&self) -> bool{
            true
        }
    }    
    async fn foo() -> (ret: (impl DummyTrait, impl DummyTrait))
        ensures
            ret.awaited() ==> ret@.0.dummy_spec(),
    {
        (true, true)
    }
    async fn bar() -> (ret:())
    {
        let future = foo();
        assert(future.awaited() ==> future@.0.dummy_spec());
        let ret_0 = future.await.0;
        assert(ret_0.dummy_spec());
    }
    fn baz() -> (ret: impl Future<Output = (impl DummyTrait, impl DummyTrait)>)
        ensures
            ret.awaited() ==> ret@.0.dummy_spec(),
    {
        foo() // Here Z3 doesn't know that the Output type of baz()'s return value equals to Output type of foo()'s
    }
```

FeizaiYiHao:
OK. fixed. Ready to be merged

PR #1787, opened 2025-06-27
Add bits lemmas

rikosellic:
This PR contains a few lemmas I found useful when verifying bit operations:
-  strengthens the post-condition of `lemma_pow2_no_overflow`
-  adds `lemma_pow2_subtracts`  which proves `pow2(m - n) = pow2(m) / pow2(n)`
-  adds `lemma_mul_pow2_no_overflow_iff_shr` which shows that `x * pow2(k) <= MAX` holds iff `x <= (MAX >> k)`

### DIFF
--- a/source/vstd/arithmetic/power2.rs
+++ b/source/vstd/arithmetic/power2.rs
@@ -17,7 +17,13 @@ use super::super::prelude::*;
 verus! {
 
 #[cfg(verus_keep_ghost)]
-use super::power::{pow, lemma_pow_positive, lemma_pow_adds, lemma_pow_strictly_increases};
+use super::power::{
+    pow,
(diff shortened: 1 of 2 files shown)


parno:
Thanks very much for the contribution.  I think the overflow lemma could be generalized a bit.  At least, [this example](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=f8e773c552463910864e4965fb710c59) goes through with a custom `max` argument, instead of hard coding `u64::MAX`.  What do you think?

rikosellic:
Thanks a lot for the suggestion — I agree this generalization makes the lemma more useful. I've updated the lemma to take a custom max argument.

PR #1797, opened 2025-07-07
add lemmas to vstd::set_lib, about Set::to_seq

WeituoDAI:

- add lemmas about Set::to_seq, it could be useful if one want to convert a set to seq and then fold_left/right over the seq.
- fix some typos (in seq.rs, there is a spec fn called `surj`, but it looks more like injective function..)


jaylorch:
This looks good to me, thanks! But, before I approve it, please resolve the conflicts that have arisen since you created this PR.

WeituoDAI:
updated for both iset and set

jaylorch (approved):
Thanks very much!

PR #1799, opened 2025-07-08
Use `constrain_type` function to set return value type in ensures cla…

FeizaiYiHao:
In this pull request, we add a new builtin function `constrain_type` to set the return value type of the function in ensures clauses. This PR will enable opaque types. The changes include:

- A new builtin function `constrain_type::<T>(_x: T, _y:T){}`
- In `verus!` macro for functions with a return value, not including closures:
  -  the ensures clause closure now take one parameter with no type annotation.
  - instead, we add a statement at the beginning of the closure with `constrain_type(self_fn(args), ret)` to constrain the type of the return value in the ensures closure.
  - Currently, we only perform a check at the Rust to VIR stage to make sure return value has different name with the function parameters. This is too late. Rustc will throw an error about `constrain_type`, and the users will be confused with the error message with a strange function `constrain_type`. Therefore, we perform an early check in the `verus!` macro.
- In Rust to VIR stage:
  - The `constrain_type` function calls in ensures clause are erased.
- In `vstd` and `state_machines_macros`, fix a few places where `verus!` is called instead `impl`, which causes `verus!` to pares `ImplItemFn` and `TraitItemFn` as `ItemFn`.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -7,7 +7,6 @@ use proc_macro2::TokenTree;
 use quote::ToTokens;
 use quote::format_ident;
 use quote::{quote, quote_spanned};
-use syn::Receiver;
 use syn::token::Comma;
 use verus_syn::BroadcastUse;
 use verus_syn::DefaultEnsures;
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -7,10 +7,13 @@ use proc_macro2::TokenTree;
 use quote::ToTokens;
 use quote::format_ident;
 use quote::{quote, quote_spanned};
+use syn::Receiver;
+use syn::token::Comma;
 use verus_syn::BroadcastUse;
 use verus_syn::DefaultEnsures;
 use verus_syn::ExprBlock;
 use verus_syn::ExprForLoop;
+use verus_syn::Generics;
 use verus_syn::parse::{Parse, ParseStream};
 use verus_syn::parse_quote_spanned;
 use verus_syn::punctuated::Punctuated;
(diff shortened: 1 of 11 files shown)


Chris-Hawblitzel:
> * In `vstd` and `state_machines_macros`, fix a few places where `verus!` is called instead `impl`, which causes `verus!` to pares `ImplItemFn` and `TraitItemFn` as `ItemFn`.

I made a pull request, https://github.com/verus-lang/verus/pull/1817 , to allow proper parsing of code inside an impl, so hopefully these fixes won't be needed or can be simplified.

FeizaiYiHao:
Yes. With these macros, we wouldn't need these fixes anymore.

Chris-Hawblitzel:
> Yes. With these macros, we wouldn't need these fixes anymore.

Ok, I merged https://github.com/verus-lang/verus/pull/1817 .

FeizaiYiHao:
OK, updated changes. They are clean now. Ready to merge now

source/builtin_macros/src/syntax.rs
```diff
@@ -930,7 +974,34 @@ impl Visitor {
         self.inside_ghost += 1; // for requires, ensures, etc.
 
         let sig_span = sig.span().clone();
-        let spec_stmts = self.take_sig_specs(&mut sig.spec, ret_pat, None, sig_span);
+
+        // In VIR there's the same check, but Rustc will complain first, and throw out
+        // some errors about "constrain_type", which are confusing and the users should not see.
+        // Instead we give an early error with nicer error msg here.
+        if let Some((p, _)) = &ret_pat {
+            for input in &sig.inputs {
```
Chris-Hawblitzel: There are two copies of this in the code.  Could you move it (or most of it) into a helper function?

Chris-Hawblitzel:
The following code results in a parse error:

```
pub exec fn f(
    mut b: bool,
) -> (result: bool)
    ensures true
{
    true
}
```

source/builtin_macros/src/syntax.rs
FeizaiYiHao: Looking... Somehow the notification didn't work...

FeizaiYiHao: Turns out one of them is for `syn::*` another is for `syn_verus`. I moved them to two different functions, but I'm not sure if there's a way to merge them into one function.

FeizaiYiHao:
OK fixed the parse error. Let me know if there are other corner cases.

Chris-Hawblitzel (approved):

dschoepe:
@tjhance Does this look good to you now as well?

PR #1802, opened 2025-07-09
veritas: add vest and the auto-generated vest-tls

y1ca1:

This PR adds [Vest](https://github.com/secure-foundations/vest) (and the Vest-generated TLS 1.3 parsers and serializers) to veritas, the Verus regression suite.

The `make` command in `prepare_script` for `vest-tls` would invoke Verus: `verus src/lib.rs --compile --export vest.verusdata`, so if `verus` is in path in the container, it should work.


parno (approved):
Looks good to me.  Thanks for adding this!

PR #1803, opened 2025-07-09
add const documentation'

elanortang:

### DIFF
--- a/source/docs/guide/src/const.md
+++ b/source/docs/guide/src/const.md
@@ -17,8 +17,7 @@ It is always ghost and cannot be used as an `exec` value.
 ## `proof` and `exec` consts
 Just as `proof` and `exec` functions can have `ensures` clauses specifying a postcondition, 
 `proof` and `exec` consts can have `ensures` clauses to tie the declaration to a `spec` expression. 
-The syntax follows the syntax of a function definition: we define the type, 
-followed by the ensures clause, followed by the declaration in curly brackets. 
+The syntax follows the syntax of a function definition. 
 
 ```rust
 {{#include ../../../../examples/guide/const.rs:exec_const_syntax}}


source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
 # const declarations
 
-`const` declarations can either be marked `spec` or left without a mode:
+In Verus, `const` declarations are treated internally as 0-argument function calls. 
+Thus just like functions, `const` declarations can be marked `spec`, `proof`, `exec`, 
+or left without a mode.
```
parno: "without **an explicit** mode"

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
 # const declarations
 
-`const` declarations can either be marked `spec` or left without a mode:
+In Verus, `const` declarations are treated internally as 0-argument function calls. 
+Thus just like functions, `const` declarations can be marked `spec`, `proof`, `exec`, 
+or left without a mode. 
+By default, a `const` without a mode has a dual `spec/exec` mode.
```
parno: "without **an explicit** mode"
"has" -> "is assigned"

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+```rust
+spec const SPEC_ONE: int = 1;
+
+spec fn spec_add_one(x: int) -> int {
+    x + SPEC_ONE
+}
+```
+
+## `proof` and `exec` consts
+<!-- `proof` and `spec` consts are like `proof` and `exec` functions with no arguments.  -->
```
parno: We can probably just delete this comment

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+Just as `proof` and `exec` functions can have `ensures` clauses specifying a post-condition,
```
parno: In the rest of the guide, I think we use "postcondition" (with no hyphen)

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+
+spec fn spec_add_one(x: int) -> int {
+    x + SPEC_ONE
+}
+```
+
+## `proof` and `exec` consts
+<!-- `proof` and `spec` consts are like `proof` and `exec` functions with no arguments.  -->
+Just as `proof` and `exec` functions can have `ensures` clauses specifying a post-condition, 
+`proof` and `exec` consts have `ensures` clauses to tie the declaration to a `spec` expression.
```
parno: "have" -> "can have"

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+
+```rust
+exec const C: u64 
+    ensures C == 7 
+{
+    7 
+}
+```
+
+Idiomatically, the ensures clause is typically on the same line as the type,
```
parno: I don't think we need this bit, since `verusfmt` will dictate the formatting.

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+}
+```
+
+Idiomatically, the ensures clause is typically on the same line as the type, 
+and if the definition is short enough, 
+the entire declaration will be on one line. 
+Note here that we can also use `assert` when defining the const. 
 
 ```rust
-{{#include ../../../../examples/guide/modes.rs:const1}}
```
parno, line 6: It would be nice to keep all of these examples (or at least all the ones that are supposed to work) in one of the files in `examples/guide`, so that they're checked by CI.

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+    1 + e()
+}
+
+exec const C: int ensures C == 7 { 7 }
 ```
 
-A `spec const` is like `spec` function with no arguments.
-It is always ghost and cannot be used as an `exec` value.
+## `spec/exec` consts
+A `const` without a mode is dual-use:
```
parno: "without *an explicit* mode"

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
 For example, as with `exec` code, its type must be compilable (e.g. `u8`, not `int`),
-and, as with `spec` code, it cannot call any `exec` or `proof` functions.
+and, as with `spec` code, it cannot call any `exec` or `proof` functions. 
+
+```rust
+const fn foo() -> u64 { 1 }
+const C: u64 = foo();  // FAILS with error "cannot call function `foo` with mode exec"
+```
+
+## Using `when_used_as_spec`
```
parno: The title leaves it a bit unclear what the higher-level purpose is.  Perhaps "Using an `exec const` in a `spec` or `proof` context" instead?

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+Similar to functions, if you want to use an `exec const` in a `spec` or `proof` context, 
+you can annotate the declaration with `[verifier::when_used_as_const(SPEC_DEF)]`.
```
parno: ", where `SPEC_DEF` is the name of a `spec const` or a `spec` function with no arguments."

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+
+fn test(){
+    assert(EIGHT == EIGHT_SPEC);
+}
+```
+
+## Trouble-shooting overflow errors
+Verus may have difficulty proving that a `const` declaration does not overflow; 
+using `[verifier::when_used_as_const(SPEC_DEF)]` 
+or with `[verifier::non_linear]` may help.
```
parno: I think we can cut "with" here.

source/docs/guide/src/const.md
```diff
@@ -1,17 +1,111 @@
...
+```
+
+## Trouble-shooting overflow errors
+Verus may have difficulty proving that a `const` declaration does not overflow; 
+using `[verifier::when_used_as_const(SPEC_DEF)]` 
+or with `[verifier::non_linear]` may help. 
+
+For example, here `[verifier::non_linear]` is added to prevent the error 
+"possible arithmetic underflow/overflow." 
+This tells Verus to perform const propagation.
```
parno: I think this last sentence should be "This allows Verus to reason about the (seemingly) non-linear expression `BAR_PLUS_ONE * BAR`, instead of giving up immediately.  See the [chapter on non-linear reasoning](appropriate-chapter-file-here.md) for more details."

parno (commented):
Thanks for putting this together!  I think it will be very helpful for Verus users.  I've left a series of small comments inline.

source/docs/guide/src/const.md
elanortang: I found it confusing to parse the one-line version of a const declaration when I was first reading Chris' PR. It might be helpful to have a breakdown of what parts are what. 

I agree that we can at least delete the "idiomatically" line since `verusfmt` will take care of it.

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: I moved all the examples to `examples/guide/const.rs`.

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done

source/docs/guide/src/const.md
elanortang: done, also I added some more explanation so that other people don't run into the same issue I was running into.

source/docs/guide/src/const.md
elanortang: oh yeah typo

source/docs/guide/src/const.md
elanortang: done

parno (approved):
Thanks for the fixes and enhancements!

PR #1804, opened 2025-07-10
Fix attribute macro: evaluation of constant value failed

ziqiaozhou:
This PR is related to https://github.com/verus-lang/verus/pull/1793 and fix the const-evaluation error for const functions. 

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -41,7 +41,7 @@ use syn::{Expr, Item, parse2, spanned::Spanned};
 use crate::{
     EraseGhost,
     attr_block_trait::{AnyAttrBlock, AnyFnOrLoop},
-    syntax::{self, mk_verus_attr_syn},
+    syntax::{self, mk_verifier_attr_syn, mk_verus_attr_syn},
     syntax_trait,
     unerased_proxies::VERUS_UNERASED_PROXY,
 };
@@ -49,26 +49,6 @@ use crate::{
 pub const VERIFIED: &str = "_VERUS_VERIFIED";
 
 pub const DUAL_SPEC_PREFIX: &str = "__VERUS_SPEC";
-
-fn mk_verifier_attr_syn(span: proc_macro2::Span, tokens: &str) -> syn::Attribute {
-    let mut path_segments = syn::punctuated::Punctuated::new();
-    path_segments.push(syn::PathSegment {
-        ident: syn::Ident::new("verifier", span),
-        arguments: syn::PathArguments::None,
-    });
-    path_segments.push(syn::PathSegment {
-        ident: syn::Ident::new(tokens, span),
-        arguments: syn::PathArguments::None,
-    });
-    let path = syn::Path { leading_colon: None, segments: path_segments };
-    let meta = syn::Meta::Path(path);
-    syn::Attribute {
-        pound_token: syn::token::Pound { spans: [span] },
-        style: syn::AttrStyle::Outer,
-        bracket_token: syn::token::Bracket::default(),
-        meta,
-    }
-}
 
 enum VerusIOTarget {
     Local(syn::Local),
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -5024,7 +5024,7 @@ macro_rules! declare_mk_rust_attr {
 declare_mk_rust_attr!(mk_rust_attr, syn_verus);
 declare_mk_rust_attr!(mk_rust_attr_syn, syn);
 
-/// Constructs #[verus::internal(tokens)]
+/// Constructs #[verus::internal(tokens)] and #[verifier::tokens]
 macro_rules! declare_mk_verus_attr {
     ($name:ident, $name2:ident, $s:ident) => {
         pub(crate) fn $name(span: Span, tokens: TokenStream) -> $s::Attribute {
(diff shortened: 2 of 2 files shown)


source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -41,15 +41,35 @@ use syn::{Expr, Item, parse2, spanned::Spanned};
...
+    });
+    let path = syn::Path { leading_colon: None, segments: path_segments };
+    let meta = syn::Meta::Path(path);
+    syn::Attribute {
+        pound_token: syn::token::Pound { spans: [span] },
+        style: syn::AttrStyle::Outer,
+        bracket_token: syn::token::Bracket::default(),
+        meta,
+    }
+}
```
Chris-Hawblitzel: I don't think we should have to have a second definition of `mk_verifier_attr_syn` -- can we just change the one that's already declared in `syntax.rs` so that it also works here?  Specifically (@tjhance , @ziqiaozhou ), can we change it to generate `#[verifier::external]` instead of `#[verifier(external)]`?  The following diff should do that, and it passes the tests:

```
diff --git a/source/builtin_macros/src/attr_rewrite.rs b/source/builtin_macros/src/attr_rewrite.rs
index 6982f730..077ae787 100644
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -41,7 +41,7 @@ use syn::{Expr, Item, parse2, spanned::Spanned};
 use crate::{
     EraseGhost,
     attr_block_trait::{AnyAttrBlock, AnyFnOrLoop},
-    syntax::{self, mk_verus_attr_syn},
+    syntax::{self, mk_verifier_attr_syn, mk_verus_attr_syn},
     syntax_trait,
     unerased_proxies::VERUS_UNERASED_PROXY,
 };
@@ -50,26 +50,6 @@ pub const VERIFIED: &str = "_VERUS_VERIFIED";
 
 pub const DUAL_SPEC_PREFIX: &str = "__VERUS_SPEC";
 
-fn mk_verifier_attr_syn(span: proc_macro2::Span, tokens: &str) -> syn::Attribute {
-    let mut path_segments = syn::punctuated::Punctuated::new();
-    path_segments.push(syn::PathSegment {
-        ident: syn::Ident::new("verifier", span),
-        arguments: syn::PathArguments::None,
-    });
-    path_segments.push(syn::PathSegment {
-        ident: syn::Ident::new(tokens, span),
-        arguments: syn::PathArguments::None,
-    });
-    let path = syn::Path { leading_colon: None, segments: path_segments };
-    let meta = syn::Meta::Path(path);
-    syn::Attribute {
-        pound_token: syn::token::Pound { spans: [span] },
-        style: syn::AttrStyle::Outer,
-        bracket_token: syn::token::Bracket::default(),
-        meta,
-    }
-}
-
 enum VerusIOTarget {
     Local(syn::Local),
     Expr(syn::Expr),
@@ -396,7 +376,7 @@ pub fn rewrite_verus_spec_on_fun_or_loop(
                 let mut const_fun = fun.clone();
                 let span = fun.sig.constness.unwrap().span();
                 const_fun.block.as_mut().stmts.retain(|stmt| !is_verus_proof_stmt(stmt));
-                const_fun.attrs.push(mk_verifier_attr_syn(span, "external"));
+                const_fun.attrs.push(mk_verifier_attr_syn(span, quote! { external }));
                 const_fun.attrs.push(mk_verus_attr_syn(span, quote! { uses_unerased_proxy }));
                 const_fun.attrs.push(mk_verus_attr_syn(span, quote! { encoded_const }));
                 const_fun.to_tokens(&mut new_stream);
diff --git a/source/builtin_macros/src/syntax.rs b/source/builtin_macros/src/syntax.rs
index 480d92f5..1d0821e4 100644
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -5064,18 +5064,9 @@ macro_rules! declare_mk_verus_attr {
                 ident: $s::Ident::new("verifier", span),
                 arguments: $s::PathArguments::None,
             });
+            path_segments.push($s::parse_quote_spanned!(span => #tokens));
             let path = $s::Path { leading_colon: None, segments: path_segments };
-            let meta = if tokens.is_empty() {
-                $s::Meta::Path(path)
-            } else {
-                $s::Meta::List($s::MetaList {
-                    path,
-                    delimiter: $s::MacroDelimiter::Paren($s::token::Paren {
-                        span: into_spans(span),
-                    }),
-                    tokens: quote! { #tokens },
-                })
-            };
+            let meta = $s::Meta::Path(path);
             $s::Attribute {
                 pound_token: $s::token::Pound { spans: [span] },
                 style: $s::AttrStyle::Outer,
```

ziqiaozhou: The syn::mk_verifier_attr_syn  will not work. The verifier(xxx) is not defined as a real attribute in builtin_macro lib, but is instead used as an placeholder inside verus macro and the verus macro will replace verifier(xxx) to verifier::xxx. I need to use verifier::external if I want to skip that replacement.

ziqiaozhou: Oh. I see. You are suggesting to update that function inside syntax.rs.

ziqiaozhou: @tjhance , I updated both mk_verifier_attr_syn and mk_verifier_attr to create #[verifier::xxx]. Currently, mk_verifier_attr is only used with "external" attribute and so I think it seems fine to do that.

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -372,6 +374,22 @@ pub fn rewrite_verus_spec_on_fun_or_loop(
                 let unverified_fun = rewrite_unverified_func(&mut fun, with.with.span());
                 unverified_fun.to_tokens(&mut new_stream);
             }
+            if fun.sig.constness.is_some() {
+                let mut const_fun = fun.clone();
+                let span = fun.sig.constness.unwrap().span();
+                // It seems that we do not need to erase anything.
+                // But just do it to be safe and consistent with verus macro.
```
tjhance, line 381: Why do you not need to erase anything?

ziqiaozhou: Because inside the execution function, it only has some proof blocks, somehow, I noticed that it does not trigger constant evaluation error. I tried to skip erasing it here and my tests still work.

tjhance: ah yeah, that makes sense. The erasure will be more important after merging in https://github.com/verus-lang/verus/pull/1792 But for the time being, it shouldn't matter.

tjhance (approved):

source/builtin_macros/src/attr_rewrite.rs
ziqiaozhou: Maybe it is because my proof macros always add #[verus::internal(const_header_wrapper)]. I guess this attribute will be deprecated. But I think my current version already erases all proofs and so should be ready to remove const_header_wrapper.

PR #1806, opened 2025-07-11
Add spec for range contains

ziqiaozhou:

### DIFF
--- a/source/vstd/std_specs/range.rs
+++ b/source/vstd/std_specs/range.rs
@@ -33,6 +33,24 @@ pub assume_specification<A: core::iter::Step>[ Range::<A>::next ](range: &mut Ra
         (*range, r) == spec_range_next(*old(range)),
 ;
 
+pub assume_specification<Idx: PartialOrd<Idx>, U>[Range::<Idx>::contains](range: &Range<Idx>, i: &U) -> (r: bool) where
+    Idx: PartialOrd<U>,
+    U: ?Sized + PartialOrd<Idx>,
+
+    ensures
+        call_ensures(<U as PartialOrd<Idx>>::lt, (i, &range.end), true) && call_ensures(
+            <Idx as PartialOrd<U>>::le,
+            (&range.start, i),
+            true,
+        ) <==> r,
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/range.rs
```diff
@@ -33,6 +33,24 @@ pub assume_specification<A: core::iter::Step>[ Range::<A>::next ](range: &mut Ra
...
+            true,
+        ) <==> r,
+        call_ensures(<U as PartialOrd<Idx>>::lt, (i, &range.end), false) && call_ensures(
+            <Idx as PartialOrd<U>>::le,
+            (&range.start, i),
+            false,
+        ) <==> !r
+;
+
+
```
Chris-Hawblitzel: A couple comments:
- I think it will be easier to work with specs like this if they are written in terms of `partial_cmp_spec`, under the condition that `obeys_partial_cmp_spec` holds, than in terms of `call_ensures`
- for the possible outcomes of (start <= i, i < end), this covers (false, false) and (true, true), but not (false, true) and (true, false)

ziqiaozhou: If using obeys_partial_cmp_spec, it should looks like
```
U::obeys_partial_cmp_spec() && <Idx as PartialOrdSpec<U>>::obeys_partial_cmp_spec() ==> (i.partial_cmp_spec(&r.end) == Some(
            core::cmp::Ordering::Less,
        ) && matches!(r.start.partial_cmp_spec(i), Some(
            core::cmp::Ordering::Less | core::cmp::Ordering::Equal,
        ))) == ret
```

which only covers the case when both Idx and U obeys_partial_cmp_spec. Both works for me. I am not sure whether it is better to only cover some cases but not others. If so, I guess I should put obeys_partial_cmp_spec as preconditions to make it easy to debug a failed case?

Chris-Hawblitzel: That looks good to me.  It's an interesting question whether to put "obeys" in the requires or ensures, but so far we've been putting them in the ensures (in https://github.com/verus-lang/verus/blob/main/source/vstd/std_specs/hash.rs , for example).

PR #1738, opened 2025-06-12
WIP: use prophecy encoding for functions that take a mutable reference as a parameter

utaal:
with @apribiso

(first step towards general mutable reference support)

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #1812, opened 2025-07-13
add several key building blocks for general mut ref support

tjhance:
This implements a number of building blocks needed for mut ref support outside of function call arguments.

Because the full project is significant, I want to move forward incrementally. The feature in this PR is incomplete & unsound, but is gated behind a flag `-V new-mut-ref`. However, by laying down the building blocks, it unlocks a number of fairly independent next steps.

The building blocks are:

 * The ability to reason about `&mut T` values in spec code just like any other type.
   * Following https://github.com/verus-lang/verus/pull/1738, we add a TypX::MutRef type to VIR. I used a slightly different AIR encoding, as there were a number of oddities in 1738, which I think may have been related to the special-casing for function calls, whereas I wanted something more uniform with the other types. The encoding I use is a quick-and-dirty Poly type (no MonoTyp) with prelude-defined spec functions.
* `mut_ref_current` and `mut_ref_future` spec functions, each with type `(&mut T) -> T`. The latter is prophecy-dependent.
* The `resolve` command and the `has_resolved` spec function. These can be applied to any type; they are usually applied to mutable references or types contain mutable references.
     * e.g., for `m: &mut T`, `has_resolved(m)` is defined as `mut_ref_current(m) == mut_ref_future(m)`.
     * The `resolve(m)` is a command that assumes `has_resolved(m)`. This is only sound to invoke after `m` has finished being mutated. Right now, we don't automate their insertion, nor do we check that the user is inserting them soundly. Both these are left for subsequent PRs.
* `BorrowMut` as a first-class node in VIR.

### How `BorrowMut` works in VIR:

We add these to ExprX:
```rust
    AssignToPlace {
        place: Place,
        rhs: Expr,
        op: Option<BinaryOp>,
    },
    BorrowMut(Place),
    BorrowMutPhaseOne(Place),
    BorrowMutPhaseTwo(Place, Expr),
    DerefMut(Expr),
```

With the new `Place` type:

```rust
pub enum PlaceX {
    Field(FieldOpr, Place),
    /// Conceptually, this is like a Field, accessing the 'current' field of a mut_ref.
    DerefMut(Place),
    Local(VarIdent),
    Temporary(Expr),
}   
```

The new `Place` type is like the existing `Loc` system. I made a separate type because: (i) A separate enum makes it easier to see what a valid Place is (ii) I wanted the benefits of a 'clean slate' and (iii) I wanted to leave the existing system alone as much as possible so the new system could exist side-by-side with the old one. To be clear, I don't intend to keep these around redundantly forever, this is just for the intermediary while we implement the feature.

`BorrowMut` lowers into a `BorrowMutPhaseOne` and `BorrowMutPhaseTwo`. (Having these as separate nodes will let us support two-phase borrows.) `BorrowMutPhaseOne` generates the new mut_ref with a prophecized value; `BorrowMutPhaseTwo` assigns to the borrowed-from Place by writing in the prophecized value. (This is easily implemented by simply lowering it to `AssignToPlace`.)

Note that because `DerefMut` is a "place", this allows us to assign to mutable references (e.g., `*u_ref = 20;`) and also implement reborrows (because we can do `MutBorrow` on a `DerefMut` place).

The examples below show several nontrivial interactions, including reborrows and the interactions of mutable borrows with field access.

### Examples

Here's some example code that verifies:

Note that Verus natively doesn't yet know the definition of `has_resolved`, so I provide it as an axiom.

```rust
broadcast axiom fn resolved_defn<T>(a: &mut T)
    ensures
        #[trigger] has_resolved(a) ==> mut_ref_current(a) == mut_ref_future(a);

fn test_no_update() {
    broadcast use resolved_defn;

    let mut u: u64 = 20;
    let u_ref: &mut u64 = &mut u; // BorrowMut(place: Local(u))

    proof { resolve(u_ref) }

    assert(u == 20);
}

fn test_basic_update() {
    broadcast use resolved_defn;

    let mut u: u64 = 20;
    let u_ref: &mut u64 = &mut u; // BorrowMut(place: Local(u))

    *u_ref = 30; // AssignToPlace DerefMut(Local(u_ref))

    proof { resolve(u_ref) }

    assert(u == 30);
}

struct Pair<A, B>(A, B);

fn test_field_update() {
    broadcast use resolved_defn;

    let mut u: Pair<u64, u64> = Pair(20, 20);
    let u_ref: &mut Pair<u64, u64> = &mut u; // BorrowMut(place: Local(u))

    u_ref.0 = 30; // AssignToPlace Field("0", DerefMut(Local(u_ref)))

    proof { resolve(u_ref) }

    assert(u.0 == 30);
    assert(u.1 == 20);
}

fn test_field_update2() {
    broadcast use resolved_defn;

    let mut u: Pair<u64, u64> = Pair(20, 20);
    let u_ref: &mut u64 = &mut u.0; // BorrowMut(place: Field("0", Local(u)))

    *u_ref = 30; // AssignToPlace DerefMut(Local(u_ref))

    proof { resolve(u_ref) }

    assert(u.0 == 30);
    assert(u.1 == 20);
}

fn test_mut_ref_in_pair() {
    broadcast use resolved_defn;

    let mut u: u64 = 20;
    let u_ref: Pair<&mut u64, u64> = Pair(&mut u, 70);

    *u_ref.0 = 30; // AssignToPlace DerefMut(Field("0", Local(u_ref)))

    proof { resolve(u_ref.0) }

    assert(u == 30);
}

fn test_reborrow() {
    broadcast use resolved_defn;

    let mut u: u64 = 20;
    let u_ref: &mut u64 = &mut u; // BorrowMut(place: Local(u))

    *u_ref = 11; // AssignToPlace DerefMut(Local(u_ref))

    let u_ref2 = &mut *u_ref; // BorrowMut(place: DerefMut(Local(u_ref)))

    let x = *u_ref2;
    assert(x == 11);
    *u_ref2 = 13; // AssignToPlace DerefMut(Local(u_ref2))

    proof { resolve(u_ref2); }

    let x = *u_ref;
    assert(x == 13);

    *u_ref = 17;

    proof { resolve(u_ref); }

    assert(u == 17);
}
```

### TODO

As I said, this PR is good to go as-is because it doesn't change existing behavior, but these are the things that will need to be done next:

 - [ ] Automate the insertion of resolve expressions and make sure they're only used soundly
 - [ ] Export the definition of the `has_resolved` predicate for builtin types and user-defined datatypes
 - [ ] Prevent the prophecy time travel paradox
 - [ ] Resolve the mode issue @Chris-Hawblitzel and I discussed
 - [ ] Flesh out the lowering from Rust->VIR for the new BorrowMut scheme, support "temporaries"
 - [ ] Support match ergonomics
 - [ ] Implement two-phase borrows where necessary

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -275,6 +275,7 @@ pub enum TypX {
     ConstBool(bool),
     /// AIR type, used internally during translation
     Air(air::ast::Typ),
+    MutRef(Typ),
 }
 
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, ToDebugSNode)]
@@ -336,9 +337,17 @@ pub enum UnaryOp {
     /// Each trigger group becomes one SMT trigger containing all the expressions in the trigger group.
     Trigger(TriggerAnnotation),
     /// Force integer value into range given by IntRange (e.g. by using mod)
-    Clip { range: IntRange, truncate: bool },
+    Clip {
+        range: IntRange,
+        truncate: bool,
+    },
     /// Operations that coerce from/to verus_builtin::Ghost or verus_builtin::Tracked
-    CoerceMode { op_mode: Mode, from_mode: Mode, to_mode: Mode, kind: ModeCoercion },
+    CoerceMode {
+        op_mode: Mode,
+        from_mode: Mode,
+        to_mode: Mode,
+        kind: ModeCoercion,
+    },
     /// Internal consistency check to make sure finalize_exp gets called
     /// (appears only briefly in SST before finalize_exp is called)
     MustBeFinalized,
@@ -363,9 +372,13 @@ pub enum UnaryOp {
     /// For an exec/proof expression e, the spec s should be chosen so that the value v
     /// that e evaluates to is immutable and v == s, where v may contain local variables.
     /// For example, if v == (n..m), then n and m must be immutable local variables.
-    InferSpecForLoopIter { print_hint: bool },
+    InferSpecForLoopIter {
+        print_hint: bool,
+    },
     /// May need coercion after casting a type argument
     CastToInteger,
+    MutRefCurrent,
+    MutRefFuture,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord, ToDebugSNode)]
@@ -418,6 +431,11 @@ pub enum UnaryOpr {
     IntegerTypeBound(IntegerTypeBoundKind, Mode),
     /// Custom diagnostic message
     CustomErr(Arc<String>),
+    /// Predicate over any type that indicates its mutable references has resolved.
+    /// For &mut T this says the prophetic value == the current value.
+    /// For primitive types this is trivially true.
+    /// For datatypes this is recursive in the natural way.
+    HasResolved(Typ),
 }
 
 /// Arithmetic operation that might fail (overflow or divide by zero)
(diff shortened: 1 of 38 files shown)


utaal:
Does this supersede #1738?

source/vir/src/prelude.rs
```diff
@@ -835,6 +864,9 @@ pub(crate) fn prelude_nodes(config: PreludeConfig) -> Vec<Node> {
             :skolemid skolem_prelude_singularmod
         )))
 
+        // REVIEW: I think we can skip the decoration?
+        (declare-fun [resolved] ([decoration] [typ] [Poly]) Bool)
```
Chris-Hawblitzel, line 867: Yes, I think that can be skipped.

source/vir/src/ast.rs
```diff
@@ -883,6 +937,28 @@ pub enum ExprX {
...
+    /// However, the two phases can be split up to do a 2-phase borrow.
+    BorrowMut(Place),
+    /// Phase 1: Create the mut_ref value with the prophetic future value
+    BorrowMutPhaseOne(Place),
+    /// Phase 2: Update the original place to be equal to the prophecized value.
+    /// The Expr argument here is the expression returned by the PhaseOne (of type &mut T)
+    BorrowMutPhaseTwo(Place, Expr),
+    DerefMut(Expr),
+    Resolve(Expr, Typ),
+    HasResolved(Expr, Typ),
```
Chris-Hawblitzel: Can HasResolved be a UnaryOpr instead of an ExprX?

source/vir/src/prelude.rs
tjhance: Unfortunately, I thought about it more, and I don't think it can ... the Ghost decoration has to work differently than the others since `Ghost<&mut T>` isn't a "real" mutable reference

source/vir/src/ast.rs
tjhance: yes

PR #1798, opened 2025-07-08
Automate generating spec types and their View/DeepView implementations

amaurremi:
## Summary
Add macros for generating spec types with View and DeepView implementations to reduce boilerplate code.

- `#[make_spec_type]` for a struct or enum `X` creates a spec type `XSpec` with the same shape (fields names/variant names) as `X`, but with all field types set to the original types' `DeepView` versions
  * for structs, supports `exclude` parameter that allows excluding specific fields (can be useful e.g. for protobuf-generated types)
  * generates straightforward View and DeepView trait implementations, with View just being defined through DeepView
- #[self_view] for a struct or enum defines `View` and `DeepView` implementations that are set to the type itself.
 
## Limitations
- Does not support generic types
- No custom View/DeepView logic
- All field types must support DeepView implementations

### #[make_spec_type]
```rust
#[make_spec_type(exclude(private_field))]
pub struct MyStruct<'a> {
    pub id: u32,
    pub data: Vec<String>,
    pub private_field: String, // Excluded from spec
}
```
will expand to
```
verus! {
pub struct MyStruct<'a> {
    pub id: u32,
    pub data: Vec<String>,
    pub private_field: String,
}

pub ghost struct MyStructSpec {
    pub id: <u32 as DeepView>::V,
    pub data: <Vec<String> as DeepView>::V,
}

#[cfg(verus_keep_ghost)]
impl<'a> DeepView for MyStruct<'a> {
    type V = MyStructSpec;

    open spec fn deep_view(&self) -> Self::V {
        Self::V {
            id: self.id.deep_view(),
            data self.data.deep_view(),
        }
    }
}

#[cfg(verus_keep_ghost)]
impl<'a> View for MyStruct<'a> {
    type V = <Self as DeepView>::V;

    open spec fn view(&self) -> Self::V {
        self.deep_view()
    }
}
} // verus!
```

### #[self_view]
```rust
#[self_type]
pub struct MyStruct<'a> {
    pub id: u32,
    pub data: Vec<String>,
    pub private_field: String,
}
```
will generate
```rust
verus! {
pub struct MyStruct<'a> {
    pub id: u32,
    pub data: Vec<String>,
    pub private_field: String,
}

#[cfg(verus_keep_ghost)]
impl<'a> DeepView for MyStruct<'a> {
    type V = Self;

    open spec fn deep_view(&self) -> Self::V {
        self
    }
}

#[cfg(verus_keep_ghost)]
impl<'a> View for MyStruct<'a> {
[...]

PR #1813, opened 2025-07-14
Add vstd/contrib and builtin_macros/src/contrib directories

Chris-Hawblitzel:
With pull requests https://github.com/verus-lang/verus/pull/1798 , https://github.com/verus-lang/verus/pull/1608 , and https://github.com/verus-lang/verus/pull/1678 , we have contributions that aren't quite core language features, but are nevertheless useful to share.  For contributions like this, this pull request adds two `contrib` directories:

- `vstd/contrib`
- `builtin_macros/src/contrib`

Furthermore, this pull request sets up a way for macros to optionally hook into `builtin_macro`'s `syn_verus` processing, so that the contributed macro can manipulate the original Verus syntax in `syn_verus` form rather than trying to work with the more verbose Rust code generated by the `verus!` macro.  As an example of using this, this pull request includes `vstd::contrib::auto_spec`, a syn_verus-based macro that transforms this:

```
#[vstd::contrib::auto_spec]
fn f(u: u8) -> u8 {
    u / 2
}
```

into this:

```
#[verifier::allow_in_spec]
fn f(u: u8) -> u8
    returns (u / 2)
{
    u / 2
}
```

`auto_spec` is similar to the `dual_spec` feature that is built into the attribute-based syntax ( https://github.com/verus-lang/verus/pull/1608 ), but unlike `dual_spec`, `auto_spec` is based on `returns` and works inside `verus!`.

So far, there are no libraries in `vstd/contrib`.  However, the name resolution for `contrib` macros goes through `vstd/contrib`, as in `#[vstd::contrib::auto_spec]` or as in:

```
use vstd::contrib::auto_spec;
#[auto_spec]
fn f(u: u8) -> u8 {
    u / 2
}
```

This naming convention is meant to make clear that macros in `contrib` are separate from the core Verus language features, even though they are packaged together in the same crate for ease of deployment.  (Note that Rust prohibits proc macros from residing in submodules.  For example, we can't declare the `autospec` macro in `builtin_macros/src/contrib`, only in `builtin_macros/src/lib.rs`.  We work around this limitation by re-exporting the `builtin_macros` `contrib` macros in `vstd::contrib`.)

### DIFF
--- a/source/builtin_macros/src/lib.rs
+++ b/source/builtin_macros/src/lib.rs
@@ -389,7 +389,7 @@ To add a contrib proc macro, complete the following steps:
   builtin_macros/src/contrib/mod.rs (example: `pub mod auto_spec;`)
 - Add a short macro declaration below, calling into your file in builtin_macros/src/contrib
   for any complex work (i.e. the macro declaration below should have a body of at most a few lines)
-- Add a "pub use" to vstd/contrib/mod.rs (example: `pub use builtin_macros::auto_spec;`)
+- Add a "pub use" to vstd/contrib/mod.rs (example: `pub use verus_builtin_macros::auto_spec;`)
 
 If your macro needs to manipulate function signatures or function bodies,
 it's generally cleaner to write this manipulation on the verus_syn representation of the function


tjhance (commented):
Can you document the steps a contributor needs to take to add a contrib macro?

Chris-Hawblitzel:
> Can you document the steps a contributor needs to take to add a contrib macro?

Sure.  Where's a good place to do this?  In the guide, for example?

parno:
> > Can you document the steps a contributor needs to take to add a contrib macro?
> 
> Sure. Where's a good place to do this? In the guide, for example?

Perhaps in `CONTRIBUTING.md`?

tjhance:
eh, i think the comments in the source code in the builtin_macros directory is fine. You already have a little bit of information, it's just not clear all the places you need to modify.

CONTRIBUTING.md might be another option

Chris-Hawblitzel:
> eh, i think the comments in the source code in the builtin_macros directory is fine. You already have a little bit of information, it's just not clear all the places you need to modify.

Ok, I added comments describing the process.

amaurremi:
Nit: Could we call the new `contrib` directories `extensions` or `extras` instead? I think `contrib` can read as second-class or experimental, while we probably would want to treat those as first-class optional features.

parno:
>  I think contrib can read as second-class or experimental, while we probably would want to treat those as first-class optional features.

I think the question is the level of support will these extensions receive?  As long as they have suitable tests, I think we can hopefully keep from breaking them as Verus changes.  However,  I think all of the examples that Chris mentions in his original post are incomplete in the sense that they'll work for some examples but not for others.  Given current resourcing, the core Verus team probably won't be able to provide fixes or enhancements for these gaps when users encounter them, so it would be helpful to signal that in the name somehow.  For myself, I think `contrib` probably strikes a good balance here.

amaurremi:
That's fair, sounds good to me.

PR #1815, opened 2025-07-15
Prepare Verus-related crates for posting to crates.io

parno:
This is a first step in the plan discussed in #1790.  Specifically, we
1. Rename our various crates (except `vstd`) to include a `verus` specifier, to be polite in our usage of the crates.io namespace. 
2. Adjust the syntax macro so that all references to `builtin` and `builtin_macros` go through the existing re-export in `vstd::prelude`.  This way, clients only need to list `vstd` as an explicit dependency in their `Cargo.toml` file.
3. Add/adjust meta-data in our various `Cargo.toml` files for crates.io purposes.

Step 1 touches a lot of files in uninteresting ways, so it may be easier to review this PR commit-by-commit, rather than using the final diff.  If the bulk renaming commits need special handling from git, let me know how that's done.

The next steps include:
1. Finding a `synstructure` version we can use (instead of a git commit hash), or removing our dependency on `synstructure`
2. Developing a script to keep our package versions numbers and dependencies in sync.

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -359,40 +359,38 @@ pub(crate) fn rewrite_verus_spec_on_fun_or_loop(
         AnyFnOrLoop::Fn(mut fun) => {
             // Note: trait default methods appear in this case,
             // since they look syntactically like non-trait functions
-            replace_block(erase, fun.block_mut().unwrap());
             let spec_attr =
                 verus_syn::parse_macro_input!(outer_attr_tokens as verus_syn::SignatureSpecAttr);
 
             fun.attrs.push(mk_verus_attr_syn(fun.span(), quote! { verus_macro }));
 
+            let mut new_stream = TokenStream::new();
+
             // Create a copy of unverified function.
             // To avoid misuse of the unverified function,
             // we add `requires false` and thus prevent verified function to use it.
             // Allow unverified code to use the function without changing in/output.
-            let mut new_stream = TokenStream::new();
             if let Some(with) = &spec_attr.spec.with {
-                let unverified_fun = rewrite_unverified_func(&mut fun, with.with.span());
-                unverified_fun.to_tokens(&mut new_stream);
+                let extra_funs = rewrite_unverified_func(&mut fun, with.with.span());
+                extra_funs.iter().for_each(|f| f.to_tokens(&mut new_stream));
             }
-            if fun.sig.constness.is_some() {
-                let mut const_fun = fun.clone();
-                let span = fun.sig.constness.unwrap().span();
-                // It seems that we do not need to erase anything.
-                // But just do it to be safe and consistent with verus macro.
-                replace_block(EraseGhost::Erase, const_fun.block_mut().unwrap());
-                const_fun.attrs.push(mk_verifier_attr_syn(span, quote! { external }));
-                const_fun.attrs.push(mk_verus_attr_syn(span, quote! { uses_unerased_proxy }));
-                const_fun.attrs.push(mk_verus_attr_syn(span, quote! { encoded_const }));
-                const_fun.to_tokens(&mut new_stream);
-                fun.sig.ident = syn::Ident::new(
-                    &format!("{VERUS_UNERASED_PROXY}{}", fun.sig.ident),
-                    fun.sig.ident.span(),
-                );
-                fun.attrs.push(mk_verus_attr_syn(span, quote! { unerased_proxy }));
-            }
+
+            // Update function signature based on verus_spec.
             let spec_stmts = syntax::sig_specs_attr(erase, spec_attr, &mut fun.sig);
+
+            // Create const proxy function if it is a const function.
+            if fun.sig.constness.is_some() {
+                let proxy = rewrite_const_ret_proxy(&mut fun);
+                fun.to_tokens(&mut new_stream);
+                fun = proxy; // Add proof and spec on proxy func.
+            }
+
+            // Add the spec/proof (requires/ensures) to the function body.
             let new_stmts = spec_stmts.into_iter().map(|s| parse2(quote! { #s }).unwrap());
             let _ = fun.block_mut().unwrap().stmts.splice(0..0, new_stmts);
+
+            // Parse and replace proof_xxx!() inside function and replace panic.
+            replace_block(erase, fun.block_mut().unwrap());
             fun.to_tokens(&mut new_stream);
             proc_macro::TokenStream::from(new_stream)
         }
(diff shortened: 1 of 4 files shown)


parno:
Also of interest to @jaybosamiya-ms

parno:
@tjhance The failing test is in `source/rust_verify_test/tests/basic.rs`, named `test_is_core`.

tjhance:
how does this effect tests? right now most tests don't import vstd for the sake of efficiency. Is that still the case?

tjhance:
in fact, does `--no-vstd` still work at all?

parno:
> how does this effect tests? right now most tests don't import vstd for the sake of efficiency. Is that still the case?

This diff changed how we refer to builtin inside the syntax macro.
```
+++ b/source/builtin_macros/src/syntax.rs
@@ -5114,9 +5114,11 @@ pub(crate) struct Builtin(pub Span);
 impl ToTokens for Builtin {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         if crate::cfg_verify_core() {
-            tokens.extend(quote_spanned! { self.0 => crate::verus_builtin });
+            tokens.extend(quote_spanned! { self.0 => crate::prelude });
+        } else if crate::cfg_verify_vstd() {
+            tokens.extend(quote_spanned! { self.0 => crate::prelude });
         } else {
-            tokens.extend(quote_spanned! { self.0 => ::verus_builtin });
+            tokens.extend(quote_spanned! { self.0 => ::vstd::prelude });
         }
     }
 }
```

This then prompted a change to `source/rust_verify_test/tests/common/mod.rs`, where I added `vstd::prelude` here:
```
 pub const USE_PRELUDE: &str = crate::common::code_str! {
+    use vstd::prelude::*;
     use verus_builtin::*;
     use verus_builtin_macros::*;
 };
```

source/builtin/Cargo.toml
```diff
@@ -2,6 +2,14 @@
 name = "verus_builtin"
 version = "0.1.0"
 edition = "2018"
+license = "MIT"
+description = "Verus Builtins: Rust definitions for core constructs in the Verus language"
```
Chris-Hawblitzel: I would say "in Verus" rather than "in the Verus language"

source/builtin_macros/Cargo.toml
```diff
@@ -2,6 +2,13 @@
 name = "verus_builtin_macros"
 version = "0.1.0"
 edition = "2018"
+license = "MIT"
+description = "Verus Builtin Macros: Various macros used by the Verus language"
```
Chris-Hawblitzel: I would say "by Verus" rather than "by the Verus language"

source/builtin/Cargo.toml
```diff
@@ -2,6 +2,14 @@
...
+documentation = "https://verus-lang.github.io/verus/verusdoc/vstd/pervasive"
+homepage = "https://github.com/verus-lang/verus"
+repository = "https://github.com/verus-lang/verus"
+readme = "../../README.md"
```
jaybosamiya-ms, line 10: Do we want to have a README specific to `builtin`/`builtin_macros` that points people at Verus and/or vstd instead?  Might be a little more convenient for folks who end up looking at these on crates.io to just directly get redirected.

source/builtin_macros/Cargo.toml
```diff
@@ -18,8 +18,8 @@ proc-macro2 = "1.0.39"
 quote = "1.0"
 synstructure = { git = "https://github.com/mystor/synstructure.git", rev = "1079497eb2bea252433dac53afe41291d8779641" }
 syn = { version = "2.0", features = ["full", "visit", "visit-mut", "extra-traits"] }
-syn_verus = { path="../../dependencies/syn", features = ["full", "visit", "visit-mut", "extra-traits"] }
-prettyplease_verus = { path="../../dependencies/prettyplease" }
+syn_verus = { version = "2.0.96", path="../../dependencies/syn", features = ["full", "visit", "visit-mut", "extra-traits"] }
+prettyplease_verus = { version = "0.2.29", path="../../dependencies/prettyplease" }
```
jaybosamiya-ms: Should these be `verus_syn` and `verus_prettyplease` (for the namespacing consistency)?

jaybosamiya-ms:
I added some comments inline, but also some more minor points:
- When specifying version numbers for the dependencies of vstd (e.g., builtin), we might want to do an explicit `=` version requirement, otherwise, the defaults are ["caret" version requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements).  Generally using caret version requirements _is_ the right move, since it allows semver-compatible updates to happen.  However, we are very _very_ particular about versions being in sync, in which case, we should be limiting it to exact version matches, rather than a caret requirement which means "the latest version that is semver-compatible with whatever was specified here".
- I think at some point in the past, there was a (brief) mention of `builtin` should probably have been called `builtins` but now we were stuck with `builtin`; if we are renaming to `verus_builtin`, this is probably the perfect time to update that if we still want to update that.  I personally think `builtins_macros` is too awkward for that, and having a `builtin_macros` alongside a `builtins` seems mildly annoying inconsistency too.  So yeah I don't think we should really use `builtins` but nonetheless, if we want to consider it, now is the time.

tjhance:
I would much rather we leave the majority of tests independent of vstd. vstd tests are substantially slower. I would also like `--no-vstd` to continue working since it's very useful in development.

Can you change the syntax macro to use `builtin` directly if no-vstd is set?

source/builtin/Cargo.toml
parno: Done.

source/builtin_macros/Cargo.toml
parno: Done.

source/builtin/Cargo.toml
parno: Do you mean we should add a tiny README.md for `builtin`/`builtin_macros` that says something to the effect of "Support crates for Verus.  See https://github.com/verus-lang/verus for more details"?

source/builtin_macros/Cargo.toml
parno: Good idea.  Done.

parno:
Thanks for all of the comments.  I've addressed several of them.

> Can you change the syntax macro to use builtin directly if no-vstd is set?

I can attempt that; I'll try to figure out how to connect that setting to the syntax macro.

parno:
Another oddity is that with the current version, I can successfully build Verus and verify vstd with vargo.  In a test Verus project that only lists `vstd` as a dependency, `cargo verus verify` succeeds, but `cargo build` fails when trying to compile `vstd`, reporting `failed to resolve: could not find vstd in the list of imported crates` in three spots:
```
   --> /Users/parno/git/verus/source/vstd/cell.rs:367:40
    |
367 |     pub fn new(val: T, Ghost(f): Ghost<spec_fn(T) -> bool>) -> (cell: Self)
    |                                        ^^^^^^^ could not find `vstd` in the list of imported crates

  --> /Users/parno/git/verus/source/vstd/map.rs:33:14
   |
33 |     mapping: spec_fn(K) -> Option<V>,
   |              ^^^^^^^ could not find `vstd` in the list of imported crates

   --> /Users/parno/git/verus/source/vstd/rwlock.rs:242:32
    |
242 | impl<V> RwLockPredicate<V> for spec_fn(V) -> bool {
    |                                ^^^^^^^ could not find `vstd` in the list of imported crates
```
which appear to be three places where, without `verus_keep_ghost`, the `::vstd::prelude::FnSpec` name "leaks" to the compiler.

parno:
@tjhance I think I've addressed your concern.  The `test_is_core` still fails with ~148 errors of the form:
```
error[E0433]: failed to resolve: unresolved import
    --> /Users/parno/git/verus/source/target/debug/test_inputs/basic-d6217e7fe140fc8d-test_is_core/../../../../vstd/seq_lib.rs:3562:17
     |
803  | /             assert_seqs_equal!(
804  | |                 self.subrange(k, self.len() as int).drop_last() ==
805  | |                 self.subrange(k, self.len() - 1)
806  | |             );
     | |_____________- in this macro invocation
...
3562 |           $crate::prelude::verus_proof_macro_exprs!($crate::vstd::seq_lib::assert_seqs_equal_internal!($($tail)*))
     |                   ^^^^^^^ unresolved import
```
I think this is because I changed some of the paths used in the macros written in `vstd`, e.g., 
```
macro_rules! assert_seqs_equal {
    [$($tail:tt)*] => {
-        ::builtin_macros::verus_proof_macro_exprs!($crate::vstd::seq_lib::assert_seqs_equal_internal!($($tail)*))
+        $crate::prelude::verus_proof_macro_exprs!($crate::vstd::seq_lib::assert_seqs_equal_internal!($($tail)*))
    };
}
```
Is there a way to write such macros that's generic as to whether we're in the i`s-core` mode?

parno:
[b259d39](https://github.com/verus-lang/verus/pull/1815/commits/b259d390ae1ee0d0b0aeab34a8b186c2d12150e5) fixed the issue with `spec_fn`, so I think the two remaining items are @jaybosamiya-ms's suggestion re: separate README files, and the issue with the `test_is_core`.

source/builtin/Cargo.toml
jaybosamiya-ms: Yeah, that's what I had in mind.  Having thought about it a bit, I don't know if that is really necessary though, since the way the current README is set up, they are told that this is Verus related, and adding a redirect that points them at literally that same README seems unnecessary to do until we see that someone was genuinely confused.

parno: Okay, sounds good.  Thanks for helping us think through this process!

Chris-Hawblitzel (approved):

PR #1826, opened 2025-07-19
seq_lib : add general lemmas about fold_left/fold_right

WeituoDAI:
add lemmas about fold_left/fold_right

- Give a more general version of  `lemma_fold_right_permutation` (I find it useful during my practice)
- prove the lemmas about the relation between fold_left and fold_right (s.reverse().fold_left  ~~ s.fold_right)
- make fold_left versions of existing fold_right lemmas

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -745,6 +745,44 @@ impl<A> Seq<A> {
         }
     }
 
+    /// [`Self::fold_left`] on the reversed sequence is equivalent to
+    /// [`Self::fold_right`] on the original sequence with corresponding folding operator
+    pub proof fn lemma_reverse_fold_left<B>(self, v: B, f: spec_fn(B, A) -> B)
+        ensures
+            self.reverse().fold_left(v, f) == self.fold_right(|a: A, b: B| f(b, a), v),
+    {
+        assert(self.reverse().reverse() =~= self);
+        let g = |a: A, b: B| f(b, a);
+        assert(f =~= |b: B, a: A| g(a, b));
+        self.reverse().lemma_reverse_fold_right(v, |a: A, b: B| f(b, a))
+    }
+
+    /// For any invariant `inv`, if the initial value `v` satisfies `inv`
+    /// and the fold_left operator preseves `inv`,
+    /// then the folding result preserves this invariant.
+    pub proof fn lemma_fold_left_preserves_inv<B>(
+        self,
+        f: spec_fn(B, A) -> B,
+        v: B,
+        inv: spec_fn(B) -> bool,
+        inv_e: spec_fn(A) -> bool,
+    )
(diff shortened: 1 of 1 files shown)


source/vstd/seq_lib.rs
```diff
@@ -820,6 +858,39 @@ impl<A> Seq<A> {
...
+        self,
+        a: A,
+        f: spec_fn(A, B) -> B,
+        v: B,
+        inv: spec_fn(B) -> bool,
+        inv_e: spec_fn(A) -> bool,
+    )
+        requires
+            forall|v: B, x: A, y: A|
+                inv(v) && inv_e(x) && inv_e(y) ==> #[trigger] f(x, f(y, v)) == f(y, f(x, v)),
```
zeldovich: Could `inv_e` be replaced by `self.contains`?  If the caller wants to weaken `self.contains` to some `inv_e`, that can be done as part of proving the implications in the requires clause, like this one and the one below.  This would simplify the caller's job for this lemma.

Similarly, could `inv(x)` be replaced by `exists |n: int| x == self.skip(n).fold_right(f, v)`?

If those definitions are workable for your use case, that means it might be fruitful to define a `commutative_foldr_in_seq(f, s)` that captures this notion of `f` being commutative with respect to sequence `s`, as opposed to in general.

WeituoDAI: I agree that we can remove `inv_e`, it must work. But if we replace `inv(v)` by something like `exists |n: int| x == self.skip(n).fold_right(f, v)`, then it may be hard for the caller to prove the commutativity. (Of course your suggestion makes the lemma more general, but here we need a trade-off)

In my practice, `lemma_fold_left_permutation_with_inv` comes together with `lemma_fold_left_preserves_inv`.

A typical use-case of these lemma : 
- Given a struct `S` with well-formed predicates `spec fn wf(self) -> bool`, and when I want to use the `lemma_fold_left_permutation` on it, I find that I cannot directly use it since the commutativity only holds when `wf` holds.

WeituoDAI: ok, I reconsider your point, for this lemma about permutation. Logically we do not need inv.. The proof will be a little harder, I will try it later.

WeituoDAI:
Following the advice of @zeldovich , I try to give a lemma like `lemma_fold_left_permutation` with weaker commutativity precondition. Anyhow, If I minimize the precondition, then the precondition seems too weak that nobody will really need it :) At last, I find that it goes far away from what I initially want. 

I  believe that using a user-defined invariant `inv : spec_fn(B) -> bool` can be a general method in real project if someone uses `Seq::fold_left` . But I doubt whether this lemma is suitable to be incorporated into vstd (I hardly see other functions in vstd take a user-defined predicates as a parameter)

I move other part of this pr into #1833

PR #1831, opened 2025-07-20
Place everywhere

tjhance:
(This depends on #1812)

OK, this is a complicated one...

As part of new_mut_refs, we need to compute the points where mutable references are resolved. An important input to this computation is knowing where all the _moves_ are. There is also another project that wants to use the move information.

Unfortunately, moves are not explicit in Rust's internal representation until the MIR level. I have come to the conclusion we will basically need to compute the moves ourselves.

Now, the tricky thing is that moves are contextual. If you write `a` for example, that doesn't necessarily mean you move `a`. If that `a` is actually a subexression of `a.f`, then that means, at most, you're moving only the field `f` (or moving nothing at all if the field is a `Copy` type). You can also consider scenarios like:

```
let (a, _) = x;   // only moves `x.0`
```

or:

```
{ f: foo, b: bar, .. x }   // moves all fields of `x` except for `f` and `b`
```

The expressions that you aren't necessarily moving from are called "place expressions", which are mostly [documented here](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions). e.g., `x`, `x.f`, `*x` etc. are all place expressions. The idea is these expressions compute a "place", and then how that place is used determines whether you move (or copy) something from it.

Therefore, I decided to explicitly compute all the place expressions. This makes a somewhat major change to the VIR's Expr type, adding `Place` nodes as a first-class part of the AST. I originally added Places in #1812 but didn't do much with them outside of mut_refs. This PR makes the Places a much more integral part of the AST.

Specifically, we add this to the ExprX enum:

```
ReadPlace(Place, ReadKind),
```

where:

```
pub enum ReadKind {
    Move,
    Copy,
    ImmutBor,
    Spec,
}
```

and:

```
pub enum PlaceX {
    Field(FieldOpr, Place),
    DerefMut(Place),
    Local(VarIdent),
    Temporary(Expr),
}
```

That means all vars and field expressions now go through the `ReadPlace`. So for example, `x.f` would previously be:

```
Expr::UnaryOpr(FieldOpr(...), ExprX::Var("x"))
```

But now it's:

```
Expr::ReadPlace(Place::Field(FieldOpr(..), PlaceX::Local("x")))
```

Also, decls now use `Place`s for their RHS and struct-update expressions use `Place` expressions after the `..`.

So, how do we compute the Place expression in rust_to_vir_expr? The `expr_to_vir` function is changed to return a `ExprOrPlace`. Note that we can always turn an Expr into a Place by making a Temporary, and we can make a Place into an Expr by making a ReadPlace (which requires us to determine the ReadKind, which can be done by querying if the type is `Copy`). This forces every caller of `expr_to_vir` to make a decision on whether they want to potentially coerce to a place or expression. 

**What this PR does:**

 * Updates the VIR Expr to use Place in a bunch of places.
 * Updates rust_to_vir to put the Place expressions in the right place.

**What this PR doesn't do:**

 * Determine which moves are "real" versus which ones are just for spec code.
 * Explicitly compute the move operations as nodes in the AST.

Both of these are reserved for future PRs.

Also note:

 * We don't need to do a similar paradigm shift in SST. Rather, we can compute all the Move nodes at the AST level.
 * I haven't changed the Loc system at all. The plan is still to replace this fully by the mut_refs work. Thus, the Loc system still uses the exact same representation it used before.


Chris-Hawblitzel:
Can you sketch out the algorithm you have in mind for computing the moves and the resolve points?

tjhance:
Well, from this change, the VIR will now have all the information it needs to compute moves directly. e.g. if we have ReadPlace(Move, a.b.c) or whatever, that's a move from a.b.c. There are some special cases like the struct ctors and patterns that can be handled on a case-by-case basis.

For the resolve points: it's safe to insert AssumeHasResolved at any point for a given place if it's guaranteed that place will never be mutated or moved from. So first, we use a dataflow analysis to compute "Is it safe to insert AssumeHasResolved here?" for every program point. With this information, we then insert AssumeHasResolved as early as possible (i.e., everywhere it's safe to insert, but where it wasn't safe to insert at the previous instruction).

Chris-Hawblitzel (approved):

PR #1833, opened 2025-07-22
vstd : add lemmas about Seq::fold_left/fold_right

WeituoDAI:
- add lemmas about transformation between fold_left and fold_right (s.reverse().fold_left ~ s.fold_right)
- add the lemma `lemma_fold_left_permutation`

### DIFF
--- a/source/vstd/seq_lib.rs
+++ b/source/vstd/seq_lib.rs
@@ -745,6 +745,18 @@ impl<A> Seq<A> {
         }
     }
 
+    /// [`Self::fold_left`] on the reversed sequence is equivalent to
+    /// [`Self::fold_right`] on the original sequence with corresponding folding operator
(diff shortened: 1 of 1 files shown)


source/vstd/seq_lib.rs
```diff
@@ -895,6 +936,28 @@ impl<A> Seq<A> {
         }
     }
 
+    /// Conversion of a sequence to multiset is equivalent to conversion of its reversion to multiset
+    proof fn lemma_reverse_to_multiset(self)
```
zeldovich: This seems like a useful lemma; perhaps it should be `pub`?

zeldovich (approved):
Looks good to me, though I don't feel like I have a strong enough sense of what should be or not be in `vstd` to unilaterally merge this.

parno (approved):
These all seem reasonably useful and general to me.  Thanks for the contribution!

PR #1837, opened 2025-07-23
Document additional assumption of `hash::obeys_key_model`

### LINKED ISSUE #1835: `HashMap::clone` allows to prove false facts with a non-standard `Clone` implementation
As described in [this issue](https://github.com/rust-lang/hashbrown/issues/629), if the `Clone` implementation of a `HashMap`'s key produces a key with a different hash, or if `key.clone() == key` does not hold, then this can result in a broken `HashMap` after `map.clone()`. These assumptions are not documented anywhere, which is why I made a [pull request](https://github.com/rust-lang/rust/pull/144330) to update the `Clone` trait documentation.

In Verus, the requirement of `key.clone() == key` for the keys of a `HashMap`/`HashSet` is (as far as I can tell) also not documented anywhere. [`HashMap::clone` is specified](https://verus-lang.github.io/verus/verusdoc/vstd/std_specs/hash/fn._verus_external_fn_specification_213__60__32_HashMap_32__58__58__32__60__32_K_44__32_V_44__32_S_32__62__32_as_32_Clone_32__62__32__58__58__32_clone.html) to return a map equal to the original one, which is not true if `key.clone() != key`. Due to this, it is possible to prove false facts about a cloned `HashMap`, as shown [here](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=0fff82729390eb3abc21c864176417a1) in the playground.

To my understanding, simply updating the `HashMap::clone` specification would not be enough to fix this issue: A `Clone` implementation that does not respect `key.clone() == key` can break very basic invariants of a `HashMap`, like every key occurring at most once in the map, so that afterwards, the specification of, e.g., `HashMap::remove` does not hold anymore. So maybe the requirement that `key.clone() == key` should be part of `obeys_key_model<Key>`, or be modelled similarly?

jaylorch:
Fixes #1835

### DIFF
--- a/source/vstd/std_specs/hash.rs
+++ b/source/vstd/std_specs/hash.rs
@@ -8,14 +8,15 @@
 ///
 /// Likewise, the specification is only meaningful when the `Key`
 /// obeys our hash table model, i.e., (1) `Key::hash` is
-/// deterministic and (2) any two `Key`s satisfying `==` are
-/// identical. We have an axiom that all primitive types and `Box`es
-/// thereof obey this model. But if you want to use some other key
-/// type `MyKey`, you need to explicitly state your assumption that it
-/// does so with
-/// `assume(vstd::std_specs::hash::obeys_key_model::<MyKey>());`.
-/// In the future, we plan to devise a way for you to prove that it
-/// does so, so that you don't have to make such an assumption.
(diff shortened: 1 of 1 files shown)


parno (approved):
Looks good.  Thanks for the documentation fix!

PR #1839, opened 2025-07-23
First draft for async function support

FeizaiYiHao:
Here's a draft on how we plan to support `Async` functions. The latest commit contains changes related to async. I'll clean it up asap. 
The high-level idea is the following: 
# In `verus!` macro:
For each async function 
```
async fn future_usize(x:&usize) -> (ret: usize) 
    ensures
        ret == x,
{
    x
}
```
We generate two functions in `verus!` macro:
A wrapper function that mimics the behavior of `Async` with the same name
```
#[verifier::external_body]
fn future_usize(x: &usize) -> (ret: impl Future<Ourput = usize>) 
    ensures
        ret.awaited() ==> {let ret = ret@; ret == x},
{
   make_future(VERUS_ASYNC_BODY_future_usize(x))
}
```
A body function that is renamed, verified but never called
```
fn VERUS_ASYNC_BODY_future_usize(x:&usize) -> (ret: usize) 
    ensures
        ret == x,
{
    x
}
```
For each `await`
```
let x = future.await;
```
it's translated into a `vstd` function
```
let x = verus_await(future);
```
# In lifetime generate
There are many Rust rules that only apply to async code and we'd like to keep them. Therefore, in lifetime generate, we reintroduce `Async` and `Await` keywords for the async functions.
```
fn VERUS_ASYNC_BODY_future_usize(x: &usize) -> (ret: usize) 
    ensures
        ret == x,
{
    x
}
```
back to
```
async fn future_usize(x: &usize) -> (ret: usize) 
    ensures
        ret == x,
{
    x
}
```
and 
```
let x = verus_await(future);
```
back to
```
let x = future.await;
```
So that we can still rely on Rustc for checking all these rules. For example:
```
let mut x = 233usize;
let future = future_usize(&x);
let x = 2333usize; //should fail because borrow of x does not end yet
let res = future.await; // borrow ends here
```

# In `vstd::future`
We define `ExFuture` trait for `core::Future`, `FutureSpecImpl` and `VerusFuture` for adding additional spec and exec functions for `Future`, and a `DummyFuture<T>` struct that acts as the concrete type behind `impl Future<Output = T>` during verification. Since the whole point of using opaque type is to hide the concrete type, I assume this is fine for verification? 

# opens invariants
Async functions straight up cannot open any invariants. They default to `opens invariants none`, and if they mark `opens invariants` with anything else, the macro throws an error.

# Uglyness 
In lifetime generate, `vstd::future::ExFuture` is indeed treated as a different new trait, and Rustc only allows `await` on `impl core::Future`. There are some type confusions between `ExFuture` generated by our wrapper functions and the `Future` generated by a real `async fn`. Therefore, I manually skip emitting lifetime checks for anything defined under `vstd::future::lifetime_ignore::*` and treat all the `ExFuture` as `core::Future`. Although this might sound less ugly, the code implementation is a bit ugly. Please let me know if there's some better solutions.


tjhance:
Quick thoughts before looking into this deeply:

 - There shouldn't be a problem with opening invariants in an async function. Requiring `opens_invariants none` would be very restrictive. We do, however, need to make sure that the program can't yield control while an invariant is open.
 - I would advise not spending much time fiddling with lifetime-generate. #1792 will probably be merging soon.

FeizaiYiHao:
> Quick thoughts before looking into this deeply:
> 
> * There shouldn't be a problem with opening invariants in an async function. Requiring `opens_invariants none` would be very restrictive. We do, however, need to make sure that the program can't yield control while an invariant is open.
> * I would advise not spending much time fiddling with lifetime-generate. [Implement lifetime-checking by forking rustc_mir_build #1792](https://github.com/verus-lang/verus/pull/1792) will probably be merging soon.

Yes. If we can mark each function with whether it has yield points, we can check and make sure each `open_invariant` does not include any yield point. But this seems to require a recursive reaserch each time we encourage an `open_invariant`.

parno:
FWIW, I think the initial use case, which Xiangdong is aiming to support, doesn't require any local/atomic invariants, so if it's non-trivial to make the policy on invariants more flexible, it might be pragmatic to get the conservative version working first, and then expand its capabilities.

tjhance:
Seems like this was superseded by the other async PR.

PR #1841, opened 2025-07-24
Update Veritas to Rust 1.88.0 and make future upgrades easier

hayley-leblanc:
This PR updates Veritas to use Rust 1.88.0, which is required by some of the repos it uses. It also makes some changes to how Veritas uses Docker that make it a bit easier to bump Veritas to future Rust versions, although they add a bit of overhead when using Veritas. Specifically:

1. This PR makes Veritas use locally-created containers. Veritas currently pulls several Docker images that have been built with specific Rust versions from a registry. Whenever Verus moves to a new Rust version, the registry images also have to be rebuilt and updated. Building the images is quite fast and the script to build them is already present, so I've updated the Veritas instructions to include a step to build the images (on a first run and/or after a version update) and updated all of the files that access the containers to use the locally-built versions rather than pulling them from the internet.
2. This PR makes Veritas create new volumes when cached versions may cause incompatibilities. Veritas creates several volumes to cache information used in each run. However, this can cause version issues when the cached Rust version is different from one that a repository expects. We saw this issue in the verified-storage repo, which contains several systems consisting of several unverified crates that must be built using `cargo` before verifying a main crate. If we used Veritas on Rust 1.85.1, we encountered version errors related to these crates (even though they have `rust-toolchain.toml` files specifying the correct version to use). Even after updating Veritas to 1.88.0, we continued to encounter these errors because the incorrect Rust version was cached somewhere in the Veritas-created volumes. This PR adds the current Rust version to the name of Rust-related volumes to ensure that new ones are created when the Rust version changes. 

If anyone has thoughts about alternative ways to do this (@utaal ?) I'd be happy to make changes.

### DIFF
--- a/tools/veritas/README.md
+++ b/tools/veritas/README.md
@@ -5,9 +5,11 @@ Veritas is a rust crater-alike tool to run a version of verus on a number of pro
 generating a report of verification success/failures and verification performance data.
 
 Running veritas requires docker (or another container runtime). 
-You can start a run with `bash run.sh path_to_run_configuration.toml`.
-There is an example run configuration in this directory.
-Running that command will start an ephemeral container and will create four
-permanent docker volumes: `verus-veritas-cargo-cache`, `verus-veritas-repo-cache`,
-`verus-veritas-rustup`, `verus-veritas-z3-cache`. These volumes cache dowloaded repositories,
-binaries, and other files, to reduce unnecessary traffic when performing multiple runs.
+
+
+### Running Veritas
+
+1. Run `bash build_images.sh` to create the Docker images Veritas uses locally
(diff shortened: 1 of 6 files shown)


utaal (approved):
This is a much needed improvement, thanks!
One note: I believe this will leave the cache volumes of older versions lying around on the user's docker host. Consider adding to the readme a note on how to delete these outdated volumes so they don't take up disk space unnecessarily.

hayley-leblanc:
Thanks Andrea! I just pushed an update to the README adding instructions on how to delete the old volumes.

I don't seem to have permissions to merge PRs even after they are approved, so someone else will need to merge this.

Chris-Hawblitzel:
> I don't seem to have permissions to merge PRs even after they are approved, so someone else will need to merge this.

I just added write permission for you, so hopefully you can merge it now.

PR #1714, opened 2025-05-30
make pow opaque

ahuoguo:
@parno spotted `pow` function is not marked as opaque even though it's revealed everywhere.

For context, Dafny standard library also [did not mark it as opaque now](https://github.com/dafny-lang/dafny/blob/master/Source/DafnyStandardLibraries/src/Std/Arithmetic/Power.dfy), even though it was marked as opaque before.

Seems like part of the reason I did not mark opaque is because I wanted to `reveal(pow)` in the spec function of `pow2`

https://github.com/verus-lang/verus/blob/20c3b088cc854504638586b1e64a9ed41e529a0d/source/vstd/arithmetic/power2.rs#L26-L37

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #1849, opened 2025-08-01
vstd: `lemma_pow2_increases`

mmcloughlin:
Add `lemma_pow2_increases` lemma, parallel to `lemma_pow_increases`.

### DIFF
--- a/source/vstd/arithmetic/power2.rs
+++ b/source/vstd/arithmetic/power2.rs
@@ -22,6 +22,7 @@ use super::power::{
     lemma_pow_positive,
     lemma_pow_adds,
     lemma_pow_strictly_increases,
+    lemma_pow_increases,
     lemma_pow_subtracts,
 };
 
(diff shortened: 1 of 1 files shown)


source/vstd/arithmetic/power2.rs
```diff
@@ -113,6 +114,18 @@ pub broadcast proof fn lemma_pow2_strictly_increases(e1: nat, e2: nat)
     lemma_pow_strictly_increases(2, e1, e2);
 }
 
+/// Proof that if `e1 <= e2` then `2^e1 <= 2^e2`.
+pub broadcast proof fn lemma_pow2_increases(e1: nat, e2: nat)
+    requires
+        e1 <= e2,
+    ensures
+        #[trigger] pow2(e1) <= #[trigger] pow2(e2),
```
mmcloughlin, line 122: This matches `lemma_pow2_strictly_increases` but I'm not sure why `#[trigger]` is on both expressions.

parno: That does seem redundant.  I suspect we could remove it both here and in the original `pow` function.

parno (commented):
I think a broader question here is whether `pow2` should be opaque.  If it's defined in terms of `pow`, which is opaque, I don't think we're gaining anything from marking `pow2` as opaque and then repeating all of these proofs.

parno:
It's possible that decision was leftover from before `pow` was [marked as opaque](https://github.com/verus-lang/verus/pull/1714).

mmcloughlin:
> I think a broader question here is whether `pow2` should be opaque. If it's defined in terms of `pow`, which is opaque, I don't think we're gaining anything from marking `pow2` as opaque and then repeating all of these proofs.

So the other possibility is removing `opaque` from `pow2` and deprecating the `pow2`-variant lemmas? Or even deleting them? If so, then users would just use the `pow` lemmas instead, which should just work if the `pow2` spec function is open?

parno:
> So the other possibility is removing opaque from pow2 and deprecating the pow2-variant lemmas? Or even deleting them? If so, then users would just use the pow lemmas instead, which should just work if the pow2 spec function is open?

Yes, I think if we remove opaque from pow2, then calling the main `pow` lemmas (or broadcasting them) should just work (although it would be good to double check that).  Then we could deprecate and eventually remove the redundant pow2 lemmas.

Issue #1854, opened 2025-08-06
Support for mixed sign and size in integer type conversions


rod-chapman:
Currently, it appears that Verus does not "know" the semantics of integer type conversions that change the signedness of the argument, or its size (either truncating or expanding the representation.) Verus currently issues a "truncating warning", and nothing can be proven about the result of the conversion.

Rust Reference 8.2.4 says that these conversions are well-defined, so they should be supported. They are also common in cryptographic code.

The work could be broken down into several PR's. For example.
1. Same size, changing sign (e.g. i32 <-> u32 and so on)
2. Expanding, same sign (e.g. u8 -> u16, i16 -> i32)
3. Truncating, same sign (e.g. u16 -> u8, i32 -> i16)
4. Expanding, mixed signedness (e.g. i8 -> u16, u16 -> i32)
5. Truncating, mixed signedness (e.h. u32 -> i8, i16 -> u8)

Did I miss any cases?

tjhance:
These semantics are not exported in Verus's default prover mode, but they are available in other modes such as 'bit_vector' or 'compute'.

Relevant docs:

 * https://verus-lang.github.io/verus/guide/bitvec.html#integer-wrapping-and-truncation
 * https://verus-lang.github.io/verus/guide/reference-as.html

tjhance:
You can also silence the warnings using the `truncate` attribute: https://verus-lang.github.io/verus/guide/reference-attributes.html#verifiertruncate

mmcloughlin:
Broadly, as @tjhance says I think Verus can reason about this with the help of `bit_vector` proofs. Some lemmas may make it easier though.

For case 1 above, some lemmas for properties of sign casts: https://github.com/verus-lang/verus/pull/1855

Chris-Hawblitzel:
Since all the conversions have simple and well-defined semantics, maybe we should support them all without the need for `bit_vector`.  @tjhance points: 

> If we're casting from iN to uN or vice versa, there is a much easier way to express the relationship that does not involve mod, just a single if statement:
>
> x as uN == (if x < 0 { x + uN::MAX + 1 } else { x })
> 
> For something fundamental like this, I think we should consider exporting it in the AIR prelude, where most of the other arithmetic axioms are. Right now, we don't export anything for truncations because we don't want to bring in nonlinear stuff by default, but an exception could be made for the truncations that don't require modulo.

[...]

PR #1855, opened 2025-08-06
vstd: signedness casts preserve modulo

mmcloughlin:
Prove that casts between `iN` and `uN` preserve the value modulo the integer
range.

Updates #1854 

### DIFF
--- a/source/vstd/bits.rs
+++ b/source/vstd/bits.rs
@@ -35,38 +35,55 @@ use super::calc_macro::*;
 
 } // verus!
 macro_rules! lemma_cast_signed_to_unsigned_preserves_mod {
-    ($name:ident, $iN:ty, $uN:ty, $range:expr) => {
+    ($name_signed_to_unsigned:ident, $name_unsigned_to_signed:ident, $iN:ty, $uN:ty, $range:expr) => {
         #[cfg(verus_keep_ghost)]
         verus! {
-        proof fn $name(s: $iN)
+        proof fn $name_signed_to_unsigned(s: $iN)
             ensures
                 (s as $uN) as int % $range == s as int % $range
         {
             assert(s >= 0 ==> s as $uN == s as int);
             assert(s < 0  ==> s as $uN == $range + (s as int)) by (bit_vector);
         }
+
+        proof fn $name_unsigned_to_signed(u: $uN)
+            ensures
+                (u as $iN) as int % $range == u as int % $range
+        {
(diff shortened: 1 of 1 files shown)


tjhance:
If we're casting from iN to uN or vice versa, there is a much easier way to express the relationship that does not involve mod, just a single if statement:

x as uN == (if x < 0 { x + uN::MAX + 1 } else { x })

For something fundamental like this, I think we should consider exporting it in the AIR prelude, where most of the other arithmetic axioms are. Right now, we don't export anything for truncations because we don't want to bring in nonlinear stuff by default, but an exception could be made for the truncations that don't require modulo.

mmcloughlin:
For context, these lemmas help with some crypto proofs that depend arithmetic modulo `2^n`,
and the fact that these casts preserve the value modulo word size. It may be useful to have them even if we extend the prelude as you describe.

To make sure I understand what you mean, we'd extend the prelude with a couple of new operators for sign casts? Verus would use them when it sees casts between `iN` and `uN`?

Chris-Hawblitzel:
> If we're casting from iN to uN or vice versa, there is a much easier way to express the relationship that does not involve mod, just a single if statement

> Verus would use them when it sees casts between iN and uN?

I think there's a simple variant that takes advantage of this optimization:

```
pub broadcast proof fn u8_to_i8(u: u8)
    requires
        u >= 0x80
    ensures
        #[trigger] (u as i8) == u - 0x100,
{
    assert(u as i8 == u - 0x100) by(bit_vector) requires u >= 0x80;
}

proof fn test() {
    broadcast use u8_to_i8;
    assert(0x20u8 as i8 == 0x20i8);
    assert(0xf0u8 as i8 == -0x10i8);
}
```

In AIR, the broadcast looks like:

```
;; Broadcast crate::u8_to_i8
(axiom (=>
  (fuel_bool fuel%u8_to_i8.)
  (forall ((u! Int)) (!
    (=>
     (uInv 8 u!)
     (=>
      (>= u! 128)
      (= (iClip 8 u!) (Sub u! 256))
    ))
    :pattern ((iClip 8 u!))
))))
```

This takes advantage of the type invariant `(uInv 8 u!)` that gets added because `u` has type `u8`.  Because of this type invariant, no mod operation is needed.  Furthermore, the broadcast only needs to prove something about the `u >= 0x80` case, because the clients already know the `u < 0x80` case from the prelude's existing axioms about `iClip`.

PR #1856, opened 2025-08-06
vstd: lemma `1 << n` is `pow2(n)`

mmcloughlin:
Provide lemmas `lemma_uN_one_shl_is_pow2` that `1 << n` is equal to `pow2(n)`. Basic fact come up in crypto proofs and general bit manipulation.


parno:
Should we generalize this to something that says `x << n` is `x * pow2(n)`?  I guess overflow gets a bit messier in that case.

rikosellic:
`lemma_uN_mul_pow2_le_max_iff_max_shr` may be helpful in the generalized case.

PR #1859, opened 2025-08-11
f32/f64 types and constants, with uninterpreted f32/f64 arithmetic

Chris-Hawblitzel:
This supports f32/f64 types and f32/f64 constant values, and declares arithmetic operations with uninterpreted semantics.  This is the first stage of floating point support as proposed in https://github.com/verus-lang/verus/discussions/1853.  This does not yet take advantage of the SMT solver's floating point theory; that can come in a later pull request.  It does include some basic properties of f32/f64 numbers to help classify floating point constants:

```
impl FloatBitsProperties for f32 {
    open spec fn to_bits_spec(&self) -> int {
        f32_to_bits(*self)
    }

    open spec fn to_bits_abs_spec(&self) -> int {
        let bits = self.to_bits_spec();
        if bits >= 0x8000_0000 {
            bits - 0x8000_0000
        } else {
            bits
        }
    }

    // Sign bit is true (may apply to zero, subnormal, normal, infinite, NaN)
    open spec fn is_sign_negative_spec(&self) -> bool {
        self.to_bits_spec() >= 0x8000_0000
    }

    open spec fn is_finite_spec(&self) -> bool {
        self.to_bits_abs_spec() < 0x7f80_0000
    }

    open spec fn is_infinite_spec(&self) -> bool {
        self.to_bits_abs_spec() == 0x7f80_0000
    }

    open spec fn is_nan_spec(&self) -> bool {
        self.to_bits_abs_spec() > 0x7f80_0000
    }
}
```

In AIR/SMT, f32/f64 are represented as `int` values in the same range as u32/u64.

Here is an example of some limited reasoning about f64 (found in examples/float.rs):

```
use vstd::std_specs::ops::AddSpec;
use vstd::float::FloatBitsProperties;

/*
Verus deliberately omits axioms about floating point from vstd,
because the desired set of useful and sound axioms may vary by project and platform.
(See https://github.com/rust-lang/rfcs/blob/master/text/3514-float-semantics.md for details
about why Rust floating point semantics are complex, may be non-deterministic, and may fall short
of desired behavior on some platforms.)
Therefore, projects that want to prove properties of about floating-point numbers may want
to define their own axioms, or even define different groups of axioms for different situations.

For example, one useful axiom is that it is always safe to add any two floats
(this assumes that the platform is correctly configured not to trap on a NaN result,
which should usually be true):
*/

pub broadcast axiom fn f64_can_add_anything(a: f64, b: f64)
    ensures
        #[trigger] a.add_req(b);

/*
The axiom above doesn't guarantee non-NaN results -- it's possible to add large positive numbers to
construct positive infinity, to add large negative numbers to construct negative infinity,
and then add negative infinity to positive infinity to construct a NaN.
As an example of verifying something slightly nontrivial,
the axioms below only permit addition of positive numbers,
and guarantee non-NaN results.
*/

pub broadcast axiom fn f64_add_positive_spec(a: f64, b: f64)
    requires
        !a.is_nan_spec(),
        !b.is_nan_spec(),
        !a.is_sign_negative_spec(),
        !b.is_sign_negative_spec(),
    ensures
        #![trigger a.add_spec(b)]
        !a.add_spec(b).is_nan_spec(),
        !a.add_spec(b).is_sign_negative_spec();

pub broadcast axiom fn f64_add_positive_exec(a: f64, b: f64)
    requires
        !a.is_nan_spec(),
        !b.is_nan_spec(),
        !a.is_sign_negative_spec(),
        !b.is_sign_negative_spec(),
    ensures
        #[trigger] a.add_req(b);

use vstd::std_specs::ops::add_ensures;

pub broadcast axiom fn f64_add_positive_ensures(a: f64, b: f64, o: f64)
    requires
        !a.is_nan_spec(),
        !b.is_nan_spec(),
        !a.is_sign_negative_spec(),
        !b.is_sign_negative_spec(),
    ensures
        #[trigger] add_ensures::<f64>(a, b, o) ==> o == a.add_spec(b);

pub broadcast group f64_add_positive {
    f64_add_positive_spec,
    f64_add_positive_exec,
    f64_add_positive_ensures,
}

fn main() {
    broadcast use f64_add_positive;
    let a: f64 = 3.1;
    let b: f64 = 2.8;
    let c = a + b;
    let d = b + c;
    let e = c + d;

    // This would fail the !b.is_sign_negative_spec() precondition:
    // let f = e + (-0.7);

    // But if we use the more permissive axiom, then we can add a negative number
    // (albeit with no guarantee about the result):
    broadcast use f64_can_add_anything;
    let f = e + (-0.7);
}
```


source/vstd/float.rs
```diff
@@ -0,0 +1,89 @@
+//! Properties of floating point values.
+use super::prelude::*;
+
+verus! {
+
+pub trait FloatBitsProperties {
+    spec fn to_bits_spec(&self) -> int;
```
tjhance: I might have expected this to return a finite width integer (e.g., f64 -> u64, f32 -> u32). Why'd you pick int?

source/builtin/src/lib.rs
```diff
@@ -1337,6 +1337,20 @@ impl_binary_op!(SpecShr, spec_shr, Self, [
...
+pub fn f32_to_bits(_f: f32) -> int {
+    unimplemented!()
+}
+
+#[cfg(verus_keep_ghost)]
+#[verifier::spec]
+#[rustc_diagnostic_item = "verus::verus_builtin::f64_to_bits"]
+pub fn f64_to_bits(_f: f64) -> int {
+    unimplemented!()
+}
```
tjhance, line 1359: Do these need to be builtins?

source/vstd/float.rs
Chris-Hawblitzel: I don't have a strong opinion on this.  Either results in the AIR type `int` with bounds ensuring it's in u32 or u64 range.  I just pushed a commit to change these to u32/u64 in case that seems nicer.

source/builtin/src/lib.rs
Chris-Hawblitzel: sst_to_air has to know about them, and sst_to_air treats them similar to how integer casts are treated with iClip/uClip (f32_to_bits/f64_to_bits actually have the semantics of no clip at all, but it's the same idea).  `builtin` is my preferred place for sst_to_air kind of things.

source/vstd/float.rs
tjhance: If they are fixed-width it will be slightly easier to use them in bit_vector, which may be a consideration if someone wants to work with the encoding.

tjhance (approved):

PR #1864, opened 2025-08-12
Automated versioning for Verus crates

parno:
This PR is the next step in the [campaign to simplify installing and using Verus](https://github.com/verus-lang/verus/discussions/1790).  It adds a new Rust tool for updating our Verus crates' versions in a coherent manner.  This PR also sets the versions for all of the crates in a style that's consistent with the auto-generated version numbers.


parno:
@jaybosamiya-ms If you have any thoughts or suggestions, they'd be most appreciated.

parno:
@Chris-Hawblitzel @tjhance After this PR, I think the last thing keeping us from pushing to crates.io is the dependence of `builtin_macros` on a particular commit of `synstructure`:
```
synstructure = { git = "https://github.com/mystor/synstructure.git", rev = "1079497eb2bea252433dac53afe41291d8779641" }
```
Would either of you be able to either change builtin to not rely on `synstructure` or to make builtin work with a numbered version of `synstructure`?

parno:
Just to be explicit, the intent is that we'd augment the GitHub `release` workflow so that it automatically calls the new version bump tool and then pushes any changed crates to crates.io.

parno:
> However, I'm not sure, in general, how local development on the Verus crates will work once we switch to crates.io versioning instead of `path` dependencies.  For example, if I edit `vstd` or `builtin_macros` locally, can we still build and test Verus without pushing to crates.io all the time?

I learned from Jay that we can address this by including both a `path` and a `version` entry for each of our Verus crate dependencies, so I've done that now.  I think that just leaves the open question about `synstructure`

source/tools/bump_crate_versions/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let now = Utc::now();
+    format!(
+        "0.0.0-{}-{:02}-{:02}-{:04}",
+        now.year(),
+        now.month(),
+        now.day(),
+        now.hour() * 100 + now.minute()
+    )
+}
```
jaybosamiya-ms: A possibly-safer option is to actually derive the date from the latest commit that you are bumping the versions on top of.

```sh
git log -1 --format=%cd --date=format:'%Y-%m-%d-%H%M%S'
```

source/tools/bump_crate_versions/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+
+fn update_cargo_verus_template(new_version: &str) {
+    let main = Path::new(CARGO_VERUS_MAIN);
+    let content = fs::read_to_string(main).expect("Failed to read cargo-verus main.rs");
+
+    // Replace the version in the template
+    let re = Regex::new("(?m)^vstd =.*$").expect("Failed to create regex");
+    let updated_content = re.replace(&content, format!("vstd = \"={}\"", new_version).as_str());
+    println!("Updated cargo-verus main.rs:\n{}", updated_content);
```
jaybosamiya-ms: Minor: might be good to confirm that the regex occurs _exactly_ once; otherwise there is a chance of things getting difficult to debug in the future.

source/tools/bump_crate_versions/src/main.rs
jaybosamiya-ms: Also, one way to ensure that it is called only once is to use a LazyCell: https://doc.rust-lang.org/std/cell/struct.LazyCell.html

parno: Thanks for the pointer -- I vaguely remembered Lazy*, but didn't have access to the docs when I was working on this on the bus :)  I've updated the code to use a LazyLock + a const.  I think this is simpler than computing the date of the latest commit.

source/tools/bump_crate_versions/src/main.rs
parno: Good point -- done!

source/tools/bump_crate_versions/src/main.rs
jaybosamiya-ms: There's always a local copy of the docs available: `rustup doc --std` (or more generally, just `rustup doc` which gives links to all the various documentation for that toolchain).  Also, `cargo doc --open` is super helpful to also pull up the docs of the dependencies too.

parno: Wild, thanks for the tip!

parno:
@jaybosamiya-ms The updater currently bumps the version for each crate that changed since the last update to its Cargo.toml file.  For each crate that depends on the modified crate, we update the version in its Cargo.toml's dependency section.  Should that in turn bump that crate's version?  Or is the assumption that if the parent depends in some important way on the child's updated behavior, then its code will change, which the updater will notice?  And if no code change was required, then it's not important to push out the new version of the parent?

jaybosamiya-ms:
(Using "A depends on B" and "B's code changed, but A's did not" as the scenario)

If A is using _equality_ versions rather than caret/tilde/no-symbol versions, then you need to bump A's version to let A's users get B's newer version.  If A is using the caret/... versions, then you don't need to bump A; folks will automatically notice that B's newer version is compatible with A, and thus attempt using that one.

So the actual version bumps should be if either code changed, or if the code in any equality dependencies changed.

PR #1865, opened 2025-08-13
Fix recursive call's termination check inside a loop

### LINKED ISSUE #1832: Fails to prove termination when recursive calls are placed inside a loop
The following function fails to prove termination unless annotated with #[verifier::loop_isolation(false)]:
```
pub fn test(x: usize)
    decreases x,
{
    if x == 0 {
        return;
    }
    for i in 0..1 {
        test(x - 1);
    }
}
```
Is this a reasonable default behavior? It seems the user can not express explicit termination-related conditions in the loop invariant, so `loop_isolation(false)` is the only solution.

rikosellic:
Fix #1832 by performing an additional termination check during air generation if loop_isolation is enabled.

### DIFF
--- a/source/rust_verify_test/tests/loops.rs
+++ b/source/rust_verify_test/tests/loops.rs
@@ -1481,10 +1481,10 @@ test_verify_one_file_with_options! {
     } => Err(err) => assert_fails(err, 4)
 }
 
-test_verify_one_file!{
-    #[test] recursive_call_in_loop verus_code! {
+test_verify_one_file! {
+    #[test] recursive_call_in_loop1 verus_code! {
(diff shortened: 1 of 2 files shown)


rikosellic:
Well, this failure is quite unexpected.

Chris-Hawblitzel:
Thanks!  One change that needs to be made is that sst_to_air can't get an Stm directly from check_termination_stm, since this Stm would not get processed as part of the normal AST -> SST -> AIR pipeline (e.g. processed by poly.rs to handle spec fn calls).  I've revised this to go through the regular AST -> SST -> AIR pipeline and I'll merge the changes together and push it into the main branch.

PR #1866, opened 2025-08-18
Suggestions for external_type_specification and assume_specification

rsofaer:
I have been using Verus to verify parts of an existing unverified Rust project, and one of the major barriers has been the time to get to '0 verified' on a piece of code by resolving compatibility issues and setting a coherent boundary of verification.  

I would like to improve this situation, starting by making it so that all of the external functions and types referenced by verified code are reported at once, with a specific specification suggestion that will correctly mark the verification boundary, rather than requiring a separate compilation for each error.  In particular, this will help with situations like `format!`, where it takes several specifications to get to a point where the actual barrier to using the macro (visibility of a type in a private module) is revealed.

### Changes:

This pull request adds suggestions for `assume_specification` and `external_type_specification` declarations.  In addition, it makes the errors produced by `vir/src/well_formed.rs` on encountering one of these verification boundary violations non-blocking.  The well_formed code now can generate a false `Datatype` or `Function` in order to progress through the code and collect as many errors as possible.

In order to cleanly gather diagnostic information into one data structure without conflicting borrows, the `WalkASTVisitor` now carries an `FnMut` for emitting diagnostic information, and I had to add a `PhantomData` marker to make the input type of that generic.

Suggestions are constructed after well_formed checking and appended to the generated errors.

### Example:
This code:
```
        verus! {
            fn foo(x: &str, y: &str) -> (res: bool)
            {
                x == y
            }
        }
```
Creates this error:
```
error: `core::cmp::impls::impl&%9::eq` is not supported (note: you may be able to add a Verus specification to this function with `assume_specification`) (note: the vstd library provides some specification for the Rust std library, but it is currently limited)
  --> /local/home/rsofaer/verus/source/target/debug/test_inputs/boundary_suggestions-178984263f1c02b7-test_assume_specification_str_eq_suggestion_made/test.rs:19:9
   |
19 |         x == y
   |         ^^^^^^
   |
           pub assume_specification<'a, 'b, A, B> [<&'b A as std::cmp::PartialEq<&B>>::eq] (_0: &&'b A, _1: &&B) -> bool
           where
           A: std::cmp::PartialEq<B> + ?Sized,
           B: ?Sized,;
```


tjhance:
> The errors occur in vir/src/well_formed.rs, after the mir information has been discarded.

This isn't correct. While it's true that the information is not *accessible* from VIR, it is still accessible from the rust_verify crate after well_formed returns. I expect you can propagate error information much more easily by simply having `well_formed` return a list of functions to construct suggestions for, without having to precompute any diagnostics.

rsofaer:
Great point, I will revise based on that.

rsofaer:
Any suggestion for getting from the vir::ast::Path we have in the well_formed crate to the original DefId?  It's not clear to me that the logic in `def_path_to_vir_path` can be reversed, since ValueNs and TypeNs become the same thing and  ForeignMod is dropped.  Are you open to adjusting the content of Path?

tjhance:
You could probably create a reverse map Path -> DefId at the rust_verify level

rsofaer:
~~Seems like I'd have to populate that reverse map either by creating a static map like `PATH_AS_RUST_NAME_MAP` or by keeping a RefCell in the Context, like the diagnostics that's already in there (`std::rc::Rc<std::cell::RefCell<Vec<vir::ast::VirErrAs>>>`), and passing it into def_id_to_vir_path to be populated (maybe optionally.  Do you have a preference?~~

Never mind I did the RefCell.  I also moved the suggestions to the error 'help', I'll update the PR once I have the tests working again.

rsofaer:
OK, the PR is updated with a Path->DefId map added to Context, the changes to erasure info removed, the shim Datatype and Function construction moved to well_formed.rs and done inline with error creation, and the suggestion construction now done after vir crate construction.  Suggestions are also moved to the help section of the error, and references to an external type are accumulated into secondary spans.

rsofaer:
I just updated again with a small adjustment to the def_paths that makes the suggestions more LLM-consumable.

PR #1872, opened 2025-08-22
Fix termination check inside loop when decreases clause contains &mut params

WeituoDAI:
Related to #1832

Thanks for the quick fix, but I find a bug when the decreases clause contains mutable parameters (i.e. old(var)).

### DIFF
--- a/source/rust_verify_test/tests/loops.rs
+++ b/source/rust_verify_test/tests/loops.rs
@@ -1600,3 +1600,26 @@ test_verify_one_file_with_options! {
         }
     } => Ok(_err) => {/* allow decreases checks warnings */ }
 }
+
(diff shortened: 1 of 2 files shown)


tjhance:
~~I think something's wrong here - previously, we had the 'snapshot' line in between the 2 'typ_invariant' loops, so type invariant assumptions would be added for both the pre-snapshot and post-snapshot state. But now that you've moved the snapshot line, it looks like we're only generating type invariants for the post-snapshot state.~~

tjhance:
actually never mind, I see that the type invariants are being added to a different vec, so the re-ordering doesn't matter.

Chris-Hawblitzel (approved):

PR #1877, opened 2025-08-28
fix ast_to_sst encoding of calls with args containing assignments

tjhance:
currently we get it backwards on the following test case:

```rust
fn foo(x: u64, y: u64) -> (ret: (u64, u64))
    ensures ret == (x, y)
{
    (x, y)
}

fn test2() {
    let mut x = 24; 
    let mut y = 30; 

    let z = foo(x, ({ x = 60; y }));

    assert(z === (24, 30)); // this should pass
    assert(z === (60, 30)); // this shouldn't
}
```


tjhance:
update: realized this is a problem for more than just calls. I fixed ExprX::Ctor as well, and then I went down the list and identified `BinaryOp`, `BinaryOpr`, and `ArrayLiteral` as needing the same fix (Update: now done).

Chris-Hawblitzel (approved):

PR #1881, opened 2025-09-01
Update the existing GitHub release workflow to also publish any updated crates to crates.io

parno:

### DIFF
--- a/source/tools/bump_crate_versions/src/main.rs
+++ b/source/tools/bump_crate_versions/src/main.rs
@@ -1,16 +1,24 @@
 use std::{fs, path::Path, process::Stdio, sync::LazyLock};
 use toml_edit::DocumentMut;
 use regex::Regex;
+use clap::Parser as ClapParser;
 
-/*
- * This tool scans for modified crates in the Verus repository and updates the version numbers
- * in their respective Cargo.toml files. In cases where one crate depends on another, we also
- * update the version of the dependency in the dependent crate's Cargo.toml.  Finally, when vstd
- * is modified, we also update the version in the cargo-verus template.  The code is optimized
- * for readability and maintainability, rather than performance.
- *
- * Usage: Run this tool from the root of the Verus repository.
- */
+
+/// This tool scans for modified crates in the Verus repository and updates the version numbers
+/// in their respective Cargo.toml files. In cases where one crate depends on another, we also
+/// update the version of the dependency in the dependent crate's Cargo.toml.  Finally, when vstd
+/// is modified, we also update the version in the cargo-verus template.  The code is optimized
+/// for readability and maintainability, rather than performance.
+///
+/// Usage: Run this tool from the root of the Verus repository.
+
+#[derive(ClapParser)]
+#[command(version, about)]
+struct Args {
+    /// If set, update the versions and dependencies, but run `cargo publish` in dry-run mode
+    #[arg(long = "dry-run")]
+    dry_run: bool,
+}
 
 // Path to cargo-verus's main file, where we have a static string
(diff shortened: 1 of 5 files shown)


source/tools/bump_crate_versions/src/main.rs
```diff
@@ -1,16 +1,24 @@
...
+/// for readability and maintainability, rather than performance.
+///
+/// Usage: Run this tool from the root of the Verus repository.
+
+#[derive(ClapParser)]
+#[command(version, about)]
+struct Args {
+    /// If set, update the versions and dependencies, but run `cargo publish` in dry-run mode
+    #[arg(long = "dry-run")]
+    dry_run: bool,
```
jaybosamiya-ms: Suggestion: instead have a `--publish` flag, that is the opposite of dry-run.

Otherwise, based on descriptions, it kinda seems like simply saying `bump_crate_versions` bumps crate versions, and `bump_crate_versions --dry-run` _also_ runs `cargo publish --dry-run` but the actual publishing should happen separately.

It would also make the CI script be more obvious that it is not just bumping versions but also publishing.

parno: Ah, good point.  I'll toggle the polarity.  Thanks for noticing that.

parno: I applied this, but then I had a change of heart and swapped to a sub-command style.  The new proposed workflow is:
```
bump_crate_versions update   # Updates the various Cargo.toml files
git commit   
bump_crate_versions publish  # Push to crates.io (unless provided with the --dry-run flag)
```
We need all three pieces, since we want to persist the version bumps, and because `cargo publish` complains if the repo is dirty.

In the GitHub release workflow, we can use the [git-auto-commit action](https://github.com/stefanzweifel/git-auto-commit-action) for that middle step.

Does that seem sane?

jaybosamiya-ms: Based on a quick read of the git-auto-commit action, it seems like a pretty broad permission to be given to a 3rd party action for what could just be done in the run steps.

Also, workflow wise, do you expect the automatic commit to get pushed to `main` or something?  I am not I'm fully comfortable with broad permissions of that form (important question to answer: can an attacker create a malicious PR to trigger bad behavior somehow?).  

If the created commit is not getting pushed and is only used to silence `cargo publish` that's equivalent to telling cargo `--allow-dirty` no?

Oh, and just as a quick sidenote: I would recommend _disabling_ the cron job in this PR. Once it is on `main`, test with worflow dispatch from the actions tab to ensure it is working as expected and then separately, enable the cronjob once it has been tested good.  Fewer chances of breakage that way.

jaybosamiya-ms: Oh and separating as subcommands is reasonable fwiw

parno: > Also, workflow wise, do you expect the automatic commit to get pushed to main or something?

Yes, I’d like to keep the version numbers in the repo consistent with the most recent release.  We can have the run step do the git commit, but sorting out the permissions to push back to the repo seemed like it might be complicated, so the action seemed like a useful packaging of that complexity.

parno: Plus that script was one of the top search results, so it must be secure, right?

[Subsequent searches](https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action/58393457#58393457) suggest that since we're using `actions/checkout@v2`, then just running `git push` might suffice after all, after setting this bit:
```
permissions:
  contents: write
```
I'll experiment with that, once I've disabled the cron job, as you suggested.

Issue #1868, opened 2025-08-19
verus_syn fails to parse `!is`


rikosellic:
When running `tools/line_count` on a file that contains Verus’s `!is` syntax, the tool aborts with a parse error.
### Example:
```
98     pub open spec fn test(a:Option<int>) -> int 
99     recommends
100         a !is None,
101    {
102        a->Some_0
103    }
```
### Observed Output:
```
[tools\line_count\src\main.rs:1530:29] &e.span().start() = LineColumn {
    line: 100,
    column: 6,
}
[tools\line_count\src\main.rs:1530:29] &e.span().end() = LineColumn {
    line: 100,
    column: 7,
}
error: failed to parse file ......rs: : expected curly braces bytes(230303..230304)
```
The abort point is a call to `verus_syn::parse2`:
```
let macro_content: File =
                        verus_syn::parse2(source_toks.clone()).map_err(|e| {
                            dbg!(&e.span().start(), &e.span().end());
                            format!(
                                "failed to parse file {}: {} {:?}",
                                input_path.display(),
                                e,
                                e.span()
                            )
                        })?;
```
### Notes
I am a bit confused, because `verus_syn` already defines `Expr::IsNot` and `Expr::HasNot`. Is there any reason why `!is` is not accepted by `verus_syn::parse2` here?
Any hint on how to fix this would be very helpful.

tjhance:
I didn't check for sure, but I suspect the reason it's failing is that it isn't calling `rejoin_tokens` (https://github.com/verus-lang/verus/blob/e87878197e3eb8962394fef3672061613369db6d/source/builtin_macros/src/syntax.rs#L4971C15-L4971C28) which the `verus!` macro always calls before parsing.

rikosellic:
@tjhance Thank you for your insight! I see that `rejoin_tokens` is a proc-macro utility internal to `builtin_macros`. Do you think the right approach for `line_count` would be to implement a similar function locally (perhaps a simplified version just handling `!is` and `!has`)?

tjhance:
I think it would probably be possible to just move `rejoin_tokens` into verus_syn.

matthias-brun:
The problem with `!is` actually extends to compilation as well (#1573), so it'll be great to have a fix. Thanks for looking into this.

PR #1882, opened 2025-09-03
Move `rejoin_tokens` to `verus_syn`

rikosellic:
Move `rejoin_tokens` into `verus_syn` to address #1868 . Removed the use of `proc_macro::Span::source().file()` when determining adjacency,  since it is unavailable in `proc_macro2::Span` and the `file` method does not appear to serve the intended purpose.


tjhance:
~~You should be able to get the `proc_macro::Span` from the `proc_macro2::Span`: https://docs.rs/proc-macro2/1.0.91/proc_macro2/struct.Span.html#method.unwrap~~ EDIT: actually this doesn't work outside of a proc-macro, never mind

Can you explain why the `file` method doesn't serve its intended purpose?

rikosellic:
The `file` method is for display purposes, but I just learned that `local_file` returns the actual path; now it works and all tests pass on my machine.

tjhance:
Thanks for the fix!

PR #1890, opened 2025-09-11
Limits for interpreter depth

### LINKED ISSUE #1889: Infinitely-recursive ExpX (symptom: hang in the assert-by-compute interpreter)
Consider this code (minimized from a case that @jaylorch came across):
```rs
use vstd::prelude::*;
fn main() {}

verus! {

spec fn a(b: int) -> nat
    decreases b,
{
    a(1)
}

fn c(d: int) {
    assert(a(d) >= 0) by (compute)
}

} // verus!
```

This sends the interpreter into an infinite recursion.  Running `gdb --args verus foo.rs`, starting execution (`r`), stopping after a few seconds (via `^C`), switching to the interpreter thread (`info threads` + `thread THREADNUM`) shows a _massive_ backtrace.  The start and end of the backtrace looks like this:

Image

Image

### Version info

```console
$ verus --version
Verus
  Version: 0.2025.09.09.28f0dd1
  Profile: release
  Platform: linux_x86_64
  Toolchain: 1.88.0-x86_64-unknown-linux-gnu
```

jaybosamiya-ms:
This PR fixes #1889 by adding a depth control that is a multiplier of the rlimit (so now, in addition to the iteration count control, there is also a recursion depth control).  It also adds a debug-only `eprintln!` print whenever an assert-by-compute has taken more than 2 seconds.  This is mostly aimed at debugging for Verus developers, rather than for reporting to users of Verus, thus is gated to being a debug-only feature.

Note: the code in this PR was primarily generated by an LLM, that I then manually read and fixed up to be more reasonable and maintainable.

### DIFF
--- a/source/rust_verify_test/tests/assert_by_compute.rs
+++ b/source/rust_verify_test/tests/assert_by_compute.rs
@@ -780,3 +780,43 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] depth_tracking_infinite_recursion verus_code! {
+        use vstd::prelude::*;
+
+        spec fn infinite_recursion(n: int) -> nat
+            decreases n,
+        {
+            infinite_recursion(1)  // Infinite recursion - should trigger depth limit
+        }
+
+        proof fn test_depth_limit() {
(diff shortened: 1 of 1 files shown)


parno:
This looks great -- thanks for putting it together!

PR #1895, opened 2025-09-16
Fix axioms for proper alignment of references, layout for slices and str

zero-to-nat:

### DIFF
--- a/source/vstd/layout.rs
+++ b/source/vstd/layout.rs
@@ -1,7 +1,7 @@
 #![allow(unused_imports)]
 
-use super::prelude::*;
 use super::group_vstd_default;
+use super::prelude::*;
 
 verus! {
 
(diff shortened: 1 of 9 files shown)


source/vstd/layout.rs
```diff
@@ -33,9 +33,9 @@ pub open spec fn valid_layout(size: usize, align: usize) -> bool {
 #[cfg_attr(not(verus_verify_core), deprecated = "is_sized is now defunct; lemmas that require V to be sized should now use the trait bound `V: Sized` instead of is_sized<V>")]
 pub uninterp spec fn is_sized<V: ?Sized>() -> bool;
 
-pub uninterp spec fn size_of<V>() -> nat;
+pub uninterp spec fn size_of<V: ?Sized>() -> nat;
```
tjhance: we shouldn't be defining sizes for Unsized types

zero-to-nat: you're right, this wouldn't make sense. thanks!

PR #1896, opened 2025-09-21
add sequence prophecies

zeldovich:
Sequence prophecy variables predict a sequence of values of type T. Their design comes from the same "Future is ours" paper that the current prophecy variable machinery was based on, and comes with the same restrictions on the type T of the sequence elements (Structural).

This commit also implements and proves the existing single-value prophecy variables in terms of sequence prophecy variables, rather than keeping them as a separate assumption/axiom.

### DIFF
--- a/source/vstd/proph.rs
+++ b/source/vstd/proph.rs
@@ -26,28 +26,68 @@ verus! {
 /// `new()` and `resolve()` are exec-mode, there is no ambiguity about which `new()`
 /// call corresponds to a particular `resolve()` value.
 pub struct Prophecy<T> {
-    v: Ghost<T>,
+    ps: ProphecySeq<T>,
 }
 
 impl<T> Prophecy<T> where T: Structural {
     /// The prophecized value.
     pub closed spec fn view(self) -> T {
-        self.v@
+        if self.ps@.len() > 0 {
+            self.ps@[0]
+        } else {
+            arbitrary()
+        }
     }
 
     /// Allocate a new prophecy variable.
     #[inline(always)]
     pub exec fn new() -> (result: Self) {
-        Prophecy::<T> { v: Ghost(arbitrary()) }
+        Prophecy::<T> { ps: ProphecySeq::<T>::new() }
     }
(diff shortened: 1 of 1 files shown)


tjhance (commented):
As I learned recently, prophecies-that-depend-on-prophecies can break the meaning of a `decreases` clause. e.g., using the fact that length of the prophecized sequence is a function of the length of the next prophecized sequence:

```rust
fn loop_forever() {
    let mut ps = ProphecySeq::<bool>::new();
    loop
        decreases ps.view().len()
    {
        ps.resolve(&false);
    }
}
```

i.e., Verus is able to prove the decreases-measure decreases, but the loop is clearly infinite.

I recently added a restriction so that values marked `prophetic` are no longer allowed in decreases-measures. So if `PropheticSeq::view` were marked `prophetic` then the above code would correctly be disallowed.

I know you were hoping to avoid using the `prophetic` attribute, but I don't see a way to add this soundly without it, without also precluding the possibility of real termination checking.

zeldovich:
What an interesting example -- I did not appreciate this subtle interaction between prophecy variables and liveness reasoning (since I've been thinking largely about safety proofs in the context of Verus).  Thanks!

Marking `ProphecySeq::view()` as `#[verifier::prophetic]` makes sense.  In that case, we'd want to keep `Prophecy::view()` as-is, and it would remain a separate axiom not provable on top of `ProphecySeq::view()`.

Another possible fix would be to represent `ProphecySeq::<T>::view()` as a (possibly infinite) map, say, `Map<nat, T>`.  Operating on this kind of `view()` might be more tedious, though (and it seems like it has to be less expressive than the `Seq<T>` version, in some way, in order to avoid the same liveness problem).

I'll play around with the target use case I had in mind (predicting the timestamps for all future transactions in a distributed system) and see which of these tweaks makes it work well (if any).

zeldovich:
Closing this PR since there's now more general support for `verifier::prophetic` prophecy variables that can support various kinds of sequences.

PR #1900, opened 2025-09-24
enforce alignment for PointsTo

tjhance:
Currently, PointsTo allows pointers to be null for ZSTs. This is based in the fact that reading and writing is permitted for null ZST pointers. However, one of the qualities of a PointsTo is that it should allow **taking references** and as Natalie noted, references are required to be non-null.

With this PR, we now enforce the non-null requirement to `PointsTo`.

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -94,7 +94,44 @@ pub ghost struct PtrData<T: core::marker::PointeeSized> {
     pub metadata: Metadata<T>,
 }
 
-/// Permission to access possibly-initialized, _typed_ memory.
+/**
+Permission to access possibly-initialized, _typed_ memory.
+
+The associated pointer ([`points_to.ptr()`](PointsTo::ptr)) is always a valid pointer for constructing
+a reference to the underlying data. That means it's always aligned to its type
+([`is_aligned`](PointsTo::is_nonnull)) and is non-null ([`is_nonnull`](PointsTo::is_nonnull)).
+
+### Notes
+
+The invariants on a `PointsTo` are a little more restrictive than is necessary for all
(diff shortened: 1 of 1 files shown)


source/vstd/raw_ptr.rs
```diff
@@ -705,10 +734,15 @@ impl PointsToRaw {
...
     /// that the domain of the `PointsToRaw` permission matches the size of `V`.
     ///
-    /// In combination with PointsToRaw::empty(),
-    /// this lets us create a PointsTo for a ZST for _any_ aligned pointer (any address and provenance, even null).
+    /// In combination with [`PointsToRaw::empty()`],
+    /// this lets us create a PointsTo for a ZST for _any_ non-null aligned pointer.
+    ///
+    /// To call this, it is necessary for the address to be non-null. This can be proved either
+    /// by showing `start != 0` or by showing the size of the type is non-zero (in which case
+    /// the non-null-ness follows from the existence of the `PointsToRaw`).
```
elanortang, line 822: 1. How does the non-null-ness follow from the existence of the `PointsToRaw`? What `PointsToRaw` axioms enforce this? 

2. Does this mean that if the size of the type is non-zero, there must be a corresponding `PointsToRaw`?

elanortang:
thanks! other than the questions I had, looks good to me!

source/vstd/raw_ptr.rs
tjhance: If the size is non-zero, then the PointsToRaw is non-empty, which means start can't be null (since you can't get a memory permission for a null address)

tjhance:
whoops, I forgot about this one. This is a legit soundness issue, I shouldn't have sat on it forbso long.

PR #1902, opened 2025-09-26
std_specs: add tracked methods for unwrap/expect on options and results

bsdinis:
These methods allow for having `tracked Result`s and `tracked Option`s.


tjhance (changes requested):
I don't think we need the `tracked_is_ok` functions, we already have `is_ok` and `is_err`

bsdinis:
> I don't think we need the `tracked_is_ok` functions, we already have `is_ok` and `is_err`

This is a good point and I updated it (note that the `Option` equivalent existed).
One annoying thing is that in the context of `OptionAdditionalFns` and `ResultAdditionalFns` function specs we need to use the deprecated `is_Some`, `is_None`, `is_Ok`, and `is_Err` (at that point we are talking about any generic trait implementer, not just `Option` and `Result`)

bsdinis:
I think the failure here is spurious?

bsdinis:
@tjhance how is this looking?

tjhance (approved):
Sorry for the long delay. I've been swamped and am just now getting around to reviews.

Anyway, this looks good to me.

bsdinis:
No worries! Thank you!

PR #1909, opened 2025-10-07
Extend EndianNat to convert from nat

zero-to-nat:

### DIFF
--- a/source/vstd/endian.rs
+++ b/source/vstd/endian.rs
@@ -194,14 +194,20 @@ impl<B: Base> EndianNat<B> {
         EndianNat { endian: self.endian, digits: self.digits.push(n), phantom: self.phantom }
     }
 
-    pub open spec fn push_least(self, n: int) -> Self {
+    pub open spec fn push_least(self, n: int) -> Self
+        recommends
+            n < B::base(),
+    {
         match self.endian {
             Endian::Little => self.push_first(n),
             Endian::Big => self.push_last(n),
         }
     }
 
-    pub open spec fn push_most(self, n: int) -> Self {
+    pub open spec fn push_most(self, n: int) -> Self
+        recommends
+            n < B::base(),
+    {
         match self.endian {
             Endian::Little => self.push_last(n),
             Endian::Big => self.push_first(n),


source/vstd/endian.rs
```diff
@@ -173,6 +173,35 @@ impl<B: Base> EndianNat<B> {
...
 
+    pub open spec fn append(self, other: Self) -> Self
+        recommends
+            self.endian == other.endian,
+    {
+        EndianNat::new(self.endian, self.digits + other.digits)
+    }
+
+    pub open spec fn push_first(self, n: int) -> Self {
+        EndianNat { endian: self.endian, digits: seq![n].add(self.digits), phantom: self.phantom }
```
elanortang, line 187: for the `push` functions, maybe add a `recommends n < B::base()` clause, so that we maintain the invariant?

source/vstd/endian.rs
```diff
@@ -173,6 +173,35 @@ impl<B: Base> EndianNat<B> {
...
+    }
+
+    pub open spec fn push_last(self, n: int) -> Self {
+        EndianNat { endian: self.endian, digits: self.digits.add(seq![n]), phantom: self.phantom }
+    }
```
elanortang, line 195: to make the spec slightly simpler (unless there was a reason it didn't work), could say `self.digits.push(n)` instead?

source/vstd/endian.rs
```diff
@@ -552,21 +581,120 @@ impl<B: Base> EndianNat<B> {
...
+
+    /// Ensures that `to_nat` correctly inverts `from_nat_with_len`,
+    /// provided that the given value can be encoded in the given base using the given length.
+    pub broadcast proof fn from_nat_to_nat(n: nat, len: nat)
+        requires
+            n < pow(B::base() as int, len),
+        ensures
+            #[trigger] Self::to_nat(Self::from_nat_with_len(n, len)) == n,
+        decreases Self::from_nat_with_len(n, len).len(),
```
elanortang, line 688: was there a reason for using the wrapper `Self::to_nat` instead of `.to_nat()`?

zero-to-nat: nope no reason, fixed!

source/vstd/endian.rs
zero-to-nat: fixed, thank you!

source/vstd/endian.rs
zero-to-nat: thanks for catching this!

zero-to-nat:
> generally looks good! left a few minor comments, but that's all.
> 
> I'm a little puzzled by why the diff between `verify-rustlib` and this branch is still so big, even after I merged `main` into `verify-rustlib`. Do you happen to know why that would be the case?

I think I had to merge `verify-rustlib` into this branch to reflect that, now the amount of changes looks more sensible

source/vstd/endian.rs
elanortang: and same thing for `push_least` and `push_most`, since it seems like these will be the main interfaces for users who don't want to think about endianness?

elanortang:
ah I see, makes sense! yeah that seems more reasonable now.

source/vstd/endian.rs
zero-to-nat: ahhh yes thank you, I forgot those!

PR #1912, opened 2025-10-10
Add basic support for spec functions (and hence constants) to bit-vector mode

parno:
We've run into a number of situations where it would be convenient to use the bit-vector prover mode without having to manually inline constants or spec functions.  This PR adds that functionality by pre-processing the asserted expression with the interpreter before handing the remaining expression off to the bit-vector mode.

### DIFF
--- a/source/vir/src/sst_elaborate.rs
+++ b/source/vir/src/sst_elaborate.rs
@@ -328,22 +328,17 @@ pub(crate) fn elaborate_function_bv<'a>(
     function: &mut FunctionSst,
 ) -> Result<(), VirErr> {
     if function.x.attrs.bit_vector {
-        if function.x.exec_proof_check.is_some() {
+        if let Some(exec_proof_check_arc) = &mut Arc::make_mut(function).x.exec_proof_check {
+            let exec_proof_check_mut = Arc::make_mut(exec_proof_check_arc);
             // Expand reqs and ens_exps using the interpreter
-            let exec_proof_check = function.x.exec_proof_check.as_ref().unwrap();
-            let reqs = expand(ctx, &fun_ssts, exec_proof_check.reqs.to_vec())?;
+            let reqs = expand(ctx, &fun_ssts, exec_proof_check_mut.reqs.to_vec())?;
             let ens_exps =
-                expand(ctx, &fun_ssts, exec_proof_check.post_condition.ens_exps.to_vec())?;
+                expand(ctx, &fun_ssts, exec_proof_check_mut.post_condition.ens_exps.to_vec())?;
 
-            // Reassemble the exec_proof_check with the expanded expressions
-            let mut new_exec_proof_check = function.x.exec_proof_check.clone();
-            let new_exec_proof_check = Arc::make_mut(new_exec_proof_check.as_mut().unwrap());
-            new_exec_proof_check.reqs = Arc::new(reqs);
-            let new_post = Arc::make_mut(&mut new_exec_proof_check.post_condition);
-            new_post.ens_exps = Arc::new(ens_exps);
-            new_exec_proof_check.post_condition = Arc::new(new_post.clone());
-            Arc::make_mut(function).x.exec_proof_check =
-                Some(Arc::new(new_exec_proof_check.clone()));
+            // Update the exec_proof_check fields directly
+            exec_proof_check_mut.reqs = Arc::new(reqs);
+            let post = Arc::make_mut(&mut exec_proof_check_mut.post_condition);
+            post.ens_exps = Arc::new(ens_exps);
         }
     }
     Ok(())


parno:
> 1. A lot of the existing bit_vector test cases are probably going to be trivialized by this and won't properly exercise the bit_vector code the way they're meant to. Can we make the interpreter step optional (even if the optionality is only exposed as some internal test-only mode)?

Yes, if we stick with the interpreter, then I can certainly do that.  However, once I was mostly done with this PR, I began to wonder if using the interpreter was the right call.  It has the side benefit that it will probably solve a fair number of assertions without hitting Z3 (and even failing that, in some cases produce simpler queries to Z3), which should lead to better and stabler automation.  OTOH, the interpreter is a pretty large, complex hammer.  Perhaps it would be better to have a custom elaborator pass that just does function inlining and nothing else.  Of course, that would put us up to (at least) four different inlining mechanisms (the new one, the interpreter, the code that implements `#[verifier::inline]`, and the code in expand-errors).

Thoughts either way?

> 2. Can you update the guide page on bit vectors to no longer recommend macros?

Indeed.  I held off b/c I wasn't sure if this was going to work :)

parno:
@Chris-Hawblitzel Do you have any preferences w.r.t. what approach we take to inlining here?

parno:
Given our discussion on Monday, I moved forward with the interpreter-based simplification.  I added a command-line option to disable the interpreter in bv proofs, and then rejiggered the bv tests to run both with and without simplification.  I also updated the docs to remove the text suggesting using macros instead of spec functions.

source/vir/src/sst_elaborate.rs
```diff
@@ -214,6 +247,65 @@ impl<'a, 'b, D: Diagnostics> Visitor<Rewrite, VirErr, NoScoper> for ElaborateVis
...
+                self.ctx.global.arch,
+                crate::ast::ComputeMode::Z3,
+                &mut self.ctx.global.interpreter_log.lock().unwrap(),
+            )
+        })
+    }
+}
+
+impl<'a> Visitor<Rewrite, VirErr, NoScoper> for ElaborateVisitorBv<'a> {
+    // This is the same as visit_func_check in sst_visitor.rs, except for the call to self.expand(..)
```
Chris-Hawblitzel: Do you need to override visit_func_check and visit_postcondition?  Would overriding just visit_exp be enough?

Chris-Hawblitzel (approved):

source/vir/src/sst_elaborate.rs
parno: Good question.  Based on one of my earlier attempts, I think overriding just `visit_exp` might do the wrong thing.  In particular, earlier I also overrode `FuncDeclSst`, so that I could expand function calls in its pre-/post-conditions.  However, when I did that, we lost information in the emitted assumption for the postcondition.  For example, instead of learning:
```
forall|a: u32| #[trigger] (a & 0) == 0
```
we woud learn:
```
forall|a: u32| #[trigger] 0 == 0
```
because the interpreter simplified the inner expression.

Given that my override of `FuncDeclSst` did get called and had this effect, I'm guessing that if we universally override `visit_exp`, then I think we'd see the same effect.  Does that seem accurate?

arsene1995:
Eagerly waiting ...
This feature would help LLM a LOT!

source/vir/src/sst_elaborate.rs
Chris-Hawblitzel: Yes, that makes sense.  In fact, if you're just editing the pre- and post-conditions, maybe it would be less boilerplate to not use the visitor at all, and instead just directly modify the individual pre- and post-condition fields of function.x.exec_proof_check.  You can use clone or Arc::make_mut to edit individual fields inside an Arc without having to rebuild the whole data structure and mention all the other fields.

parno: Ah, interesting.  I battled with the borrow checker until I came up with something reasonably slim (and that still passes the regression tests).

parno:
Given the discussion at today's Verus meeting and the green CI, I'm going to merge this in.

arsene1995:
Hi, 

I tried the latest version of Verus with this patch. 
The example in ```examples/guide/nonlinear_bitvec.rs``` works; however, bit_vector prover still fails on more complicated examples, such as this [code](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=6576f062b745584bdfaec7419907c10c)

(I understand that the playground verus is not updated yet. I am only using it to put this example.)

This example is extracted from nrkernel. If you replace ```MASK_DIR_ADDR``` on Line 82 with mask_dir_addr, the proof would go through. However, if we replace any normal variable with constant variable between line 82--86 the proof would fail.

The old verus would complain 
```
error: unsupported for bit-vector: expression conversion Call(Fun(Fun(Path(Some("main"), ["MASK_DIR_ADDR_SPEC"])), None), [], [])
  --> /playground/src/main.rs:82:23
   |
82 |         assert(addr & MASK_DIR_ADDR == addr) by (bit_vector)
   |                       ^^^^^^^^^^^^^
```

The latest version would complain
```
error: unsupported for bit-vector: expression conversion Call(Fun(Fun(Path(Some("main"), ["MAX_PHYADDR_WIDTH"])), None), [], [])
  --> tst_good.rs:10:62
   |
10 | pub spec const MASK_ADDR_SPEC: usize = bitmask_inc!(12usize, MAX_PHYADDR_WIDTH - 1);
   |                                                              ^^^^^^^^^^^^^^^^^
```

I am curious which aspect of this example failed the latest version of bit_vector.

parno:
It's because `MAX_PHYADDR_WIDTH` is marked `#[verifier(external_body)]`, so it shows up as an uninterpreted function, which isn't supported by the bit-vector solver at the moment.

arsene1995:
> It's because `MAX_PHYADDR_WIDTH` is marked `#[verifier(external_body)]`, so it shows up as an uninterpreted function, which isn't supported by the bit-vector solver at the moment.

Aha! I see!! Thanks!

PR #1914, opened 2025-10-15
std_specs: add map lemmas to infer contains_value after insert

### LINKED ISSUE #1908: map: `contains_value` only seems to trigger if correct `contains_key` was asserted, even for trivial cases
Consider the following piece of code:

```rust
use vstd::{map,prelude,set}::*;

proof fn works() {
  let m = map![10int => 20int];
  assert(m.contains_key(10));
  assert(m.contains_value(20));
}

proof fn fails_but_why() {
  let m = map![10int => 20int];
  assert(m.contains_value(20)); // FAILS
}
```

The first function checks but the second one does not. Is this a fundamental limitation given the way maps are modeled or is it a bug? 

Verus version: `Version: 0.2025.10.07.f3c426a.dirty`

[2025-10-07-09-53-25.zip](https://github.com/user-attachments/files/22746329/2025-10-07-09-53-25.zip)

bsdinis:
Fixes #1908

### DIFF
--- a/source/vstd/map_lib.rs
+++ b/source/vstd/map_lib.rs
@@ -413,6 +413,46 @@ impl<K, V> Map<K, V> {
         };
         assert(lhs == rhs);
     }
+
+    /// Inserting a value in a map means it contains the value that was inserted.
+    ///
+    /// ## Example
+    /// ```rust
+    /// proof fn example() {
+    ///     let a = map![1int => 2int];
+    ///     assert(a.contains_value(2));
+    /// }
+    /// ```
+    pub broadcast proof fn lemma_insert_contains_value(self, k: K, v: V)
+        ensures
+            #[trigger] self.insert(k, v).contains_value(v),
+    {
+        assert(self.insert(k, v).contains_key(k));
+    }
+
+    /// Inserting a value in a map where that key is already present ensures
+    /// The contains value sees (only) the updated value;
+    ///
+    /// ## Example
+    /// ```rust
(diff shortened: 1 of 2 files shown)


source/vstd/map_lib.rs
```diff
@@ -413,6 +413,45 @@ impl<K, V> Map<K, V> {
...
+    /// }
+    /// ```
+    pub broadcast proof fn lemma_insert_contains_value(self, k: K, v: V)
+        ensures
+            #[trigger] self.insert(k, v).contains_value(v),
+    {
+        assert(self.insert(k, v).contains_key(k));
+    }
+
+    /// Inserting a value in a map where that key already cont
```
parno: It looks like this comment got cut off.

source/vstd/map_lib.rs
```diff
@@ -413,6 +413,45 @@ impl<K, V> Map<K, V> {
...
+    ///
+    /// ## Example
+    /// ```rust
+    /// proof fn example() {
+    ///     let a = map![1int => 2int, 1int => 3int];
```
parno: Can we change this example to `let a = map![1int => 2int].insert(1int, 3int)`?  I think that makes the intended semantics clearer (compared to expecting the reader to understand that the `map!` macro desugars into a series of insertions).

bsdinis:
Done, will wait until EOD to see if someonehas other feelings about this PR and then ship it

PR #1921, opened 2025-10-21
Add encodings and type representations for transmutes

zero-to-nat:
This PR adds encodings for `bool`, `()`, unsigned integers, signed integers, and raw pointers. This also defines helpers for deriving encodings on structs/enums which are annotated with `#[repr(transparent)]` and `#[repr(<primitive>)]`.

In `vstd/type_representation.rs`:
- Definitions for `AbstractByte` (moved from transmute.rs)
- Uninterp spec fns `abs_encode`, `abs_decode`, `can_be_encoded` (moved from transmute.rs)
- `AbstractByteRepresentation`, `AbstractByteRepresentationUnsized` traits are defined to provide concrete specifications for `abs_encode` and `abs_decode`
- `AbstractByteRepresentation` impls for `bool`, `u8...usize`, `i8...isize`
- `AbstractByteEncoding<T>` trait is defined to contain reusable definitions for encodings
- `AbstractByteRepresentation` impls for `*const T` and `*mut T` from the single `AbstractByteEncoding` for raw pointers. This encoding is conditional on the pointer metadata having a suitable encoding -- see comments for explanation.
- `PrimitiveRepresentation`, `TransparentRepresentation`, `ScalarRangeRepresentation` traits are defined to aid in definiting `AbstractByteEncoding`s for corresponding structs/enums

In `vstd/transmute.rs`:
- Standardized pre and post conditions for transmute which use `abs_encode`, `abs_decode`, `can_be_encoded`
- Test cases to validate encoding specs:
    - `transmute_{type}_refl_unique` for types `bool`, `()`, unsigned ints, signed ints, and raw pointers to sized types
    - `transmute_{type}_refl` for raw pointers to unsized types
    - `transmute_usize_mut_ptr`

In `vstd/layout.rs`:
- Helper lemmas/axioms

Notes:
- I tried to structure things to allow for a decent amount of code reuse, but this was only based on my limited exposure to the handful of types in `std` that we need to define encodings for. If you have any suggestions or further insights, please let me know!

### DIFF
--- a/source/rust_verify_test/tests/opaque_types.rs
+++ b/source/rust_verify_test/tests/opaque_types.rs
@@ -211,7 +211,7 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
-    #[test] test_opaque_type_assume_spec_ok verus_code! {
+    #[ignore] #[test] test_opaque_type_assume_spec_ok verus_code! {
         use vstd::prelude::*;
         trait DummyTrait{
             type Output;


source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+                },
+        )
+    }
+
+    /// Returns the Provenance shared by all of the (initialized) bytes in the sequence.
+    /// If some bytes are uninitialized, or if some bytes do not share the same provenance (including having no provenance),
+    /// then `Provenance::null()` is returned.
+    pub open spec fn shared_provenance(bytes: Seq<Self>) -> Provenance
+        recommends
+            bytes.len() > 1,
```
elanortang: do you need this recommends? Bc it seems like you account for the 0 and 1 cases

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    // Required properties for a valid encoding.
+    /// Any encoding should match the size of this type.
+    proof fn encoding_size(v: Self, b: Seq<AbstractByte>)
+        ensures
+            Self::encode(v, b) ==> b.len() == size_of::<Self>(),
+            Self::decode(b, v) ==> b.len() == size_of::<Self>(),
+    ;
+
+    /// Every value should have at least one encoding.
+    proof fn encoding_exists(tracked v: &Self) -> (b: Seq<AbstractByte>)
```
elanortang, line 144: why does this take in a tracked reference?

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    axiom fn valid_encoding(v: Self, b: Seq<AbstractByte>);
+
+    axiom fn can_be_encoded();
+}
+
+// We utilize the EndianNat type to transform an unsigned int to its byte representation.
+/// Convert the given `EndianNat`` representation to `AbtractByte`s with the given provenance.
+pub open spec fn endian_to_bytes(endian: EndianNat<u8>, prov: Option<Provenance>) -> Seq<
+    AbstractByte,
+> {
```
elanortang: could add a `recommends endian.wf()` so there is no overflow in the integer cast

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    endian.digits.map_values(|e| AbstractByte::Init(e as u8, prov))
+}
+
+/// Convert the given `AbstractByte` representation to its `EndianNat` representation (provenance is ignored).
+pub open spec fn bytes_to_endian(bytes: Seq<AbstractByte>) -> EndianNat<u8> {
+    EndianNat::<u8>::new_default(bytes.map_values(|b: AbstractByte| b.byte() as int))
```
elanortang, line 304: could add a `recommends AbstractByte::all_init(bytes)`

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+        EndianNat::from_nat_len,
+        EndianNat::from_nat_with_len_wf,
+        EndianNat::from_nat_with_len_endianness,
+    ;
+
+    let endian = EndianNat::<u8>::from_nat_with_len(n, len);
+    let bytes = endian_to_bytes(endian, prov);
+    assert(endian.wf());
+    assert(endian.len() == len);
+    assert(endian.len() == bytes.len());
```
elanortang: does the proof still go through if you take out these asserts? (also seems like 224 and 225 can be combined?)

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+
+                    unsigned_int_max_bounds();
+                }
+
+                axiom fn valid_encoding(v: Self, b: Seq<AbstractByte>);
+
+                axiom fn can_be_encoded();
+            }
+
+            pub broadcast proof fn $lemma_name(v: $int, bytes: Seq<AbstractByte>)
```
elanortang, line 421: add a comment saying what this functions purpose is?

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+            EndianNat::<u8>::from_nat_with_len(v@.addr as nat, size_of::<usize>()),
+            Some(v@.provenance),
+        );
+        let suffix = Seq::<AbstractByte>::new(
+            (size_of::<*mut T>() - size_of::<usize>()) as nat,
+            |i| AbstractByte::Uninit,
+        );
+        let b = prefix.add(suffix);
+        assert(b.subrange(0, size_of::<usize>() as int) == prefix);
+        assert(b.len() == size_of::<*mut T>());
```
elanortang: are these asserts needed for the proof?

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    (i64);
+    (i128);
+    (isize);
+}
+
+/// This trait defines an `AbstractByte` encoding over the given type `T`.
+/// This definition enables reuse of the same underlying representation for `AbstractEncoding` across several concrete types.
+/// For example, enums can use the same representation as a primitive type through the use of the `#[repr]` attribute.
+/// See: https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr
+pub trait TypeRepresentation<T> {
```
elanortang: I think it would be good to explain in more detail how this trait relates to the `AbstractEncoding` trait - we implement this trait over a placeholder struct as a possible implementation of `AbstractEncoding` (since we can only implement `AbstractEncoding` once over a given type, and then make it concrete for a particular type by implementing `AbstractEncoding` in terms of `TypeRepresentation` functions.

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    (const, const_ptr_sized_encode, const_ptr_unsized_encode);
+}
+// Composite types: The representations for these types are derived from the encoding on another type.
+
+
+/// This trait is used to define a `TypeRepresentation` for the given type using the `AbstractEncoding` for a primitive type.
+/// The corresponding `TypeRepresentation` is implemented on the struct `PrimitiveRepresentationEncoding<Primitive, Self>`.
+/// This models the type representation for fieldless enums with the `#[repr(<primitive>)]` annotation.
+/// [Primitive representations](https://doc.rust-lang.org/reference/type-layout.html#primitive-representation-of-field-less-enums)
+pub trait PrimitiveRepresentation<Primitive: AbstractEncoding + PrimitiveInt> where Self: Sized {
```
elanortang: A type implements this trait if its internal representation is the same as some primitive type (such as `AlignmentEnum`).

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+    proof fn encoding_invertible(v: T, b: Seq<AbstractByte>) {
+        Primitive::encoding_invertible(T::to_primitive(v), b);
+    }
+}
+
+/// This trait is used to define a `TypeRepresentation` for the given type using the `AbstractEncoding` for the only nonzero-sized field on that type.
+/// The corresponding `TypeRepresentation` is implemented on the struct `TransparentRepresentationEncoding<Primitive, Self>`.
+/// This models the type representation for structs and enums with the `#[repr(transparent)]` annotation.
+/// [Transparent representations](https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation)
+pub trait TransparentRepresentation<Inner: AbstractEncoding> where Self: Sized {
```
elanortang: A type implements this trait if ... (bring over some of the documentation from the link)

source/vstd/type_representation.rs
```diff
@@ -0,0 +1,849 @@
...
+pub uninterp spec fn can_be_encoded<T: ?Sized>() -> bool;
+
+/// Can 'value' be encoded to the given bytes?
+pub uninterp spec fn abs_encode<T: ?Sized>(value: &T, bytes: Seq<AbstractByte>) -> bool;
+
+/// Can the given bytes always be decoded to the given value?
+pub uninterp spec fn abs_decode<T: ?Sized>(bytes: Seq<AbstractByte>, value: &T) -> bool;
+
+/// This trait is defines the given type's `AbstractByte` encoding. This definition is for `Sized` types only: see `AbstractEncodingUnsized` for `?Sized` types.
+pub trait AbstractEncoding where Self: Sized {
```
elanortang: typo - "trait is defines"

elanortang (commented):
overall great job! I left some comments

source/vstd/type_representation.rs
zero-to-nat: thanks! you are right, it is unnecessary

source/vstd/type_representation.rs
zero-to-nat: Thank you, this turned out to not be necessary! This was a remnant from when the encoding itself could impose range restrictions on niche types, and so we required a tracked argument in order to invoke the type invariant and prove that the value that we wanted to encode always fell within the valid range. However, we found that this restriction on the encoding is not needed, because the type invariant itself is enough to forbid the "bad transmutes" (e.g., NonZero -> 0, 0 -> NonZero)

source/vstd/type_representation.rs
zero-to-nat: One assert was indeed not necessary; the remaining two are needed for triggers on the broadcast lemmas used here.

source/vstd/type_representation.rs
zero-to-nat: The subrange one is necessary: it seems that Verus needs it to trigger facts about subrange which it does not know from `b = prefix.add(suffix)`. The len assertion is not necessary.

source/vstd/type_representation.rs
zero-to-nat: I added some more comments, let me know if you think it is still unclear

elanortang: that helps, thanks!

source/vstd/type_representation.rs
elanortang: looks good

PR #1927, opened 2025-10-26
fix: skip consuming tokens in for-loop for mod::method in verus_syn

### LINKED ISSUE #1926: `line_count` error caused by `verus_syn` for `for` loop
I'm using `line_count` to count the verified LoC. However, when counting LoC for file containing `for` loop with `mod::method`, it fails due to the `verus_syn` error.

## Minimal PoC:
```rust
pub fn ok_to_parse() {
    for paddr in Self::read(va) {
        ;
    }
}

pub fn fail_to_parse() {
    for paddr in largest_pages::read(va) {
        ;
    }
}
```

## Reproduce
In `source/tools/line_count` run
`$ cargo run -- --one-file test.rs`

## Error Log
```
Running `/root/verus/source/target/debug/line_count --one-file /root/test.rs`
[tools/line_count/src/main.rs:1439:9] &e.span().start() = LineColumn {
    line: 8,
    column: 31,
}
[tools/line_count/src/main.rs:1439:9] &e.span().end() = LineColumn {
    line: 8,
    column: 32,
}
error: failed to parse file /root/test.rs: expected an expression
```

Marsman1996:
Fix #1926 

~~I fork the `input` here to get `expr_name` without consuming tokens, since original syn does not get `expr_name` here.~~

Since the code logic is designed for `iter: ` in the for-loop but mistakely deal with `mod::`, I add a judge for the second `:`: if there is `:` in the third token, then we will skip the token consume.

### DIFF
--- a/dependencies/syn/src/expr.rs
+++ b/dependencies/syn/src/expr.rs
@@ -2515,11 +2515,12 @@ pub(crate) mod parsing {
             let pat = Pat::parse_multi_with_leading_vert(input)?;
 
             let in_token: Token![in] = input.parse()?;
-            let expr_name = if input.peek2(Token![:]) && input.peek(Ident) {
-                Some(Box::new((input.parse()?, input.parse()?)))
-            } else {
-                None
-            };
+            let expr_name =
+                if input.peek2(Token![:]) && !input.peek3(Token![:]) && input.peek(Ident) {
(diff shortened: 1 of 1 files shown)


Marsman1996:
Well, now I understand why this token has to be consumed...

PR #1932, opened 2025-10-30
fix: set `is_impl_fn` for `ImplItemFn`

### LINKED ISSUE #1916: `#[verus_spec]` macro errors within `impl` blocks.
# Description

The following code does not compile

```rust

pub struct Foo;

impl Foo {
    #[verus_spec(ret =>
        with
            Tracked(c): Tracked<&mut DekoPointsTo<()>>
        requires
            true,
        ensures
            ret == 1,
    )]
    fn test(a: u64, b: u64) -> u64 {
        1
    }
}
```

Moving the function body outside `Foo` makes Rust compiler happy, however. This should have something to do with the proc_macro bug. Also a sidenote, adding a trailing comma after arguments in `with` block causes macro expansion error.

Marsman1996:
Fix #1916

This issue is caused by Verus handles `ImplItemFn` together with `ItemFn` and always set `is_impl_fn` as false, which makes function inside a `impl` fails the verification with `verus_spec`.

### DIFF
--- a/source/builtin_macros/src/attr_block_trait.rs
+++ b/source/builtin_macros/src/attr_block_trait.rs
@@ -1,4 +1,41 @@
-use syn::{Attribute, Block, ExprForLoop, ExprLoop, ExprWhile, ImplItemFn, ItemFn, TraitItemFn};
+use quote::ToTokens;
+use syn::spanned::Spanned;
+use syn::{
+    Attribute, Block, ExprForLoop, ExprLoop, ExprWhile, ImplItemFn, ItemFn, Signature, TraitItemFn,
+};
+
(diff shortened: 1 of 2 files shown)


Marsman1996:
Well,
```rust
#[verus_spec(sum => 
     requires 
         x < 100, 
         y < 100, 
     ensures 
         sum < 200, 
)]
fn my_exec_fun(x: u32, y: u32) -> u32 
{ 
    x + y 
}
```
is parsed as `ImplItemFn` too...

Maybe we need to identify whether it is an `ImplItemFn` before the  `rewrite_verus_spec`

Marsman1996:
Updated:

To precisely identify between the `ImplItemFn` and `ItemFn`, I parse the file again and match the token string.

PR #1934, opened 2025-10-31
Fix `#[verus_spec]` for methods in `Impl` blocks

### LINKED ISSUE #1916: `#[verus_spec]` macro errors within `impl` blocks.

rikosellic:
Another attempt to fix #1916. In #1932, we found it impossible for `verus_spec` to determine whether a function is inside an Impl block (it is always set to `false` now, i.e., not in a block).  My solution is to apply `[verus_verify]` at the `Impl` block level, which will insert an internal `verus_impl_method_marker` into each method, and `verus_spec` can use to identify whether the function belongs to an `Impl` block.

However, providing a clear error message when verification fails is tricky. Some functions may still verify successfully even without `[verus_verify]` at the block level, for example, if `dual_spec` applies or if the specification does not include an `ensures` clause.

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -155,9 +155,15 @@ pub(crate) fn rewrite_verus_attribute(
     if let syn::Item::Impl(ref mut impl_block) = item {
         for impl_item in &mut impl_block.items {
             if let syn::ImplItem::Fn(ref mut method) = impl_item {
-                method.attrs.push(syn::parse_quote! {
-                    #[allow(unused, verus_impl_method_marker)]
+                let has_verus_spec = method.attrs.iter().any(|attr| {
+                    attr.path().get_ident().map_or(false, |ident| ident == "verus_spec")
                 });
+
+                if has_verus_spec {
+                    method.attrs.push(syn::parse_quote! {
+                        #[allow(unused, verus_impl_method_marker)]
+                    });
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


rikosellic:
#1932 now provides a different approach to fix the same issue. You may choose whichever you prefer.

hiroki-chen:
BTW, any interest in fixing #1936 along with this PR?

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -364,6 +375,35 @@ pub(crate) fn rewrite_verus_spec_on_fun_or_loop(
...
+                if let Some(ident) = attr.path().get_ident() {
+                    if ident == "allow" {
+                        if let syn::Meta::List(meta_list) = &attr.meta {
+                            return !meta_list
+                                .tokens
+                                .to_string()
+                                .contains("verus_impl_method_marker");
+                        }
+                    }
+                }
```
ziqiaozhou: Could you simplify the code?

For example,


```
let is_hidden_impl_marker = |attr: &syn::Attribute| {
    attr.path().get_ident().map_or(false, |ident| {
        ident == "allow" && matches!(&attr.meta, syn::Meta::List(meta_list)
            if meta_list.tokens.to_string().contains("verus_impl_method_marker"))
    })
};

// Check if the function has the impl method marker
let is_impl_fn = fun.attrs.iter().any(&is_hidden_impl_marker);

// Remove the marker attribute (internal use only)
fun.attrs.retain(|attr| !is_hidden_impl_marker(attr));
```

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -151,6 +151,17 @@ pub(crate) fn rewrite_verus_attribute(
...
+    // Special handling for impl blocks, add marker attribute to each method for `#[verus_spec]`.
+    if let syn::Item::Impl(ref mut impl_block) = item {
+        for impl_item in &mut impl_block.items {
+            if let syn::ImplItem::Fn(ref mut method) = impl_item {
+                method.attrs.push(syn::parse_quote! {
+                    #[allow(unused, verus_impl_method_marker)]
+                });
+            }
+        }
+    }
```
ziqiaozhou, line 169: What if we have
```
#[verus_verify]
impl X {
   fn unverified() {}
}
```

verus_impl_method_marker should be applied only when the function has a verus_spec attribute.

ziqiaozhou (changes requested):
Thanks for fixing the bug. 

Please add a test 
```
#[verus_verify]
impl X {
   fn unverified() {}
}
```

source/builtin_macros/src/attr_rewrite.rs
rikosellic: Thanks!

source/builtin_macros/src/attr_rewrite.rs
rikosellic: That's a good point, I've fixed it in the latest commit.

hiroki-chen:
I don't know what exactly is causing the trouble but I just found that upgrading to the latest `verus_builtin_macro` (working commit is `bf6ffb4a`) panics my execution code.

PR #1937, opened 2025-10-31
Upgrade verus to use rustc-1.91.0

dschoepe:
This makes Verus compatible with the latest stable Rust toolchain release (as of 10/31).

### DIFF
--- a/source/rust_verify_test/tests/traits.rs
+++ b/source/rust_verify_test/tests/traits.rs
@@ -2683,11 +2683,16 @@ test_verify_one_file! {
 }
 
 // This test should fail due to conflicting trait implementations, but currently
-// fails due to extern types being an experimental feature. Once extern type
-// types are stable and supported by verus, the expected error should be updated
-// here.
-test_verify_one_file! {
-    #[test] test_conflicting_sizedness_constraints verus_code! {
+// fails due to extern types not being supported by verus. Once extern type are
+// supported, the expected error should be updated here.
+test_verify_one_file_with_options! {
+    #[test] test_conflicting_sizedness_constraints ["no-auto-import-verus_builtin"] => code! {
+        #![cfg_attr(verus_keep_ghost, feature(extern_types))]
+        #![cfg_attr(verus_keep_ghost, feature(sized_hierarchy))]
+
+        use vstd::prelude::*;
+
+        verus! {
         extern "C" {
             type E;
         }
@@ -2709,8 +2714,8 @@ test_verify_one_file! {
             assert(<E as T>::f() == 4);
             assert(false);
         }
-
-    } => Err(err) => assert_rust_error_msgs(err, &["extern types are experimental", "type annotations needed", "type annotations needed"])
+        }
+    } => Err(err) => assert_vir_error_msg(err, "The verifier does not yet support the following Rust feature: foreign types")
 }
 
 test_verify_one_file! {


source/rust_verify_test/tests/adts_opaque.rs
```diff
@@ -71,6 +71,7 @@ test_verify_one_file! {
 const M1: &str = verus_code_str! {
     mod M1 {
         use verus_builtin::*;
+        use vstd::prelude::*;
```
parno: Do you know why this was needed?

source/vstd/raw_ptr.rs
```diff
@@ -178,9 +182,10 @@ impl<T: ?Sized> View for *const T {
 
 /// Compares the address and metadata of two pointers.
 ///
-/// Note that this does NOT compare provenance, which does not exist in the runtime
+/// Note that this DOES not compare provenance, which does not exist in the runtime
```
parno: Above, "DOES not" swapped to "does NOT".  Here it appears to have gone the other way.  Of the two forms, I think "does NOT" is probably better, since it emphasizes the negation.

parno (commented):
From a quick look through, this generally looks good to me, modulo the upcoming discussion of how to handle changes to vstd et al. in the context of the `cargo verus new` regression tests.

For the version number updates, did you use the [bump_crate_versions](https://github.com/verus-lang/verus/tree/main/source/tools/bump_crate_versions) tool?

parno:
Also, thanks very much for putting this together.  It looks like it was a lot of work!

source/vstd/raw_ptr.rs
dschoepe: This must have been by accident, I agree the `does NOT` formulation is clearer.

source/rust_verify_test/tests/adts_opaque.rs
dschoepe: I think this is a stale change from debugging an issue with this test that was ultimately unrelated to the imports. I'll revert this part.

dschoepe:
> For the version number updates, did you use the [bump_crate_versions](https://github.com/verus-lang/verus/tree/main/source/tools/bump_crate_versions) tool?

No, I wasn't aware of this tool, I'll try it with that and see what changes.

> It looks like it was a lot of work!

Thanks! Figuring out what happened to sizing was the worst of it. Most the changes to the rustc crate forks were handled by Travis' script, though it took some time to see which were genuinely new changes rather than things that were intentionally removed in the initial fork too.

dschoepe:
It looks like the bump_crate_versions script either produced incorrect versions or I used it wrong: https://github.com/verus-lang/verus/actions/runs/19058469777/job/54433456722

parno:
> It looks like the bump_crate_versions script either produced incorrect versions or I used it wrong: 

Ah, I see the issue.  The line counting tool specifies dependencies on `verus_syn` and `verus_prettyplease`.  For each one, it specifies both a version number and a path.  When the tool is compiled locally (including on CI), it's supposed to use the path, while if it were to be published on crates.io, it's supposed to use the version number.  The problem you ran into is that apparently cargo still expects the version number it finds in the path to match the version number in the dependency entry.  Since the line counting tool still had the old version number, that check failed.  

I guess this means we need to add a special case in the version bumper to reason about the line counting tools dependencies as well.

source/rust_verify_test/tests/cargo.rs
```diff
@@ -104,15 +105,47 @@ fn run_vanilla_cargo_for_dir(dir: &str) {
...
+// If vstd is modified as part of a change, the tests should use the local
+// version rather than what's on crates.io. This is not a great solution to
+// handle this though..
+fn adjust_version(mut project_path: PathBuf) {
+    project_path.push("Cargo.toml");
+    let mut toml_table = parse_toml_file(&project_path);
+    let Some(Value::Table(dependencies)) = toml_table.get_mut("dependencies") else {
+        panic!("no dependencies");
+    };
+    println!("toml table: {dependencies:#?}");
```
dschoepe: old debug print statement, will remove

parno:
> I guess this means we need to add a special case in the version bumper to reason about the line counting tools dependencies as well.

I added this to `main`.

dschoepe:
> > I guess this means we need to add a special case in the version bumper to reason about the line counting tools dependencies as well.
> 
> I added this to `main`.

I can confirm that this works now. Updated the PR to use the script instead of doing the version bumping manually

parno:
> I can confirm that this works now. Updated the PR to use the script instead of doing the version bumping manually

Cool, thanks for trying it out!

Chris-Hawblitzel:
I just added a PR to update syn: https://github.com/verus-lang/verus/pull/1964 .  With this, hopefully the edits to the syn directories in the 1.91.0 PR won't be necessary.

dschoepe:
> I just added a PR to update syn: #1964 . With this, hopefully the edits to the syn directories in the 1.91.0 PR won't be necessary.

Thanks, that did indeed fix the failing tests, but I kept the added lifetimes to avoid build-time warnings.

Chris-Hawblitzel:
> Thanks, that did indeed fix the failing tests, but I kept the added lifetimes to avoid build-time warnings.

I don't think it should emit the warnings anymore, because the new syn version added `#![allow(mismatched_lifetime_syntaxes)]` and `#![allow(clippy::elidable_lifetime_names)]`.  If we can avoid the added lifetimes, we'll have fewer diffs with syn, to minimize merge conflicts.

source/rust_verify/src/fn_call_to_vir.rs
```diff
@@ -170,15 +165,16 @@ pub(crate) fn fn_call_to_vir<'tcx>(
...
-    unsupported_err_unless!(
-        bctx.ctxt
-            .tcx
-            .impl_of_method(f)
-            .and_then(|method_def_id| bctx.ctxt.tcx.trait_id_of_impl(method_def_id))
-            .is_none(),
-        expr.span,
-        "call of trait impl"
-    );
+    // TODO(1.91.0): is this still needed? (impl_trait_id no longer returns Option)
```
Chris-Hawblitzel: I think this can safely be deleted.  It looks like it goes all the way back to December 2021 ( https://github.com/verus-lang/verus/commit/1ac48b55e0a80ccff4e16e4ee910225ab807e67e ) and was meant to check that we weren't calling a trait function.  But we've since added support for traits, and the check doesn't appear to be doing anything now.

source/vir/src/traits.rs
```diff
@@ -770,7 +773,12 @@ pub(crate) fn trait_bound_to_air(
     }
     match trait_id {
         TraitId::Path(path) => Some(ident_apply(&crate::def::trait_bound(path), &typ_exprs)),
-        TraitId::Sized => {
+        // We treat both MetaSized and PointeeSized as unsized in AIR for now.
+        // This may have to become more precise at some point in the future, but
+        // collapsing the two should be sound since they don't lead to different
+        // verification obligations.
```
Chris-Hawblitzel, line 779: I think this is fine for now but in the comment, an important reason it is sound is that trait_conflicts.rs catches conflicting trait implementations.  I think we should extend the comment and add a test case for this.  The only test I can think of, unfortunately, isn't ideal because it relies on a feature currently unsupported by Verus (foreign types):

```
extern "C" {
    type E;
}

trait T: core::marker::PointeeSized {
    spec fn f() -> int;
}

impl<A: core::marker::MetaSized> T for A {
    spec fn f() -> int { 3 }
}

impl T for E {
    spec fn f() -> int { 4 }
}

proof fn f() {
    assert(<E as T>::f() == 3);
    assert(<E as T>::f() == 4);
    assert(false);
}
```

If Verus were extended with support for foreign types, it would soundly give an error that there are "conflicting implementations of trait `T2_T` for type `E`".  If this error were ignored with `--no-lifetime`, it would unsoundly generate these conflicting axioms and incorrectly prove `assert(false)`:

```
;; Function-Axioms crate::impl&%0::A::f
(axiom (fuel_bool_default fuel%impl&%0.f.))
(axiom (=>
  (fuel_bool fuel%impl&%0.f.)
  (forall ((A&. Dcr) (A& Type)) (!
    (= (T.f.? A&. A&) (I 3))
    :pattern ((T.f.? A&. A&))
    :qid internal_T.f.?_definition
    :skolemid skolem_internal_T.f.?_definition
))))

;; Function-Axioms crate::E::f
(axiom (fuel_bool_default fuel%impl&%1.f.))
(axiom (=>
  (fuel_bool fuel%impl&%1.f.)
  (= (T.f.? $ TYPE%E.) (I 4))
))
```

I think we should add this test or something better, although I can't think of anything better.  For now, the test would have to match the error "The verifier does not yet support the following Rust feature: foreign types".

dschoepe: I tried adding the test, but it seems to fail prior to getting to Verus:

```
    #[test] test_conflicting_sizedness_constraints ["no-auto-import-verus_builtin"] => code! {
        #![feature(extern_types)]

        use vstd::prelude::*;

        verus! {

        extern "C" {
            type E;
        }

        trait T: core::marker::PointeeSized {
            spec fn f() -> int;
        }

        impl<A: core::marker::MetaSized> T for A {
            spec fn f() -> int { 3 }
        }

        impl T for E {
            spec fn f() -> int { 4 }
        }

        proof fn f() {
            assert(<E as T>::f() == 3);
            assert(<E as T>::f() == 4);
            assert(false);
        }
        }
    } => Err(err) => assert_rust_error_msgs(err, &["extern types are experimental"])
}
```

This fails with the following:

```
error[E0283]: type annotations needed: cannot satisfy `E: T`
  --> /Volumes/workplace/verus-191/source/target/debug/test_inputs/traits-f985bf3c5acba558-test_conflicting_sizedness_constraints/test.rs:19:12
   |
19 | impl T for E {
   |            ^
   |
note: multiple `impl`s satisfying `E: T` found
  --> /Volumes/workplace/verus-191/source/target/debug/test_inputs/traits-f985bf3c5acba558-test_conflicting_sizedness_constraints/test.rs:15:1
   |
15 | impl<A: core::marker::MetaSized> T for A {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
19 | impl T for E {
   | ^^^^^^^^^^^^


error[E0283]: type annotations needed: cannot satisfy `E: T`
  --> /Volumes/workplace/verus-191/source/target/debug/test_inputs/traits-f985bf3c5acba558-test_conflicting_sizedness_constraints/test.rs:24:13
   |
24 |     assert(<E as T>::f() == 3);
   |             ^
   |
note: multiple `impl`s satisfying `E: T` found
  --> /Volumes/workplace/verus-191/source/target/debug/test_inputs/traits-f985bf3c5acba558-test_conflicting_sizedness_constraints/test.rs:15:1
   |
15 | impl<A: core::marker::MetaSized> T for A {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
19 | impl T for E {
   | ^^^^^^^^^^^^


```

Leaving out the `feature` attribute, results in a Rust warning that extern types are experimental. Do you have an idea on how to fix that? We could just leave it in with the "extern types are experimental" error as the expected result, and we'll have to deal with this once they stabilize anyway. That's what's in the current version of the PR now.

Chris-Hawblitzel: > Leaving out the `feature` attribute, results in a Rust warning that extern types are experimental. Do you have an idea on how to fix that?

I think it should work if both `#![feature(sized_hierarchy)]` and `#![feature(extern_types)]` are present.

dschoepe: You're right, that fixed it. I tried moving `E` outside of the `verus!` block but that still doesn't work with `external_type_specification`.

Chris-Hawblitzel (approved):

PR #1939, opened 2025-11-02
fix: parse input in `verus_spec` ends with `,`

### LINKED ISSUE #1936: `#[verus_spec]` errors when trailing comma is added in `with` clause.
The following code does not parse:

```rust

#[verus_spec(res =>
    with
        Tracked(bar): Tracked<&mut ()>,
    requires
        true,
    ensures
        true,
)]
fn foo() {}
```

This generates

```text
error: expected `:`
   |
11 |         true,
   |         ^^^^
```

I mentioned this in #1916 but think another separate issue might be clearer for tracking. It'd be more ergonomic to allow trailing commas though (e.g. for code refactoring, proof-reading.).

Marsman1996:
Fix #1936

If another `FnArg` or `Expr` cannot be parsed, we treat the comma as trailing and stop.

### DIFF
--- a/dependencies/syn/src/verus.rs
+++ b/dependencies/syn/src/verus.rs
@@ -2448,7 +2448,13 @@ impl parse::Parse for WithSpecOnFn {
             if !input.peek(Token![,]) {
                 break;
             }
+            let fork = input.fork();
+            let _comma: Token![,] = fork.parse()?;
+            let has_next_input = fork.parse::<FnArg>().is_ok();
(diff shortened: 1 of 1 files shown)


ziqiaozhou (approved):
Thanks for improving verus_spec to treat a trailing comma as the end.

hiroki-chen:
Seems that this PR didn't fix trailing commas when we have `->` in `with`. See the above PR for more information.

PR #1940, opened 2025-11-03
Add specification for Vec::is_empty

dschoepe:


dschoepe:
The failing test seems to fail due to an extra output about the test running longer than expected [here](https://github.com/verus-lang/verus/actions/runs/19046223570/job/54394992376?pr=1940#step:7:1649).

PR #1946, opened 2025-11-04
ghost_map: add points to for single key submaps

bsdinis:
Having a singleton `GhostSubmap` is a fairly common pattern.

This commit makes using these singleton `GhostSubmap`s (a `PointsTo`) much more ergonomic (you can call `key` and `value` on it, as opposed with dealing with the underlying `Set`).

This PR introduces an API change on the `GhostMapAuth`.

`GhostMapAuth::insert` takes a `Map`, but `Map::insert` takes a key and a value.

This makes the signature of `insert` consistent.

`proof fn insert(tracked &mut self, k: K, v: V) -> (tracked result: PointsTo<K, V>)`

And the old `insert` is renamed to `insert_map`.

### DIFF
--- a/source/vstd/tokens/map.rs
+++ b/source/vstd/tokens/map.rs
@@ -1,10 +1,74 @@
+//! Maps that support ownership of keys
+//!
+//! - [`GhostMapAuth<K, V>`] represents authoritative ownership of the entire map;
+//! - [`GhostSubmap<K, V>`] represents client ownership of a submap;
+//! - [`GhostPointsTo<K, V>`] represents client ownership of a single key-value pair.
+//!
+//! Updating the authoritative `GhostMapAuth<K, V>` requires a `GhostSubmap<K,
+//! V>` containing the keys being updated.
+//!
+//! `GhostSubmap<K, V>`s can be combined or split.
+//! Whenever a `GhostSubmap<K, V>` can be used, we can instead use a `GhostPointsTo<K, V>` (but not vice-versa).
+//!
+//! ### Example
+//!
+//! ```
+//! fn example_use() {
+//!     let tracked (mut auth, mut sub) = GhostMapAuth::new(map![1u8 => 1u64, 2u8 => 2u64, 3u8 => 3u64]);
+//!
+//!     // Allocate some more keys in the auth map, receiving a new submap.
+//!     let tracked sub2 = auth.insert_map(map![4u8 => 4u64, 5u8 => 5u64]);
+//!     proof { sub.combine(sub2); }
+//!
+//!     // Allocate a single key in the auth map, receiving a points to
+//!     let tracked pt1 = auth.insert(6u8, 6u64);
+//!     proof { sub.combine_points_to(pt1); }
+//!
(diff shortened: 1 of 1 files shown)


zeldovich (commented):
Supporting `PointsTo` seems like a sensible special-case to support.  Maybe this should be named something other than `PointsTo`, though, to avoid confusion with the heap points-to in Verus?  Maybe `GhostPointsTo`?

tjhance:
Is it possible for the new PointsTo type to implement the [KeyValueToken trait](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/trait.KeyValueToken.html) and the [UniqueKeyValueToken trait](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/trait.UniqueKeyValueToken.html)? You could then define GhostSubmap via the [`MapToken`](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/struct.MapToken.html). I would ideally like to start making all our token libraries a bit more consistent & unified.

bsdinis:
> Is it possible for the new PointsTo type to implement the [KeyValueToken trait](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/trait.KeyValueToken.html) and the [UniqueKeyValueToken trait](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/trait.UniqueKeyValueToken.html)? You could then define GhostSubmap via the [`MapToken`](https://verus-lang.github.io/verus/verusdoc/vstd/tokens/struct.MapToken.html). I would ideally like to start making all our token libraries a bit more consistent & unified.

I think generally yes, but some things are less obvious:
`agree` in particular seems to have particularly different semantics here.

I'll look into it

bsdinis:
@tjhance looking into it a bit more, I don't think this is possible.

Fundamentally, it seems that VerusSync's token system doesn't have the notion of authoritative/split ownership as the `GhostMapAuth`/`GhostSubmap` do. In particular, the `agree` method on the `GhostSubmap` has a completely different signature:

```
proof fn agree(tracked &self: GhostSubmap<...>, auth: &GhostMapAuth<...>)
    requires self.id() == auth.id(),
    ensure self@ <= auth@
    ;
```

This is completely different from the Token agree signature (both at the type and semantic level).
```
proof fn agree(tracked &self, tracked other: &Self)
  requires 
    self.instance_id() == other.instance_id(), 
    self.key() == other.key(),
  ensures
    self.value() == other.value(),
```

In fact, this is impossible to prove for `GhostPointsTo` (except for vacuously true `self.agree(self)`). We can never have two `GhostPointsTo` owning the same key.


I wonder if instead the best thing to do here is to instead take the `GhostMapAuth` out of the `tokens` module and come up with a better home for it.

bsdinis:
@tjhance very gentle ping on this

tjhance (approved):
This looks fine to me.

I do think it should be possible to implement that trait (note that the agree lemma takes shared refs (&) as arguments) but I need to think longer about how I want these pieces fitting together in the end.

PR #1955, opened 2025-11-06
Fix attr impl item

ziqiaozhou:
Impl Item Fn is not properly rewritten by verus_spec attribute: see unit test at 38ea20ed8f53cfb1fc0dceb967599d90190db844

This is due to the use of builtin::constrain_type https://github.com/verus-lang/verus/pull/1799

verus_spec applies to items and an ItemFn could be a function inside or outside impl item and only verus_verify applied at ItemImpl will be able to tell verus_spec that the item is inside impl.

This PR adds a hidden verus_spec_in_impl, which is a version of verus_spec that believe the ItemFn is inside impl. 

When verus_verify sees an impl Item, it will change verus_spec to verus_spec_in_impl for all ImplItemFn inside it.


ziqiaozhou:
I just realized that there is another better fix from #1934 to resolve it.

PR #1963, opened 2025-11-08
Fix trailing commas in `with` for `#[verus_spec]`.

hiroki-chen:
Previous there was an issue #1936 for tracking this and #1939 made an initial attempt to fix it. However the fix is incomplete as we still have the following problem if `with` contains `Output` clauses which again make `syn` confused.

```rust
#[verus_spec(r =>
    with Tracked(aaa): Tracked<&mut ()>, // <- trailing comma
        -> Tracked(bar): Tracked<()>, // <- trailing comma
    requires true,
)]
fn bar() {
    ()
}
```

This is another attempt to fix this issue and one new test case has been added.

### DIFF
--- a/source/rust_verify_test/tests/syntax_attr.rs
+++ b/source/rust_verify_test/tests/syntax_attr.rs
@@ -999,7 +999,7 @@ test_verify_one_file! {
             requires
                 x < 100,
             ensures
-            )]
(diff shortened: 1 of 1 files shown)


hiroki-chen:
Closed with #1967

PR #1965, opened 2025-11-11
Into from

ziqiaozhou:
I thought IntoSpec and FromSpec were already included in vstd, but they weren’t, so I added the spec I used.

### DIFF
--- a/source/vstd/std_specs/convert.rs
+++ b/source/vstd/std_specs/convert.rs
@@ -1,0 +1,77 @@
+#![allow(unused_imports)]
+use super::super::prelude::*;
+
+use core::convert::{From, Into};
+
+verus! {
+
+#[verifier::external_trait_specification]
+#[verifier::external_trait_extension(FromSpec via FromSpecImpl)]
+pub trait ExFrom<T>: Sized {
+    type ExternalTraitSpecificationFor: core::convert::From<T>;
+
+    spec fn obeys_from_spec() -> bool;
+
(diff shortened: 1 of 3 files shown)


Chris-Hawblitzel:
This is a (minor) breaking change for Vest ( https://github.com/secure-foundations/vest.git , @y1ca1 ) and perhaps other projects.  (Vest was the only benchmark in veritas that broke.)  A patch for Vest is:

```
diff --git a/vest/src/errors.rs b/vest/src/errors.rs
index 27d37e9..a7b2898 100644
--- a/vest/src/errors.rs
+++ b/vest/src/errors.rs
@@ -33,6 +33,24 @@ pub enum Error {
     Serialize(SerializeError),
 }

+impl vstd::std_specs::convert::FromSpecImpl<ParseError> for Error {
+    open spec fn obeys_from_spec() -> bool {
+        true
+    }
+    open spec fn from_spec(e: ParseError) -> Self {
+        Error::Parse(e)
+    }
+}
+
+impl vstd::std_specs::convert::FromSpecImpl<SerializeError> for Error {
+    open spec fn obeys_from_spec() -> bool {
+        true
+    }
+    open spec fn from_spec(e: SerializeError) -> Self {
+        Error::Serialize(e)
+    }
+}
+
```

We should probably make an announcement on Zulip before merging.

Chris-Hawblitzel (approved):

PR #1966, opened 2025-11-11
Add RangeInclusive::contains and new

ziqiaozhou:
Fields in RangeInclusive are private and so defined a view for RangeInclusive.

### DIFF
--- a/source/vstd/std_specs/range.rs
+++ b/source/vstd/std_specs/range.rs
@@ -20,6 +20,42 @@ pub struct RangeInclusiveView<Idx> {
     pub exhausted: bool,
 }
 
+pub trait ContainsSpec<Idx, U> where Idx: PartialOrd<U>, U: ?Sized + PartialOrd<Idx> {
+    spec fn obeys_contains() -> bool;
+
+    spec fn contains_spec(&self, i: &U) -> bool;
+}
+
+impl<Idx, U> ContainsSpec<Idx, U> for RangeInclusive<Idx> where
+    Idx: PartialOrd<U>,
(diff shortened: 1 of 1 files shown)


source/vstd/std_specs/range.rs
```diff
@@ -47,6 +64,23 @@ pub assume_specification<Idx: PartialOrd<Idx>, U>[ Range::<Idx>::contains ](
             ==> ret == (r.start.is_le(i) && i.is_lt(&r.end)),
 ;
 
+pub assume_specification<Idx: PartialOrd<Idx>, U>[ RangeInclusive::<Idx>::contains ](
+    r: &RangeInclusive<Idx>,
+    i: &U,
+) -> (ret: bool) where Idx: PartialOrd<U>, U: ?Sized + PartialOrd<Idx>
+    ensures
+        (U::obeys_partial_cmp_spec() && <Idx as PartialOrdSpec<U>>::obeys_partial_cmp_spec())
+            ==> ret == (r@.start.is_le(i) && i.is_le(&r@.end)),
```
Chris-Hawblitzel: What about the case where `exhausted == true`, as in: https://doc.rust-lang.org/src/core/ops/range.rs.html#1088 ?

ziqiaozhou: 45ba2be737768c44a0265e9d3194677d97778552 should cover this. I defined contains_spec and obeys_contains so that I do not need to repeat the logic outside the spec function outside of the vstd.

Chris-Hawblitzel (approved):

PR #1967, opened 2025-11-11
Fix trailing commas in with for #[verus_spec].

ziqiaozhou:
This supersedes #1963.

@hiroki-chen, thanks for fixing the trailing comma issue. Based on your PR #1963, I simplifies the change and makes it more general to cover all possible spec keyword after with.  

BTW, I did not try it yet but will the latest rustfmt always inserts a trailing comma and cause the problem you mentioned at #1936


hiroki-chen:
AFAIK formatter won't format what's inside `#[verus_spec]`

Issue #1969, opened 2025-11-11
Possible bug with proof fns looking at implementation details


bsdinis:
I managed to include two lemmas that, depending on the contents of their proof make another proof fail. I think this qualifies as a bug.

## Context

I was working on adding support for a lemma on `Set`s that roughly has the following signature:

```rust
proof fn lemma_map_size_bound<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
     requires
         forall|a: A| x.contains(a) ==> y.contains(#[trigger] f(a)),
         forall|b: B| (#[trigger] y.contains(b)) ==> exists|a: A| x.contains(a) && f(a) == b,
         y == x.map(f),
         x.finite(),
     ensures
         y.finite(),
         y.len() <= x.len(),
{}
```

For which I added a complementary lemma on `Map` 
```rust
     pub proof fn lemma_image_size(self)
         requires
             self.dom().finite(),
         ensures
             self.values().finite(),
             self.values().len() <= self.dom().len(),
```

## The Bug

Depending on the proof implementation of these lemmas, another completely different `proof fn` ([`find_unique_minimal_ensures`](https://verus-lang.github.io/verus/verusdoc/vstd/set/struct.Set.html#method.find_unique_minimal_ensures)) fails to verify.

Note: neither lemma was added to a broadcast group.

For comparison, these two commits show when the [verification failure is introduced](https://github.com/bsdinis/verus/commit/3f30cf1008f12de5b77ece71b0f9248ad07b3377) and [how the contents of the proof change the verification result](https://github.com/bsdinis/verus/commit/336cdf887c370b7c86e1efb4274945abe9005e0b). The first commit fails verification on the aforementioned unrelated `proof fn` and the second commit passes verification.

WeituoDAI:
I meet similar problem sometimes. I am not sure about the reason, but adding `#[verifier::spinoff_prover]` might help.

parno:
[...]

PR #1970, opened 2025-11-11
Add lemmas regarding the length of `Set::map` for non-injective `f` and for `Map::values`

bsdinis:
This PR adds three lemmas `Set::lemma_map_size_bound`, `Map::lemma_values_len` and
`Map::lemma_injective_values_len`. It also weakens the preconditions on `Set::lemma_map_size` (that previously only worked on injective functions -- now it works as long as the function is injective _in the mapping domain_).

`Set::lemma_map_size_bound` allows for reasoning for `s.map(f)` when `f` is not injective in the mapping domain. In particular, it allows to state `s.map(f).len() <= s.len()` for _any_ `f`.

The `Map` lemmas allow to reason about the length of `m.values().len()` (in particular that `m.values().len() <= m.dom().len()`) and the fact that this reaches equality for the case if the map is injective.

This required a weakening of the precondition on `Set::lemma_map_size`.
Previously, this lemma required the mapping function to be injective everywhere, even
outside the mapping domain. Now, it does not, matching the Dafny
function this was based on. Note that:

1. Sometimes it is impossible for a function to be injective in the
   mapping domain (if the mapped to type has a lower cardinality);
2. It is not necessary -- having the weaker precondition changes nothing
   about the original proof.

### DIFF
--- a/source/vstd/set_lib.rs
+++ b/source/vstd/set_lib.rs
@@ -950,13 +950,11 @@ pub proof fn lemma_subset_equality<A>(x: Set<A>, y: Set<A>)
 
 /// If an injective function is applied to each element of a set to construct
 /// another set, the two sets have the same size.
-// the dafny original lemma reasons with partial function f
 pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
     requires
-        injective(f),
-        forall|a: A| x.contains(a) ==> y.contains(#[trigger] f(a)),
-        forall|b: B| (#[trigger] y.contains(b)) ==> exists|a: A| x.contains(a) && f(a) == b,
         x.finite(),
+        injective_on(f, x),
+        x.map(f) == y,
     ensures
         y.finite(),
         x.len() == y.len(),
@@ -970,6 +968,7 @@ pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
         }
     } else {
         let a = x.choose();
+        assert(x.remove(a).map(f) == y.remove(f(a)));
         lemma_map_size(x.remove(a), y.remove(f(a)), f);
         assert(y == y.remove(f(a)).insert(f(a)));
     }
--- a/source/vstd/map_lib.rs
+++ b/source/vstd/map_lib.rs
@@ -4,6 +4,8 @@ use super::map::{Map, assert_maps_equal, assert_maps_equal_internal};
 use super::pervasive::*;
 #[allow(unused_imports)]
 use super::prelude::*;
+#[allow(unused_imports)]
+use super::relations::*;
 use super::set::*;
 #[cfg(verus_keep_ghost)]
 use super::set_lib::*;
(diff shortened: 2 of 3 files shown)


bsdinis:
Note: This would be a good PR to add some notion of partial injectivity (or domain_injectivity) and to standardize naming of the lemmas

source/vstd/set_lib.rs
```diff
@@ -950,10 +950,10 @@ pub proof fn lemma_subset_equality<A>(x: Set<A>, y: Set<A>)
 
 /// If an injective function is applied to each element of a set to construct
 /// another set, the two sets have the same size.
-// the dafny original lemma reasons with partial function f
 pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
     requires
-        injective(f),
+        forall|a1: A, a2: A|
```
parno: This seems useful enough that I think it's worth factoring into a spec function, e.g.,
`spec fn injective_on<X,Y>(r: spec_fn(X) -> Y, s: Set<X>) -> bool` in `relations.rs`.

source/vstd/set_lib.rs
```diff
@@ -975,6 +975,80 @@ pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
...
+            assert(x.difference(pre).contains(pre_wit));
+        }
+
+        assert(x == x.difference(pre).union(pre));
+        assert(y == y.remove(f(xx)).insert(f(xx)));
+        lemma_map_size_bound(x.difference(pre), y.remove(f(xx)), f);
+    }
+}
+
+proof fn lemma_find_preimages<A, B>(x: Set<A>, y: Set<B>, img: B, f: spec_fn(A) -> B) -> (pre: Set<
```
parno: I don't see this used anywhere, and it doesn't have a `pub` or a doc comment.  Could you please either add those, or cut this if you didn't mean to include it?

source/vstd/set_lib.rs
```diff
@@ -975,6 +975,80 @@ pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
...
+    A,
+>)
+    requires
+        x.finite(),
+        forall|a: A| x.contains(a) ==> y.contains(#[trigger] f(a)),
+        forall|b: B| (#[trigger] y.contains(b)) ==> exists|a: A| x.contains(a) && f(a) == b,
+        exists|a: A| x.contains(a) && f(a) == img,
+    ensures
```
parno: From the lemma's name, I expected to see something saying `forall|a: A| pre.contains(a) ==> f(a) == img`.

source/vstd/set_lib.rs
```diff
@@ -975,6 +975,80 @@ pub proof fn lemma_map_size<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
     }
 }
 
+/// If any function is applied to each element of a set to construct
+/// another set, the constructed set's length is at most the original's
+pub proof fn lemma_map_size_bound<A, B>(x: Set<A>, y: Set<B>, f: spec_fn(A) -> B)
+    requires
+        forall|a: A| x.contains(a) ==> y.contains(#[trigger] f(a)),
```
parno: This pair of quantifiers shows up regularly, e.g., here, in `lemma_subset_equality` above, and `lemma_find_preimages` below.  It'd be nice to name this property.  One option would be to not take `y` as an argument, and instead just talk about `x.map(f)`.  If it's important to take `y`, then perhaps these two quantifiers could be replaced with `y == x.map(f)`.  

If that doesn't provide the triggering we want, then we could resort to something like `maps_to(f, x, y) -> bool` with the two quantifiers as the body.

parno (commented):
Thanks for the contribution!  I've added some comments/questions.

source/vstd/set_lib.rs
bsdinis: Oh, good catch. I had it separately but then folded into the main proof fn then

source/vstd/set_lib.rs
bsdinis: I agree, was just keeping in line with the original lemma

bsdinis: (will change)

bsdinis: Interestingly, making the precondition `x.map(f) == y` triggers the same bug as #1969

bsdinis: Okay, interestingly, both ways to do it trigger #1969

bsdinis: @parno I left the quantifiers as they were. I'm really confused about the bug (even just adding the spec fn `maps_to` without using was triggering the bug). I'm pretty confident it is not a bug in these proofs but a bug in Verus itself. Happy to leave a not here to the fact that there is a better way to do this and come back at a later time (when the bug is fixed) to cleanup

bsdinis: Figured out that the problem was not the preconditions themselves but the `maps_to`. Having the precondition be `x.map(f) == y` as you suggested works. Still puzzled by the bug

bsdinis: @parno this should be resolved now

parno (approved):
Looks good to me.  Thanks for cleaning up the spec functions.

Issue #1975, opened 2025-11-14
Bug: spinoff_prover breaks proof blocks in recursive functions


parno:
This code  verifies without complaint:
```
use vstd::prelude::*;

verus! {

spec fn count(n: nat) -> nat
    decreases n,
{
    proof {
        assert(true);
    }

    if n == 0 {
        0
    } else {
        count((n - 1) as nat)
    }
}

// #[verifier::spinoff_prover]
proof fn test() {
    assert(count(0) == 0);
}

}

```

However, if you uncomment the `spinoff_prover` attribute, we get an error about the proof block inside of `count`:
```
error: proof blocks inside spec code is currently supported only for recursion
  --> test.rs:8:11
   |
8  |       proof {
   |  ___________^
9  | |         assert(true);
10 | |     }
   | |_____^
```
It seems like something is going wrong in the code path for spinoff_prover w.r.t. identifying recursive functions.

PR #1979, opened 2025-11-14
Fix : spinoff_prover breaks the proof blocks in recursive fn

WeituoDAI:
Fix the bug noted in #1975, related to #1969. This is a temporary solution and may need to be modified when proof-in-spec gets more progress. I add the example from #1975 to the test set.

### DIFF
--- a/source/vir/src/ast_to_sst_func.rs
+++ b/source/vir/src/ast_to_sst_func.rs
@@ -229,49 +229,49 @@ fn func_body_to_sst(
     let proof_body_stm = check_state.finalize_stm(ctx, &proof_body_stm)?;
     check_state.finalize();
 
-    let termination_check =
-        if crate::recursion::fun_is_recursive(ctx, function) && verifying_owning_bucket {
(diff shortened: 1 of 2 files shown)


source/vir/src/ast_to_sst_func.rs
```diff
@@ -229,49 +229,49 @@ fn func_body_to_sst(
...
+            body: termination_stm,
+            local_decls: Arc::new(termination_decls),
+            local_decls_decreases_init: termination_inits,
+            statics: Arc::new(vec![]),
+            reqs: Arc::new(vec![]),
+            unwind: UnwindSst::NoUnwind,
         };
+        Some(termination_check)
+    } else {
+        if function.x.decrease.len() > 0 && !is_recursive {
```
WeituoDAI, line 265: In fact, I change nothing but this condition (The code is rearranged by Verusfmt)

parno (approved):
It took me a bit to understand the change, but it now that I do, it looks reasonable to me.  Thanks for the quick fix!

PR #1981, opened 2025-11-15
Some initial support for associated constants in traits

Chris-Hawblitzel:
This adds some limited initial support for associated constants in traits.  For example:

```
        trait T<A, B> {
            const C: usize;
            const S: &str;
            const E: usize;
        }

        impl<Z: U> T<u8, Z> for bool {
            const C: usize = 3;
            const S: &str = "ha";

            #[verifier::external_body]
            const E: usize = 4;
        }
```

For now, the only supported impls of associated consts are literals like "3" or external_body expressions.  More general support will need the appropriate unerased proxies.  Also, the only supported mode is dual-mode spec/exec, and there isn't yet any support for declaring ensures clauses on associated constants.

Up to now, we've treated constants and functions separately in the HIR -> VIR translation, and then merged them in ast_simplify.  For associated constants, it turned out to be less code duplication to just use the function translation during HIR -> VIR, rather than adding trait handling to the existing HIR -> VIR translation of constants.  This might be a sign that we should just merge the HIR -> VIR translation of functions and constants, although I've left the existing non-trait constant handling as it was for now.


jaylorch:
Thanks! This works for me, as long as the assigned constant is a literal and not something like `MyEnum::MyValue as u8`. (It would be nice if it could also support the latter some day.)

jaylorch:
I don't feel comfortable approving the PR since it touches a whole lot of the Verus code base I'm not familiar with. I've leave that to a more qualified reviewer. I assume I was only added as a reviewer to make sure it fit my needs, since I was the one who asked Chris for this feature.

source/builtin_macros/src/unerased_proxies.rs
```diff
@@ -94,9 +94,30 @@ impl crate::syntax::Visitor {
...
-    fn impl_item_needs_unerased_proxy(&self, impl_item: &ImplItem) -> bool {
+    fn impl_item_needs_unerased_proxy(&self, impl_item: &ImplItem, for_trait: bool) -> bool {
         match impl_item {
-            ImplItem::Const(impl_item_const) => !is_external(&impl_item_const.attrs),
+            ImplItem::Const(impl_item_const) => {
+                if for_trait {
+                    // For the moment, allow associated const impls only if they are external_body,
+                    // or are simple literals,
+                    // and don't generate unerased_proxy.
+                    // TODO: generate unerased_proxy for associated const impls
```
tjhance, line 104: I'm a little confused - how does this work at all without the unerased proxies?

Chris-Hawblitzel: It only works because the only supported associated constants in this pull request are simple enough that there's nothing that would have to go into an unerased proxy: no requires, no ensures, and no complex body (only literals or external_body).  Then the associated constants are translated through the existing HIR->VIR machinery for functions rather than for constants, since the HIR->VIR translation for functions supports traits.  I wanted to wait to deal with unerased proxies because those seem harder, so I limited what went into this initial pull request.

JakeGinesin:
Is this planning to be merged @parno @Chris-Hawblitzel?

PR #1988, opened 2025-11-17
Add verifier::assume(externals_available_without_declaration)

Chris-Hawblitzel:
This implements one of the configuration options for extended static checking ( https://github.com/verus-lang/verus/discussions/1987 ).  Specifically, when the user sets `#[verifier::assume(externals_available_without_declaration)]`, Verus code can call external functions without declaring the external functions to Verus.  Such external functions are treated as exec functions with empty specifications (i.e. `requires true, ensures true`).  Any use of such an undeclared external function is illegal if `--no-cheating` is specified.

Other external items, like consts, datatypes, and traits, are not covered by this pull request but could be added in future pull requests.


dschoepe:
Would it make sense to allow this annotation at the module or crate level as well to avoid having to repeat this for every function?

Chris-Hawblitzel:
> Would it make sense to allow this annotation at the module or crate level as well to avoid having to repeat this for every function?

Yes, this is allowed.  I'm not sure how to write an official test for it, but I've tried `#![verifier::assume(externals_available_without_declaration)]` on a crate and it applies across the whole crate.  Similarly, it applies to modules.  You can selectively opt in and out by using `[verifier::assume(externals_available_without_declaration)]` and `[verifier::deny(externals_available_without_declaration)]` at different levels.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1352,12 +1352,16 @@ pub(crate) fn expr_to_vir_with_adjustments<'tcx>(
             if let Some((fun, typ_args)) = f {
                 let autospec_usage =
                     if bctx.in_ghost { AutospecUsage::IfMarked } else { AutospecUsage::Final };
+                let call_target_attrs = vir::ast::CallTargetAttrs {
+                    autospec: autospec_usage,
+                    assume_external_allowed: false,
```
parno, line 1468: What’s the intuition for when you can directly set this to false?

source/rust_verify_test/tests/verifier_assume_allow.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        assume_specification[ f3 ](u: u8)
+            requires
+                u > 10,
+        ;
+
+        #[verifier::assume(externals_available_without_declaration)]
+        fn g0() {
+            f0();
+            f3(5); // FAILS
```
parno, line 28: Why does this fail?

parno (approved):
Looks reasonable to me.

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: In this case, we're calling some fixed vstd function (like `vstd::raw_ptr::cast_array_ptr_to_slice_ptr`) so it doesn't make sense to anticipate a case where the function isn't found

tjhance (approved):

source/rust_verify_test/tests/verifier_assume_allow.rs
tjhance: The `assume_specification` for `f3` requires u > 10.

parno: Ah, thanks, I thought it was for some more exotic reason.

source/rust_verify/src/rust_to_vir_expr.rs
parno: Okay, that makes sense. Thanks!

PR #1990, opened 2025-11-18
Allow `rustc_diagnostic_item` on `const` functions

### LINKED ISSUE #1989: Error with `rustc_diagnostic_item` on a `const fn`
The following code:
```
verus! {
    #[rustc_diagnostic_item = "Crabcakes"]
    pub const fn my_diagnostic_item() -> usize
    {
        0
    }
}
```
produces the following error:
```
error: duplicate diagnostic item in crate `test`: `Crabcakes`
    --> src\main.rs:1049:1
     |
1049 | pub const fn my_diagnostic_item() -> usize
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
note: the diagnostic item is first defined here
    --> src\main.rs:1049:1
     |
1049 | pub const fn my_diagnostic_item() -> usize
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It seems that for each `const fn`, Verus synthesizes two related functions and assigns the attributes to both of them. But this causes problems when there's an attribute that can't be duplicated, like `rustc_diagnostic_item`.

Perhaps `rustc_diagnostic_item` should be a special case:  Attributes of that specific type should only be assigned to one of the functions.

jaylorch:
Fixes #1989.

Uses code synthesized by Claude Sonnet 4.5 in Github Copilot.

### DIFF
--- a/source/builtin_macros/src/contrib/exec_spec.rs
+++ b/source/builtin_macros/src/contrib/exec_spec.rs
@@ -1805,8 +1805,8 @@ pub fn exec_spec(input: TokenStream) -> TokenStream {
         .collect::<Result<Vec<_>, _>>();
 
     match res {
-        Ok(ts) => quote! {
-            ::verus_builtin_macros::verus! { #(#ts)* }
+        Ok(ts) => quote_vstd! { vstd =>
+            #vstd::prelude::verus! { #(#ts)* }
         }
         .into(),
         Err(err) => err,
--- a/source/builtin_macros/src/is_variant.rs
+++ b/source/builtin_macros/src/is_variant.rs
@@ -123,8 +123,8 @@ fn attribute_is_variant_internal(
                 &syn::Fields::Unit => quote! {},
             };
 
-            quote_spanned_builtin! { verus_builtin, variant_ident.span() =>
-                ::verus_builtin_macros::verus! {
+            quote_spanned_builtin_builtin_macros! { verus_builtin, verus_builtin_macros, variant_ident.span() =>
+                #verus_builtin_macros::verus! {
                     #[cfg(verus_keep_ghost)]
                     #[allow(non_snake_case)]
                     #[verus::internal(spec)]
--- a/source/builtin_macros/src/struct_decl_inv.rs
+++ b/source/builtin_macros/src/struct_decl_inv.rs
@@ -76,8 +76,8 @@ fn struct_decl_inv_main(sdi: SDI) -> parse::Result<TokenStream> {
 
     output_wf(&sdi, &mut stream, wf_body_stream);
 
-    Ok(quote! {
-        ::verus_builtin_macros::verus!{
+    Ok(quote_vstd! { vstd =>
+        #vstd::prelude::verus!{
             #stream
         }
     })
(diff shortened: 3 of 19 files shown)


source/builtin_macros/src/unerased_proxies.rs
```diff
@@ -379,6 +379,14 @@ impl crate::syntax::Visitor {
                     &format!("{}{}", VERUS_UNERASED_PROXY, &item_fn.sig.ident),
                     item_fn.sig.span(),
                 );
+                // Remove rustc_diagnostic_item attribute to avoid duplicate diagnostic item errors
+                item_fn.attrs.retain(|attr| {
+                    if let Some(last_seg) = attr.path().segments.last() {
```
tjhance: can you factor this out into 'is_rustc_diagnostic_item_attribute' function or similar?

And please check that `path().segments` has length 1, as well.

tjhance: Actually, it's probably easiest to use the `get_ident` or `is_ident` function (https://docs.rs/syn/latest/syn/struct.Path.html#method.get_ident)

jaylorch: The use case is for a project about writing GPU kernels in Rust. In that use case, it doesn't much matter how Verus treats the `rustc_diagnostic_item` attribute since it will only be used when compiling. All that's necessary is that Verus doesn't choke on such an attribute. But you have a good point that in other scenarios where it does matter, it would be better to leave the attribute on the non-erased version.

Thanks for the suggestions! I've incorporated them into the latest commit.

ziqiaozhou: This is an ad-hoc solution for rustc_diagnostic_item attribute, but I can imagine that there are also other Rust attribute we did not used may also get a similar issue if the attribute can only be used once with the same parameter in a project.

Would that make sense to do a general solution if the unerased_proxy is a long-term solution for const?
```
#[verifier::no_dup_attribute(xxx)]
#[xxx(...)]
const fn foo() {}
```

BTW, I think I also need to apply similar patch to attr_rewrite.rs for verus_spec since verus_spec does not reuse the unerased_proxy here. No need to make change in this PR. I will find time to change verus_spec to reuse the unerased_proxy code here.

jaylorch: If someone runs into other unduplicable attributes, we can generalize the solution later. I'm inclined to have this ad-hoc solution now because if it turns out to be sufficient, it has the advantage of not requiring any more documentation. It just works without the developer needing to know about a feature like `#[verifier::no_dup_attribute]`.

jaylorch: I updated the name of the function from `is_rustc_diagnostic_item_attribute` to `attribute_cannot_be_on_two_different_functions` so that, in future, if we find additional such attributes and/or decide to let the user manually specify such attributes with `#[verifier::no_dup_attributes(xxx)]`, we have a place to put those future updates.

jaylorch: FWIW, now there's practically no code left in this PR that wasn't written by a human. :-)

Issue #1984, opened 2025-11-16
`#[verus_spec]` for cross-crate reference


hiroki-chen:
The symbols generated for `#[verus_spec]` annotated function in crate A will be missing if crate B is trying to reference it.

```rust

// crate_a/src/lib.rs

#[verus_spec(r =>
    with
        Tracked(bar): Tracked<()>,
    ensures
        true,
)]
pub fn foo() {}

// crate_b/src/lib.rs
#![feature(proc_macro_hygiene)]

use vstd::prelude::*;

fn bar() {
    {
        #[verus_spec(with Tracked(()))]
        liba::foo()
    };
}
```

```txt
error[E0425]: cannot find function `_VERUS_VERIFIED_foo` in crate `liba`
 --> libb/src/lib.rs:8:15
  |
8 |         liba::foo()
  |               ^^^ not found in `liba`

For more information about this error, try `rustc --explain E0425`.
```

I'm not sure if this is an intended behavior or a bug?

hiroki-chen:
For what it's worth, I attach the expanded rust files below:

```rust

#[verus::internal(verus_macro)]
#[verus::internal(external_body)]
pub fn foo() {
    ::vstd::prelude::requires([false]);
    ::core::panicking::panic("not implemented");
}
#[verus::internal(verus_macro)]
pub fn _VERUS_VERIFIED_foo(verus_tmp_bar: Tracked<()>) {
    #[verus::internal(header_unwrap_parameter)]
    let bar;

    #[verifier::proof_block]
    { bar = verus_tmp_bar.get() };
    ::vstd::prelude::ensures([true]);
}
....


use std::prelude::rust_2024::*;
use vstd::prelude::*;
pub fn foo() {}
// missing exported
```


```rust
#[macro_use]
extern crate std;
#[prelude_import]
use std::prelude::rust_2024::*;
use vstd::prelude::*;
fn bar() {
    {
        liba::_VERUS_VERIFIED_foo(#[verifier::ghost_wrapper] ::vstd::prelude::tracked_exec(#[verifier::tracked_block_wrapped] ()))
    };
}
```

Thus we may have to export symbols to the outside to make ensure that `verus_spec` can bridge them up and seems now we only have `foo` exported.

hiroki-chen:
So this is what happens during verification and compilation:

1. In crate A we generated verus only symbols `_VERUS_VERIFIED_`.
2. After verification this symbol is erased.
[...]

PR #1992, opened 2025-11-18
verus_spec: Support cross-crate call to function with ghost/tracked vars

ziqiaozhou:
* Erase all when erase_all()
* Keep function stub and erase proof statements when in erase()

It should fix the bug reported in #1984

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -820,8 +820,13 @@ fn rewrite_unverified_func(
     );
     unverified_fun.attrs_mut().push(mk_verus_attr_syn(span, quote! { external_body }));
     if let Some(block) = unverified_fun.block_mut() {
-        // If not erase mode, we erase the function body to avoid using
-        // proof code, since verus_spec may not be properly erased proofs.
+        // For an unverified function, if it is in keep mode,
+        // we erase the function body to avoid using
+        // proof code, since we do not need to verify anything in unverified
+        // function and we never pass ghost/tracked to unverified function
+        // and so it may cause errors due to undefined vars.
+        // Since the body is erased only in keep mode, we still
+        // see correct body in generated executable in erase mode.
         if erase.keep() {
             block.stmts.clear();
             block.stmts.push(precondition_false);


hiroki-chen (commented):
The change fixes the cross-crate call but unfortunately the executable code is broken.

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -795,7 +795,11 @@ fn rewrite_const_ret_proxy(const_fun: &mut syn::ItemFn) -> syn::ItemFn {
 // function body, to enable seamless use of unverified call to the function in
 // verification.
 // If the function is const, it will be rewritten to a proxy function and a verified function.
-fn rewrite_unverified_func(fun: &mut syn::ItemFn, span: proc_macro2::Span) -> Vec<syn::ItemFn> {
+fn rewrite_unverified_func(
```
hiroki-chen, line 798: Specifically when B tries to call A's function exported, the stub is "unimplemented!()" and we will panic inside B instead of calling the real function's body.

ziqiaozhou: When you compile the code, it should in EraseAll mode (no verus_keep_ghost or verus_keep_ghost_body) and so it should just return the original code without reaching this rewrite.

ziqiaozhou: Oh. I see. It may cause problem if using `cargo verus build` instead of cargo build to build it

hiroki-chen: I tried but the function's body is indeed replaced with unimplemented and the caller panicked

ziqiaozhou: Both `cargo build` and `cargo verus build` should work properly with the unverified function call used in generated binary. Unfortunately, I cannot add a unit test but I tried it locally.

hiroki-chen: While this fixes the problem, it may break backward compatability:

```rust
#[verus_spec(r =>
    with
        Tracked(bar): Tracked<()>,
    ensures
        true,
)]
pub fn foo() {
    let abcd = Tracked(bar);
}

error[E0425]: cannot find value `bar` in this scope
  --> liba/src/lib.rs:12:24
   |
12 |     let abcd = Tracked(bar);
   |                        ^^^ not found in this scope
```

Imo a lot of the old codebase still relies on `Tracked`; should this also be supported?

For example when you borrow pointers you'd wrap tracked variables inside `Tracked` to make the code compile.

ziqiaozhou: I see. I need to erase function body in keep mode.

```
#[verus_spec(r =>
    with
        Tracked(bar): Tracked<()>,
    ensures
        true,
)]
pub fn foo() {
    let abcd = Tracked(bar);
}
```
Did you use it inside verus!? It should never work if it is outside verus!.

hiroki-chen: The code was wrapped in `verus!` macro.

hiroki-chen:
Now seems the issue has been fixed, thanks!

hiroki-chen:
Just a nitpick:

```rust
use A::foo;

#[verus_spec(with xxx) foo();
                                         ^^^ not found in this scope

consider importing this function

use A::_VERUS_VERIFIED_foo;
```

causes some trouble because we are importing `foo` rather than `_VERUS_VERIFIED_foo`. Simple workaround could be always to use the fully-qualified path to replace the loacl `foo` call.

ziqiaozhou:
> Just a nitpick:
> 
> ```rust
> use A::foo;
> 
> #[verus_spec(with xxx) foo();
>                                          ^^^ not found in this scope
> 
> consider importing this function
> 
> use A::_VERUS_VERIFIED_foo;
> ```
> 
> causes some trouble because we are importing `foo` rather than `_VERUS_VERIFIED_foo`. Simple workaround could be always to use the fully-qualified path to replace the loacl `foo` call.

Yes. That is a problem even when it is used not cross crate. I usually will do `use A::*` guarded by a cfg(verus_keep_ghost) or just use full path. We may get a better solution if verus tool can support real ghost function parameter from the a verus-builtin function instead of via the macro tricks to create two version of function with different signatures. It might be feasible based on @Chris-Hawblitzel's feedback. But it takes time to make it happen.

PR #1993, opened 2025-11-18
support async functions

FeizaiYiHao:
OK. Finally, here's the PR for async functions. 
# Changes in Macros
We add two functions to make sure async functions cannot `open_invariants`, must name their return value, and each expression in the ensures clause must have the format `return_value.awaited() ==> { xxx }`. 

# Changes in Vstd
We add `future.rs` to define the `Future` trait and three functions for `impl Future<Output = T>` returned from async functions:
 
`spec fn view(&self) -> T;` returns the value from calling `await()` on the future.

`spec fn awaited(&self) -> bool;`: This is an uninterp spec function that only becomes `True` when `await()` returns

`fn exec_await(self) -> (ret: T);`: `.await()` are rewritten into `.exec_await()` during HIR to VIR. I'm not sure if this is a good idea. I'll explain later...

# Changes in HIR to VIR
## Async function
`async_body_to_vir()` looks into the hir function body of the async function and extracts the actual function body from the desugared async function (Rustc desugars the async function body into a special async closure). 

In the special async closure, all the function parameters are rebound with the same name. Therefore, `handle_async_func` looks into the closure and records the bindings.

In `mode.rs`, we use this info to resolve the rebindings in async functions. 

## Await
Each await expression (`await()`) is treated as a call to `self.exec_await()`. And `exec_await` defines all the necessary postconditions of (`await()`). However, a user might just call `exec_await()` instead of `await()`, which can cause problems because first `exec_await()` doesn't really exist, and second `exec_await()` can be called in a non-async function.

# Changes in AIR
Not much change is needed.

1. For checking postconditions, to use the `T` in `impl Future<Output = T>` for revealing opaque types if `T` is an opaque type
2. We add a few AIR `Stmt`s to tell the solver that when the returned future is called `await()` on, the output is equal to the return value of the function body (as if the function is not async). 

Let me know if `exec_await` is a good idea, and anything that can be improved. Thank you!

I'll add more test cases and comments once we settle on a design. 

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -12,6 +12,7 @@
     register_tool(verifier)
 )]
 
+use core::future::Future;
 use core::marker::PhantomData;
 
 #[cfg(verus_keep_ghost)]
@@ -244,6 +245,13 @@ pub fn with_triggers<A, B>(_triggers_tuples: A, body: B) -> B {
 #[verifier::spec]
 pub fn constrain_type<T>(_x: T, _y: T) -> bool {
     true
+}
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::get_future_output_type"]
+#[verifier::spec]
+pub fn get_future_output_type<T>(_x: impl Future<Output = T>) -> T {
+    unimplemented!()
 }
 
 // example: forall with three triggers [f(x), g(y)], [h(x, y)], [m(y, x)]:
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -541,6 +541,7 @@ impl Visitor {
         fn_ident: &Ident,
         generics: Option<impl ToTokens>,
         inputs: (Option<impl ToTokens>, impl ToTokens), // optional self and args
+        is_async_fn: bool,                              // is the function an async function
     ) -> Vec<Stmt> {
         let requires = self.take_ghost(&mut spec.requires);
         let recommends = self.take_ghost(&mut spec.recommends);
@@ -719,7 +720,11 @@ impl Visitor {
                                         }
                                     }
                                 };
-                                quote_spanned_builtin!(verus_builtin, token.span => #verus_builtin::constrain_type(#ret_val_ident, #receiver_token#fn_ident#generics_token(#args)))
+                                if is_async_fn {
+                                    quote_spanned_builtin!(verus_builtin, token.span => #verus_builtin::constrain_type(#ret_val_ident, #verus_builtin::get_future_output_type(#receiver_token#fn_ident#generics_token(#args))))
+                                } else {
+                                    quote_spanned_builtin!(verus_builtin, token.span => #verus_builtin::constrain_type(#ret_val_ident, #receiver_token#fn_ident#generics_token(#args)))
+                                }
                             };
                             let contrain_typ_expr = Expr::Verbatim(constrain_type);
                             spec_stmts.push(Stmt::Expr(
@@ -1061,6 +1066,7 @@ impl Visitor {
             &sig.ident,
             verus_generic_to_tokens(&sig.generics),
             verus_inputs_to_tokens(&sig.inputs),
+            sig.asyncness.is_some(),
         );
         if !self.erase_ghost.erase() {
             if !(self.rustdoc && sig.constness.is_some()) {
@@ -5060,6 +5066,7 @@ pub(crate) fn sig_specs_attr(
         &sig.ident,
         generic_to_tokens(&sig.generics),
         inputs_to_tokens(&sig.inputs),
+        sig.asyncness.is_some(),
     ));
     spec_stmts
 }
@@ -5573,7 +5580,7 @@ fn check_return_idents(
 }
 
 /// In VIR there's the same check, but Rustc will complain first, and throw out
-/// some errors about "constrain_type", which ar confusing and the users should not see.
+/// some errors about "constrain_type", which are confusing and the users should not see.
 /// Instead we give an early error with nice error msg here.
 fn check_verus_return_idents(
     ret_pat: &Pat,
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -2877,6 +2877,15 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
                 }
             }
         }
+        // This a desugered `await` expression
+        ExprKind::Match(
+            Expr { hir_id: _, kind: ExprKind::Call(_expr, call_args), span: _ },
+            _arms,
+            rustc_hir::MatchSource::AwaitDesugar,
+        ) => {
+            let vir_expr = expr_to_vir_consume(bctx, &call_args[0], modifier)?;
+            mk_expr(ExprX::Await(vir_expr))
+        }
         ExprKind::Match(expr, arms, _match_source) => {
             let vir_place = expr_to_vir_place(bctx, expr, modifier)?;
             let mut vir_arms: Vec<vir::ast::Arm> = Vec::new();
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -3,18 +3,19 @@ use crate::automatic_derive::AutomaticDeriveAction;
 use crate::config::Vstd;
 use crate::context::{BodyCtxt, Context, ContextX, HeaderSetting};
 use crate::resolve_traits::{ResolutionResult, ResolvedItem};
-use crate::rust_to_vir_base::mk_visibility;
 use crate::rust_to_vir_base::{
-    check_fn_opaque_ty, check_generics_bounds_no_polarity, def_id_to_vir_path, no_body_param_to_var,
+    check_fn_opaque_ty, check_generics_bounds_no_polarity, def_id_to_vir_path, local_to_var,
+    no_body_param_to_var,
 };
+use crate::rust_to_vir_base::{mk_visibility, qpath_to_ident};
 use crate::rust_to_vir_expr::{ExprModifier, expr_to_vir_consume, pat_to_mut_var};
 use crate::rust_to_vir_impl::ExternalInfo;
 use crate::util::{err_span, err_span_bare};
 use crate::verus_items::{BuiltinTypeItem, VerusItem};
 use crate::{unsupported_err, unsupported_err_unless};
 use rustc_hir::{
-    Attribute, Body, BodyId, Crate, ExprKind, FnDecl, FnHeader, FnSig, Generics, HeaderSafety,
-    HirId, MaybeOwner, Param, Safety,
+    Attribute, Body, BodyId, Crate, Expr, ExprKind, FnDecl, FnHeader, FnSig, Generics,
+    HeaderSafety, HirId, MaybeOwner, Param, Safety,
 };
 use rustc_middle::ty::{
     AdtDef, BoundRegion, BoundRegionKind, BoundVar, Clause, ClauseKind, ConstKind, GenericArg,
@@ -27,19 +28,22 @@ use rustc_span::def_id::DefId;
 use rustc_span::symbol::Ident;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use std::vec;
 use vir::ast::{
     BodyVisibility, Fun, FunX, FunctionAttrsX, FunctionKind, FunctionX, ItemKind, KrateX, Mode,
     OpaqueTypes, Opaqueness, ParamX, Path, Typ, TypDecoration, TypX, VarIdent, VirErr, Visibility,
 };
 use vir::ast_util::{air_unique_var, unit_typ};
-use vir::def::{RETURN_VALUE, VERUS_SPEC};
+use vir::def::{RETURN_VALUE, Spanned, VERUS_SPEC};
 use vir::sst_util::subst_typ;
 
+#[derive(Debug)]
 enum FnOrConstSigEnum<'tcx> {
     Fn(&'tcx FnSig<'tcx>),
     ConstVar(Typ),
 }
 
+#[derive(Debug)]
 pub(crate) struct FnOrConstSig<'tcx> {
     span: Span,
     sig: FnOrConstSigEnum<'tcx>,
@@ -68,6 +72,13 @@ impl<'tcx> FnOrConstSig<'tcx> {
                 _ => Safety::Unsafe,
             },
             FnOrConstSigEnum::ConstVar(_) => Safety::Safe,
+        }
+    }
+
+    fn asyncness(&self) -> rustc_hir::IsAsync {
+        match &self.sig {
+            FnOrConstSigEnum::Fn(sig) => sig.header.asyncness,
+            FnOrConstSigEnum::ConstVar(_) => rustc_hir::IsAsync::NotAsync,
         }
     }
 
@@ -259,9 +270,11 @@ fn handle_autospec<'tcx>(
                     ignore_outside_new_mut_ref: functionx.attrs.ignore_outside_new_mut_ref,
                     tracked_swap: false,
                     tracked_take_option: false,
+                    is_async: false,
                 }),
                 body: Some(ret_clause.clone()),
                 extra_dependencies: functionx.extra_dependencies.clone(),
+                async_ret: functionx.async_ret.clone(),
             },
         );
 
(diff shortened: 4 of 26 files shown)


FeizaiYiHao:
One alternative design for `exec_await()` is to carry the `await` expression all the way to the AIR stage and hardcode the effect of `await`.

source/builtin_macros/src/syntax.rs
```diff
@@ -5384,3 +5392,80 @@ fn check_verus_return_ident(
     }
     None
 }
+
+///
+fn check_async_fn_open_invariants(sig: &mut Signature) -> Option<Stmt> {
```
tjhance: This sort of check should be performed in Verus itself (e.g., `vir/src/well_formed.rs`)

source/vstd/future.rs
```diff
@@ -0,0 +1,42 @@
...
+pub trait ExFuture {
+    type ExternalTraitSpecificationFor: core::future::Future;
+
+    type Output;
+}
+
+pub trait FutureAdditionalSpecFns<T>: Future<Output = T> {
+    spec fn view(&self) -> T;
+
+    spec fn awaited(&self) -> bool;
```
tjhance, line 20: These functions need to be marked `#[verifier::prophetic]` for soundness.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2352,6 +2354,66 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
...
+                            hir_id: _,
+                            kind: ExprKind::Path(QPath::LangItem(LangItem::IntoFutureIntoFuture, _)),
+                            span: _,
+                        },
+                        call_args,
+                    ),
+                span: _,
+            },
+            _arms,
+            _match_source,
```
tjhance: Please check that that the `match_source` is `AwaitDesugar`.

FeizaiYiHao:
Thank you @tjhance. Let me spend some time on these changes...

parno:
@FeizaiYiHao I just wanted to check in to see if you were able to make progress on this and whether there's anything we can help with.  I just talked with a large Verus user who mentioned that this is the biggest feature they're missing, so they were very excited to hear that you're working on it.

FeizaiYiHao:
> @FeizaiYiHao I just wanted to check in to see if you were able to make progress on this and whether there's anything we can help with. I just talked with a large Verus user who mentioned that this is the biggest feature they're missing, so they were very excited to hear that you're working on it.

Sorry, I got distracted by other stuffs. Will send updates in a few days for you all to review. Stay tuned 😉.

parno:
No worries -- thanks for continuing to push it along!

FeizaiYiHao:
> Did a quick pass through. I have a few high-level points:
> 
> 1. I think I would prefer if `await` were an actual VIR node. Due to its nontrivial semantics, some of our internal analyses may need to know about it.
> 2. I think it would be cleaner if signatures on async functions were written on the non-wrapped type. So to take your example, the user would write:
> 
> ```
>     async fn foo() -> (ret: (impl DummyTrait, impl DummyTrait))
>         ensures
>             ret.0.dummy_spec(),
>     {
>         (true, true)
>     }
> ```
> 
> And the body would be checked much like an ordinary function.
> 
> Verus can then automatically create the specification for the for the Future version of the function-signature:
> 
> ```rust
> fn foo_desugared() -> (ret: Future<(impl DummyTrait, impl DummyTrait)>)
>         ensures
>             ret.awaited() ==> ret@.0.dummy_spec(),
> ```
> 
> which all callers would use.

@tjhance  Do we want to have `await()` in vir, sst, and air? Or we just do checks in vir?

tjhance:
I think what makes the most sense would be

AST: Have a dedicated node Await
ast_to_sst: Replace it with a function call

FeizaiYiHao:
> Did a quick pass through. I have a few high-level points:
> 
> 1. I think I would prefer if `await` were an actual VIR node. Due to its nontrivial semantics, some of our internal analyses may need to know about it.
> 2. I think it would be cleaner if signatures on async functions were written on the non-wrapped type. So, to take your example, the user would write:
> 
> ```
>     async fn foo() -> (ret: (impl DummyTrait, impl DummyTrait))
>         ensures
>             ret.0.dummy_spec(),
>     {
>         (true, true)
>     }
> ```
> 
> And the body would be checked much like an ordinary function.
> 
> Verus can then automatically create the specification for the for the Future version of the function-signature:
> 
> ```rust
> fn foo_desugared() -> (ret: Future<(impl DummyTrait, impl DummyTrait)>)
>         ensures
>             ret.awaited() ==> ret@.0.dummy_spec(),
> ```
> 
> which all callers would use.

@tjhance @parno 
 ```
     async fn foo() -> (ret: (impl DummyTrait, impl DummyTrait))
         ensures
             ret.0.dummy_spec(),
     {
         (true, true)
     }
 ```
has a problem. The type of `ret` at the function return is already a `future`. Do we want to follow the principle in which we have consistent types for both `ret` at the function return and `ret` in the ensure clause? 

After spending some time revisiting this design (and I tried it around the summer when we were working on the opaque type PR), I think my original design is cleaner (at least in terms of the internal changes needed to Verus). We add a few spec functions in vstd, do a well-formed check to the postconditions of the `async` function, and rewrite `await` into a function call in vir_to_sst.

If we want to stick to the above design, we'd need to modify the `verus` macro to rewrite `ret` in the ensures clause to `ret@` anyway, otherwise Rust type check will not pass, which I think is better to left to the users to write manually? 

Let me know what you guys think.

tjhance:
It would be good to get the others' opinions, but I still think that the following:

```
    async fn foo() -> (ret: u64)
        ensures ret > 0
```

is cleaner than:

```rust
    async fn foo() -> (ret: u64)
        ensures ret.awaited() ==> ret@ > 0
```

I'm sure we can work through any implementation challenges; it's more important to have a clean design for users.

FeizaiYiHao:
How about 
```
async fn foo() -> (ret: u64)
        ensures ret@ > 0
```
?

At least here the types are consistent?

FeizaiYiHao:
Required changes are not very hard to make. We just need to rewrite `ret` to `ret@` in Verus macro before we send anything to Rustc.

parno:
FWIW, I like Travis's clean version
```
    async fn foo() -> (ret: u64)
        ensures ret > 0
```
but I think @FeizaiYiHao's proposal for
```
async fn foo() -> (ret: u64)
        ensures ret@ > 0
```
seems like a reasonable compromise, although it might be non-obvious to the developer if they initially write Travis' version that they can fix it by adding an `@`.

> We just need to rewrite `ret` to `ret@` in Verus macro before we send anything to Rustc.

That makes me a bit nervous, since I'm not sure we can be confident of finding all of the right `ret`s to replace.  E.g., if the user writes:
```
    async fn foo() -> (ret: u64)
        ensures ret > 0 && { let ret = 5; ret < 10 }
```
then we want to replace the first `ret` but not the next two.

FeizaiYiHao:
You are right, rewriting `ret` in Verus macro is difficult. Alternatively we can change the `constraint_type` function to set `ret` type differently.  Let me know which approach we'd want the most. They are not very hard to implement.

tjhance:
If the user writes `ret: u64` then it looks like `ret` has type `u64`. IMO, anything else will be confusing. Furthermore, we've talked before about adding pattern syntax to the return value, and I'd like that to extend here as well.

---

We control the encoding, so we can make `ret` whatever type we want. Right now the type is determined by the line:

```
::vstd::prelude::constrain_type(ret, test(a))
```

but for async functions, we can change that to:

```
::vstd::prelude::constrain_type(ret, get_future_output_type(test(a))
```

or something similar.

FeizaiYiHao:
Current plan:
In Macro: use `constain_typ` + `get_future_output_type` to set the type of the return variable in the ensure clause to the `Output` type (without the Future).

For async function body checks, since the above approach, we can perform the checks almost 100% the same as normal functions.

For `await`, we keep it as a separate vir node and rewrite it to a function call in vir_to_sst, since the open invariants check happens at sst stage and we set the function to have `open_invaraints any`, this will prevent any function with less than `open_invariants any` from calling `await`.

Function postconditions (Function specs axioms) are a bit tricky. The current plan is to handle it during vir_to_sst too. 
If we have 
```
async fn foo() -> (ret: u64)
        ensures ret > 0
```
we generate sst that is conceptually equivalent to the following code

```
async fn foo() -> (VERUS_INTERNAL_ASYNC_ret: impl Future<Output = u64>)
        ensures 
            VERUS_INTERNAL_ASYNC_ret.awaited() ==> {let ret = VERUS_INTERNAL_ASYNC_ret@; ret > 0}),
```
This seems to be clean. 

Let me know what you think @parno @tjhance

FeizaiYiHao:
Update: The Above change is implemented. Rebasing to the newest Verus version

FeizaiYiHao:
OK. @tjhance Take a look when you have time? If it looks good I'll start on async trait functions and test case. lmk!

FeizaiYiHao:
@tjhance @parno In case GitHub notification didn't work. Pin.

source/builtin/src/lib.rs
```diff
@@ -246,6 +247,13 @@ pub fn constrain_type<T>(_x: T, _y: T) -> bool {
     true
 }
 
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::get_future_type"]
+#[verifier::spec]
+pub fn get_future_type<T>(_x: impl Future<Output = T>) -> T {
```
tjhance: nitpick, but could we call this future_output_type or something with output in it?

source/vir/src/ast_to_sst.rs
```diff
@@ -2546,6 +2547,56 @@ pub(crate) fn expr_to_stm_opt(
...
+                                        Arc::new("exec_await".to_string()),
+                                    ]),
+                                }),
+                            }),
+                            typs: Arc::new(vec![expr.typ.clone(), e.typ.clone()]),
+                            impl_paths: Arc::new(vec![]),
+                            is_trait_default: false,
+                        },
+                        Arc::new(FunX {
+                            path: Arc::new(crate::ast::PathX {
```
tjhance: You can use the `path!` macro to make this look a lot cleaner.

source/vir/src/ast_to_sst.rs
```diff
@@ -2546,6 +2547,56 @@ pub(crate) fn expr_to_stm_opt(
...
+                &expr.typ,
+                ExprX::Call(
+                    CallTarget::Fun(
+                        crate::ast::CallTargetKind::DynamicResolved {
+                            resolved: Arc::new(FunX {
+                                path: Arc::new(crate::ast::PathX {
+                                    krate: Some(Arc::new("vstd".to_string())),
+                                    segments: Arc::new(vec![
+                                        Arc::new("future".to_string()),
+                                        Arc::new("impl&%0".to_string()),
```
tjhance: Use `rustc_diagnostic_item` to give this a stable name without an `impl&%0` identifier

source/vir/src/sst_to_air.rs
```diff
@@ -3270,6 +3268,62 @@ pub(crate) fn body_stm_to_air(
...
+//     );
+//     let eq = ExprX::Binary(
+//         air::ast::BinaryOp::Eq,
+//         exp_to_expr(ctx, ret_exp, expr_ctxt)?,
+//         exp_to_expr(ctx, &SpannedTyped::new(&stm.span, &ret, call), expr_ctxt)?,
+//     );
+
+//     *ret_op = Some(async_rete.clone());
+//     Ok(vec![Arc::new(StmtX::Assume(eq.into()))])
+// }
```
tjhance, line 3336: delete this if you don't plan to use it

FeizaiYiHao:
@tjhance The following two functions produce two different opaque types, but they should be the same since conceptually they are the same.
```
    #[verifier(external)]
    async fn negate_bool(b: bool, x: u8) -> bool {
        !b
    }

    async fn negate_bool_requires_ensures(b: bool, x: u8) -> (ret_b: bool)
        requires x != 0,
        ensures ret_b == !b
    {
        negate_bool(b, x).await
    }
```
My plan is to emit an axiom in SMT that states these two types are equivalent. Any idea?

FeizaiYiHao:
> @tjhance The following two functions produce two different opaque types, but they should be the same since conceptually they are the same.
> 
> ```
>     #[verifier(external)]
>     async fn negate_bool(b: bool, x: u8) -> bool {
>         !b
>     }
> 
>     #[verifier(external_fn_specification)]
>     async fn negate_bool_requires_ensures(b: bool, x: u8) -> (ret_b: bool)
>         requires x != 0,
>         ensures ret_b == !b
>     {
>         negate_bool(b, x).await
>     }
> ```
> 
> My plan is to emit an axiom in SMT that states these two types are equivalent. Any idea?

Actually await. There's something changes to make for opaque types for assume_specification.

parno:
@FeizaiYiHao, now that #2138 has landed, where does that leave this PR?

FeizaiYiHao:
> @FeizaiYiHao, now that #2138 has landed, where does that leave this PR?

@parno, We should be really close. The hardest part is probably rebasing. I'll try to get this ready in one to two days.

parno:
Awesome!  Thanks for working on it.

FeizaiYiHao:
@parno @tjhance Pushed all the changes. We don't support trait async functions as we don't support trait functions that return opaque types. Otherwise, everything is there.   

@tjhance I couldn't get rid of `impl&%0`. Not sure why. Maybe because the default impl has a different SMT encoding for the types than the impl function. I could take another look at it.

tjhance:
@Chris-Hawblitzel says he is going to take a look at the `impl&%0` issue

Chris-Hawblitzel:
> @tjhance I couldn't get rid of `impl&%0`. Not sure why. Maybe because the default impl has a different SMT encoding for the types than the impl function. I could take another look at it.

> @Chris-Hawblitzel says he is going to take a look at the impl&%0 issue

Actually, after looking at it, I think you don't need it anyway.  Since you already declare the `ensures` that you want in the trait:

```
pub trait FutureAdditionalSpecFns<T>: Future<Output = T> {
    spec fn view(&self) -> T;

    spec fn awaited(&self) -> bool;

    fn exec_await(self) -> (ret: T)
        ensures
            self.awaited() == true,
            ret == self@,
        opens_invariants any
    ;
}
```

you don't need to additionally declare the same `ensures` in the impl:

```
    #[verifier::external_body]
    fn exec_await(self) -> (ret: V)
        ensures
            self.awaited() == true,
            ret == self@,
    {
        unimplemented!()
    }
```

So you can remove the impl ensures and then just use `Dynamic` instead of `DynamicResolved`:

```
diff --git a/source/vir/src/ast_to_sst.rs b/source/vir/src/ast_to_sst.rs
index cc9f15aa6..90833ae83 100644
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -2827,12 +2827,7 @@ pub(crate) fn expr_to_stm_opt(
                 &expr.typ,
                 ExprX::Call(
                     CallTarget::Fun(
-                        crate::ast::CallTargetKind::DynamicResolved {
-                            resolved: fun!("vstd" => "future", "impl&%0", "exec_await"),
-                            typs: Arc::new(vec![expr.typ.clone(), e.typ.clone()]),
-                            impl_paths: Arc::new(vec![]),
-                            is_trait_default: false,
-                        },
+                        crate::ast::CallTargetKind::Dynamic,
                         fun!("vstd" => "future", "FutureAdditionalSpecFns", "exec_await"),
                         Arc::new(vec![e.typ.clone(), expr.typ.clone()]),
                         Arc::new(vec![crate::ast::ImplPath::TraitImplPath(
diff --git a/source/vstd/future.rs b/source/vstd/future.rs
index 326fb5951..0d4554acd 100644
--- a/source/vstd/future.rs
+++ b/source/vstd/future.rs
@@ -30,11 +30,7 @@ impl<V, T: Future<Output = V>> FutureAdditionalSpecFns<V> for T {
     uninterp spec fn awaited(&self) -> bool;

     #[verifier::external_body]
-    fn exec_await(self) -> (ret: V)
-        ensures
-            self.awaited() == true,
-            ret == self@,
-    {
+    fn exec_await(self) -> (ret: V) {
         unimplemented!()
     }
 }
```

FeizaiYiHao:
@Chris-Hawblitzel 

Thank you! I'll push the fixes soon.

FeizaiYiHao:
Sorry for the delay. Pushed the fix for the `impl&%0` issue. Also added `verifier::prophetic` for `view` and `awaited` in `FutureAdditionalSpecFns`

parno:
Thanks for implementing the fix!  @FeizaiYiHao Does that mean all of the requested changes have been made at this point?

FeizaiYiHao:
> Thanks for implementing the fix! @FeizaiYiHao Does that mean all of the requested changes have been made at this point?

Yes. I believe so. You can double-check all the features in `source/rust_verify_test/tests/async_functions.rs`, in case I miss anything (it's been a long time, so I'm not 100% sure I remember everything...).

source/vir/src/ast.rs
```diff
@@ -1586,6 +1632,12 @@ pub struct FunctionX {
     /// Extra dependencies, only used for for the purposes of recursion-well-foundedness
     /// Useful only for trusted fns.
     pub extra_dependencies: Vec<Fun>,
+    /// The async function body is desugared into a special async closure, in which
+    /// the parameters of the function are rebound again in the closure with the same name but different
+    /// identifier, here we record and later resolve them
+    pub async_params_mode_binding: Option<Arc<Vec<(VarIdent, Mode)>>>,
+    /// The return type of the async function i.e., impl Future<Output>.
+    pub async_ret: Option<Param>,
```
tjhance, line 1693: Could we package these into one field with one Option?

FeizaiYiHao: Np.

FeizaiYiHao:
> I noticed a small problem, which is that Verus allows you to call `exec_await`:
> 
> ```
>         use vstd::future::FutureAdditionalSpecFns;
>         use vstd::prelude::*;
>         verus!{
>         pub async fn bar(x: &mut usize) -> (ret: ())
>             ensures
>                 *x == 2333,
>         {
>             *x = 2333;
>         }
> 
>         fn foo(){
>             let mut x = 233;
>             let future = bar(&mut x);
>             future.exec_await();
>             assert(x == 2333);
>         }
>         }
>         fn main() { }
> ```
> 
> Other than this I think it looks good.

@tjhance Let's add some checks in `well_form.rs`? Or is there some way to make it inaccessible for the users that I'm not aware of?

tjhance:
If you make it a top level function, rather than a trait function, you can mark it private:

```
    // Do not call this function. Call the regular Rust `await` keyword instead.
    fn exec_await<F: Future>(future: F) -> (ret: F::Output)
        ensures
            future.awaited() == true,
            ret == future@,
        opens_invariants any
    ;
```

See how exec_nonstatic_call works:

https://github.com/verus-lang/verus/blob/cf420a595bbfc82381b12b96521ee7d74b1fd324/source/vstd/pervasive.rs#L125

https://github.com/verus-lang/verus/blob/cf420a595bbfc82381b12b96521ee7d74b1fd324/source/rust_verify/src/rust_to_vir_func.rs#L1930

FeizaiYiHao:
@tjhance fixed. Take a look?

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -2017,11 +2215,36 @@ pub(crate) fn check_item_fn<'tcx>(
         attrs: fattrs,
         body,
         extra_dependencies: header.extra_dependencies,
+        async_params_mode_binding_and_ret: if let Some(async_ret) = async_ret {
+            Some((Arc::new(vec![]), async_ret))
```
Chris-Hawblitzel: Is the `Vec` in `async_params_mode_binding_and_ret` used?  It seems to always be set to `vec![]`.

FeizaiYiHao: Nope. Rebinding is done directly in `handle_async_func()` now. Forget to kill this dead code. Good catch

FeizaiYiHao: OK pushed @Chris-Hawblitzel

FeizaiYiHao:
OK. Time to rebase...

FeizaiYiHao:
@Chris-Hawblitzel @tjhance OK ready to be merged

PR #1997, opened 2025-11-20
Apply poly wrappers to decreases check from loops

### LINKED ISSUE #1372: Panic when `decreases` clause is on a set with generic
I wasn't sure if using a set directly in the decreases clause (rather than using `s.len()`) is valid. It seems to work for fixed types but Verus panics when the set elements are a generic type:

```rust
use vstd::prelude::*; verus! {

spec fn foo<A>(s: Set<A>) -> int
    decreases s when s.finite()
{
    if s.is_empty() {
        0
    } else {
        foo(s.remove(s.choose()))
    }
}

} // verus!
```

```
thread '<unnamed>' panicked at vir/src/sst_to_air.rs:168:29:
abstract datatype should be boxed Datatype(Path(Path(Some("vstd"), ["set" :: "Set"])), [TypParam("A")], [])
thread '<unnamed>' panicked at rust_verify/src/verifier.rs:2123:29:
worker thread panicked
```

### LINKED ISSUE #1996: Verus panics on "ill-typed AIR code" when decreases clause is given an Option
This code
```
use vstd::prelude::*;

verus! {

fn main() {
    let mut x = Some(true);

    loop 
        decreases x,
    {
        match x {
            Some(b) => {
                x = None;
            }
            None => {
                break;
            }
        }
    }
}

}
```
produces invalid AIR:
```
thread '<unnamed>' (20508260) panicked at rust_verify/src/verifier.rs:833:21:
internal error: generated ill-typed AIR code: error 'in call to check_decrease_height, argument #1 has type "core!option.Option." when it should have type "Poly"' in expression '(check_decrease_height x@ decrease%init0%0 false)'
```

This seems related to but distinct from the existing issues about decreases-related panics (#1845, #1372, and #1222).

parno:
When creating a decreases check for loops that use a non-integer decreases expression, apply poly wrappings.

Fixes #1996, Fixes #1372.


jaybosamiya-ms:
Quick note about the PR description: you need to put the `#` signs for GitHub to automatically pick things up and link them such that it'll automatically close the issues as fixed when you merge this. https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

parno:
> Quick note about the PR description: you need to put the `#` signs for GitHub to automatically pick things up and link them such that it'll automatically close the issues as fixed when you merge this. https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

Ah, good point -- thanks!  I was concentrating on getting the right incantation to close two items and entirely forgot the hash sign.

PR #2001, opened 2025-11-21
Expand some of our internal documentation

parno:
1. Expand `CODE.md` with a more thorough overview of Verus's internals.
2. Add `docs/CARGO-VERUS.md`, since (I've found that) the interactions between all of the components are pretty obscure and difficult to debug.
3. Expand the documentation of SST to better match the level of documentation for AST.  This should be helpful both for us, and for projects working to translate SST to other backends, e.g., Lean.


source/CODE.md
```diff
@@ -1,33 +1,377 @@
...
+constructs.
+
+**Key files:**
+- `main.rs`: Entry point, argument parsing, output formatting
+- `driver.rs`: Orchestrates `rustc` invocations for verification and compilation
+- `verifier.rs`: Core `Verifier` struct and verification logic
+- `context.rs`: `Context<'tcx>` carrying `rustc` state
+- `config.rs`: Command-line argument definitions (`Args` struct)
+- `attributes.rs`: Verus attribute parsing (`#[verifier(...)]`)
+- `lifetime.rs`: Lifetime and borrow checking integration
```
tjhance: these files need to be renamed. They used to do lifetime-checking + the trait conflict checker on the side. Then we rearchitected the lifetime system and moved it else. So now these files only do trait conflict checking

source/CODE.md
```diff
@@ -1,33 +1,377 @@
...
+
+### Stage 3: VIR Simplification
+
+After VIR construction, several passes run:
+
+1. `ast_simplify::simplify_krate()`: Optimizations and normalizations
+2. `modes::check_crate()`: Validate Spec/Proof/Exec mode usage
+3. `well_formed::check_crate()`: Structural validation
+4. `prune::prune_krate_for_module_or_krate()`: Remove unused definitions
+5. `traits::merge_external_traits()`: Resolve trait implementations
```
tjhance: These are out of order. well_formed comes before modes and simplify I think is last. Not sure about the others.

source/CODE.md
```diff
@@ -1,33 +1,377 @@
...
+- `proof` functions and blocks
+- `Ghost<T>` and `Tracked<T>` wrapper contents
+- Specifications (requires, ensures, invariants, decreases)
+- Assertions and assumptions
+
+### Erasure Mechanism
+
+The `ErasureInfo` structure tracks what needs to be erased:
+- `ErasureModes`: Classifies each expression/statement by mode
+- `ResolvedCalls`: Records how function calls should be handled
```
tjhance: these are for lifetime-checking erasure, not compilation

source/CODE.md
```diff
@@ -1,33 +1,377 @@
...
+### Debugging Verification Issues
+
+1. Use `--log-all` to generate intermediate representations
+2. Check `.verus-log/` for VIR, AIR, and SMT-LIB files
+3. Run a single test at a time when logging
+
+### Understanding a Verification Failure
+
+1. Find the assertion in `sst_to_air.rs` that generates the failing query
+2. Check the SMT-LIB output to understand what's being asked
```
tjhance, line 364: You can also run AIR as a standalone binary.

source/CODE.md
parno: To clarify, you mean that all of the `lifetime*.rs` files now do only trait-checking, except for `lifetime_generate.rs`?  Should I rename the others as part of this PR or just update the description for now?  Also, is it accurate to say that `lifetime_generate` handles lifetime checking for ghost code, or does it encompass all code?

source/CODE.md
parno: Thanks, that's a good tip; I've added it.

source/CODE.md
parno: Okay, removed.

source/CODE.md
parno: Okay sorted (as best I can) into:
```
1. `prune::prune_krate_for_module_or_krate()`: Remove unused definitions
2. `traits::merge_external_traits()`: Resolve trait implementations
3. `well_formed::check_crate()`: Structural validation
4. `modes::check_crate()`: Validate Spec/Proof/Exec mode usage
5. `ast_simplify::simplify_krate()`: Optimizations and normalizations
```

source/CODE.md
tjhance: all files titled lifetime*.rs, including lifetime_generate, are no longer related to lifetime-checking. i think it would be a good idea to rename them. (i should have done it a while ago, but i forgot)

parno: Ah I see.  Okay, I'll rename them.  

It looks like in `lifetime.rs` still includes lifetime-related functionality. For example,  there's the `check` function that calls `mir_borrowck` and also a `check_tracked_lifetimes` that gets invoked from the verifier.  If it's not obsolete, should we leave the name alone?

parno: Or perhaps you just meant the `lifetime_*.rs` files?

tjhance: OK, it looks like `crate::lifetime::check` is used for something related to lifetimes. It's called from the CompilerCallbacksEraseMacro. So it's lifetime checking for exec code only, I think.

However, I am certain that `emit_check_tracked_lifetimes`, despite the name, is now only used for trait conflict checking.

tjhance: Sorry to make your documentation life difficult >_< I left this file in a bad shape that doesn't make any sense.

tjhance: I think that the `check` function should be moved to driver.rs, and everything else in lifetime.rs, lifetime_ast.rs, lifetime_emit.rs, and lifetime_generate.rs should be renamed to trait_conflicts or somesuch.

parno: Okay, I moved `check` and the `rustc_driver` callbacks that invoke it into `driver.rs` (814e6ccac5d86b03) and then renamed the `lifetime*.rs` files to `trait_check*.rs` (56b8fb162c).  There are still function and constant names that refer to lifetimes, but I thought I'd leave that for a separate PR.

PR #2006, opened 2025-11-23
Add lemma_seq_two_subranges_index and group_seq_lemmas_expensive

Chris-Hawblitzel:
This pull request adds one new inexpensive lemma `lemma_seq_two_subranges_index` to the default `Seq` broadcast group and declares a new broadcast group `group_seq_lemmas_expensive ` (not part of the default) to hold various new `Seq` lemmas about `push`, `update`, `subrange`, and `add`.

## Background and performance measurements

The file `seq.rs` includes axioms for basic operations on Seq like `push`, `update`, `subrange`, and `add`.  For example, the key axiom for `subrange` relates an element in a subrange to the element in the original sequence, triggering on `s.subrange(j, k)[i]`:

```
pub broadcast axiom fn axiom_seq_subrange_index<A>(s: Seq<A>, j: int, k: int, i: int)
    requires
        0 <= j <= k <= s.len(),
        0 <= i < k - j,
    ensures
        #[trigger] s.subrange(j, k)[i] == s[i + j],
;
```

The trigger on `s.subrange(j, k)[i]` is natural, since someone using `s.subrange(j, k)` probably wants to know something about elements of `s.subrange(j, k)`.  However, we could also reverse the direction of the triggering by using `s[i]` as the trigger for `i` and then looking up `s.subrange(j, k)[i - j]` in response:

```
pub broadcast proof fn lemma_seq_subrange_index_alt<A>(s: Seq<A>, j: int, k: int, i: int)
    requires
        0 <= j <= k <= s.len(),
        0 <= i - j < k - j,
    ensures
        (#[trigger] s.subrange(j, k))[i - j] == #[trigger] s[i],
{
    broadcast use axiom_seq_subrange_index;
}
```

Right now, Verus only declares the forward direction version (the trigger on `s.subrange(j, k)[i]`), not the reverse direction version.  There's a reason for this: adding in all the reverse direction lemmas for `push`, `update`, `subrange`, and `add` slows down the SMT solving.  The "original" and "expensive" columns of this table show the smt-run times for Verus as-is and for Verus with all the reverse direction lemmas for `push`, `update`, `subrange`, and `add` added to the default broadcast group in `Seq`:

| original	| two_subranges	| subrange_alt	| expensive	| benchmark |
|	---	|	---	|	---	|	---	| --- |
|	1938	|	1932	|	1930	|	1946	| node-replication |
|	5706	|	5690	|	18403	|	10091	| ironsht |
|	35508	|	34587	|	62521	|	279594	| capybaraKV |
|	3070	|	2508	|	4195	|	4687	| vest
|	68577	|	66193	|	66720	|	68105	| vest-dsl |
|	87363	|	86102	|	86592	|	459029*	| mimalloc |
|	197489	|	194964	|	194507	|	160943	| anvil-vdeployment_admission_controller |

(The * for mimalloc indicates two timeouts in mimalloc with rlimit 100.)  Adding all of these lemmas slows down many of the benchmarks dramatically.  Even just adding the one reverse lemma for `subrange` alone (subrange_alt, showing adding just `lemma_seq_subrange_index_alt` to the default `Seq` broadcast) results in a significant slowdown.  (I'm also slightly worried about the forward and reverse lemmas introducing a matching loop involving `i + j - j + j - j ...`, although I didn't observe any such loops in the profiler.  It may be that Z3 is simplifying the arithmetic expressions early enough to avoid matching loops.)

By contrast, the following more limited lemma results in no appreciable slowdown (column two_subranges in the table above):

```
// Less expensive, more limited alternative to lemma_seq_subrange_index_alt
pub broadcast proof fn lemma_seq_two_subranges_index<A>(s: Seq<A>, j: int, k1: int, k2: int, i: int)
    requires
        0 <= j <= k1 <= s.len(),
        0 <= j <= k2 <= s.len(),
        0 <= i < k1 - j,
        0 <= i < k2 - j,
    ensures
        #[trigger] s.subrange(j, k1)[i] == (#[trigger] s.subrange(j, k2))[i],
{
    broadcast use axiom_seq_subrange_index;
}
```

## Example where reverse direction helps

For most part, the forward triggers work reasonably well.  Here's a typical example of a proof that succeeds naturally with just the forward triggers:

```
fn test_forall(src: &Vec<u32>, dst: &mut Vec<u32>, lo: usize, hi: usize)
    requires
        lo <= hi,
        hi <= src.len(),
        hi <= old(dst).len(),
    ensures
        src@.subrange(lo as int, hi as int) == dst@.subrange(lo as int, hi as int),
{
    for n in lo..hi
        invariant
            lo <= hi,
            hi <= src.len(),
            hi <= dst.len(),
            forall|i: int| lo <= i < n ==> dst[i] == src[i],
    {
        dst[n] = src[n];
    }
}
```

However, a seemingly equivalent version of this using extensional equality of subranges fails:

```
fn test_subrange(src: &Vec<u32>, dst: &mut Vec<u32>, lo: usize, hi: usize)
    requires
        lo <= hi,
        hi <= src.len(),
        hi <= old(dst).len(),
    ensures
        src@.subrange(lo as int, hi as int) == dst@.subrange(lo as int, hi as int),
{
    for n in lo..hi
        invariant
            lo <= hi,
            hi <= src.len(),
            hi <= dst.len(),
            src@.subrange(lo as int, n as int) =~= dst@.subrange(lo as int, n as int),
    {
        dst[n] = src[n];
    }
}
```

Making it work with the forward triggers alone requires extra assertions to manually trigger the reverse direction:

```
fn test_subrange(src: &Vec<u32>, dst: &mut Vec<u32>, lo: usize, hi: usize)
    requires
        lo <= hi,
        hi <= src.len(),
        hi <= old(dst).len(),
    ensures
        src@.subrange(lo as int, hi as int) == dst@.subrange(lo as int, hi as int),
{
    for n in lo..hi
        invariant
            lo <= hi,
            hi <= src.len(),
            hi <= dst.len(),
            src@.subrange(lo as int, n as int) =~= dst@.subrange(lo as int, n as int),
    {
        dst[n] = src[n];
        assert forall|i: int| lo <= i + lo < n implies #[trigger] src@.subrange(lo as int, n + 1)[i] == dst@.subrange(lo as int, n + 1)[i] by {
            assert(src@.subrange(lo as int, n + 1)[i] == src@.subrange(lo as int, n as int)[i]);
        }
    }
}
```

It's not obvious at first glance why the assertion above is necessary, but this smaller example demonstrates how an equality can fail without the reverse trigger:

```
proof fn test_seq(s: Seq<int>, d1: Seq<int>)
    requires
        s.len() == 10,
        d1.len() == 10,
{
    let d2 = d1.update(4, s[4]);
    assume(s.subrange(0, 4) == d1.subrange(0, 4));
//    assert(s.subrange(0, 5)[3] == s.subrange(0, 4)[3]);
    assert(s.subrange(0, 5)[3] == d2.subrange(0, 5)[3]); // FAILS without uncommenting previous assertion
}
```

The issue is that when we try to compare `d2.subrange(0, 5)[3]` to `s.subrange(0, 5)[3]`, we look up element `3` in `s.subrange(0, 5)` but the assumptions we need are actually in a different subrange `s.subrange(0, 4)`.  The forward trigger takes us from `s.subrange(0, 5)[3]` to `s[3]`, but the reverse trigger is need to go back from `s[3]` to `s.subrange(0, 4)[3]`.

Adding the reverse trigger lemmas fixes this, but (as the experiments above show) is expensive.  As a less expensive alternative, this pull request adds `lemma_seq_two_subranges_index`, which also fixes the problem and makes the loop above verify with no extra asserts.

## Trigger weights (the original goal of the pull request, which I abandoned)

I had hoped that using Z3 quantifier trigger weights would provide a middle ground here between no reverse trigger lemmas and full-blown reverse trigger lemmas.  The hope was that weights would allow the reverse triggered quantifiers but limit their instantiations.  Dafny, for example, uses weights on some of its quantifiers for `take` and `drop`, which are the Dafny analogues to Verus subrange (see the uses of `:weight` in https://github.com/dafny-lang/dafny/blob/master/Source/DafnyCore/DafnyPrelude.bpl ).

Weights are designed to cut off excessive quantifier instantiations.  You can think of Z3 as having a budget for quantifier instantiations set by a flag called `eager_threshold` (which we set to 100).  The weight of a quantifier counts towards this budget, so that a weight of 25 would reach a threshold of 100 in roughly 4 instantiations.  I had hoped that simply by putting a high weight on these reverse triggered quantifiers, we could allow a few instantiations without significantly slowing down verification.

However, after some experimentation, I became disenchanted with weights.  The problem is that whenever the threshold is reached, verification becomes unstable and starts to appear to fail in arbitrary ways.  Furthermore, it's not just the quantifiers with explicit weights that suffer in this situation:  all quantifiers are at risk of hitting the threshold once the high-weight quantifiers have used up most of the budget.  For example, I tried adding just one extra lemma `lemma_seq_subrange_index_alt`, with a weight of 50 (exactly half the threshold of 100), and a huge number of previously succeeding proofs started failing because two instantiations of the extra lemma used up all the budget, leaving nothing for all the rest of the quantifiers we normally rely on.  I tried setting the weight for the extra lemma to 52 or 53 (more than half the threshold of 100), and this behaved better than 50, leaving existing proofs working, but it behaved very unstably for any proof relying on the extra lemma:  it would work fine for simple examples, but in more complicated examples, seemingly obvious assertions would start and stop verifying without any clear reason.  Because of this, I gave up on using weights for the reverse triggered Seq quantifiers.  Given all our efforts to reduce instability, I don't want to add another source of instability.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -16,29 +16,29 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`fmt`** (macOS ARM64)
+1. **`fmt`** (linux)
    - Validates Rust code formatting with `rustfmt`
    - Validates `vstd` formatting with `verusfmt`
 
-2. **`test-and-release-macos`** (macOS ARM64)
-   - Runs full test suite across multiple feature configurations:
-     - Default features (full test suite)
-     - `record-history` (limited tests)
-     - `no-std` and `no-alloc` (vstd variants)
-     - `singular` (computer algebra system integration)
-     - `cvc5` (alternate SMT solver)
-   - Builds release binary for ARM64 macOS
-   - Generates vstd API documentation
-   - Uploads `verus-arm64-macos.zip` artifact
+2. **`full-test`** (macOS ARM64)
+   - Runs full test suite.
+
+3. **`basic-test`** (macOs x64, Windows x64, and Linux x64)
+   - Runs basic tests only
+
+3. **`smoke-test`** (macOs ARM64)
+   - Checks that `verus` builds with esoteric configurations
+   - Runs minimal tests relevant to the configuration
+
+4. **`build-docs`** (linux)
+   - Builds the `verusdoc` artifact
    - Uploads `verusdoc` artifact for documentation deployment
 
-3. **`smoke-test-and-release-...`**
-    - Jobs of each of macOS x86, Windows x64, and Linux x64
-   - Runs basic smoke tests only
-   - Builds a release binary for the target
-   - Uploads the `verus-....zip` artifact
+5. **`build-release`** (macOS ARM64, macOs x64, Windows x64, and Linux x64)
+   - Builds release binary artifacts for every supported platform
+   - Uploads `verus-<arch>-<os>.zip` artifacts
 
-4. **`release`** (Ubuntu)
+6. **`release`** (linux)
    - **Only runs on push to `main`** (not PRs)
    - Downloads all platform artifacts
    - Extracts version information from `version.txt`
@@ -48,9 +48,6 @@ and publishing for the Verus project.
    - Uploads all platform binaries to the rolling release
    - Deletes old rolling release tags and assets
    - Publishes the updated rolling release
-
-5. **`docs`** (macOS ARM64)
-   - Validates documentation by running `rustdoc` and `verusdoc` disallowing warnings
 
 **Output:**
 - Continuous binary distribution via the Rolling Release
@@ -134,7 +131,7 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`build`** (Ubuntu)
+1. **`build`** (Linux)
    - Sets up mdbook for building documentation
    - Builds user guide: `source/docs/guide` → `/_site/guide`
    - Builds state machines guide: `source/docs/state_machines` → `/_site/state_machines`
@@ -143,7 +140,7 @@ and publishing for the Verus project.
    - Builds Verus landing page using Jekyll
    - Uploads combined site artifact
 
-2. **`deploy`** (Ubuntu)
+2. **`deploy`** (Linux)
    - Deploys the built site to GitHub Pages
    - Makes documentation available at the GitHub Pages URL
 
@@ -170,7 +167,7 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`bump-crate-versions`** (Ubuntu)
+1. **`bump-crate-versions`** (linux)
    - Runs version bump tool: `source/tools/bump_crate_versions` with update command
    - Commits version changes if needed
    - Pushes changes to `main`
@@ -200,8 +197,8 @@ crate-updates.yml (weekly/manual) → Updates crate versions → Publishes to cr
 ## Platform Support
 
 All workflows build and test Verus on:
-- **Linux**: `x86_64` (Ubuntu 22.04)
-- **macOS**: ARM64 (macOS 14) and `x86_64` (macOS 15)
+- **Linux**: `x86_64` (ubuntu-latest)
+- **macOS**: ARM64 (macOS latest) and `x86_64` (macOS 15)
 - **Windows**: `x86_64` (latest)
 
 ARM64 macOS receives full testing; other platforms run smoke tests for efficiency.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,67 +14,175 @@ permissions:
   contents: write
 
 jobs:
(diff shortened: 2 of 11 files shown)


Chris-Hawblitzel:
> Can we do something to make `group_seq_lemmas_expensive` more discoverable? E.g., adding it to the Guide somewhere? Otherwise, we'll have to rely on oral dissemination, which doesn't seem very scalable.

Yes, I would like to take a pass at adding more about SMT solving and SMT failures to the guide in general, and I think this could fit in as an example.

PR #2020, opened 2025-12-01
Basic support for dyn

Chris-Hawblitzel:
This implements the type `dyn T` for trait T, following the design from https://github.com/verus-lang/verus/discussions/1047 (made much easier thanks to https://github.com/verus-lang/verus/pull/1529 ).  In particular, it extends the existing Rust "dyn compatibility" with rules for proof functions and spec functions, as described in https://github.com/verus-lang/verus/discussions/1047:
- proof functions must either be explicitly non-dispatchable or have a `tracked` self
- spec functions must either be explicitly non-dispatchable or have no ensures clauses

This pull request doesn't support additional auto traits like `dyn T + Send + Sync`; I want to leave this for later in order to look into our handling of Send, Sync, etc. first.  (It's important for trait dispatch that `dyn T + Send + Sync` have a different type id from `dyn T`, for example.)

`dyn T` is implemented at the AIR level as `Poly`, with a new unsized decoration `$dyn`, a type id `DYN%T`, and coercion `to_dyn%T` from values to `dyn T` values.  For example, for:

```
trait T {
    spec fn f(&self, i: int) -> int;
}
impl T for u8 {
    spec fn f(&self, i: int) -> int { 3 }
}
```

this AIR is generated if `dyn T` is reachable after pruning:

```
(declare-const DYN%scratch!T. Type)
(declare-fun to_dyn%scratch!T. (Dcr Type Poly) Poly)

(axiom (forall ((self! Poly) (i! Poly)) (!
   (=
    (T.f.? $dyn DYN%T. (to_dyn%T. $ (UINT 8) self!) i!)
    (T.f.? $ (UINT 8) self! i!)
   )
   :pattern ((T.f.? $dyn DYN%T. (to_dyn%T. $ (UINT 8) self!) i!))
)))
```

### DIFF
--- a/source/rust_verify_test/tests/traits_dyn.rs
+++ b/source/rust_verify_test/tests/traits_dyn.rs
@@ -345,6 +345,56 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] dyn_cycle4 verus_code! {
+        use vstd::prelude::*;
+        trait T {
+            spec fn f(&self) -> S;
+        }
+        struct S(Box<dyn T>);
+
+        proof fn p(s: &S)
+            ensures
+                false,
+            decreases s
+        {
+            p(&(s.0).f())
+        }
+
+        proof fn test()
+            ensures
+                false,
+        {
+            p(&arbitrary());
+        }
+    } => Err(err) => assert_vir_error_msg(err, "found a cyclic self-reference")
+}
+
+test_verify_one_file! {
+    #[test] dyn_cycle5 verus_code! {
+        use vstd::prelude::*;
+        trait T<A> {
+            spec fn f(&self) -> A;
+        }
+        struct S(Box<dyn T<S>>);
+
+        proof fn p(s: &S)
+            ensures
+                false,
+            decreases s
+        {
+            p(&(s.0).f())
+        }
+
+        proof fn test()
+            ensures
+                false,
+        {
+            p(&arbitrary());
+        }
+    } => Err(err) => assert_vir_error_msg(err, "non-positive position")
+}
+
+test_verify_one_file! {
     #[test] dyn_unsupported1 verus_code! {
         trait T {}
         fn test(d: &(dyn T + Send)) {


tjhance:
What's your plan for dealing with the coherence issue with blanket impls? (rust issue 57893)? Will the trait-conflict-checker catch any issues?

Chris-Hawblitzel:
> What's your plan for dealing with the coherence issue with blanket impls? (rust issue 57893)? Will the trait-conflict-checker catch any issues?

Right now, the pull request prohibits this up front.  Any trait T with a blanket impl for an unsized type parameter gets marked `DynCompatible::RejectUnsizedBlanketImpl` and is rejected if the user tries to use `dyn T`, with the message "Verus disallows dyn for trait T because it has an unsized blanket impl (see https://github.com/rust-lang/rust/issues/57893 )".

tjhance:
I think those aren't the only problematic cases, e.g., we have this example from [this comment](https://github.com/rust-lang/rust/issues/57893#issuecomment-528993546):

```rust
trait TraitA { type Item: ?Sized; }

trait TraitB<T> { }

impl<X: TraitA> TraitB<X> for X::Item { }

impl TraitA for () {
    type Item = dyn TraitB<()>;
}
```

Chris-Hawblitzel:
> ```rust
> trait TraitA { type Item: ?Sized; }
> 
> trait TraitB<T> { }
> 
> impl<X: TraitA> TraitB<X> for X::Item { }
> 
> impl TraitA for () {
>     type Item = dyn TraitB<()>;
> }
> ```

Ok, I updated trait_conflicts.rs to check for https://github.com/rust-lang/rust/issues/57893 in general.  The obvious blanket impl still gets checked by well_formed.rs up front, but more obscure cases get caught in trait_conflicts.rs.  This also includes some improvements to make trait_conflicts.rs more precise about sized structs.

source/vir/src/recursive_types.rs
```diff
@@ -103,6 +103,13 @@ fn check_well_founded_typ(
             }
             true
         }
+        TypX::Dyn(..) => {
+            // This does not support decreases; it is a base case.
+            // For spec mode, this is like a collection of all the spec fns in the trait
+            // (the proof and exec fns are not callable from spec mode),
+            // so spec mode default values are formed like spec_fn default values.
+            true
```
tjhance, line 111: Does this really not depend on the type args of the trait?

Chris-Hawblitzel: I believe that in the other recursive_types checks, such as check_positive_uses and build_datatype_graph,
it is necessary to follow the type args of TypX::Dyn, but that it is not necessary in check_well_founded/check_well_founded_typ. The reason I believe this is that check_well_founded is a very narrow, specialized check that just makes sure that it's safe to access fields of structs/enums and conclude that field is smaller than the overall datatype, for the purposes of decreases checking. For example, we need to prevent things like this:

```
struct S(Box<S>);

proof fn rec(s: S)
    ensures false
    decreases s
{
    rec(*s.0);
}

proof fn bad()
    ensures false
{
    rec(arbitrary());
}
```

This example is unsafe because there is no finite-size value of S that `arbitrary()` can produce.

So check_well_founded just walks through fields of structs/enums, ensuring that there is always a way to construct a finite `arbitrary()` struct/enum value. For this, dyn is a base case even if the dyn type arguments mention the struct/enum recursively, because the arbitrary dyn value doesn't need to recursively contain smaller struct/enum values. Instead, it simply contains `arbitrary()` expressions that evaluate to struct/enum values when the dyn spec functions are called. (More formally, this is the way spec functions in the Verus arXiv paper work: the defaults_to value for a spec function is a lambda whose body contains the equivalent of `arbitrary()`, and is thus a base case for defaults_to.) The `arbitrary()` expressions in the lambdas may evaluate to larger struct/enum values that the outer struct/enum, but this is ok, because decreases clauses don't guarantee that dyn functions return smaller values than the outer struct/enum.

PR #2022, opened 2025-12-01
ci: make release conditional on being on the root (non-fork) repo

bsdinis:
This change to the CI workflow makes running release actions conditional on the action being run on the main `verus-lang/verus` repo. This allows CI's on forks to pass.

The rationale here is that if any fork actually wants to run release actions they need to maintain a patch anyway to release to their endpoints, so this shouldn't be a bad thing to add.


bsdinis:
You can check [my fork](https://github.com/bsdinis/verus/tree/main) to see how it runs

Issue #1382, opened 2025-01-10
Binary release incompatible with glibc version


hayley-leblanc:
Not sure if this is really an issue or something that you can/want to fix, but the version of glibc required by the GitHub Linux release binary is not supported on some current/recent Linux distros. I use Pop! OS 22.04, which is based on Ubuntu 22.04, and I get this error trying to run the Verus binary from the release:
```
./verus: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by ./verus)
```
`ldd --version` outputs `ldd (Ubuntu GLIBC 2.35-0ubuntu3.8) 2.35`. 

I know 22.04 isn't the newest Ubuntu version, but it is the latest stable Pop! OS version. I haven't tested this on other distros, but based on a bit of Googling it looks like the latest stable version of Debian is also behind glibc 2.38, so I suspect it won't work there either.

I initially ran into this issue trying to use verus-analyzer, which pulls in the current Verus binary release by default.

jaybosamiya-ms:
I suspect this is due to the use of `ubuntu-latest` here: https://github.com/verus-lang/verus/blob/08df1651967ac1d796c5da05681eebe618f38c9b/.github/workflows/ci.yml#L216-L217

Until _very_ recently, `ubuntu-latest` meant Ubuntu 22.04 on GitHub Actions. However, GitHub is slowly rolling out an upgrade from 22.04 to 24.04 for `ubuntu-latest` (see https://github.com/actions/runner-images/issues/10636), and it seems like Verus's upgrade has now finished its place in the roll-out.

Since until some days ago, our CI was using 22.04, I don't see it as being a major issue to simply change the above to `ubuntu-22.04` to force it to use 22.04, thereby supporting more distros with the release.  GitHub will have support for `ubuntu-22.04` as a runner for ~2 more years, by which time I expect most distros to have moved on, so we can bump it to 24.04 then.

A heavier-handed alternative is to build our binaries as static binaries (possibly using musl) which might allow us to stay on ubuntu-latest while supporting even _much_ older Linux distros. I suspect this is unnecessary, and simply setting CI to `ubuntu-22.04` rather than `ubuntu-latest` is probably the right move.

As a point of comparison, verusfmt uses 20.04 in its release process, in order to maintain support for a large set of distros. I'll probably have to change that to 22.04 when GitHub Actions phases out its 20.04 runners, but generally speaking, release builds using an older version works great as a way to support things.

[...]

PR #2024, opened 2025-12-03
ci: refactor for maintainability and optimize for latency

bsdinis:
Based on today's conversation around CI, this PR does the following:
- makes the ci a bit more maintainable (fully exploiting the strategy.matrix and separating building and testing)
- moves to cargo-nextest
- uses dtolnay's rust toolchain action (which seems to be the default + is a bit faster)
- compiles tests in release mode (so they run faster) -- reusing the artifacts for release when appropriate

Moving to compiling tests in release mode was a win (building and testing and building the release for MacOs moved from [28m + 4m](https://github.com/verus-lang/verus/actions/runs/19868369188/job/56937145250) to [5m + 13m](https://github.com/bsdinis/verus/actions/runs/19875116544/job/56960520568).

Sadly, Github CI is not exactly a performance isolated platform, so there is a lot of variance going around. In particular:
- Compiling macos_x86-64 in release mode takes a while. 
- nextest was not a clear win. 

I still think nextest it's worthwhile to put it in, as it shows the time each test took (for instance, I found out that the cargo tests are by far the slowest -- almost all of them take between 30s and >1min).

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -16,29 +16,29 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`fmt`** (macOS ARM64)
+1. **`fmt`** (linux)
    - Validates Rust code formatting with `rustfmt`
    - Validates `vstd` formatting with `verusfmt`
 
-2. **`test-and-release-macos`** (macOS ARM64)
-   - Runs full test suite across multiple feature configurations:
-     - Default features (full test suite)
-     - `record-history` (limited tests)
-     - `no-std` and `no-alloc` (vstd variants)
-     - `singular` (computer algebra system integration)
-     - `cvc5` (alternate SMT solver)
-   - Builds release binary for ARM64 macOS
-   - Generates vstd API documentation
-   - Uploads `verus-arm64-macos.zip` artifact
+2. **`full-test`** (macOS ARM64)
+   - Runs full test suite.
+
+3. **`basic-test`** (macOs x64, Windows x64, and Linux x64)
+   - Runs basic tests only
+
+3. **`smoke-test`** (macOs ARM64)
+   - Checks that `verus` builds with esoteric configurations
+   - Runs minimal tests relevant to the configuration
+
+4. **`build-docs`** (linux)
+   - Builds the `verusdoc` artifact
    - Uploads `verusdoc` artifact for documentation deployment
 
-3. **`smoke-test-and-release-...`**
-    - Jobs of each of macOS x86, Windows x64, and Linux x64
-   - Runs basic smoke tests only
-   - Builds a release binary for the target
-   - Uploads the `verus-....zip` artifact
+5. **`build-release`** (macOS ARM64, macOs x64, Windows x64, and Linux x64)
+   - Builds release binary artifacts for every supported platform
+   - Uploads `verus-<arch>-<os>.zip` artifacts
 
-4. **`release`** (Ubuntu)
+6. **`release`** (linux)
    - **Only runs on push to `main`** (not PRs)
    - Downloads all platform artifacts
    - Extracts version information from `version.txt`
@@ -48,9 +48,6 @@ and publishing for the Verus project.
    - Uploads all platform binaries to the rolling release
    - Deletes old rolling release tags and assets
    - Publishes the updated rolling release
-
-5. **`docs`** (macOS ARM64)
-   - Validates documentation by running `rustdoc` and `verusdoc` disallowing warnings
 
 **Output:**
 - Continuous binary distribution via the Rolling Release
@@ -134,7 +131,7 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`build`** (Ubuntu)
+1. **`build`** (Linux)
    - Sets up mdbook for building documentation
    - Builds user guide: `source/docs/guide` → `/_site/guide`
    - Builds state machines guide: `source/docs/state_machines` → `/_site/state_machines`
@@ -143,7 +140,7 @@ and publishing for the Verus project.
    - Builds Verus landing page using Jekyll
    - Uploads combined site artifact
 
-2. **`deploy`** (Ubuntu)
+2. **`deploy`** (Linux)
    - Deploys the built site to GitHub Pages
    - Makes documentation available at the GitHub Pages URL
 
@@ -170,7 +167,7 @@ and publishing for the Verus project.
 
 **Jobs:**
 
-1. **`bump-crate-versions`** (Ubuntu)
+1. **`bump-crate-versions`** (linux)
    - Runs version bump tool: `source/tools/bump_crate_versions` with update command
    - Commits version changes if needed
    - Pushes changes to `main`
@@ -200,8 +197,8 @@ crate-updates.yml (weekly/manual) → Updates crate versions → Publishes to cr
 ## Platform Support
 
 All workflows build and test Verus on:
-- **Linux**: `x86_64` (Ubuntu 22.04)
-- **macOS**: ARM64 (macOS 14) and `x86_64` (macOS 15)
+- **Linux**: `x86_64` (ubuntu-latest)
+- **macOS**: ARM64 (macOS latest) and `x86_64` (macOS 15)
 - **Windows**: `x86_64` (latest)
 
 ARM64 macOS receives full testing; other platforms run smoke tests for efficiency.


tjhance:
> compiles tests in release mode (so they run faster) -- reusing the artifacts for release when appropriate

Do you mean the tests themselves, or the rust_verify binary? The tests themselves do very little work other than spawning a process and checking the output. I have a hard time believing you'd get a big win from building the tests in release mode.

Regarding the rust_verify binary itself, building in release mode can definitely speed it up, but there are also some checks that only run when `debug_assertions` is enabled. We should make sure these are enabled for the tests.

bsdinis:
> Regarding the rust_verify binary itself, building in release mode can definitely speed it up, but there are also some checks that only run when `debug_assertions` is enabled. We should make sure these are enabled for the tests.

Hmm, good point. I'll look into this part. OTOH I'm not sure of a way of running tests with the release binary and the debug tests. But if there is one I'll find it. I think the general advice on it is that it is still worth it to compile tests in release though (modulo the debug assertions bit).

bsdinis:
@tjhance related to just building `rust_verify` in release and the tests in `debug`. Is this actually possible with `vargo`. I think it might recompile `rust_verify` if the flags are different

tjhance:
I don't think you need to bend over backwards to build the tests with a different profile. I was just wondering what you meant when you said you changed it to "compile tests in release mode".

bsdinis:
> I don't think you need to bend over backwards to build the tests with a different profile. I was just wondering what you meant when you said you changed it to "compile tests in release mode".

So this diff does build everything in release mode (my current tinkering is to enable debug asserts when running the tests).

There is some evidence that this can help:
1. We run _a lot_ of tests. Having `rust_verify` built in release mode helps run them faster.
2. We can hopefully reuse the release build `rust_verify` as the release artifact
3. I was trying to follow your suggestion of trying to have `rust_verify` in release and the tests in debug, but I'm not sure it's trivial

bsdinis:
Okay, after much tinkering, here's what I got to:
- run the full-tests in macos with release and debug assertions. This shaves between 10-20 mins
- wall clock time as reported by github runs seems to be similarly lower (between 10-20 mins)

Ways to move forward:
- debug the performance of the cargo tests; I struggle to understand how they can end up (in some cases) at around 6mins
- failing that, consider not running the cargo tests on the mac runner but do them on a linux runner. This way we can parallelize with the full tests
- introduce an action to setup the environment. we download z3 for every job, seems excessive (but is not the biggest problem)
- failing all, it seems that the next step would be to get a self-hosted runner for the main tests

@tjhance does the way the test run make sense to you?

parno:
> debug the performance of the cargo tests; I struggle to understand how they can end up (in some cases) at around 6mins

From what I've seen the slowness of those tests seems to be due to a combination of (a) downloading and compiling the various package dependencies, and (b) for the crates that get verified in the tests, verifying `vstd` (since it shows up as a dependency and it's `Cargo.toml` file is marked for verification).  I'm not sure why it seems to be so variable on the GitHub runners; perhaps it's a matter of network contention?

bsdinis:
> > debug the performance of the cargo tests; I struggle to understand how they can end up (in some cases) at around 6mins
> 
> From what I've seen the slowness of those tests seems to be due to a combination of (a) downloading and compiling the various package dependencies, and (b) for the crates that get verified in the tests, verifying `vstd` (since it shows up as a dependency and it's `Cargo.toml` file is marked for verification). I'm not sure why it seems to be so variable on the GitHub runners; perhaps it's a matter of network contention?

Great insight on the cargo tests! 

In terms of why github runners are variable it's just that they are (from my understanding) almost akin to spot instances (very low specs). I think we should definitely work to optimize these tests (but maybe in a different PR?)

parno:
> I think we should definitely work to optimize these tests (but maybe in a different PR?)

Yes, I think it would be fine to do that in another PR.

tjhance:
It all makes sense to me. Sounds like this is going to be a big improvement.

bsdinis:
@parno not sure if something is missing or if I should merge it

parno:
Sorry, I'm hoping to review it soon; the end of the semester is quite busy, unfortunately.

bsdinis:
No worries and no rush!

.github/workflows/ci.yml
```diff
@@ -14,81 +14,185 @@ permissions:
...
         run: |
-          . "$HOME/.cargo/env"
           . ../tools/activate
           vargo fmt -- --check
+
       - name: check cargo fmt for vargo
         working-directory: ./tools/vargo
+        run: cargo fmt -- --check
+
+  # check if it builds and runs on our main supported platforms
```
parno: This comment seems out of date now, right?  This is only running the tests for one platform (macos)

.github/workflows/ci.yml
```diff
@@ -102,225 +206,122 @@ jobs:
...
-          cd ..; zip -r verus-x86-linux.zip ./verus-x86-linux
-      - name: run verus release
-        working-directory: ./source
+
+      - name: create archive
         run: |
-          ls -Al ./target-verus/release
-          ./target-verus/release/verus ../examples/test.rs
-      - name: upload verus release artifact
+          ./target-verus/release/verus --version --output-json > ./target-verus/release/version.json
```
parno, line 308: Any idea why this line was previously commented out with a TODO?

.github/workflows/ci.yml
```diff
@@ -102,225 +206,122 @@ jobs:
...
+          cp -R ./target-verus/release ../${{ matrix.release_name }}
+          cd ..;
+          if [ ${{ matrix.os }} == "windows-latest" ]; then
+            7z a ${{ matrix.release_name }}.zip ./${{ matrix.release_name }}
```
parno, line 312: Do we know that 7z will be available on a Windows runner?

.github/workflows/crate-updates.yml
```diff
@@ -19,8 +19,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Rust toolchain
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: setup rust
```
parno: Sorry, this is very minor, but can we keep this capitalized ("Setup Rust") to be consistent with the other names?

parno (commented):
Thanks very much for putting this together.  It seems like it both cleans up the process and should make it run more efficiently too.

I notice that some of the jobs that used to run on macos (e.g., `docs`) now run on Linux runners.  Is there an advantage to preferring Linux?  

Also, would you mind editing the CI section of the `.github/workflows/README.md` to reflect the changes?  It will probably be shorter and simpler as a result of your changes.

.github/workflows/ci.yml
bsdinis: no, not at all

bsdinis: but seems like a good idea

.github/workflows/ci.yml
bsdinis: yes: https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md

.github/workflows/crate-updates.yml
bsdinis: good point!

.github/workflows/ci.yml
bsdinis: correct, fixed it.

bsdinis:
> I notice that some of the jobs that used to run on macos (e.g., `docs`) now run on Linux runners. Is there an advantage to preferring Linux?

Nothing really confirmed. However, I assume there are more linux runners than macos ones and hopefully we have better luck with scheduling these jobs. Since it's just verifying the state of the repo / building platform independent artifacts, figured it would be a better choice. 

> Also, would you mind editing the CI section of the `.github/workflows/README.md` to reflect the changes? It will probably be shorter and simpler as a result of your changes.

done

bsdinis:
Some more notes on macos vs. linux: https://blog.pradumnasaraf.dev/macos-vs-ubuntu (not very substantiated).
According to Github docs:
 - Linux comes with 4 vCPUs, 16GB ram
 - MaCOS arm comes with 3 vCPU (M1), 7GB ram.
 
 This might all be irrelevant and the M1 performance might compensate for things.

parno:
Interesting, thanks for the data and links.  If something fails on the Linux runner, hopefully it will still reproduce for those of us on Mac.  And I optimistically hope that we'll spend far more time running CI than dealing with problems it turns up, so let's optimize for CI performance.

Thanks also for updating the docs!

parno (approved):

bsdinis:
Yeah, I'd be extremely surprised if there was any delta in behaviour on `fmt` and `docs` between MacOs and Linux.

tangruize:
Changing from ubuntu 22.04 to ubuntu-latest causes the latest binary release to be incompatible with glibc version on older ubuntu releases. This causes [verus-analyzer](https://github.com/verus-lang/verus-analyzer) to fail since it downloads the latest verus release. A similar issue: https://github.com/verus-lang/verus/issues/1382

bsdinis:
Thanks for catching this. Ill look more into it tomorrow. Im ondering mostly if its better to keep this as is or to fix verus-analyzer to work with a more recent glibc, since I think its fine to support only the latest ubuntu LTS

parno:
If it was just a matter of updating verus-analyzer's runner to ubuntu-latest, that seems easy enough.  But if people are using older versions of Ubuntu (like @tangruize appears to be?) or other OSes like in #1382, then I think it's worth trying to accomodate that, especially since it seems like a simple change to our CI/release process.

bsdinis:
@tangruize opened #2051 which should fix this in the interim

PR #2025, opened 2025-12-03
Add the "real" type for real-number arithmetic

Chris-Hawblitzel:
This adds support for real numbers in specs, using a new `real` type (similar to the existing `int` and `nat` types) and `+`, `-`, `*`, `/`, and comparison operations on reals, as well as real literals.  As with integers, real arithmetic is linear by default and is nonlinear only with assert-by-nonlinear_arith.

Examples:

```
proof fn test(i: int, x: real, y: real) {
    assert(x == 0.2real ==> x / 2.0 == 0.1real);
    assert(x > 0.0 ==> x / 3.0 > 0.0);
    assert(x <= 1.0 ==> x / 3.0 < 1.0);
    let z: real = x + 1.1;
    assert(z > x);
    let q: real = 5u8 as real;
    assert(q == 5real);
    assert((i + 1) as real == i as real + 1.0);
    assert(x >= 0.0 ==> x / 3.0 > 0.0); // FAILS
}

proof fn test_nonlinear(a: real, b: real, q: real) {
    assert(a >= 0.0 && b >= 0.0 && a * a + b * b == (a * b + 1.0) * q ==> q >= 0.0)
        by (nonlinear_arith);
}
```

### DIFF
--- a/examples/experimental_new_mut_ref/README.md
+++ b/examples/experimental_new_mut_ref/README.md
@@ -1,0 +1,1 @@
+Examples that work under the experimental `-V new-mut-ref` feature.
--- a/examples/experimental_new_mut_ref/hash_table_entry.rs
+++ b/examples/experimental_new_mut_ref/hash_table_entry.rs
@@ -1,0 +1,173 @@
+use vstd::prelude::*;
+use vstd::std_specs::hash::*;
+
+use core::borrow::Borrow;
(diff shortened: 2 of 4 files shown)


ahuoguo:
The newly supported theory breaks verusfmt (see https://github.com/verus-lang/verusfmt/issues/167).

PR #2026, opened 2025-12-03
Update `raw_ptr`

elanortang:
~~Changes:~~
- ~~Removed `phy()` from `PointsTo<[T]>` and inlined calls to `spec_size_of_val` and `spec_align_of_val`.~~
- ~~Added `MapPointsTo<T>` along with axioms for converting to and from `PointsTo<[T]>` and `PointsToRaw`.~~
- ~~Added additional `Provenance` spec fns along with axioms/type invariants.~~

Updated list of changes: 
1. Removed `phy()` from `PointsTo<[T]>` and inlined calls to `spec_size_of_val` and `spec_align_of_val`.
2. Changed `opt_value` to `mem_contents` as a more accurate function name. 
3. Changed `usize` types to `nat` or `int`, wherever possible and appropriate. 
4. Added additional `Provenance` spec fns along with broadcast axioms. 
5. Added the `MapPointsTo<T>` struct along with 
    a. Axioms for converting to and from `PointsTo<[T]>` and `PointsToRaw`. 
    b. Helper spec fns. 
    c. An implementation defining a type invariant, spec fn methods, and proofs of allowed operations/properties. 

### DIFF
--- a/source/vstd/transmute.rs
+++ b/source/vstd/transmute.rs
@@ -16,22 +16,22 @@ verus! {
 broadcast use {group_layout_axioms, group_vstd_default};
 
 /// Generic precondition on transmute.
-pub open spec fn transmute_pre<T, U>(src: T, dst: Tracked<U>) -> bool {
+pub open spec fn transmute_pre<T, U>(src: T, dst: U) -> bool {
     &&& forall|bytes|
         #![trigger abs_encode::<T>(&src, bytes)]
-        #![trigger abs_decode::<U>(bytes, &(dst@))]
-        abs_encode::<T>(&src, bytes) ==> abs_decode::<U>(bytes, &(dst@))
+        #![trigger abs_decode::<U>(bytes, &dst)]
+        abs_encode::<T>(&src, bytes) ==> abs_decode::<U>(bytes, &dst)
     &&& abs_can_be_encoded::<T>()
 }
 
 /// Generic postcondition on transmute.
-pub open spec fn transmute_post<U>(dst_ghost: Tracked<U>, dst: U) -> bool {
-    dst_ghost@ == dst
+pub open spec fn transmute_post<U>(dst_ghost: U, dst: U) -> bool {
+    dst_ghost == dst
 }
 
 /// Generic precondition on transmute for `?Sized` types.
-pub open spec fn transmute_pre_unsized<T: ?Sized, U: ?Sized>(src: &T, dst: Tracked<&U>) -> bool {
-    &&& forall|bytes| #[trigger] abs_encode::<T>(src, bytes) ==> abs_decode::<U>(bytes, dst@)
+pub open spec fn transmute_pre_unsized<T: ?Sized, U: ?Sized>(src: &T, dst: &U) -> bool {
+    &&& forall|bytes| #[trigger] abs_encode::<T>(src, bytes) ==> abs_decode::<U>(bytes, dst)
     &&& abs_can_be_encoded::<T>()
 }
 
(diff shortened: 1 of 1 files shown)


source/vstd/raw_ptr.rs
```diff
@@ -81,14 +85,33 @@ impl Provenance {
...
+    /// The alignment of the pointer's allocation. Must be a power of 2 bounded by `isize::MAX + 1`.
+    pub uninterp spec fn alignment(&self) -> usize;
+
+    /// The ID of the `Allocator` instance used to allocate this memory.
+    pub uninterp spec fn alloc_id(&self) -> AllocId;
+
+    /// Since `self.alignment()` returns a `usize`, `Alignment` invariants do not follow directly from the type.
+    /// We enforce them via a type invariant, instead.
+    // Q: What is the difference between making something an axiom and making it a type invariant?
+    #[verifier::type_invariant]
```
tjhance: This doesn't make sense as a type invariant. Type invariants should probalby just be disallowed for external_body types. This should probably be an axiom instead.

source/vstd/raw_ptr.rs
```diff
@@ -280,6 +306,7 @@ impl<T> PointsTo<T> {
     ///
     /// Note: If both S and T are non-zero-sized, then this implies the pointers
     /// have distinct addresses.
+    // Q: Why does this take in a &mut reference?
```
tjhance: If it took a & shared reference for both, then you could pass the same PointsTo in for both arguments.

source/vstd/raw_ptr.rs
```diff
@@ -516,6 +547,210 @@ impl PointsTo<str> {
...
+    /// Guarantee that the `PointsTo` for any non-zero-sized type points to a non-null address.
+    ///
+    // ZST pointers *are* allowed to be null, so we need a precondition that size != 0.
+    // See https://doc.rust-lang.org/std/ptr/#safety
+    pub axiom fn is_nonnull(tracked &self)
+        requires
+            size_of::<T>() * self.ptr()@.metadata != 0,
+        ensures
+            self.ptr()@.addr != 0,
+    ;
```
tjhance: I don't think you should have any axioms on this MapPointsTo type. Any property of it ought to be provable from the base axioms of PointsTo.

This axiom (`is_nonnull`) in particular doesn't make any sense to me. What if the map is empty? How does the map of `PointsTo` objects relate to `self.ptr()` anyway?

source/vstd/raw_ptr.rs
elanortang: Gotcha! Will add that to the documentation.

source/vstd/raw_ptr.rs
elanortang: Makes sense! I put it as a type invariant because that was Bryan's original solution to making sure we still had access to the `Alignment` invariants, but I will make it an axiom instead.

source/vstd/raw_ptr.rs
elanortang: The `PointsTo` objects must correspond to addresses that fall within `[self.ptr()@.addr, self.ptr()@.addr + self.ptr()@.metadata)`.

elanortang: Shouldn't I still have the `into_slice` axiom? I'm not sure that one is provable from the base axioms. 

Would the `is_nonnull` axiom make more sense if I added the requires that the map is non-empty?

elanortang:
Updated list of changes: 
1. Removed `phy()` from `PointsTo<[T]>` and inlined calls to `spec_size_of_val` and `spec_align_of_val`.
2. Changed `opt_value` to `mem_contents` as a more accurate function name. 
3. Changed `usize` types to `nat` or `int`, wherever possible and appropriate. 
4. Added additional `Provenance` spec fns along with broadcast axioms. 
5. Added the `MapPointsTo<T>` struct along with 
    a. Axioms for converting to and from `PointsTo<[T]>` and `PointsToRaw`. 
    b. Helper spec fns. 
    c. An implementation defining a type invariant, spec fn methods, and proofs of allowed operations/properties.

tjhance (approved):

PR #2028, opened 2025-12-04
new-mut-ref: handle Temporary places correctly

tjhance:
This handles the PlaceX::Temporary node correctly

Sometimes we need to work with unnamed "temporary places", given by the `PlaceX::Temporary` place, e.g., we might take a mutable reference of a computed expression:

```rust
let x = &mut foo(); // equivalent to `let mut tmp = foo(); let x = &mut tmp;`
```

or we might update a mutable reference returned by some function without assigning it

```rust
*func() = y; // equivalent to `let mut tmp = func(); *tmp = y;`
```

Furthermore, these temporary places might need resolving (just like any other place). For example, suppose `pair()` returns `(&mut A, &mut A)`. If we write:

```rust
let x = pair().0;
```

Then `pair().1` goes unused and requires immediate resolution.

In the resolution analysis, we assign a TempId to each PlaceX::Temporary node, and in the main part of the analysis, we treat these places just like any other places. After we compute the set of resolutions to add, we find all temporary nodes that are mentioned in the set. We then replace those temporaries with named Local nodes.

Specifically, we can replace the place node `Temorary(expr)` with `WithExpr({ tmp = expr; }, Local("tmp"))` for a fresh `tmp` variable. We place the declaration `let tmp;` at a larger scope. The `WithExpr` is a new kind of Place we add to the AST to allow Place expressions to have side-effects.

Later stages also have to handle temporaries; e.g., ast_to_sst will introduce a temporary variable on the fly if it encounters a Temporary node that needs to be mutated (though we could probably "optimize" the AIR a bit to remove this.)

### DIFF
--- a/source/vir/src/resolution_inference.rs
+++ b/source/vir/src/resolution_inference.rs
@@ -141,7 +141,7 @@ main part of the analysis, we treat these places just like any other places.
 After we compute the set of resolutions to add, we find all temporary nodes that
 are mentioned in the set. We then replace those temporaries with named Local nodes.
 
-Specifically, we can replace the place node `Temorary(expr)` with
+Specifically, we can replace the place node `Temporary(expr)` with
 `WithExpr({ tmp = expr; }, Local("tmp"))` for a fresh `tmp` variable.
 We place the declaration `let tmp;` at a larger scope.
 */


source/vir/src/resolution_inference.rs
```diff
@@ -111,19 +111,54 @@ into the expression, we always check that the local variable it refers to is act
...
+
+Then `pair().1` goes unused and requires immediate resolution. (This last example
+also illustrates that temporaries are nontrivial even when they are not mutated.)
+
+In our analysis, we assign a TempId to each PlaceX::Temporary node, and in the
+main part of the analysis, we treat these places just like any other places.
+After we compute the set of resolutions to add, we find all temporary nodes that
+are mentioned in the set. We then replace those temporaries with named Local nodes.
+
+Specifically, we can replace the place node `Temorary(expr)` with
```
Chris-Hawblitzel: small typo

PR #2029, opened 2025-12-05
test: add integration tests for `cargo verus verify`

mkovaxx:
## Summary

- add integration `tests` to confirm the behavior of `cargo-verus verify`
- add the `fake-cargo` helper binary inside `tests/src` to capture args and env vars that `cargo-verus` passes to `cargo`
- gate `fake-cargo` behind an `integration-tests` feature, so that it is absent from release builds

### DIFF
--- a/source/cargo-verus/tests/verify.rs
+++ b/source/cargo-verus/tests/verify.rs
@@ -1,0 +1,314 @@
+#[cfg(not(feature = "integration-tests"))]
+compile_error!("Enable the `integration-tests` feature to run these tests.");
+
+#[path = "src/utils.rs"]
+mod utils;
+
+use utils::*;
+
+#[test]
+fn crate_optin_workdir() {
+    let package_name = "foo";
+    let verify_crate_prefix = format!("__VERUS_DRIVER_VERIFY_{package_name}-0.1.0-");
+    let project_dir = MockPackage::new(package_name).lib().verify(true).materialize();
+
+    let (status, data) = run_cargo_verus(|cmd| {
+        cmd.current_dir(&project_dir).arg("verify");
+    });
+
+    assert!(status.success());
+
+    assert_eq!(data.args, vec!["build"]);
+
+    data.assert_env_has("RUSTC_WRAPPER");
+    data.assert_env_sets("__CARGO_DEFAULT_LIB_METADATA", "verus");
+    data.assert_env_sets("__VERUS_DRIVER_VIA_CARGO__", "1");
+    data.assert_env_sets_key_prefix(&verify_crate_prefix, "1");
+}
+
+#[test]
+fn crate_optin_manifest() {
+    let package_name = "foo";
+    let verify_crate_prefix = format!("__VERUS_DRIVER_VERIFY_{package_name}-0.1.0-");
+    let package_dir = MockPackage::new(package_name).lib().verify(true).materialize();
+
+    let manifest_path = package_dir.path().join("Cargo.toml");
+
+    let (status, data) = run_cargo_verus(|cmd| {
+        cmd.arg("verify");
+        cmd.arg("--manifest-path").arg(&manifest_path);
+    });
+
+    assert!(status.success());
+
+    assert_eq!(
+        data.args,
+        vec!["build", "--manifest-path", manifest_path.to_str().expect("manifest path to string")]
+    );
+
+    data.assert_env_has("RUSTC_WRAPPER");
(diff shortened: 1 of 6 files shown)


parno:
Just as a quick note, the existing test suite has some infrastructure and sample tests for `cargo verus`.  They're created in `source/rust_verify_test/tests/cargo.rs`, using macros from `source/rust_verify_test_macros/src/examples.rs`, and the sample projects live in `source/rust_verify_test/tests/cargo-tests`.  We're not wedded to that infrastructure, however.  I just wanted to mention it so we could potentially unify them with your tests.

bsdinis:
One other note: `cargo-verus` tests are currently being a big bottleneck in CI. I'm planning on looking at it in a couple weeks time. If you find a way to make sure your tests don't have to use the network (assuming that `vargo-build` has been run already) and don't re-verify vstd that would be awesome. This doesn't have to be the default, but would be nice if we could toggle this _fast execution_ mode relatively easily

mkovaxx:
@parno Thanks for the pointer! I looked at the verification tests, and I think we should keep them separate. It might make sense to share the test fixtures themselves, but the high-level goals are different.

@bsdinis That's a helpful piece of info, thanks for chiming in! I think it's very likely these new `cargo-verus` tests do not need network access at all.

These tests are meant to confirm integration between `cargo-verus` and `cargo` itself, without actually running any of the heavy machinery. This is where the helper `fake-cargo` binary comes in: it allows capturing the args and env vars that `cargo-verus` passes to `cargo`, but without actually running `cargo` or `verus` in any serious capacity. It only ever invokes `cargo metadata`, which in this situation avoids relying on the network because the test fixtures are entirely local, using `path` deps, etc.

mkovaxx:
@parno @tjhance This is ready for review. I'm not adding it to the CI flow just yet, not to interfere with #2024.

I just want the basic machinery in place so I can rely on it locally when making changes to `cargo-verus`.

source/cargo-verus/tests/verify.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    data.assert_env_has("RUSTC_WRAPPER");
+    data.assert_env_sets("__CARGO_DEFAULT_LIB_METADATA", "verus");
+    data.assert_env_sets("__VERUS_DRIVER_VIA_CARGO__", "1");
+    data.assert_env_sets_key_prefix(&verify_optin_prefix, "1");
+    data.assert_env_has_no_key_prefix(&verify_optout_prefix);
+    data.assert_env_has_no_key_prefix(&verify_unset_prefix);
+    // TODO: `cargo-verus` should be fixed in a follow-up change to have
+    //       the correct behavior, i.e. not mark `hasdeps` to be verified
+    data.assert_env_sets_key_prefix(&verify_hasdeps_prefix, "1");
```
mkovaxx, line 265: Note that the current behavior here is likely incorrect, which I plan to fix in a follow-up change.

tjhance:
I think I know too little about this to be of any use, I'll leave it to @parno

parno:
I'll try to take a look, but I'm traveling for a workshop this week, so it may be a bit delayed.

mkovaxx:
> I think this looks good, and the use of `fake_cargo` is a nice way to get more visibility into the process. My only minor question is that I wasn't clear on the intended meaning of `fixture`. Is it in contrast to a project that's part of a workspace? If so, perhaps "singleton" or "standalone" might be clearer.

@parno Thanks for the review! Yes, the `fixtures` directory contains small Cargo project mocks that the tests rely on. As the `cargo-verus` test suite grows, I intend to replace them with a small DSL so that each test can construct its own file system environment for `cargo-verus` to run inside, instead of the current shared `fixtures`.

mkovaxx:
@parno I've eliminated the static `fixtures` directory in favor of a small DSL that allows each test to construct its own isolated fixture.

For example, here's how to construct a workspace with 4 crates inside:
- `optin`: explicitly opts in to verification
- `optout`: explicitly opts out of verification
- `unset`: has no verification setting
- `hasdeps`: depends on `optin` and opts in to verification

```rust
let workspace_dir = MockWorkspace::new()
    .member(MockPackage::new("optin").lib().verify(true))
    .member(MockPackage::new("optout").lib().verify(false))
    .member(MockPackage::new("unset").lib())
    .member(MockPackage::new("hasdeps").lib().dep("optin").verify(true))
    .materialize();
```

All files are created inside a fresh temporary directory, which is automatically cleaned up if the test succeeds.

parno:
Cool, I think the DSL looks nice, and it should make it easier to expand the test suite as we go.

PR #2030, opened 2025-12-08
std_specs/convert: Add spec for TryFrom and TryInto

ziqiaozhou:

### DIFF
--- a/source/vstd/std_specs/convert.rs
+++ b/source/vstd/std_specs/convert.rs
@@ -106,6 +106,15 @@ pub assume_specification<T, U: TryFrom<T>>[ <T as TryInto<U>>::try_into ](a: T) 
         call_ensures(U::try_from, (a,), ret),
 ;
 
+pub assume_specification<T, U: Into<T>>[ <T as TryFrom<U>>::try_from ](a: U) -> (ret: Result<
+    T,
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel:
One question: does this also cover trivial try_into cases, such as converting u8 to u16 via try_into?

```
fn test() {
    let _: u16 = 7u8.into();
    let _: u16 = 7u8.try_into().unwrap();
    let _: u8 = 7u16.try_into().unwrap();
}
```

Chris-Hawblitzel:
> One question: does this also cover trivial try_into cases, such as converting u8 to u16 via try_into?

More generally, I'm wondering about this from https://doc.rust-lang.org/src/core/convert/mod.rs.html#827 :

```
impl<T, U> const TryFrom<U> for T
where
    U: [const] Into<T>,
{
    type Error = Infallible;

    #[inline]
    fn try_from(value: U) -> Result<Self, Self::Error> {
        Ok(U::into(value))
    }
}
```

ziqiaozhou:
> fn test() {
>     let _: u16 = 7u8.into();
>     let _: u16 = 7u8.try_into().unwrap();
>     let _: u8 = 7u16.try_into().unwrap();
> }

I did not cover this since I did not use that. Now the new commit covers it.

PR #2032, opened 2025-12-09
vstd cell restructuring

tjhance:
This makes various clean-ups to the vstd cell libraries.

**PCell:** There are now two version of `PCell`:
  * `PCell`/`PointsTo` : cells where the contents are always initialized. (Trusted)
  * `PCellUn`/`PointsToUn` : cells where the contents my be uninitialized. (Verified)

The first one is much simpler than the existing `PCell` library, so it simplifies the trusted core and hopefully makes exposition easier. It's probably useful for some people who don't need the optional-initialization aspect, and it allows the type parameter to be unsized, addressing a recent zulip request.

The second one has the same interface as the previous `PCell` library, so we can migrate existing code by just switching to use it.

The one element here I wasn't sure about was the name for the second library. Between 'cryptic' and 'unwieldy' I chose cryptic, with "un" standing for "maybe Uninitialized". Open to other suggestions here. I'd like to get the name right so we don't need to do another rename later ...

**MaybeUninit**  to support the verification of `PCellUn` I also added external specs for `MaybeUninit`.

**InvCell** I updated this to use the new trait-style for specifying invariants. I'd like to start consolidating these traits going forward (right now every invariant-based container type declares its own trait) so rather than creating a new trait like `InvCellInvariantPredicate` or something, I just created a new `Predicate` trait and gave it its own module (predicate.rs).

**Migration strategy**  This PR leaves the original cell.rs file untouched, but marks it deprecated. To migrate off the deprecated API people will need to rename their types to the new ones. InvCell has the biggest changes with the addition of a type parameter.

### DIFF
--- a/source/vstd/cell.rs
+++ b/source/vstd/cell.rs
@@ -23,7 +23,7 @@ verus! {
 
 broadcast use {super::map::group_map_axioms, super::set::group_set_axioms};
 
-/// **Now deprecated** See [`pcell::PCell`] or [`pcell_maybe_uninit::PCellUn`] instead
+/// **Now deprecated** See [`pcell::PCell`] or [`pcell_maybe_uninit::PCell`] instead
 ///
 /// `PCell<V>` (which stands for "permissioned call") is the primitive Verus `Cell` type.
 ///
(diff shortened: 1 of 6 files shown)


tjhance:
Maybe I should just call them PCell / PointsTo, but with a short module name.

`un::PCell`, `un::PointsTo`

parno:
> Maybe I should just call them PCell / PointsTo, but with a short module name.

Does it need to be short?  Maybe something longer (e.g., `uninit`) would be acceptable, since I would guess in most cases, you'll only use one or the other, so you can just put `use uninit::*` at the top and not care about the extra characters.

Chris-Hawblitzel:
From the veritas tests, there's a failure here:

```
thread 'rustc' (11076) panicked at vir/src/modes.rs:902:13:
assertion failed: !record.temporary_modes.contains_key(&place.span.id)
```

This shows up in node-replication ( https://github.com/parno/verified-nrkernel.git , verified-node-replication/src/lib.rs ) and mimalloc.  The other tests finished successfully.

tjhance:
that was fixed: https://github.com/verus-lang/verus/issues/2034

I don't think a veritas run will be on any utility on this PR, since the existing pcell file is unchanged (except for the deprecation warning)

Chris-Hawblitzel (approved):

PR #2033, opened 2025-12-09
refactor: cargo-verus to use clap and reorganize internals

mkovaxx:
- refactor `cargo-verus` to use `clap` for parsing its command-line arguments
- reorganize internals to better reflect execution flow

### DIFF
--- a/source/cargo-verus/src/main.rs
+++ b/source/cargo-verus/src/main.rs
@@ -32,13 +32,13 @@ pub fn main() -> Result<ExitCode> {
             Ok(ExitCode::SUCCESS)
         }
         VerusSubcommand::Verify(cmd) => {
-            subcommands::run_build("build", &cmd.cargo_opts, &cmd.verus_args, true)
+            subcommands::run_cargo("build", &cmd.cargo_opts, &cmd.verus_args, true)
         }
         VerusSubcommand::Build(cmd) => {
-            subcommands::run_build("build", &cmd.cargo_opts, &cmd.verus_args, false)
+            subcommands::run_cargo("build", &cmd.cargo_opts, &cmd.verus_args, false)
         }
         VerusSubcommand::Check(cmd) => {
-            subcommands::run_build("check", &cmd.cargo_opts, &cmd.verus_args, true)
+            subcommands::run_cargo("check", &cmd.cargo_opts, &cmd.verus_args, true)
         }
     }
 }
--- a/source/cargo-verus/src/subcommands.rs
+++ b/source/cargo-verus/src/subcommands.rs
@@ -87,29 +87,29 @@ verify = true
     Ok(())
 }
 
-pub fn run_build(
-    cargo_subcommand: &str,
+pub fn run_cargo(
+    subcommand: &str,
     cargo_options: &CargoOptions,
     verus_args: &[String],
-    warn_if_not_verified: bool,
+    warn_if_nothing_verified: bool,
 ) -> Result<ExitCode> {
-    let cargo_args = build_cargo_args(cargo_options, false);
+    let cargo_args = make_cargo_args(cargo_options, false);
     let mut common_verus_driver_args: Vec<String> =
         vec!["--VIA-CARGO".to_owned(), "compile-when-not-primary-package".to_owned()];
 
-    if !warn_if_not_verified {
+    if !warn_if_nothing_verified {
         common_verus_driver_args.extend_from_slice(&[
             "--VIA-CARGO".to_owned(),
             "compile-when-primary-package".to_owned(),
         ]);
     }
 
-    let metadata_args = build_cargo_args(cargo_options, true);
+    let metadata_args = make_cargo_args(cargo_options, true);
     let metadata = fetch_metadata(&metadata_args)?;
 
     common_verus_driver_args.extend(verus_args.iter().cloned());
     let (mut command, verified_something) =
-        build_cargo_command(cargo_subcommand, &cargo_args, common_verus_driver_args, &metadata)?;
+        make_cargo_command(subcommand, &cargo_args, common_verus_driver_args, &metadata)?;
 
     let exit_status = command
         .spawn()
@@ -117,7 +117,7 @@ pub fn run_build(
         .wait()
         .context("Failed to wait for cargo")?;
 
-    if warn_if_not_verified && !verified_something {
+    if warn_if_nothing_verified && !verified_something {
         eprint!(
             "{}",
             "\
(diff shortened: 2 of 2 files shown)


mkovaxx:
@jaybosamiya-ms I think you made changes to `cargo-verus` before, so I figured it's a good idea to run these changes by you.

source/cargo-verus/src/cli.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub unstable_flags: Vec<String>,
+
+    #[arg(
+        value_name = "CARGO_OPTIONS",
+        num_args = 0..,
+        allow_hyphen_values = true,
+        help = "Options forwarded to 'cargo build' or 'cargo check'"
+    )]
+    pub cargo_args: Vec<String>,
+}
```
bsdinis, line 91: should this be spelled out or passed directly to cargo?

source/cargo-verus/src/subcommands.rs
```diff
@@ -0,0 +1,312 @@
...
+    args
+}
+
+fn build_cargo_command(
+    cargo_subcommand: &str,
+    cargo_args: &[String],
+    common_verus_driver_args: Vec<String>,
+    metadata: &cargo_metadata::Metadata,
+) -> Result<(Command, bool)> {
+    // TODO: use the "+ ... toolchain" argument?
```
bsdinis, line 215: I wonder if we should leave this to the user -- we should definitely pass it through when supplied and respect the `rust-toolchain.toml`

source/cargo-verus/src/main.rs
```diff
@@ -8,515 +8,42 @@
...
-pub const VERUS_DRIVER_IS_BUILTIN: &str = " __VERUS_DRIVER_IS_BUILTIN_";
-pub const VERUS_DRIVER_IS_BUILTIN_MACROS: &str = " __VERUS_DRIVER_IS_BUILTIN_MACROS_";
-pub const VERUS_DRIVER_VERIFY: &str = "__VERUS_DRIVER_VERIFY_";
-pub const VERUS_DRIVER_VIA_CARGO: &str = "__VERUS_DRIVER_VIA_CARGO__";
+mod cli;
+mod metadata;
+mod subcommands;
+use crate::cli::{CargoVerusCli, VerusSubcommand};
 
 pub fn main() -> Result<ExitCode> {
```
bsdinis, line 21: do we need to return the exit code explicitely? why not return Ok(())

source/cargo-verus/src/subcommands.rs
mkovaxx: That comment was simply moved over with the function from here:
https://github.com/verus-lang/verus/pull/2033/files#diff-7966b5abd1d1284a64baee6705a5fafb62467f8ffb41e697b993565ce043ef6aL253

I think it's best to leave addressing it to a follow-up change.

mkovaxx: It's line 253 in the previous state of `main.rs`.

You likely need to make GitHub first load the diff for `main.rs` then the URL will work in the same tab.

source/cargo-verus/src/main.rs
mkovaxx: I believe the purpose of this is to propagate the Cargo error code back up, so that `cargo-verus` behaves identically to `cargo` in terms of exit codes.

Related part in `subcommands.rs`:
https://github.com/verus-lang/verus/pull/2033/files#diff-cba6ae1dc6bde1b31adf9e4ab7ae581df8b27a86e5630f601ef68e71836cc860R133-R138

source/cargo-verus/src/subcommands.rs
bsdinis: yup, no worries on this -- was ultimately jotting down what i believe the correct behaviour should probably be

source/cargo-verus/src/cli.rs
mkovaxx: Good question!

I think we can trim these down a bit.

The only options strictly necessary are ones that `cargo metadata` supports. We list those out, so we know that anything else in `cargo_args` is not to be forwarded to `cargo metadata`, only to other `cargo` invocations.

Based on the previous state of `main.rs`, the options supported by `cargo metadata` are `--frozen`, `--locked`, `--offline`, `--config`, `--manifest-path`, `-Z`.

Please let me know if we're missing anything!

source/cargo-verus/src/main.rs
bsdinis: Thank you for pointing me to the thing I clearly missed!
I think this is fair enough for this change. I wonder if anyone particularly cares about cargo exit codes in practice.

source/cargo-verus/src/cli.rs
mkovaxx: Later we should revise that list to also include options in the current `--help` menu of `cargo metadata`, as follows:

```
Usage: cargo metadata [OPTIONS]

Options:
      --filter-platform <TRIPLE>  Only include resolve dependencies matching the given target-triple
      --no-deps                   Output information only about the workspace members and don't fetch dependencies
      --format-version <VERSION>  Format version [possible values: 1]
  -v, --verbose...                Use verbose output (-vv very verbose/build.rs output)
  -q, --quiet                     Do not print cargo log messages
      --color <WHEN>              Coloring [possible values: auto, always, never]
      --config <KEY=VALUE|PATH>   Override a configuration value
  -Z <FLAG>                       Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
  -h, --help                      Print help

Feature Selection:
  -F, --features <FEATURES>  Space or comma separated list of features to activate
      --all-features         Activate all available features
      --no-default-features  Do not activate the `default` feature

Manifest Options:
      --manifest-path <PATH>  Path to Cargo.toml
      --lockfile-path <PATH>  Path to Cargo.lock (unstable)
      --locked                Assert that `Cargo.lock` will remain unchanged
      --offline               Run without accessing the network
      --frozen                Equivalent to specifying both --locked and --offline
```

mkovaxx:
@bsdinis I'm stuck trying to figure out why the `ci / fmt` check is failing. Neither `cargo fmt` nor `verusfmt` output any changes when I run them locally. Any idea what might be going wrong here?

source/cargo-verus/src/main.rs
bsdinis: For reference, I surveyed `cargo-semver-checks`, `cargo-audit`, `cargo-criterion` and `cargo-nextest` and none seemed to care particularly about the return code (i.e., return something like Result<()> from main). [`cargo-q`](https://github.com/guuzaa/cargo-q/blob/main/src/main.rs) a cargo plugin that calls other cargo subprocesses also doesn't make it a big deal of returning exactly the same error code.

As long as errors return a NOK exit code I think we should be fine.

bsdinis: (for future ref)

bsdinis:
> @bsdinis I'm stuck trying to figure out why the `ci / fmt` check is failing. Neither `cargo fmt` nor `verusfmt` output any changes when I run them locally. Any idea what might be going wrong here?

It's showing me this
```
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{Context, Result, anyhow, bail};
 use colored::Colorize;
 
 use crate::cli::CargoOptions;
Diff in /Users/runner/work/verus/verus/source/cargo-verus/src/subcommands.rs:9:
-use crate::metadata::{fetch_metadata, make_package_id, MetadataIndex};
+use crate::metadata::{MetadataIndex, fetch_metadata, make_package_id};
 
```

from the action script.

I routinely run into this -- seems like our autoformatter has very strong opinions about order of imports
(which is annoying if like me you have format on save and it disagrees with verus)

bsdinis:
The way I usually go about this is by running `vargo fmt` after closing my editor and pushing again

mkovaxx:
> The way I usually go about this is by running `vargo fmt` after closing my editor and pushing again

@bsdinis Thanks, that did the trick! :) I didn't realize `vargo fmt` was meant to be different from `verusfmt`.

jaybosamiya-ms:
I haven't yet reviewed the code (might take me a little while to find the spare cycles), but some quick comments:
- it is awesome to see the move to `clap`!  I've mentioned this to @Chris-Hawblitzel in the past that I wanted us to move to `clap`, and it is good to see this, thank you for working on it!  I think one of the points he'd mentioned is that he didn't want fragmentation in command-line-parsing dependencies in the Verus repo though, so maybe we should queue up a change to `clap` for the other bits that do command-line parsing too (not _just_ `cargo-verus`).
- `vargo fmt` invokes both rustfmt _and_ verusfmt internally.  It should _technically_ be sufficient to just run `verusfmt` on most projects, but there were some opinionated decisions made with `vargo fmt`, which is why you might see mild differences in running `verusfmt` by itself, and running it with `vargo fmt`.  Most other projects (other than this repo) should just use `verusfmt` directly.

It's on my TODO to review the changes in this PR, but please don't be blocked on me, feel free to have someone else review the changes instead too if we want to get this in sooner to unblock other next steps.

hiroki-chen:
I'm sorry but this PR breaks the workflow of `cargo-verus` and `cargo verus` as the extra args to verus are forwarded to cargo which should be forwarded to `rust_verify` instead.

So this would fail:

```sh
cargo verus build --release -- --expand-errors --rlimit=100
error: unexpected argument '--expand-errors ' found
error: unexpected argument '--rlimit' found
```

mkovaxx:
> I'm sorry but this PR breaks the workflow of `cargo-verus` and `cargo verus` as the extra args to verus are forwarded to cargo which should be forwarded to `rust_verify` instead.
> 
> So this would fail:
> 
> ```shell
> cargo verus build --release -- --expand-errors --rlimit=100
> error: unexpected argument '--expand-errors ' found
> error: unexpected argument '--rlimit' found
> ```

@hiroki-chen Thanks for reporting this! A fix is in progress. :)

mkovaxx:
@hiroki-chen Fix here: #2050.

PR #2035, opened 2025-12-11
new-mut-ref: add slice/array indexing to Place

tjhance:
Presently, we handle array/slice index in rust_to_vir_expr by immediately redirecting to a function defined in vstd. This method makes it awkward to track places correctly, so the cleanest thing to do is add an `Index` variant to the `Place` enum. This PR:

 * Adds `PlaceX::Index`, which indexes into a slice or array. (This is only used when new-mut-ref is enabled.)
 * Extends the binary op `ArrayIndex` to work for slices as well as arrays (and renames it to just `Index`)
 * Add a unary op `Length` (needed to emit the bounds-check preconditions)
 * Allows indexing in an SST loc (in order to handle assignments like `x[0] = ...`)
   * To support these assignments we also add `spec_slice_update` and `spec_array_update` verus items to vstd, but these are only emitted in sst_to_air, there are no VIR nodes for them.
 
Indexing of course has a precondition; it's a little annoying that we have to handle this precondition in so many different places due to the # of different ways a Place node might get lowered. To consolidate these into one place I use the new `place_preconditions.rs` file. This system will also form the template for correctly handling union fields.

I had the whole mode system worked out for indices into exec/tracked places (partially implemented) but when I went to write tests I remembered that the `[ ]` syntax gets turned into `spec_index` in ghost code anyway. So indexing only works for exec places (as is the case today).

### DIFF
--- a/source/vir/src/sst_to_air.rs
+++ b/source/vir/src/sst_to_air.rs
@@ -867,7 +867,6 @@ pub(crate) fn new_user_qid(ctx: &Ctx, exp: &Exp) -> Qid {
 
 pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<Expr, VirErr> {
     let typ_to_ids = |typ| typ_to_ids(ctx, typ);
-    let exp_typ = &exp.typ;
 
     let result = match &exp.x {
         ExpX::Const(c) => {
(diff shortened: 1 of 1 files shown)


source/vir/src/sst_to_air.rs
```diff
@@ -1072,18 +1072,18 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<
                 ident_apply(&crate::def::to_dyn(trait_path), &args)
```
Chris-Hawblitzel, line 1071: There was a conflicting change here that caused a stack overflow and test failure, so in ToDyn, `exp` can be changed to `e` and `exp_typ` can be changed to `&exp.typ` and `exp_typ` can be removed.

Issue #2036, opened 2025-12-15
decrease_when allows non-terminating spec function


WeituoDAI:
This is about the `decrease...when...` clause of recursive spec fn.

I find that the function below is allowed
```
spec fn g(x:int) -> int
  decreases x when x > 1
{
  g(x-1)
}
```

After checking the vir, I know that verus only checks that if the when-condition is meet, then the decrease-measure decreases; but does not check that the decrease-measure in the recursive call still meets the when-condition.

I am not sure if it is the expected behavior, although currently there is no soundness problem. In my personal experience, requiring the check of the when-condition can help user write reasonable recursive function. (so the when-condition works like the pre-condition)

I implemented the check of when-condition in my own fork, then a test case in examples/syntax.rs reports an error:
```
spec fn dec0(a: int) -> int
    decreases a,
    when a > 0
    via dec0_decreases
{
    if a > 0 {
        dec0(a - 1)
    } else {
        0
    }
}
```

Here it only checks that the decrease-measure decreases (the when-condition does not work like  a pre-condition here)

And in fact for this testcase, we cannot do any proof about dec0 because any call to it must recursively call a dec0(0) which is uninterp since it does not meet the when-condition. We should remove the part of "when" and "via", then everything works well.

PR #2039, opened 2025-12-17
Check the when-condition for recursive call of spec fn

WeituoDAI:

Related to #2036 .

I notice that the decrease_when condition is not checked when it is recursively called inside a spec fn. For example, the definition below is allowed currently.
```
spec fn h(x:int) -> int
    decreases x when x > 1
{
    h(x-1)
}
```

And an example found in `examples/syntax.rs` : 
```
spec fn dec0(a: int) -> int
    decreases a,
    when a > 0
    via dec0_decreases
{
    if a > 0 {
        dec0(a - 1)
    } else {
        0
    }
}
```
In fact, we cannot do any meaningful proof about this `dec0`, since any call of `dec0(a)` (a>0) will recursively call `dec0(0)` at last, but it is uninterp since it does not meet the when-condition (If I make no mistake, the decrease_when is implemented as such).

It does not lead to soundness problem, but it must be a bug. For any recursive function `f`, We know `f(x)` is uninterp if `x` does not meet the when-condition; if inside its body, there is a recursive call `f(y)` where `y` does not meet the when-condition, then the function can only return uninterp..

In my eyes, decrease_when should really work like a pre-condition here.

So I (1) add the check of when-condition for any recursive call inside the body of spec function (2) change some related testcase


WeituoDAI:
conflict resolved

PR #2041, opened 2025-12-17
warn about implicit mut-deref in spec equality

tjhance:
Presently, Verus equality operators allow comparing `&mut T` to `T` or `&mut T` to `&mut T`. In the new-mut-ref world, the first one should be disallowed and the second one may have different semantics than it does today. I think it's best to go ahead and disable these implicit derefences.

This PR adds a deprecation warning if the user writes `a == b` where either `a` or `b` is a mutable reference.


parno:
> the second one may have different semantics than it does today.

Because `a == b` will compare the tuples of current and prophesied values?

tjhance:
> Because a == b will compare the tuples of current and prophesied values?

Yes. (stronger than that even, it will compare their abstract identifiers, so they only compare equal if they're the same borrow)

parno:
That does seem like something the developer didn't intend.  Adding the warning now is a good idea.

PR #2042, opened 2025-12-18
Add specification for `is_empty` on slice

jaylorch:


tjhance (commented):
Why does this need to be a `VerusItem`/`rustc_diagnostic_item`?

jaylorch:
> Why does this need to be a `VerusItem`/`rustc_diagnostic_item`?

Ah, good point. I was just copying `len`, but that's not a good reason. I'll fix it.

PR #2048, opened 2025-12-20
`try_broadcasts`: automatic search for relevant broadcast lemmas

dschoepe:
Completing proofs involving data structures such as sequences, sets, and maps often involves finding the right lemmas that exist in the standard library. Since `vstd` provides a large number of lemmas that is expected to grow over time, identifying which lemmas may be useful during a proof development becomes a tedious, manual task, in addition to potentially large number of project specific lemmas.

This adds a `#[verifier::try_broadcasts]` feature, inspired by the [Sledgehammer tool](https://isabelle.in.tum.de/doc/sledgehammer.pdf) provided by the [Isabelle](https://isabelle.in.tum.de/) theorem prover.

Assisted-By: codex:openai-gpt5.6-sol


source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+the function yields a proof that no longer requires the Sledgehammer attribute:
+
+```
+{{#include ../../../../examples/guide/sledgehammer.rs:set_example_done}}
+```
+
+The `#[verifier::sledgehammer]` attribute can be applied to `proof` and `exec`
+mode functions. After finding a proof with Sledgehammer, the
+`#[verifier::sledgehammer]` attribute should afterwards be removed from the
+function to avoid rerunning the proof search on every Verus invocation.
```
jaybosamiya-ms: Minor comment: it'd be nice if the sledgehammer tool could automatically recognize that it is no longer needed and tell you to remove it.  I haven't tested the branch, so if it already supports/does this, that's great!  If  not, would be good to document something to the note of "future versions of Verus will automatically detect such cases" or similar.

source/docs/guide/src/try_broadcasts.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Note:** Sledgehammer can only discover proofs using lemmas marked as
+`broadcast`. Additionally, Sledgehammer's proof attempts may time out, even
+though a proof with the right subset of relevant lemmas would succeed.
+
+## Proof Minimization
+
+By default, Sledgehammer will minimize a proof; in the above example,
+Sledgehammer initially found a proof that used 5 lemmas, but then determined
+that only one lemma is actually necessary. However, in some case, minimizing a
```
jaybosamiya-ms: Minor typo: "some case" -> "some cases"

jaybosamiya-ms (commented):
Awesome to see this!  I haven't reviewed the code, but I read through the doc and have a couple of minor comments included inline, and a couple of higher level ones here:

On naming: since this is looking for broadcasts to use, it feels a bit like tuning a radio to the valid frequencies/broadcasts, so maybe a name like `autoscan` could work 📻

In terms of output: rustc often uses a special syntax to mark "this is how code should change for an auto-fix", it may be nice to generate the final "note" with that so that tools like verus-analyzer and verus-mode.el can pick that up and automatically insert it in.  I for one would be interested in introducing a single keybinding to verus-mode.el that automatically introduces the ~`#[verifier::sledgehammer]`~ `#[verifier::autoscan]`, runs on that one function, picks up the results, and auto-rewrites the result with that.  The particular rewrite syntax output does not necessarily need to be in this PR, imho, but just documenting it here as something that might be worth doing.

source/docs/guide/src/sledgehammer.md
dschoepe: Good point, I'll add something like that.

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+Since `vstd` provides a large number of lemmas that is expected to grow over
+time, identifying which lemmas may be useful during a proof development becomes
+a tedious, manual task.
+
+To simplify this task, Verus provides *Sledgehammer*, inspired by the [eponymous
+tool](https://isabelle.in.tum.de/doc/sledgehammer.pdf) provided by
+the [Isabelle](https://isabelle.in.tum.de/) theorem prover.
+
+Sledgehammer is an *experimental* feature that attempts to automatically complete
+a proof attempt by automatically using relevant lemmas using the
```
parno: Perhaps we can cut one of the two "automatically"s in this sentence?

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+This function fails to prove without additional annotations. Suppose however,
+that elsewhere in the project, someone already proved a helpful lemma:
+
+```rust
+{{#include ../../../../examples/guide/sledgehammer.rs:set_example_lemma}}
+```
+
+The proof of `union_three_sets` automatically succeeds when adding `broadcast
```
parno: I would rephrase as:
> automatically succeeds when `broadcast use union_len;` is added to its body.

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+use union_len;` to its body. However, this requires identifying both knowing the
```
parno: Cut "identifying"

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+name of this lemma, and enabling it via `broadcast use`.
+
+To simplifying finding relevant lemmas, Verus provides the
```
parno: "simplifying" -> "simplify"

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+lemma as follows:
+
+```rust
+{{#include ../../../../examples/guide/sledgehammer.rs:set_example_sledgehammer}}
+```
+
+Running Verus on this example now prints the following message:
+
+```
+note: Sledgehammer found proof with 5 lemmas, minimizing..
```
parno: I think we want "found proof" -> "found a proof"

source/docs/guide/src/sledgehammer.md
```diff
@@ -0,0 +1,142 @@
...
+are either are defined in or referenced by the current crate. For example,
+absent any other references to `vstd::Seq::lemma_flatten_push`, Sledgehammer
+fails to find a proof in the following example:
+
+```rust
+{{#include ../../../../examples/guide/sledgehammer.rs:requires_vstd_lemma}}
+```
+
+However, the `vstd::Seq::lemma_flatten_push` would be sufficient to complete the
+proof. When referencing this lemma anywhere in the current project, Sledgehammer
```
parno: "When referencing this lemma" -> "When this lemma is referenced"

source/rust_verify/src/attributes.rs
```diff
@@ -346,6 +346,7 @@ pub(crate) enum Attr {
     EncodedStatic,
     // The "const _: () = { ... }" generated by synstructure for #[derive(Structural)]
     StructuralConstWrapper,
+    Sledgehammer(bool), // if `true` is passed, sledgehammer will minimize proof
```
parno: "will minimize proof" -> "will try to minimize its suggested proof"

source/rust_verify/src/try_broadcasts.rs
```diff
@@ -0,0 +1,801 @@
...
+    pub(crate) used_axioms: Option<Vec<air::ast::Ident>>,
+}
+
+impl VerificationOutcome {
+    fn new() -> Self {
+        VerificationOutcome { any_invalid: false, any_timeout: false, used_axioms: None }
+    }
+
+    #[allow(dead_code)]
+    fn success(&self) -> bool {
```
parno: Why is this dead?

source/rust_verify/src/sledgehammer.rs
```diff
@@ -0,0 +1,801 @@
...
+    verifier: &'a mut Verifier,
+    reporter: &'a R,
+    krate: &'a Krate,
+    source_map: Option<&'a SourceMap>,
+    bucket_id: &BucketId,
+    gctx: GlobalCtx,
+) -> Result<(Option<Krate>, GlobalCtx), VirErr> {
+    let Some(mut sh) = Sledgehammer::<'a, R>::new(verifier, reporter, krate, source_map, bucket_id)
+    else {
+        // TODO: report error here
```
parno: Lingering TODO here

source/rust_verify/src/sledgehammer.rs
```diff
@@ -0,0 +1,801 @@
...
+    reporter: &'a R,
+    krate: &'a Krate,
+    source_map: Option<&'a SourceMap>,
+    bucket_id: BucketId,
+    target_func: Function,
+    fun_to_function: HashMap<Fun, Function>,
+    do_minimize: bool,
+}
+
+pub(crate) fn sledgehammer<'a, R: Diagnostics + 'a>(
```
parno: Since there's already a constructor for Sledgehammer, should the style be to call that and then make this a function on the returned Sledgehammer?  Or is your preference to keep the call site limited to a single call to this function?

source/rust_verify/src/try_broadcasts.rs
```diff
@@ -0,0 +1,801 @@
...
+    /// to iteratively remove different subsets of the lemmas used originally.
+    /// To improve performance, we track how many times we unsuccessfully tried
+    /// removing a particular lemma.
+    fn minimize(
+        &mut self,
+        guess: Guess,
+        global_ctx: GlobalCtx,
+    ) -> Result<(Guess, GlobalCtx), SledgehammerErr> {
+        let mut min_state = MinimizeState::from(&guess);
+        self.minimize_outer(guess, 2, &mut min_state, global_ctx)
```
parno, line 337: What does 2 signify here?

source/rust_verify/src/sledgehammer.rs
```diff
@@ -0,0 +1,801 @@
...
+        reporter: &impl Diagnostics,
+        krate: &Krate,
+        source_map: Option<&SourceMap>,
+        bucket_id: &BucketId,
+        global_ctx: GlobalCtx,
+        outcome: Option<&mut VerificationOutcome>,
+    ) -> Result<GlobalCtx, VirErr> {
+        let prev_errors = self.count_errors;
+        let prev_verified = self.count_verified;
+        // TODO: the following is copied from verify_bucket_outer; it would be nicer
```
parno: Is it copied because the original logic in `verify_bucket_outer` isn't cleanly factored out of its current setting?  If so, we should probably fix that, so that the code here and there doesn't get out of sync.

source/rust_verify/src/try_broadcasts.rs
parno, line 1: I'm not sure how well we've stuck to this, but I think we've generally tried to sort files topologically, such that the callees precede callers.

parno (approved):
Thanks for putting this together.  I added a bunch of mostly minor comments.  For this one:

> Lemmas from other crates are only considered if they are referenced by the current crate; otherwise they are pruned prior to verification.

In your experience, is this going to be a significant limitation?  If so, should we consider broadening the scope (either automatically or based on another Sledgehammer option)?  We could do it in a couple of ways (e.g., detecting a sledgehammer attribute and disabling pruning, or caching a copy of the pre-pruned data prior to pruning).

source/rust_verify/src/config.rs
```diff
@@ -57,6 +57,32 @@ pub struct LogArgs {
     pub log_call_graph: bool,
 }
 
+impl Clone for LogArgs {
```
Chris-Hawblitzel: You can just derive Clone for LogArgs and VirLogOption.  It appears that what was blocking this was a spurious use of clone in decl_derive for ToDebugSNode that can be fixed as follows:

```
diff --git a/source/vir_macros/src/lib.rs b/source/vir_macros/src/lib.rs
index cbf4d3a98..3cbfb503b 100644
--- a/source/vir_macros/src/lib.rs
+++ b/source/vir_macros/src/lib.rs
@@ -91,7 +91,7 @@ fn to_node_inner(
                 let field_name = field_name.to_string();
                 stmts.push(quote!(nodes.push(::sise::Node::Atom(format!(":{}", #field_name.to_string())));));
             }
-            stmts.push(quote!(nodes.push(crate::printer::ToDebugSNode::to_node(#bi, opts.clone()));));
+            stmts.push(quote!(nodes.push(crate::printer::ToDebugSNode::to_node(#bi, opts));));
         }
         stmts.into_iter().collect::<proc_macro2::TokenStream>()
     });
```

source/rust_verify/src/try_broadcasts.rs
```diff
@@ -0,0 +1,801 @@
...
+                } else {
+                    krate_fun.clone()
+                }
+            })
+            .collect::<Vec<_>>();
+        cloned_krate.functions = new_funcs;
+        // Since we inserted broadcasts that may not have been there before, the
+        // call graph of the program may have changed. To guard against
+        // introducing cyclic references in `proof fn`s, we recompute the call
+        // graph by creating a new `GlobalCtx`.
```
Chris-Hawblitzel, line 560: Can you add a failing test to rust_verify_test to test that sledgehammer doesn't fail to update the call graph (in the case of cycles being introduced)?

source/rust_verify/src/verifier.rs
```diff
@@ -1967,6 +1982,10 @@ impl Verifier {
             reporter
                 .report_now(&note_bare(format!("verifying {bucket_name}{functions_msg}")).to_any());
         }
+        let (sh_result, new_ctx) =
+            sledgehammer(self, reporter, krate, source_map, bucket_id, global_ctx)?;
```
Chris-Hawblitzel: Rather than always calling this (which carries some per-bucket performance overhead, like computing fun_to_function), could we precompute a bool for the whole `current_vir_crate` to see if it has any sledgehammer attributes, and only call into sledgehammer if this bool is true?

Chris-Hawblitzel (approved):

Chris-Hawblitzel:
> * Lemmas from other crates are only considered if they are referenced by the current crate; otherwise they are pruned prior to verification.

If any function in `current_vir_crate` uses sledgehammer, should we have `prune.rs` automatically retain all broadcasts from imported crates?

source/rust_verify/src/sledgehammer.rs
dschoepe: I think I'd prefer to keep the extra code in `verifier.rs` minimal.

source/docs/guide/src/sledgehammer.md
dschoepe: Unfortunately this would require adding some more heuristics to detect which `broadcast`s are already used, which can be finicky if they occur inside `assert(..) by { .. }` blocks or on individual branches. I'll add a note about it.

source/rust_verify/src/sledgehammer.rs
dschoepe: Yes, that's the reason. I'll try to fix that in the next version.

source/rust_verify/src/config.rs
dschoepe: That's much nicer, thanks for looking into this.

source/rust_verify/src/verifier.rs
dschoepe: Good call, done.

source/rust_verify/src/try_broadcasts.rs
dschoepe: I added filtering to avoid trying lemmas that would create a cycle, and a test case for that instead.

source/rust_verify/src/try_broadcasts.rs
dschoepe: That's the initial number of parts to divide the set of lemmas into. The search procedure will increment this if neither half results in a working proof.

source/rust_verify/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![feature(rustc_private)]
 #![feature(box_patterns)]
+// #![feature(get_mut_unchecked)]
```
parno, line 3: Should these commented lines be removed?

source/rust_verify/src/verifier.rs
```diff
@@ -1897,21 +1940,25 @@ impl Verifier {
...
                 (rlimit_count.0 + spunoff_rlimit_count.0, rlimit_count.1 + spunoff_rlimit_count.1)
             }),
+            verification_outcome,
+            name_ctxt: ctx.name_ctxt.clone(),
         })
     }
 
-    fn verify_bucket_outer(
+    /// Verifies the given bucket but does not emit logs or update timers.
+    /// TODO: Find a better name for this
```
parno: Perhaps reduce this to a `//` comment, so it's not a documentation-level comment?

parno (approved):
Looks reasonable to me -- thanks for making the various updates!

PR #2049, opened 2025-12-20
[cargo-verus] feat: `focus` subcommand

mkovaxx:
- Introduces the `focus` subcommand, which behaves the same as `verify` for top-level packages, but excludes dependencies from verification.
- All packages reachable from the top-level selection are still processed with `verus`, but deps are treated with the `--no-verify` option.
- In `focus` mode, output artifacts are separated to prevent e.g. unverified `.vir` files from mistakenly being picked up by subsequent complete (`build`, `verify`, `check`) runs.
- Separation is achieved by overriding Cargo's `--target-dir`, appending `/verus-partial/`.

### DIFF
--- a/source/cargo-verus/src/subcommands.rs
+++ b/source/cargo-verus/src/subcommands.rs
@@ -124,10 +124,11 @@ pub fn run_cargo(cfg: CargoRunConfig) -> Result<ExitCode> {
 
     let cargo_args = {
         let mut options = cfg.options.cargo_opts;
-        if !cfg.verify_deps && options.target_dir.is_none() {
+        if !cfg.verify_deps {
             // Ensure that partially verified artifacts are separated from complete results
-            options.target_dir =
-                Some(metadata.target_directory.clone().into_std_path_buf().join("verus-partial"));
+            let target_dir =
+                options.target_dir.unwrap_or(metadata.target_directory.clone().into_std_path_buf());
+            options.target_dir = Some(target_dir.join("verus-partial"));
         }
 
         let for_cargo_metadata = false;


mkovaxx:
@parno @Chris-Hawblitzel @tjhance Quick question about this feature.

My understanding is that when `focus` is invoked on a single package, there should be exactly one `__VERUS_DRIVER_VERIFY_{package}` entry in the env, and its corresponding `__VERUS_DRIVER_ARGS_FOR_{package}` entry should not have any `import-dep-if-present` settings. (And similarly for each selected package in a workspace.) Is that correct?

parno:
> there should be exactly one __VERUS_DRIVER_VERIFY_{package} entry in the env,
> __VERUS_DRIVER_ARGS_FOR_{package} entry should not have any import-dep-if-present settings.

Hmm, I'm not sure about these (or perhaps I'm misunderstanding the question).  For any package that has a `verify = true` entry in its Cargo.toml, I think we still need to run Verus on it (with `--no-verify` when we're not processing the selected crate) to generate the `.vir` meta-data file.  And we'll still need to tell Verus about those `.vir` files.

mkovaxx:
Note from Bryan: We need to ensure that the generated `.vir` artifacts are not picked up by subsequent `cargo verify` runs, by e.g. renaming or emitting into a different `TARGET_DIR`.

source/cargo-verus/src/subcommands.rs
```diff
@@ -206,16 +233,33 @@ fn make_cargo_args(opts: &CargoOptions, for_cargo_metadata: bool) -> Vec<String>
...
+    verus_args: &'a [String],
+    warn_if_nothing_verified: bool,
+}
+
+fn make_cargo_command(cfg: CargoCommandConfig) -> Result<(Command, bool)> {
+    let mut common_verus_driver_args: Vec<String> =
+        vec!["--VIA-CARGO".to_owned(), "compile-when-not-primary-package".to_owned()];
+
+    if !cfg.warn_if_nothing_verified {
+        common_verus_driver_args.extend_from_slice(&[
```
parno: It looks like we're adding the same arguments here as we did on line 248 above, which doesn't seem right.

source/cargo-verus/tests/test_focus.rs
```diff
@@ -0,0 +1,128 @@
...
+    assert_eq!(
+        data.args,
+        vec!["build", "--manifest-path", manifest_path.to_str().expect("manifest path to string")]
+    );
+
+    data.assert_env_has("RUSTC_WRAPPER");
+    data.assert_env_sets("__CARGO_DEFAULT_LIB_METADATA", "verus");
+    data.assert_env_sets("__VERUS_DRIVER_VIA_CARGO__", "1");
+    data.assert_env_sets_key_prefix(&verify_crate_prefix, "1");
+    data.assert_env_has_no_key_prefix(&verify_for_crate_prefix);
```
parno, line 42: Why don't we expect to find this prefix?

source/cargo-verus/src/subcommands.rs
```diff
@@ -89,13 +91,41 @@ verify = true
...
+    /////////////////////////////////////////////////
+    // Phase 2: run Verus via `cargo {subcommand}` //
+    /////////////////////////////////////////////////
+
+    let cargo_args = {
+        let for_cargo_metadata = false;
+        make_cargo_args(cargo_options, for_cargo_metadata)
+    };
+
+    let mut common_verus_driver_args: Vec<String> = vec![];
```
mkovaxx: @parno Moved `common_verus_driver_args` here so it's easier to see what's going on.

mkovaxx:
@parno Apologies for the mess here. I made a mistake committing some changes to this branch that I intended to only experiment with. Cleaned those up now and going to send them as a follow-up PR instead.

source/cargo-verus/src/subcommands.rs
```diff
@@ -87,37 +89,71 @@ verify = true
...
     let mut common_verus_driver_args: Vec<String> =
         vec!["--VIA-CARGO".to_owned(), "compile-when-not-primary-package".to_owned()];
 
-    if !warn_if_nothing_verified {
+    if cfg.compile_primary {
         common_verus_driver_args.extend_from_slice(&[
             "--VIA-CARGO".to_owned(),
             "compile-when-primary-package".to_owned(),
         ]);
     }
```
mkovaxx, line 146: @parno This is fixing an earlier mistake. For reference, original behavior from before the `clap` refactor:
https://github.com/verus-lang/verus/blob/d5f430f1cf26268603a9d9a3f8d8371920d997aa/source/cargo-verus/src/main.rs#L221-L231

source/cargo-verus/src/subcommands.rs
```diff
@@ -87,37 +89,71 @@ verify = true
...
-    subcommand: &str,
-    cargo_options: &CargoOptions,
-    verus_args: &[String],
-    warn_if_nothing_verified: bool,
-) -> Result<ExitCode> {
-    let cargo_args = make_cargo_args(cargo_options, false);
+pub struct CargoRunConfig {
+    pub subcommand: &'static str,
+    pub options: VerifyCommand,
+    pub compile_primary: bool,
```
mkovaxx, line 95: The "new" `compile_primary` flag is the negative of `just_verify` from before the `clap` refactor:
https://github.com/verus-lang/verus/blob/d5f430f1cf26268603a9d9a3f8d8371920d997aa/source/cargo-verus/src/main.rs#L202-L211

source/cargo-verus/tests/test_focus.rs
mkovaxx: That env var (` __VERUS_DRIVER_ARGS_FOR_`) is only set if the package has deps or special settings (`--is-core`, `--is-vstd`, `--no-vstd`, `--no-verify`). In this test case, the package has neither.

The other subcommands have the same behavior. Is that incorrect?

Relevant lines for context:
https://github.com/verus-lang/verus/blob/ed8ac9440bda1f77b878e32fe996373331e4c6fd/source/cargo-verus/src/subcommands.rs#L303-L335

source/cargo-verus/src/subcommands.rs
```diff
@@ -87,37 +89,78 @@ verify = true
...
+    // Phase 2: run Verus via `cargo {subcommand}` //
+    /////////////////////////////////////////////////
+
+    let cargo_args = {
+        let mut options = cfg.options.cargo_opts;
+        if !cfg.verify_deps && options.target_dir.is_none() {
+            // Ensure that partially verified artifacts are separated from complete results
+            options.target_dir =
+                Some(metadata.target_directory.clone().into_std_path_buf().join("verus-partial"));
+        }
```
mkovaxx, line 132: @parno This ensures that the `target_dir` setting of `focus` is separate from complete verification runs.

source/cargo-verus/tests/test_focus.rs
```diff
@@ -0,0 +1,173 @@
...
+    let hasdeps = "hasdeps";
+
+    let workspace_dir = MockWorkspace::new()
+        .members([
+            MockPackage::new(optin).lib().verify(true),
+            MockPackage::new(optout).lib().verify(false),
+            MockPackage::new(unset).lib(),
+            MockPackage::new(hasdeps)
+                .lib()
+                .deps([MockDep::workspace(optin), MockDep::workspace(optout)])
```
mkovaxx, line 123: @parno In this test, `hasdeps` now depends on both `optin` and `optout` to check that `focus` respects the `verify = false` setting on `optout`.

source/cargo-verus/tests/test_focus.rs
parno: Ah, I see.  Thanks.  I was hazy on its purpose.

source/cargo-verus/src/subcommands.rs
parno: Makes sense -- good catch.

source/cargo-verus/src/subcommands.rs
parno: Good, thanks for adding this.

parno: One additional question here: Suppose a team decides they want to locate the target directory somewhere else, so they always pass `--target custom_target_dir` when using cargo verus, both with full verification and with partial verification.  It looks like the logic here (specifically the check for `options.target_dir.is_none()`) would collapse the two sets of results together, right?  I think we might want to always create a `verus-partial` sub-directory of the `--target` the user sets (or whatever it defaults to if the user doesn't specify).

mkovaxx: Yes, that's correct. I was thinking that once `--target-dir` is overridden, it might be a good idea to simply respect that setting. The scenario you describe also makes sense, and it seems like that alternative behavior would result in fewer surprises, so I'll update the code to always put partial artifacts under `verus-partial`.

parno: Okay, thanks, I think taking the conservative approach here makes sense.

PR #267, opened 2022-08-24
Add a simplified version of the quantifier instantiation profiler (--use-internal-profiler)

utaal:
It only counts instantiations, and reports counts for builtin/internal quantifiers.

It is intended, mainly, to find potential issues with internal quantifiers, not as a tool for the end user.

Sample output:

```
note:  1. Quantifier constructor_accessor_axiom, instantiations: 144041
          at: datatype#225, instantiations: 31253
          at: datatype#63, instantiations: 14099
          at: datatype#54, instantiations: 14099
          at: datatype#57, instantiations: 14099
          at: datatype#60, instantiations: 14099
          at: datatype#235, instantiations: 8070
          at: datatype#232, instantiations: 8070
          at: datatype#21, instantiations: 3790
          at: datatype#216, instantiations: 2628
          at: datatype#219, instantiations: 2628
          at: datatype#138, instantiations: 1673
          at: datatype#126, instantiations: 1673
          at: datatype#144, instantiations: 1673
          at: datatype#135, instantiations: 1673
          at: datatype#141, instantiations: 1673
          at: datatype#147, instantiations: 1673
          at: datatype#129, instantiations: 1673
          at: datatype#132, instantiations: 1673
          at: datatype#69, instantiations: 1649
          at: datatype#96, instantiations: 1319
          at: datatype#99, instantiations: 1319
          at: datatype#93, instantiations: 1319
          at: datatype#203, instantiations: 1180
          at: datatype#200, instantiations: 1180
          at: datatype#36, instantiations: 1126
          at: datatype#39, instantiations: 1126
          at: datatype#42, instantiations: 1126
          at: datatype#45, instantiations: 1126
          at: datatype#27, instantiations: 692
          at: datatype#82, instantiations: 348
          at: datatype#85, instantiations: 348
          at: datatype#193, instantiations: 270
          at: datatype#163, instantiations: 270
          at: datatype#181, instantiations: 270
          at: datatype#187, instantiations: 270
          at: datatype#190, instantiations: 270
          at: datatype#178, instantiations: 270
          at: datatype#175, instantiations: 270
          at: datatype#169, instantiations: 270
          at: datatype#184, instantiations: 270
          at: datatype#172, instantiations: 270
          at: datatype#166, instantiations: 270
          at: datatype#75, instantiations: 261
          at: datatype#15, instantiations: 168
          at: datatype#209, instantiations: 164
          at: datatype#110, instantiations: 79
          at: datatype#113, instantiations: 79
[...]

PR #2052, opened 2025-12-21
use smt-scope in profiler

ahuoguo:
As discussed in https://github.com/verus-lang/verus/discussions/345, z3tracer is super slow. smt-scope (an updated version of axiom profiler) is a library for analyzing quantifier instantiation and matching loops, and is under maintenance by the Viper group. I see a great improvement in parsing the smt log file (from ~10 mins to seconds for the matching loop example).

Here are some questions that are not resolved yet. Please let me know what y'all think to proceed:

- [x] It appears to me that the cost for each node on the instantiation graph is computed the same way: https://github.com/viperproject/smt-scope/issues/140, but they disagree on examples. For example, running the example in the tutorial https://verus-lang.github.io/verus/guide/profiling.html will produce:
```
note: Log analysis complete for (profile rerun) profile::trigger_forever2

note: Profile statistics for profile::trigger_forever2

note: Observed 17,963 total instantiations of user-level quantifiers

note: Cost * Instantiations: 2269911826 (Instantiated 8,981 times - 49% of the total, cost 252746) top 1 of 3 user-level quantifiers.
  --> profile.rs:14:93
   |
14 | ...   forall|x: nat, y: nat| f(x + 1, 2 * y) && f(2 * x, y + x) || f(y, x) ==> #[trigger] f(x, y),
   |       ------------------------------------------------------------------------------------^^^^^^^
   |       |
   |       Triggers selected for this quantifier

note: Cost * Instantiations: 397732566 (Instantiated 8,981 times - 49% of the total, cost 44286) top 2 of 3 user-level quantifiers.
  --> profile.rs:13:32
   |
13 |         forall|x: nat, y: nat| h(x, y) == f(x, y),
   |         -----------------------^^^^^^^----^^^^^^^
   |         |
   |         Triggers selected for this quantifier

note: Cost * Instantiations: 3 (Instantiated 1 times - 0% of the total, cost 3) top 3 of 3 user-level quantifiers.
  --> profile.rs:12:24
   |
12 |         forall|x: nat| g(x),
   |         ---------------^^^^
   |         |
   |         Triggers selected for this quantifier

verification results:: 0 verified, 1 errors
```
instead of 
```
note: Log analysis complete for (profile rerun) profile::trigger_forever2

note: Profile statistics for profile::trigger_forever2

note: Observed 17,963 total instantiations of user-level quantifiers

note: Cost * Instantiations: 3193014930 (Instantiated 8,981 times - 49% of the total, cost 355530) top 1 of 3 user-level quantifiers.
  --> profile.rs:14:93
   |
14 | ...   forall|x: nat, y: nat| f(x + 1, 2 * y) && f(2 * x, y + x) || f(y, x) ==> #[trigger] f(x, y),
   |       ------------------------------------------------------------------------------------^^^^^^^
   |       |
   |       Triggers selected for this quantifier

note: Cost * Instantiations: 647637872 (Instantiated 8,981 times - 49% of the total, cost 72112) top 2 of 3 user-level quantifiers.
  --> profile.rs:13:32
   |
13 |         forall|x: nat, y: nat| h(x, y) == f(x, y),
   |         -----------------------^^^^^^^----^^^^^^^
   |         |
   |         Triggers selected for this quantifier

note: Cost * Instantiations: 7 (Instantiated 1 times - 0% of the total, cost 7) top 3 of 3 user-level quantifiers.
  --> profile.rs:12:24
   |
12 |         forall|x: nat| g(x),
   |         ---------------^^^^
   |         |
   |         Triggers selected for this quantifier

verification results:: 0 verified, 1 errors
```

Maybe I've done something wrong, or the cost calculation do not agree. But they seem to be on the same order of magnitude, at least.

- [x] I provided minimum support for the other two flags: `-V use-internal-profiler`, which seems to print all instatiation, and `-V capture-all-profiles`, which writes the instatiation graph to a file. I think the latter might be replaced by some feature in `smt-scope`. But I will leave it as a future improvement.

### DIFF
--- a/source/air/src/profiler.rs
+++ b/source/air/src/profiler.rs
@@ -248,7 +248,7 @@ impl Profiler {
         for cost in &self.quantifier_stats {
             let count = cost.instantiations;
             let msg = format!(
-                "Instantiated heyhey{}heyhey {} times ({}% of the total)\n",
+                "Instantiated {} {} times ({}% of the total)\n",
                 cost.quant,
                 count,
                 100 * count / self.total_instantiations()
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -393,7 +393,6 @@ pub fn parse_args_with_imports(
     const EXTENDED_NO_SOLVER_VERSION_CHECK: &str = "no-solver-version-check";
     const EXTENDED_SPINOFF_ALL: &str = "spinoff-all";
     const EXTENDED_CAPTURE_PROFILES: &str = "capture-profiles";
-    const EXTENDED_USE_INTERNAL_PROFILER: &str = "use-internal-profiler";
     const EXTENDED_CVC5: &str = "cvc5";
     const EXTENDED_ALLOW_INLINE_AIR: &str = "allow-inline-air";
     const EXTENDED_USE_CRATE_NAME: &str = "use-crate-name";
@@ -412,10 +411,6 @@ pub fn parse_args_with_imports(
         (
             EXTENDED_CAPTURE_PROFILES,
             "Always collect prover performance data, but don't generate output reports",
-        ),
-        (
-            EXTENDED_USE_INTERNAL_PROFILER,
-            "Use an internal profiler that shows internal quantifier instantiations",
         ),
         (EXTENDED_CVC5, "Use the cvc5 SMT solver, rather than the default (Z3)"),
         (EXTENDED_ALLOW_INLINE_AIR, "Allow the POTENTIALLY UNSOUND use of inline_air_stmt"),


ahuoguo:
@parno Can you take a look?

parno:
Yep, will do, as soon as I can.  Classes start today, so I've been pretty swamped.  This is on my queue though.

source/air/src/profiler.rs
```diff
@@ -92,103 +86,137 @@ impl Profiler {
...
+        inst_graph: &InstGraph,
+    ) -> Result<InstantiationGraph, ProfilerError> {
+        // Convert smt-scope's graph structure to our InstantiationGraph format
+        let mut edges: HashMap<(u64, usize), HashSet<(u64, usize)>> = HashMap::new();
+        let mut names: HashMap<(u64, usize), String> = HashMap::new();
+        let mut nodes: HashSet<(u64, usize)> = HashSet::new();
+
+        // Iterate through all instantiations
+        for (inst_idx, inst) in parser.instantiations().iter_enumerated() {
+            let inst_id = (usize::from(inst_idx) as u64, 0); // Convert InstIdx to (u64, usize) tuple
```
parno: What does the second element of the tuple represent?

source/air/src/profiler.rs
```diff
@@ -218,168 +246,50 @@ impl Profiler {
...
-        }
-
-        fn pop(&mut self, _num: u64, _current_level: u64) -> RawResult<()> {
-            Ok(())
-        }
-
-        fn resolve_lit(&mut self, _i: u64, _lit: Literal) -> RawResult<()> {
-            Ok(())
-        }
+    pub fn profile(path: &std::path::Path) -> Vec<(String, u64, Vec<(String, u64)>)> {
```
parno: It seems like there's a fair bit of overlap here with the code above.  If the only difference is whether to filter out non-user quantifiers, can we add that as a toggle above and avoid duplication?  Or are there additional semantic differences?

source/docs/guide/src/profiling.md
```diff
@@ -43,36 +43,46 @@ If we run the profiler on the example above, we'll see something along the lines
...
 
-note: Observed 27,184 total instantiations of user-level quantifiers
+Z3 4.12.5
+note: Log analysis complete for (profile rerun) trigger_loops::trigger_forever2
 
-note: Cost * Instantiations: 5391549700 (Instantiated 13,591 times - 49% of the total, cost 396700) top 1 of 3 user-level quantifiers.
-  --> ../examples/trigger_loops.rs:68:78
+note: Profile statistics for trigger_loops::trigger_forever2
+
+note: Observed 17,963 total instantiations of user-level quantifiers
```
parno, line 58: It looks like we're seeing slightly more than half the number of user instantiations as we did with the old profiler.  Do you know why that might be the case?  If anything, I might have expected the newer code to discover/record more.

source/docs/guide/src/profiling.md
```diff
@@ -43,36 +43,46 @@ If we run the profiler on the example above, we'll see something along the lines
...
-  --> ../examples/trigger_loops.rs:64:1
+  --> ../examples/trigger_loops.rs:66:7
    |
-64 | fn trigger_forever2() {
-   | ^^^^^^^^^^^^^^^^^^^^^
+66 | proof fn trigger_forever2()
+   |       ^^^^^^^^^^^^^^^^^^^^^
 
-Analyzing prover log...
-[00:00:39] ████████████████████████████████████████████████████████████████████████████████ 1153/1153K lines
```
parno, line 52: Is it the case that the new profiler is fast enough that we don't need a progress bar?

source/air/src/profiler.rs
```diff
@@ -46,43 +53,30 @@ impl Profiler {
         message_interface: std::sync::Arc<dyn crate::messages::MessageInterface>,
         filename: &std::path::Path,
         description: Option<&str>,
-        progress_bar: bool,
         diagnostics: &impl Diagnostics,
-        compute_instantiation_graph: bool,
```
parno, line 51: Why don't we need this flag any more?

parno (commented):
Thanks for working on this -- it'll be nice to have a more recent profiler and even better, a much faster profiler!

Have you tried this on some code examples that time out in practice and compared the ranking of the quantifiers each version produces?  It's not vital to match the counts exactly, but if the produce different orderings, that might be more concerning.

As a minor note, we usually try to topologically sort the functions, so that callees precede their callers (e.g., `make_instantiation_graph` should precede `parse`).  It seems like that wasn't previously the case in this file, but we might as well fix it while we're overhauling everything.

source/docs/guide/src/profiling.md
ahuoguo: I believe so

source/air/src/profiler.rs
ahuoguo: Rather than a model, I always get an InstGraph from smt-scope

source/air/src/profiler.rs
ahuoguo: the data from `[instance]`

source/air/src/profiler.rs
ahuoguo: I was hesitant about moving this above since it's a rather unstable feature that's rarely used

source/docs/guide/src/profiling.md
ahuoguo: The old profiler has the same number of instantiations. The output in the `profiling.md` is quite outdated. (I don't know why there's a difference between the number instatiations, maybe there were some encoding changes...)

Here's the old profiler output:
```
error: function body check: Resource limit (rlimit) exceeded
  --> profile.rs:10:7
   |
10 | proof fn trigger_forever2()
   |       ^^^^^^^^^^^^^^^^^^^^^

note: Analyzing prover log for (profile rerun) profile::trigger_forever2 ...

[00:00:03] ████████████████████████████████████████████████████████████████████████████████ 1035/1035K lines
note: Log analysis complete for (profile rerun) profile::trigger_forever2

note: Profile statistics for profile::trigger_forever2

note: Observed 17,963 total instantiations of user-level quantifiers

note: Cost * Instantiations: 3193014930 (Instantiated 8,981 times - 49% of the total, cost 355530) top 1 of 3 user-level quantifiers.
  --> profile.rs:14:93
   |
14 |         forall|x: nat, y: nat| f(x + 1, 2 * y) && f(2 * x, y + x) || f(y, x) ==> #[trigger] f(x, y),
   |         ------------------------------------------------------------------------------------^^^^^^^
   |         |
   |         Triggers selected for this quantifier

note: Cost * Instantiations: 799650278 (Instantiated 8,981 times - 49% of the total, cost 89038) top 2 of 3 user-level quantifiers.
  --> profile.rs:13:32
   |
13 |         forall|x: nat, y: nat| h(x, y) == f(x, y),
   |         -----------------------^^^^^^^----^^^^^^^
   |         |
   |         Triggers selected for this quantifier

note: Cost * Instantiations: 7 (Instantiated 1 times - 0% of the total, cost 7) top 3 of 3 user-level quantifiers.
  --> profile.rs:12:24
   |
12 |         forall|x: nat| g(x),
   |         ---------------^^^^
   |         |
   |         Triggers selected for this quantifier

verification results:: 0 verified, 1 errors
```

source/docs/guide/src/profiling.md
parno: Exciting!

source/air/src/profiler.rs
parno: Could you add a comment indicating what that data is accumulating?

source/air/src/profiler.rs
parno: If it's rarely used, we could consider cutting it entirely.  If we want to keep it, I think we should try to consolidate the code, rather than having to remember to make updates in two places if we make fixes/improvements.

source/docs/guide/src/profiling.md
parno: > The old profiler has the same number of instantiations.

Oh good, that's reassuring and a good sanity check on the new implementation.

ahuoguo:
> Have you tried this on some code examples that time out in practice and compared the ranking of the quantifiers each version produces? It's not vital to match the counts exactly, but if the produce different orderings, that might be more concerning.

Aha! thanks for urging me to do this. Actually found some bugs (The `QuantID`s are not unique, and I need to merge them).

I tried on this function in mimaloc:

```
verus --triggers-mode=silent --no-auto-recommends-check --rlimit 2 --extern libc=build/liblibc.rlib ./verus-mimalloc/lib.rs --verify-only-module page_organization::PageOrg --verify-function take_page_from_unused_queue_ll_inv_valid_unused --profile
```

Here's the output after the fix, the ordering is mostly the same (the last two are different)
[old_profiler.txt](https://github.com/user-attachments/files/24809886/old_profiler.txt)
[smt-scope.txt](https://github.com/user-attachments/files/24809888/smt-scope.txt)

source/air/src/profiler.rs
ahuoguo: Based on the description in https://github.com/verus-lang/verus/pull/267, I think we can remove `use-internal-profiler`

parno: Indeed, looking back, it seems like performance was the main motivation, so now that you've solved that, we can remove this.  It might be worth keeping a toggle in the main profiler to enable reporting internal quantifiers too, in case a developer needs that.

parno:
> Here's the output after the fix, the ordering is mostly the same (the last two are different) [old_profiler.txt](https://github.com/user-attachments/files/24809886/old_profiler.txt) [smt-scope.txt](https://github.com/user-attachments/files/24809888/smt-scope.txt)

Okay, those look pretty close.  Why does the new output include the Z3 version info (`Z3 4.12.5`)?

ahuoguo:
> > Here's the output after the fix, the ordering is mostly the same (the last two are different) [old_profiler.txt](https://github.com/user-attachments/files/24809886/old_profiler.txt) [smt-scope.txt](https://github.com/user-attachments/files/24809888/smt-scope.txt)
> 
> Okay, those look pretty close. Why does the new output include the Z3 version info (`Z3 4.12.5`)?

That's spit out by smt-scope. I don't want to block smt-scope output in case it reported error

source/air/src/profiler.rs
ahuoguo: I'm going to remove this for now, as I might do more quantifier analysis using smt-sccope with later PRs.

parno:
> That's spit out by smt-scope. I don't want to block smt-scope output in case it reported error

Ah, I see.  That seems reasonable.

source/air/src/profiler.rs
parno: Okay, sounds good.

source/air/src/profiler.rs
```diff
@@ -203,7 +248,7 @@ impl Profiler {
         for cost in &self.quantifier_stats {
             let count = cost.instantiations;
             let msg = format!(
-                "Instantiated {} {} times ({}% of the total)\n",
+                "Instantiated heyhey{}heyhey {} times ({}% of the total)\n",
```
parno: This wasn't what I was expecting to see...

source/rust_verify/src/config.rs
```diff
@@ -810,7 +808,6 @@ pub fn parse_args_with_imports(
             extended.contains_key(EXTENDED_CAPTURE_PROFILES)
         },
         spinoff_all: extended.contains_key(EXTENDED_SPINOFF_ALL),
-        use_internal_profiler: extended.contains_key(EXTENDED_USE_INTERNAL_PROFILER),
```
parno, line 813: Are there other parts that should be removed?  E.g., I don't see `EXTENDED_USE_INTERNAL_PROFILER` being removed.

source/air/src/profiler.rs
ahuoguo: Oops...

source/rust_verify/src/config.rs
ahuoguo: I think this is the only one that I missed

PR #2055, opened 2025-12-23
infer mutability in ast_to_sst

tjhance:
Ignore user-provided 'mut' annotations; always determine mutability based on whether a mutation actually occurs. This is needed for new-mut-ref, but this PR applies to all code.

We remove `init_not_mut`; a variable which is uninitialized-but-assigned-later is just considered to be mutable.

This changes the behavior in some cases related to spec variables, so look over the tests carefully.

depends on https://github.com/verus-lang/verus/pull/2043

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -767,6 +767,7 @@ pub struct PatternBinding {
     pub by_ref: ByRef,
     pub typ: Typ,
     /// Marked mutable at the source level? Use None for variables introduced in lowering passes.
+    /// TODO: This field can probably be removed.
     pub user_mut: Option<bool>,
     /// True if the type of this variable is copy.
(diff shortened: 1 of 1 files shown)


source/vir/src/ast_simplify.rs
```diff
@@ -453,7 +448,7 @@ fn simplify_one_expr(
             // which comes after our lifetime checking pass.
             match scope_map.get(x) {
                 None => Err(error(&expr.span, "Verus Internal Error: cannot find this variable")),
-                Some(entry) if !entry.is_mut && entry.init => {
+                Some(entry) if entry.user_mut == Some(false) && entry.init => {
```
Chris-Hawblitzel, line 451: Is this the only thing left that uses user_mut?  If so, it wouldn't bother me if we removed user_mut entirely.

Chris-Hawblitzel (approved):

PR #2009, opened 2025-11-24
Add more APIs into `std_specs`

hiroki-chen:
This PR does three things:

1. Implement `ExDefault` indicated in #1951.
2. Implement slice APIs:
  - Sorted-related: `binary_search_*`, `is_sorted_`.
  - Slice manipulations `reverse`, `rotate_`, `swap`, etc.
  - Misc: `contains`, `starts_with`, `ends_with`, etc.
3. Move `unreachable_unchecked` into the correct place `control_flow`.

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #2064, opened 2026-01-01
add default trait, some option specs and string deref

sssxks:
changes:

- Added default trait (by using DefaultSpec via DefaultSpecImpl)
  - cherry picked from https://github.com/verus-lang/verus/pull/2009/ (credit: @hiroki-chen)
  - implemented DefaultSpecImpl for a few more std types like `Option<T>`, `&str`, `PhantomData<T>`, `(U, T)`, `(V, U, T)`
  - model Default::default() for String, Vec, VecDeque, HashSet, HashMap using Skolem constants.
  - added tests, including a usage for implementing this on custom types (may need newtype because of orphan rule)

- Added option and string specs
  - Option: cloned, and_then, ok_or_else, unwrap_or_default, unwrap_or_else
  - String: Deref<Target=str>
  - test covering these behavior.
 
- remove obsolete boundary suggestion tests for Option::and_then.

- fix verusdoc panic on omitted -> () and preserve return-annotation

### DIFF
--- a/source/rust_verify_test/tests/boundary_suggestions.rs
+++ b/source/rust_verify_test/tests/boundary_suggestions.rs
@@ -124,6 +124,38 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] test_assume_specification_foreign_suggestion_made code! {
+        use vstd::prelude::*;
+
+        verus! {
+            pub fn bar<T>(o: Option<T>)-> Option<T> {
+                o.inspect(|_x: &T| {})
+            }
+        }
+    } => Err(err) => assert_help_error_msg(err, "pub assume_specification<T, F> [std::option::Option::<T>::inspect] (_0: std::option::Option<T>, _1: F) -> std::option::Option<T>")
+}
+test_verify_one_file! {
+    #[test] test_assume_specification_foreign_suggestion_correct code! {
+        use vstd::prelude::*;
+
+        verus! {
+            #[verifier::external_trait_specification]
+            pub trait ExDestruct {
+                type ExternalTraitSpecificationFor: core::marker::Destruct;
+            }
+
+            pub assume_specification<T, F> [std::option::Option::<T>::inspect] (_0: std::option::Option<T>, _1: F) -> std::option::Option<T>
+            where
+            F: std::ops::FnOnce(&T,) -> () + core::marker::Destruct,;
+
+            pub fn bar<T>(o: Option<T>)-> Option<T> {
+                o.inspect(|_x: &T| {})
+            }
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
     #[test] test_assume_specification_str_eq_suggestion_made code! {
         use vstd::prelude::*;
 
--- a/source/rust_verify_test/tests/default_trait.rs
+++ b/source/rust_verify_test/tests/default_trait.rs
@@ -112,6 +112,17 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+  #[test] test_default_box_vec_u32_ok verus_code! {
+    use vstd::prelude::*;
+    fn test() {
+      let b: Box<Vec<u32>> = Box::default();
+      let v: Vec<u32> = *b;
+      assert(v@.len() == 0);
+    }
+  } => Ok(())
+}
+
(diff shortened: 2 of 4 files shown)


sssxks:
@tjhance CI is green, could you take a look?

source/rust_verify_test/tests/boundary_suggestions.rs
```diff
@@ -123,35 +123,6 @@ test_verify_one_file! {
     } => Ok(())
 }
 
-test_verify_one_file! {
```
tjhance, line 126: We should fix these tests rather than deleting them, e.g. switch out `and_then` for a function that's unlikely to be supported in the future

source/vstd/std_specs/option.rs
```diff
@@ -212,6 +213,62 @@ pub assume_specification<T, U, F: FnOnce(T) -> U>[ Option::<T>::map ](a: Option<
...
+    option: Option<T>,
+    err: F,
+) -> (res: Result<T, E>)
+    requires
+        option.is_none() ==> err.requires(()),
+    ensures
+        option.is_some() ==> res == Ok::<T, E>(option.unwrap()),
+        option.is_none() ==> {
+            &&& res.is_err()
+            &&& exists|v: E| #![auto] err.ensures((), v) && res == Err::<T, E>(v)
```
tjhance: please use explicit #[trigger] annotations rather than auto for user-facing vstd specs

source/vstd/std_specs/core.rs
```diff
@@ -206,3 +221,122 @@ pub fn index_set<T, Idx, E>(container: &mut T, index: Idx, val: E) where
 #[verifier::external_body]
 #[verifier::accept_recursive_types(T)]
 pub struct ExAssertParamIsClone<T: Clone + PointeeSized>(core::clone::AssertParamIsClone<T>);
+
```
tjhance: As a matter of code organization, we should put these specs in their own file.

tjhance (changes requested):
It should be possible to write a more general specification for `Default` that doesn't rely on `obeys_default_spec`. For example, rather than:

```rust
pub assume_specification<T: core::default::Default>[ <Box<
    T,
> as core::default::Default>::default ]() -> (res: Box<T>)
    ensures
        T::obeys_default_spec() ==> *res == T::default_spec(),
;
```

We could say:

```rust
pub assume_specification<T: core::default::Default>[ <Box<
    T,
> as core::default::Default>::default ]() -> (res: Box<T>)
    ensures
        call_ensures(T::default, (), (*res,))
;
```

I would prefer to use the generally applicable specs.

I'm also a little unsure if it's really correct to consider `Vec::<T>::default` to be deterministic in the manner as described by `obeys_default_spec`. In the future, for example, we might want to modify the specification of `Vec` to expose 'address' and 'capacity' information. Are those always identical for a newly constructed vec? (I actually don't know the answer to this off the top of my head.)

sssxks:
@tjhance Thanks for your suggestion, does this look good to you now?

removes the `external_trait_extension(DefaultSpec via DefaultSpecImpl)`, why:
- after I swapped for call_ensures, all code outside `default.rs` naturally doesn't use DefaultSpec or DefaultSpecImpl anymore
  - take a detour into call_ensures and f.ensures, they seems interchangeable for exec fn, so I choose `f.ensures` because it looks nicer
- then I added the test `test_default_custom_type_ensures_in_smart_ptrs`, and i realized user are forced to impl DefaultSpecImpl and setting obeys_default_spec() to false for their new custom type implementing Default. this is extra obligation and does nothing meaningful
- so no trait spec extension is better, and it works - not needed actually.

for the trigger thing you mentioned: after taking a detour into logic, i simplified `exists v. A(v) && res == v` with `A(res)`, so it doesn't require trigger anymore.

mkovaxx:
@sssxks @tjhance This change seems to be causing some tests to fail.

Here's how I'm running the tests:
```
vargo nextest run --release -p rust_verify_test
```

The tests pass at the [previous commit](https://github.com/verus-lang/verus/commit/d101191a04c9833b085c1e90773a32605ee17095), but fail at the current [`HEAD`](https://github.com/verus-lang/verus/commit/5552af05bc614205517b8e1a395fa7520c07b6fc).

PR #2067, opened 2026-01-07
Only specify AtomicU64/AtomicI64 when supported

jaylorch:


parno (approved):
Thanks for tracking this down and fixing it!

PR #2070, opened 2026-01-08
Clean up erasure computation

tjhance:
~~(Depends on #2062)~~

This PR makes mode-checking (modes.rs) more aggressive about marking things as spec. Right now if you write something like `let ghost y = x;` (where `x` is an exec variable), then mode-checking will consider `x` an exec usage which is immediately coerced to a spec usage. This PR changes mode-checking to consider it a spec usage to begin with. Therefore:

 * The computed `var_modes` now contains a mode assignment which is optimal for performing erasure. This allows us to dramatically simplify erasure as we no longer have to do redundant mode computations during erasure
 * The mode-checking is now optimal for resolution analysis and fixes some previous incompleteness issues.
 * Overall, it's now a lot easier to be sure that all the different components agree on the mode assignments without inconsistency.

To implement this feature, I added an `expect` mode argument to the main recursive functions in `vir/src/modes.rs`. It seems to overlap a little in quality with `outer_mode`, but I didn't want to touch `outer_mode` or mess with its existing checks. I wanted to implement `expect` in an obviously-sound way, so the only effect it can ever have is to insert mode coercions, but all the _checks_ are still done as they were before.

### DIFF
--- a/source/rust_verify_test/tests/modes.rs
+++ b/source/rust_verify_test/tests/modes.rs
@@ -1634,3 +1634,29 @@ test_verify_one_file! {
         }
     } => Err(err) => assert_vir_error_msg(err, "expression has mode spec, expected mode proof")
 }
+
+test_verify_one_file! {
+    #[test] tracked_var_consumed_in_tuple_in_match_scrutinee verus_code! {
+        struct S;
+
+        proof fn test_match(tracked t: S, g: S) {
+            // This test infers the mode of `t` as tracked here, thus consuming it:
+            // In principle we could infer it as spec (since the tuple `(t, g)` as a whole is spec)
+            // and this used to work before we changed the way erasure is computed from
+            // the mode-checking results.
+            //
+            // Supporting this fundamentally requires a second pass
+            // (or otherwise a more sophisticated way of handling mode constraints).
+            // The issue is we don't know that we want to upcast `t` to spec until we see `g`
+            // forcing the tuple to be spec.
+            //
+            // However, it also seems fine to just not support this.
+            match (t, g) {
+                (t, g) => {}
+            }
+            match (t, g) {
+                (t, g) => {}
+            }
+        }
+    } => Err(err) => assert_rust_error_msg(err, "use of moved value: `t`")
+}
--- a/source/rustc_mir_build/src/thir/cx/expr.rs
+++ b/source/rustc_mir_build/src/thir/cx/expr.rs
@@ -244,7 +244,7 @@ impl<'tcx> ThirBuildCx<'tcx> {
 
         let kind = crate::verus_expr::apply_adjustment_post(self, hir_expr, adjustment, kind);
 
-        Expr { temp_lifetime, ty: adjustment.target, span, kind }
+        Expr { temp_scope_id, ty: adjustment.target, span, kind }
     }
 
     /// Lowers a cast expression.


source/vir/src/modes.rs
```diff
@@ -1301,11 +1383,19 @@ fn check_expr_handle_mut_arg(
...
-                Ok(*to_mode)
+                let mode = check_expr(
+                    ctxt,
+                    record,
+                    typing,
+                    param_mode,
+                    Expect(expect.join(*from_mode)),
+                    e1,
+                )?;
+                if !mode_le(mode, *from_mode) { Ok(mode) } else { Ok(*to_mode) }
```
Chris-Hawblitzel: This essentially adds a new rule to allow some coercion expressions to work on spec inputs, producing spec outputs. For example, this used to be disallowed, and now it is allowed:

```
proof fn test(x: int) {
    let y = Tracked(x);
}
```

I think this is fine. However, can we be more limited in this and more explicit about the intent? For example, can we have an `allow_spec` field in `UnaryOp::CoerceMode` and allow spec-in, spec-out coercions when `allow_spec` is true (and no other additional coercions relative to the previous version of the mode checker)?

source/vir/src/modes.rs
```diff
@@ -1684,37 +1783,53 @@ fn check_expr_handle_mut_arg(
...
+
+            let mode1 = check_place(
+                ctxt,
+                record,
+                typing,
+                outer_mode,
+                e1,
+                PlaceAccess::Read,
+                scrutinee_expect,
+            )?;
```
Chris-Hawblitzel, line 1833: It looks like this is no longer allowed:

```
struct S;

proof fn test_match(tracked t: S, g: S) {
    match (t, g) {
        (t, g) => {}
    }
    match (t, g) {
        (t, g) => {}
    }
}
```

This used to be allowed because `(t, g)` has type `(S, S)` with mode spec and is therefore erased.  Should we care about this?  Are there any other places in the mode checker where this behavior changes?

tjhance: hmm I didn't think about this case, yeah. this situation seems to fundamentally require multiple passes

tjhance: what do you think of doing backtracking? i.e., for any match:

1. run with Expect(Proof)
2. If it returns Spec, re-run with Expect(Spec)

(in principle this could blow up exponentially, but it require nested matches in scrutinees or something, which would be a pretty unusual case)

Chris-Hawblitzel: In principle, I think backtracking is fine and the performance isn't a concern.  As a matter of engineering, I'm a little worried about side effects to `Record` and `Typing` and the higher potential for bugs in the mode checker if we process the same expressions more than once.

I contrived this example to point out that there's a change in the checking, but I doubt that it actually matters to anyone.  At least, it didn't show up in any of our tests or in the Veritas benchmarks.  Maybe we could just make this the new semantics of mode/lifetime checking.  I don't think it would be surprising to people if `match (t, g)` consumed `t`, since `match t` consumes `t` and `match (t1, t2)` consumes `t1` and `t2` for tracked `t1` and `t2`.  And the workaround is easy: just assign (t, g) to a temporary spec-mode variable.

source/vir/src/modes.rs
tjhance: Is there any situation where `allow_spec` would be false?

Chris-Hawblitzel: Would it be false for `ModeCoercion::BorrowMut`?  This case doesn't include the `if !mode_le...` exception.  The other cases all look like it would be true.

tjhance: thinking about this more, it seems like ModeCoercion is used in two fundamentally different ways: as a "constructor" and as a "field accessor". This new logic is meant to reflect the constructor case, but doesn't really make sense for the field case. I updated the `ModeCoercion` to make a distinction between these two cases.

source/vir/src/modes.rs
tjhance: That sounds find to me. I added a test to document the new behavior.

PR #2072, opened 2026-01-09
Spec for is_multiple_of

jaylorch:


jaylorch:
@parno, thanks for the suggestion about `returns`. I've updated the PR to use that.

parno:
Cool, I'm glad that works.  Looks good to me.

PR #2075, opened 2026-01-12
vargo: add clippy to vargo and ci

bsdinis:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,29 @@ jobs:
         working-directory: ./tools/vargo
         run: cargo fmt -- --check
 
+  # run linter (clippy)
+  clippy:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.93.0
+          components: clippy
+
+      - name: clippy check codebase
+        working-directory: ./source
+        run: |
+          . ../tools/activate
+          vargo clippy -- -D warnings
+
+      - name: clippy check vargo
+        working-directory: ./tools/vargo
+        run: cargo clippy -- -D warnings
+
   # check if it builds and runs on macos, running the full test suite
   # (in release mode because it makes a huge difference)
   full-test:
@@ -228,7 +251,7 @@ jobs:
   # build the documentation
   # will complain if there are warnings
   build-docs:
-    needs: [fmt]
+    needs: [fmt, clippy]
     runs-on: ubuntu-22.04
     steps:
       - name: checkout


.github/workflows/ci.yml
```diff
@@ -39,10 +39,39 @@ jobs:
...
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.92.0
+          components: clippy
+
+      - name: get z3
```
parno: Do we need z3 for clippy?  Could we tell vargo not to check for Z3 when run with the clippy sub-command?

.github/workflows/ci.yml
```diff
@@ -39,10 +39,39 @@ jobs:
...
+        working-directory: ./source
+        run: |
+          ./tools/get-z3.sh
+          echo z3 version `./z3 --version`
+
+      - name: clippy check codebase
+        working-directory: ./source
```
parno, line 56: Won't this break until we actually run+apply clippy on the full Verus source code?

source/Cargo.toml
parno, line 1: How did you arrive at this particular list of clippy lints?  Can we document it in the code?

source/Cargo.toml
```diff
@@ -25,3 +25,136 @@ exclude = [
 # do not modify the following two lines
 [workspace.metadata.vargo]
 tag = "workspace"
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
```
parno, line 30: Not sure it matters, but it looks like for some of the packages in the workspace, this is a strict superset of what's needed.  Probably still worth it if it simplifies the overall setup.

.github/workflows/ci.yml
bsdinis: No. Clippy by default runs on check mode. Right now, this check passes, because I added a bunch of blanket lint suppressions on the Cargo.toml.

This way:
1. Not-suppressed infractions will be stopped by CI
2. We can remove lint suppressions one by one, fix them, and further infractions (of that lint) cannot be added back in (preventing regressions).

.github/workflows/ci.yml
bsdinis: I couldn't get it to run without it. I'll give it a shot.

source/Cargo.toml
bsdinis: Painful, running and re-running until no more breaking happened. As I mentioned above, I think the overall goal should be to remove all suppressions.

Because of this, I'm unsure of what documenting in the code would look like. There are thousands of lint errors that are being fired, so it might be a low signal to noise ratio.

source/Cargo.toml
bsdinis: My intuition is that lints should be configured at the workspace level to make it uniform. (As an aside, I also think we should set package versions at the workspace level too -- would minimize having to download duplicate versions of the same crate and we wouldn't miss updates).

.github/workflows/ci.yml
bsdinis: Shot successful

parno: Excellent, thanks for giving it a try!

.github/workflows/ci.yml
parno: Ah, thanks for explaining.  Now I understand why that giant list of seemingly random clippy lints are included.  This seems like a sensible migration strategy to me.  We may want to chat about the overall lint-imposition strategy in a future Verus meeting, to make sure everyone is on board with it.

source/Cargo.toml
parno: Makes sense now.  I think a comment to this effect would be helpful, e.g.,:
> For now, this list suppresses all clippy complaints remaining in the code base.  Over time, our goal is to eliminate the need for them.

source/Cargo.toml
parno: Seems reasonable to me.  On the crate versioning, that sounds like a good topic for a Verus meeting, to make sure there aren't reasons for crate-level customization.

.github/workflows/ci.yml
bsdinis: Yup, sounds good. Want to signal ahead of time that there might be concrete instances where a particular lint might not be useful. In those cases, it would still be the case we would remove the lint from the suppression list and annotate the concrete instances with `#[allow(clippy::lint)]` (and hopefully a comment explaining why it makes sense)

parno: Makes sense to me.

tjhance:
FWIW I find a lot of clippy's lints kind of nitpicky. In my opinion we should only disallow the more important categories like 'correctness', 'suspicious', and 'perf'.

bsdinis:
> FWIW I find a lot of clippy's lints kind of nitpicky. In my opinion we should only disallow the more important categories like 'correctness', 'suspicious', and 'perf'.

That makes sense. I feel like this can be decided on a lint by lint basis (i.e., if it's too annoying we can just keep it allowed, with a comment explaining that). 

I did however sort and group them by category though, so we can prioritize accordingly.

bsdinis:
Expanded the comment on `Cargo.toml` to reflect the policy as discussed in the meeting today https://github.com/verus-lang/verus/blob/7b3eeae725424368210aa1411515c1364decec14/source/Cargo.toml#L36-L56

parno:
> Expanded the comment on `Cargo.toml` to reflect the policy as discussed in the meeting today

Looks great! Thanks for capturing that.

PR #2081, opened 2026-01-17
add support for opaque types defined by assume specification

FeizaiYiHao:
This PR adds support for external functions that return opaque types (as well as nested opaque types).

In VIR, for each opaque type constructor, we record the corresponding external opaque type from the external function.

For example:
```
trait Dummy Tr{
   pub type Output;
}

pub fn external_fn() -> (ret: impl Tr<Output = impl Tr>);

assume_specification<T> [ external_fn ]() -> (ret: (impl Tr<Output = impl Tr>));
```
`external_fn` defines two opaque types: `impl Tr<Output = impl Tr>` and the type of `Output` -- another `impl Tr`. Let's call them `ExOpaque1` and `ExOpaque2`.
`assume_specification` also defines two opaque types: `impl Tr<Output = impl Tr>`  and `impl Tr`. Let's call them `Opaque1` and `Opaque2`.
For `Opaque1`, field `OpaqueTypeX::external_fn_opaque_type` is `Some(ExOpaque1)` and For `Opaque2`, it's `Some(ExOpaque2)`.

In `opaque_ty_to_air.rs`, we generate axioms in SMT to establish equivalency between the external opaque and assume specification opaque types.

This patch will enable async functions with assume specification.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_base.rs
+++ b/source/rust_verify/src/rust_to_vir_base.rs
@@ -2121,7 +2121,7 @@ pub(crate) fn ty_remove_references<'tcx>(
 /// Add the OpaqueDef to vir if the function returns an opaque type.
 /// If the opaque type is defined by assume specification,
 /// check the opaque type defined by the original external function too.
-/// Note: We have checked that the return types of the external function and 
+/// Note: We have checked that the return types of the external function and
 /// assume specification match exactly.
 pub(crate) fn check_fn_opaque_ty<'tcx>(
     ctxt: &Context<'tcx>,
@@ -2130,18 +2130,17 @@ pub(crate) fn check_fn_opaque_ty<'tcx>(
     span: Span,
     external_def_id: Option<&DefId>,
 ) -> Result<Option<OpaqueType>, VirErr> {
-
     let ty = ctxt.tcx.fn_sig(fn_def_id).skip_binder().output().skip_binder();
     let external_ty = if let Some(external_def_id) = external_def_id {
-        if ctxt.tcx.def_kind(external_def_id).is_fn_like(){
+        if ctxt.tcx.def_kind(external_def_id).is_fn_like() {
             Some(ctxt.tcx.fn_sig(external_def_id).skip_binder().output().skip_binder())
-        }else{
+        } else {
             None
         }
-    }else{
+    } else {
         None
     };
-    opaque_def_to_vir(ctxt, opaque_types, fn_def_id, &ty,span, external_ty.as_ref())
+    opaque_def_to_vir(ctxt, opaque_types, fn_def_id, &ty, span, external_ty.as_ref())
 }
 
 pub(crate) fn opaque_def_to_vir<'tcx>(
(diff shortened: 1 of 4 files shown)


FeizaiYiHao:
@tjhance @Chris-Hawblitzel These are some changes that I overlooked for the opaque type PR. Would you take a look when you have time?

tjhance:
I don't think it should be necessary to encode 2 different opaque types into VIR. The opaque types defined by the assume_specification proxy function are really completely irrelevant and shouldn't be included in the VIR to begin with.

FeizaiYiHao:
> I don't think it should be necessary to encode 2 different opaque types into VIR. The opaque types defined by the assume_specification proxy function are really completely irrelevant and shouldn't be included in the VIR to begin with.

The function postcondition use the opaque types defined by the proxy. Maybe we could maintain a hashmap or something to replace the opaque types defined by the proxy with the original opaque types, but I'm not sure what the complications are...

FeizaiYiHao:
@tjhance pin.

tjhance:
I think the easiest thing to do would be to replace the fake opaque type with the real one when lowering (i.e., in `mid_ty_to_vir`)

FeizaiYiHao:
> I think the easiest thing to do would be to replace the fake opaque type with the real one when lowering (i.e., in `mid_ty_to_vir`)

@tjhance It's a bit hard to do it I think. We don't know which opaque type is the fake opaque type of which, especially for nested opaque types.

I'm thinking the following:

1. In `syntex.rs`, instead of calling `constrain_types(ret, assume_specfication_function(args)`, we do  `constrain_types(ret, external_function(args)` to set the return value type to the real opaque type in the ensures clause. 
2. When handling the external function in VIR, we use the real opaque type instead of the fake one for the function return type.

After these two places, I think we have eliminated all the occurrences of the fake opaque type in the pipeline.

FeizaiYiHao:
@tjhance I think there are two ways to handle this now:

The first one is to call the external function in `contraint_type` and rewrite `assume_specification` return value in VIR. However, this might change the existing behavior slightly. In current Verus, if `assume_specification` uses the wrong return type, since in the ensures clause uses the return type of `assume_specification`, it'll trigger a type error in the ensures clause (if the ensures clause use the return value as if its type is the correct type from the external function). But if we do this modification, the type error wouldn't trigger, we'll catch this later in the VIR check when we check if the return types of `assume_specification` and external function match. 

The second one is to replace the return type in VIR as you mentioned.

Let me know which one you think it's better.

FeizaiYiHao:
@tjhance I tried solution one. The only problem is for this case
```
#[cfg(verus_keep_ghost)]
#[verifier::when_used_as_spec(ptr_null_mut)]
pub assume_specification<
    T: core::marker::PointeeSized + core::ptr::Pointee<Metadata = ()>,
>[ core::ptr::null_mut ]() -> (res: *mut T)
    ensures
        res == ptr_null_mut::<T>(),
    opens_invariants none
    no_unwind
;

```
`core::ptr::null_mut` does not have a type prarameter.
If we do `constrain_type(res, core::ptr::null_mut())`, Rust cannot infer the type of `res`, since ` core::ptr::null_mut()` can be any type. The biggest problem is that the error message is impossible for users to understand... 

So I guess we need to stick to the second solution.

PR #2087, opened 2026-01-19
Fix rustdoc generation failure for `const fn` with specifications

rikosellic:
When generating documentation using rustdoc with  `VERUSDOC=1`, the process aborts if the codebase contains `const fn` with `requires` or `ensures` clauses, leading to errors like
```
error[E0015]: cannot call non-consrustdoct function `vstd::prelude::ensures::<{closure@....rs:21:13: 21:20}>` in constant functions    
```

This happens because `builtin_macros` expands these specifications inside the function body before running rustdoc. I've added a condition to skip the injection if we are running in `rustdoc` mode and if the function is `const`. It will not affect the verification or the pre/post-conditions of `const fn`s in the documentation. The condition generation logic happens before the code expansion that modifies the function body.


parno:
This looks reasonable to me, but it'd be great to get a quick look from @tjhance

PR #2090, opened 2026-01-20
clippy: lift supression of correctness group lints

bsdinis:
Lints enforced:
- uninit
- unused_io_amount
- never_loop
- derived_hash_with_manual_eq
- absurd_extreme_comparisons

### DIFF
--- a/source/Cargo.toml
+++ b/source/Cargo.toml
@@ -55,9 +55,6 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 #  - If an already enforced lint is being pedantic for a particular instance, where
 #    fixing it would be overly complex, consider also `allow`ing it on that particular instance.
 
-# correctness
-uninit_assumed_init = "allow"
-
 # perf
 cloned_ref_to_slice_refs = "allow"
 cmp_owned = "allow"
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -404,7 +404,9 @@ impl<A> Ghost<A> {
     #[cfg(verus_keep_ghost)]
     #[rustc_diagnostic_item = "verus::verus_builtin::Ghost::view"]
     #[verifier::spec]
+    #[allow(clippy::uninit_assumed_init)]
     pub fn view(self) -> A {
+        // SAFETY: this is never called in an execution, so it's safe to do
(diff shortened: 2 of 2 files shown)


tjhance:
I'm not a huge fan of the 'absurd_extreme_comparisons' lint. For example, the old version of `valid_unicode_scalar` was clearer, even if it was technically redundant.

bsdinis:
> I'm not a huge fan of the 'absurd_extreme_comparisons' lint. For example, the old version of `valid_unicode_scalar` was clearer, even if it was technically redundant.

Will annotate that case, your point it makes sense -- I think in general unless it's for clarity it might make sense to keep it on (might catch bugs where some name confusion makes a person make an absurd comparison)

source/rust_verify/src/rust_to_vir_base.rs
```diff
@@ -2043,7 +2043,7 @@ fn check_generics_bounds_main<'tcx>(
             }
         }
     }
-    for x in accept_recs.keys() {
+    if let Some(x) = accept_recs.keys().next() {
```
Chris-Hawblitzel: I think a `for` loop is clearer in the intent here: every x is considered an error.  Potentially we could report all of them.

source/vir/src/headers.rs
```diff
@@ -597,7 +597,7 @@ pub fn make_trait_decls(methods: Vec<Function>) -> Result<Vec<Function>, VirErr>
             *method = make_trait_decl(method, &spec_method)?;
         }
     }
-    for (_, extra_spec) in specs.iter() {
+    if let Some((_, extra_spec)) = specs.iter().next() {
```
Chris-Hawblitzel: I think a for loop is clearer in the intent here: every item is considered an error. Potentially we could report all of them.

source/rust_verify/src/rust_to_vir_base.rs
bsdinis: I changed this to indeed report every error. In fact here it was just a bunch of items that we had to comma separate.

I still think the lint overall adds value -- it is weird for a loop construct to never loop

source/vir/src/headers.rs
bsdinis: I changed it to report all errors -- had to add a small helper to `vir/messages.rs`

PR #2091, opened 2026-01-20
clippy: lift suppression of suspicious group lints (1/2)

bsdinis:
Lints enforced:
- unnecessary_get_then_check
- suspicious_doc_comments
- redundant_locals
- non_canonical_partial_ord_impl
- non_canonical_clone_impl
- manual_unwrap_or_default
- ineffective_open_options
- empty_line_after_doc_comments
- crate_in_macro_def

### DIFF
--- a/source/Cargo.toml
+++ b/source/Cargo.toml
@@ -75,7 +75,6 @@ vec_init_then_push = "allow"
 arc_with_non_send_sync = "allow"
 # do not remove: see https://github.com/verus-lang/verus/pull/2091#discussion_r2710412679 for a discussion
 manual_unwrap_or_default = "allow"
-unnecessary_get_then_check = "allow"
 
 # style
 assign_op_pattern = "allow"
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -718,10 +718,10 @@ pub fn parse_args_with_imports(
         log_dir: matches.opt_str(OPT_LOG_DIR),
         log_all: matches.opt_present(OPT_LOG_ALL),
         log_args: LogArgs {
-            log_vir: log.get(LOG_VIR).is_some(),
-            log_vir_simple: log.get(LOG_VIR_SIMPLE).is_some(),
-            log_vir_poly: log.get(LOG_VIR_POLY).is_some(),
-            log_vir_sst: log.get(LOG_VIR_SST).is_some(),
+            log_vir: log.contains_key(LOG_VIR),
+            log_vir_simple: log.contains_key(LOG_VIR_SIMPLE),
+            log_vir_poly: log.contains_key(LOG_VIR_POLY),
+            log_vir_sst: log.contains_key(LOG_VIR_SST),
             vir_log_option: {
                 if let Some(oo) = log.get(LOG_VIR_OPTION) {
                     let Some(oo) = oo else {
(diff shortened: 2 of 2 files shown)


source/builtin_macros/src/rustdoc.rs
```diff
@@ -39,9 +39,8 @@ use verus_syn::{
     ReturnType, Signature, TraitItemFn, Type, TypeGroup, TypePath,
 };
 
-/// Check if VERUSDOC=1.
-
 #[cfg(verus_keep_ghost)]
+/// Check if VERUSDOC=1.
```
tjhance, line 42: docs after attributes looks weird to me. Which lint is this?

source/vir/src/closures.rs
```diff
@@ -8,12 +8,10 @@ use crate::messages::error;
 ///  1. The closure does not mutate any variable from outside the closure.
 ///     Such closures are currently unsupported.
 ///
-/// TODO make this check as well:
-///
 ///  2. If a variable is referenced from spec mode but not actually captured in
 ///     tracked/exec mode, then that variable cannot be mutable.
 ///     (This is actually easy to support, but we expect it might be confusing to the user.)
-
+// TODO make the second check as well:
```
tjhance: This leaves the rustdoc comment very misleading.

source/builtin/src/lib.rs
```diff
@@ -532,6 +532,8 @@ impl<A> Tracked<A> {
     }
 }
 
+// Ghost is a ghost construct, so it is never _actually_ cloned
```
tjhance: This is not really true. `Ghost` is an exec type that contains a ghost value. The `clone` isn't erased.

source/rust_verify/src/attributes.rs
```diff
@@ -811,10 +811,7 @@ pub(crate) fn parse_attrs_opt(
     attrs: &[Attribute],
     diagnostics: Option<&mut Vec<VirErrAs>>,
 ) -> Vec<Attr> {
-    match parse_attrs(attrs, diagnostics) {
-        Ok(attrs) => attrs,
-        Err(_) => vec![],
-    }
+    parse_attrs(attrs, diagnostics).unwrap_or_default()
```
tjhance: I prefer explicit matches more often than not. Here, for example, it's a lot more obvious that we're matching on an Err and continuing rather than bubbling up the Err (which is somewhat unusual, so good to be clear about it).

source/builtin_macros/src/rustdoc.rs
bsdinis: This is empty lines after docs

bsdinis: In this case I think the doc can come before

source/vir/src/closures.rs
bsdinis: makes sense, I think I fixed it

source/builtin/src/lib.rs
bsdinis: I see. I think the fix here is to implement `clone()` as `*self`. Non canonical for clippy means that `a.clone() != a.copy()`

source/rust_verify/src/attributes.rs
bsdinis: Here I'm not super sure, I feel like it's not idiomatic to reimplement available monadic interfaces on `Option/Result`. For me, looking at how it was before makes me pause and look for some non-standard use (because I'm expecting normal cases to use the existing interfaces). Moreover, I have to question both what happens on Ok and Err (does the value change in the Ok arm? why is it vec![] in Err?). `unwrap_or_default` tells me directly that if `parse_attrs` is Ok then I just get the result of that, otherwise I get the default value.

source/builtin_macros/src/rustdoc.rs
bsdinis: fixed

source/rust_verify/src/attributes.rs
tjhance: I personally don't like the monadic interfaces. Here, it's hard to tell what types this code is even operating on. Both the Err and the Vec are hidden from view.

I don't object to these functions out-of-hand (e.g. I don't mind if other devs use 'em), but I really don't want a lint that enforces their use.

bsdinis: Sounds fair. Changed it to comment the suppression

Chris-Hawblitzel (approved):

PR #2092, opened 2026-01-20
clippy: lift suppression of arc_with_non_send_sync (suspicious)

bsdinis:
This change highlighted a pattern that is worth considering (removing)

We have multiple types that follow the structure:

```rust
struct FooX { .. }
type Foo = Arc<FooX>
```

However, it may ocurr that there is only one such reference as I believe was the case in this change (at least I couldn't find calls to clone from context) Moreover, the type may not even be suitable to be sent across threads (in this case, this was fundamentally because of a rustc_hir type that was !Send and !Sync).

We should avoid having this pattern, as it causes extra confusion and has negative runtime performance (extra refcounts + extra indirection + synchronization). If we need to have the object live across threads, we have to do it and additionally have internal synchronization. Otherwise, we should prefer either having it be an Rc/Box or a plain T.

### DIFF
--- a/source/rust_verify/src/context.rs
+++ b/source/rust_verify/src/context.rs
@@ -1,13 +1,18 @@
 use crate::{erase::ResolvedCall, verus_items::VerusItems};
 use rustc_hir::Attribute;
+use rustc_hir::Crate;
+use rustc_hir::HirId;
 use rustc_hir::def_id::LocalDefId;
-use rustc_hir::{Crate, HirId};
 use rustc_middle::ty::{TyCtxt, TypeckResults};
 use rustc_mir_build_verus::verus::BodyErasure;
 use rustc_span::SpanData;
 use rustc_span::def_id::DefId;
+use std::cell::RefCell;
 use std::collections::HashMap;
+use std::ops::DerefMut;
+use std::rc::Rc;
 use std::sync::Arc;
+use std::sync::atomic::AtomicU64;
 use vir::ast::{Ident, Mode, Path, Pattern, VirErr};
 use vir::messages::AstId;
 
@@ -23,8 +28,7 @@ pub struct ErasureInfo {
 
 type ErasureInfoRef = std::rc::Rc<std::cell::RefCell<ErasureInfo>>;
 
-pub type Context<'tcx> = Arc<ContextX<'tcx>>;
-#[derive(Clone)]
+pub type Context<'tcx> = Rc<ContextX<'tcx>>;
 pub struct ContextX<'tcx> {
     pub(crate) cmd_line_args: crate::config::Args,
     pub(crate) tcx: TyCtxt<'tcx>,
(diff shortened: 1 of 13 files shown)


tjhance:
The Arc<FooX> pattern is ubiquitous for our core AST structures. We commonly clone these values, and they do move across threads at points. So, it's not trivial to find an alternative.

bsdinis:
> The Arc pattern is ubiquitous for our core AST structures. We commonly clone these values, and they do move across threads at points. So, it's not trivial to find an alternative.

That makes sense. However, for this case you _couldn't_ send this across threads if you wanted to, it would be an error, so having it be an Rc makes it more clear. Might change it to a RefCell though, to avoid `make_mut` (which for the reasons I mentioned above makes it hard to reason about the code)

tjhance:
Oh, this change is fine, of course. I just wanted to give you a heads-up about the way we use this pattern elsewhere.

bsdinis:
Gotcha! I still might separately make a pass for make_mut -- it is hard to reason about the value of an `Arc` if it can be fully cloned to change something (we should probably have mutexes or something for interior mutability)

bsdinis:
Also just noticed my comment about `make_mut` is in a comment in the code :/, sorry for the confusion:

https://github.com/verus-lang/verus/blob/1acae13cbec434695ee2856303d4f81de495c144/source/rust_verify/src/rust_to_vir.rs#L420-L427

Chris-Hawblitzel:
> Gotcha! I still might separately make a pass for make_mut -- it is hard to reason about the value of an `Arc` if it can be fully cloned to change something (we should probably have mutexes or something for interior mutability)

A large amount of our code is purely functional or nearly purely functional, using purely immutable data values and `Arc` for garbage collection.  Rather than modifying existing values, we commonly allocate new immutable values, possibly with some fields copied from old immutable data, as you would do in a functional language like Haskell with purely immutable data.  For this style, `make_mut` is a common optimization that, if possible, reuses memory that would otherwise be immediately reclaimed anyway.  I believe every use of `make_mut` should be in service of this style.  I agree that we should not try to use `make_mut` to imitate actual semantic mutability. (I was a latecomer to `make_mut` because it sounded suspicious to me at first, but I've since become a convert.)

source/rust_verify/src/context.rs
```diff
@@ -23,23 +28,43 @@ pub struct ErasureInfo {
...
+    pub(crate) diagnostics: Rc<RefCell<Vec<vir::ast::VirErrAs>>>,
     pub(crate) no_vstd: bool,
     pub(crate) arch_word_bits: Option<vir::ast::ArchWordBits>,
     pub(crate) crate_name: Ident,
     pub(crate) vstd_crate_name: Ident,
-    pub(crate) name_def_id_map:
-        std::rc::Rc<std::cell::RefCell<std::collections::HashMap<Path, DefId>>>,
-    pub(crate) next_read_kind_id: std::rc::Rc<std::cell::Cell<u64>>,
+    pub(crate) name_def_id_map: Rc<RefCell<std::collections::HashMap<Path, DefId>>>,
+    pub(crate) next_read_kind_id: AtomicU64,
```
Chris-Hawblitzel, line 45: Why does this use `AtomicU64` if ContextX can't be shared between threads?

source/rust_verify/src/rust_to_vir.rs
```diff
@@ -417,7 +418,15 @@ pub fn crate_to_vir<'a, 'tcx>(
...
-        let ctxt = Arc::make_mut(ctxt);
+        // TODO(bsdinis): This code (and all code using either Rc::make_mut or Arc::make_mut) is suspicious,
+        // worth a second look at some point
+        //
+        // make_mut duplicates the data if the reference isn't unique, meaning the updates don't
+        // propagate to the original
+        // - Either this is a no-op for other references
+        // - Or there is a single reference, in which case it might make more sense to just use a
+        // reference without ref counting
+        let ctxt = Rc::make_mut(ctxt);
```
Chris-Hawblitzel, line 429: I pushed a fix for this to https://github.com/verus-lang/verus/tree/refs/heads/context_no_make_mut

source/rust_verify/src/context.rs
bsdinis: That's a great point -- it probably doesn't

source/rust_verify/src/rust_to_vir.rs
bsdinis: Thank you!

PR #2097, opened 2026-01-21
resource: add a ResourceAlgebra construction for Resources

bsdinis:
Currently, the only way we support resource algebras is through PCMs (being pedantic, via unitary resource algebras).
This PR implements an interface for Resource Algebras (a more general version of the PCM). This will hopefully be useful to implementing Iris style resource algebra composers, but in any case gives more flexibility to someone cooking up a Resource.

The first 3 commits mostly move things around/change them slightly (e.g., renaming the functions on the PCM).
The last commit actually changes things significantly (this way the diff is more self contained).

It is worthwhile noting that this PR provides a (single) canonical implementation of the ResourceAlgebra trait for an implementor of PCM, which ensures consistency.

I also perused the veritas repos and didn't find anyone implementing PCM directly, but the user changes for user code is fixing the imports + the names in the _trait implementation_ of the PCM trait (then it gets access to the old names via the ResourceAlgebra implementation).


source/vstd/resource/pcm.rs
```diff
@@ -7,62 +7,84 @@ verus! {
...
 broadcast use super::super::set::group_set_axioms;
 
-/// See [`Resource`] for more information.
+/// A Partial Commutative Monoid is a special type of `ResourceAlgebra`, where all elements have
+/// the same core (which belongs in the carrier), the unit. For this reason, they are also called
+/// unitary resource algebras[^note].
+///
+/// [^note]: This is slightly misleading. PCMs are partial in the sense that the operation is not
+/// defined for certain inputs. Because Verus does not support partial functions, we model that
+/// partiality with the validity predicate, which does make it a unitary resource algebra.
```
tjhance, line 29: I've been meaning to fix this for a while. Since you're moving it, can you go ahead and rename it to UnitalRA or something?

source/vstd/resource/algebra.rs
```diff
@@ -0,0 +1,355 @@
...
+        requires
+            incl::<RA>(target, self.value()),
+        ensures
+            out.loc() == self.loc(),
+            out.value() == target,
+    ;
+
+    /// Validate two items at once. If we have two resources at the same location then its
+    /// composition is valid.
+    // TODO(bsdinis): would be interesting to prove this
```
tjhance: We've proved that the existing PCM resource is sound, and it wouldn't be hard to extend to non-unital RAs. See my thesis for a sketch ([Sec 6.4.7](https://tjhance.github.io/travis_hance_thesis.pdf))

source/vstd/resource/algebra.rs
```diff
@@ -0,0 +1,355 @@
...
+                RA::op(self.value(), other.value()),
+                set_op::<RA>(new_values, other.value()),
+            ),
+        ensures
+            out.loc() == self.loc(),
+            new_values.contains(out.value()),
+    ;
+
+    /// Create a shared reference to the core from a shared reference to a resource
+    pub proof fn duplicate_shared_core<'a>(tracked &'a self) -> (tracked r: &'a Self)
```
tjhance: Why is the return value a shared reference? The core proposition should be duplicable, i.e., you can just return an owned `Self`.

source/vstd/resource/lib.rs
```diff
@@ -359,30 +362,71 @@ pub proof fn validate_multiple<P: PCM>(
...
 
+// TODO(bsdinis): generate these with a macro
+/// Validates that the three given resources have values that combine
+/// to form a valid value. Although `r1`, `r2` and `r3` are mutable, they
+/// don't change. (They change during the function but are restored to
+/// the way they were by the time the function returns.)
+pub proof fn validate_3<RA: ResourceAlgebra>(
+    tracked r1: &mut Resource<RA>,
+    tracked r2: &mut Resource<RA>,
+    tracked r3: &mut Resource<RA>,
```
tjhance: You should be able to always leave one argument as a & reference

source/vstd/resource/algebra.rs
```diff
@@ -0,0 +1,355 @@
...
+
+/// A Resource Algebra operating on a type T
+///
+/// This construction is based off the [Iris from the Ground Up](https://people.mpi-sws.org/~dreyer/papers/iris-ground-up/paper.pdf)
+/// report.
+///
+/// A striking difference is that we mandate that all elements have a core belonging to the set,
+/// obviating the need for lifting the set (in this case, the type T) into a T ⨄ {⊥ }, so that bottom
+/// can serve as the unit. This could be emulated by adding the bottom element (which can serve as
+/// the core in the lifting for elements of T without a core)
```
tjhance, line 19: I don't understand why this change is made. Won't this be overly restrictive in building RAs? What's the point of adding non-unital RAs if users have to add a unit in practice just to satisfy the core axioms?

tjhance (changes requested):
* Let's consolidate the 'resource' and 'tokens' directories into one. Also, keep 'storage_protocol.rs' in the same directory as 'pcm.rs'
* We already have a new type for representing ghost locations (InstanceId), so please consolidate them

I also think RA should be proved in terms of PCM rather than vice versa; see my comment below.

source/vstd/resource/algebra.rs
bsdinis: This is a bit subtle, maybe I can improve the comment.
In general resource algebras the core is a partial function. Because verus does not support partial functions, we would have to have a bottom element to be the "core for values without a core", if this makes sense. It's certainly not meant to be true that forall a. a.core == bot

tjhance:
I think I would prefer flipping this around, leaving PCM<sup>1</sup> axiomatized, and proving RA sound in terms of PCM. The existing PCM is a simpler definition, so it's safer to axiomatize, and it already has a lot more vetting.

In Iris, the 'core' operator is essential for allowing propositions to be persistent. But Verus doesn't have a notion of persistent proposition, it only has duplicability. So everything interesting about core can just be proved from more fundamental operations.

<sub><sup>1</sup>Yeah I know this should really be UnitalRA</sub>

source/vstd/resource/algebra.rs
tjhance: The natural interpretation of Iris's core operator in Verus would be a function `M -> Option<M>`.

> Because verus does not support partial functions, we would have to have a bottom element to be the "core for values without a core"

Like I said, this seems overly restrictive, as now you often need to add a bottom element to the carrier set just to satisfy this requirement.

source/vstd/resource/lib.rs
bsdinis: I plan to do so, this was mostly matching the proof for validate_4.

source/vstd/resource/pcm.rs
bsdinis: I wonder if it's broken. My main concern is that UnitalRA might be a bit too esoteric, whereas PCM is (I believe a better understood term). But since we're kind of matching Iris's description it may make more sense.

source/vstd/resource/algebra.rs
bsdinis: That is a good point, will see how it looks -- it might make things a bit complicated in terms of having to check whether things are options or not everywhere but should be fine

source/vstd/resource/algebra.rs
bsdinis: good point, fixed

source/vstd/resource/lib.rs
bsdinis: done

source/vstd/resource/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+pub mod seq;
+pub mod set;
+pub mod storage_protocol;
+
+pub use lib::*;
+
+verus! {
+
+#[verifier::accept_recursive_types(P)]
+pub tracked struct Resource<P> {
```
tjhance, line 63: This needs to retain its `external_body`, or else Verus will interpret this type as literally being isomorphic to `PhantomData`, which is entirely nonsensical.

source/vstd/resource/algebra.rs
```diff
@@ -0,0 +1,263 @@
...
+        ensures
+            out.0.loc() == self.loc(),
+            out.1.loc() == self.loc(),
+            out.0.value() == left,
+            out.1.value() == right,
+    ;
+
+    pub proof fn dummy() -> (tracked out: Self) {
+        Resource { p: core::marker::PhantomData }
+    }
```
tjhance: Case in point: this function should not be possible to implement (as it is possible for an RA to have no valid elements, so you could derive a contradiction from `dummy` and `validate`)

tjhance (changes requested):

bsdinis:
Closed in favour of #2141

PR #2098, opened 2026-01-21
[verus_syn] fix: relax parens around braces constraint

mkovaxx:
Update `Specification::parse` so that in the context of a `fn` definition, parentheses are no longer needed around block expressions.

### DIFF
--- a/dependencies/syn/src/verus.rs
+++ b/dependencies/syn/src/verus.rs
@@ -729,17 +729,17 @@ pub mod parsing {
                 match ctx {
                     Context::Expr => {
                         if input.peek2(token::Brace) || Self::peek2_spec_keyword(input) {
-                            // Missing parentheses.
+                            // Missing parentheses around block.
                             return Err(input.error("This block looks like the closure/loop body, but it is followed immediately by another block or clause. If you meant this block to be part of a clause, try parenthesizing it."));
                         }
                     }
                     Context::Item => {
                         if !exprs.trailing_punct() && input.peek2(Token![,]) {
-                            // Missing comma before.
+                            // Missing comma before block.
                             return Err(input.error("This block looks like it might be part of the clause. If you meant it to be part of the clause, put a comma before it."));
                         }
                         if input.peek2(token::Brace) || Self::peek2_spec_keyword(input) {
-                            // Missing comma after.
+                            // Missing comma after block.
                             return Err(input.error("This block looks like it might be part of the clause. If you meant it to be part of the clause, put a comma after it."));
                         }
                     }


dependencies/syn/src/verus.rs
```diff
@@ -716,33 +702,49 @@ pub mod parsing {
...
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by an 'invariant_except_break' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-                if input.peek2(Token![invariant]) {
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by an 'invariant' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-                if input.peek2(Token![decreases]) {
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by a 'decreases' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-            }
+            // TODO: Consider failing with an error if `exprs` is empty.
```
mkovaxx: Context here: `verus-analyzer` expects at least one clause here, see:
https://github.com/verus-lang/verus-analyzer/blob/9f987672429c5ebe4e763173ce17b2c2a1fb362a/crates/syntax/rust.ungram#L760-L767

mkovaxx: There's an example of this (`requires` with no clauses) in `vstd`:
https://github.com/verus-lang/verus/blob/1dc37af91e58031cd182ee0d5c02f50032e6dfa5/source/vstd/seq_lib.rs#L1147-L1151

mkovaxx: A few more examples in tests:
https://github.com/verus-lang/verus/blob/1dc37af91e58031cd182ee0d5c02f50032e6dfa5/source/rust_verify_test/tests/opaque_types.rs#L370-L373

https://github.com/verus-lang/verus/blob/1dc37af91e58031cd182ee0d5c02f50032e6dfa5/source/rust_verify_test/tests/opaque_types.rs#L436-L439

Chris-Hawblitzel:
I think this should be legal (it is under the PR):

```
fn f()
    requires
        { true },
    ensures
        { true },
{
}
```

but this should still be illegal:

```
fn f()
    requires
        { true } // illegal; comma required
    ensures
        { true } // illegal; comma required
{
}
```

Here are a couple real-world examples where we could eliminate the `(` and `)` from the `({` and `}),`:

https://github.com/microsoft/verified-storage/blob/main/pmemlog/src/logimpl_v.rs#L380

```
    /// Proves that a write to data that does not touch any metadata is crash safe.
    pub proof fn lemma_data_write_is_safe<Perm>(pm: Seq<u8>, bytes: Seq<u8>, write_addr: int, perm: &Perm)
        where
            Perm: CheckPermission<Seq<u8>>,
        requires
            UntrustedLogImpl::recover(pm) is Some,
            pm.len() > contents_offset,
            contents_offset <= write_addr < pm.len(),
            perm.check_permission(pm),
            permissions_depend_only_on_recovery_view(perm),
            ({
                // write must be a valid write and not overlap the live log
                let live_header = spec_get_live_header(pm);
                let physical_head = spec_addr_logical_to_physical(live_header.metadata.head as int, live_header.metadata.log_size as int);
                let physical_tail = spec_addr_logical_to_physical(live_header.metadata.tail as int, live_header.metadata.log_size as int);
                &&& physical_head <= physical_tail ==> {
                    &&& write_addr + bytes.len() <= live_header.metadata.log_size + contents_offset
                    &&& write_addr < physical_head ==> write_addr + bytes.len() <= physical_head
                    &&& (physical_tail <= write_addr || write_addr < physical_head)
                }
                &&& physical_tail < physical_head ==> {
                    &&& physical_tail <= write_addr <= write_addr + bytes.len() < physical_head
                }
            }),
        ensures
            UntrustedLogImpl::recover(pm) is Some,
            forall |chunks_flushed| {
                let new_pm = #[trigger] update_contents_to_reflect_partially_flushed_write(
                    pm, write_addr, bytes, chunks_flushed);
                perm.check_permission(new_pm)
            },
            ({
                let new_pm = update_contents_to_reflect_write(pm, write_addr, bytes);
                perm.check_permission(new_pm)
            }),
            update_data_view_postcond(pm, bytes, write_addr),
    {
        let new_pm = update_contents_to_reflect_write(pm, write_addr, bytes);
        lemma_append_data_update_view(pm, bytes, write_addr);
        lemma_same_log_state(pm, new_pm);

        assert forall |chunks_flushed| {
            let new_pm = #[trigger] update_contents_to_reflect_partially_flushed_write(
                pm, write_addr, bytes, chunks_flushed);
            perm.check_permission(new_pm)
        } by {
            let new_pm = update_contents_to_reflect_partially_flushed_write(
                pm, write_addr, bytes, chunks_flushed);
            lemma_append_data_update_view_crash(pm, bytes, write_addr, chunks_flushed);
            lemma_same_log_state(pm, new_pm);
            lemma_same_permissions(pm, new_pm, perm);
        }
    }
```

https://github.com/verus-lang/verus/blob/main/source/vstd/storage_protocol.rs#L284

```
    pub axiom fn validate_with_shared(tracked self: &mut Self, tracked x: &Self) -> (res: (
        P,
        Map<K, V>,
    ))
        requires
            old(self).loc() == x.loc(),
        ensures
            *self == *old(self),
            ({
                let (q, t) = res;
                { P::rel(P::op(P::op(self.value(), x.value()), q), t) }
            }),
    ;
```

The last example should also allow the last comma to optionally be replaced with a semicolon:

```
    pub axiom fn validate_with_shared(tracked self: &mut Self, tracked x: &Self) -> (res: (
        P,
        Map<K, V>,
    ))
        requires
            old(self).loc() == x.loc(),
        ensures
            *self == *old(self),
            {
                let (q, t) = res;
                { P::rel(P::op(P::op(self.value(), x.value()), q), t) }
            };
```

dependencies/syn/src/verus.rs
Chris-Hawblitzel: I would prefer to allow empty requires and ensures.  Partly this is for convenience: it's common to comment out preconditions and postconditions when debugging proofs, and sometimes this leaves requires and ensures that are temporarily empty.  Partly it's because of Rust precedent: empty `where` clauses are legal.

```
fn f()
where
{
}
```

mkovaxx: @Chris-Hawblitzel Thanks for the input! I'll leave this as it is.
I've filed an issue on `verus-analyzer` for updating its grammar to allow empty clauses:
https://github.com/verus-lang/verus-analyzer/issues/71

mkovaxx:
@Chris-Hawblitzel Thanks for giving those examples!

I've made the changes to require the terminal comma, and that actually makes the grammar simpler.

On the other hand, it seems that allowing the function definition to end (without a function body) in a semicolon makes the grammar context-sensitive.

Two examples back-to-back.

First, the hypothetical simplification of the body-less function def from above ([original](https://github.com/verus-lang/verus/blob/5bf8e7db4caab958572d357826fa0d1bd05b2857/source/vstd/storage_protocol.rs#L284-L296)):
```rust
    pub axiom fn validate_with_shared(tracked self: &mut Self, tracked x: &Self) -> (res: (
        P,
        Map<K, V>,
    ))
        requires
            old(self).loc() == x.loc(),
        ensures
            *self == *old(self),
            {
                let (q, t) = res;
                { P::rel(P::op(P::op(self.value(), x.value()), q), t) }
            };
```

Compare that to this closure with a spec ([source](https://github.com/verus-lang/verus/blob/1650a05d64219fdd1d150f32d0d7d5b19b8defcd/examples/std_test/result.rs#L17-L22)):
```rust
    let op = |x: i32| -> (y: i32)
        requires
            x < 100000,
        ensures
            y > x,
        { x + 1 };
```

The last `'{'` `Expr` `'}'` behaves differently between those two examples:
- In the first one, it is part of the `ensures` clause.
- In the second, it is the closure body.

For `Specification::parse` to correctly handle both of those cases, it would need to know whether it's parsing a function or a closure spec.

How should we resolve that context sensitivity?
Would it be okay to make `Specification::parse` context-dependent like that?

Chris-Hawblitzel:
> On the other hand, it seems that allowing the function definition to end (without a function body) in a semicolon makes the grammar context-sensitive.

You're right, I hadn't thought about nested functions and closures when I proposed allowing `;`.  For nested functions, there are potential ambiguities due to semicolons that are allowed to appear as statements:

```
fn t() {
    fn f() {}
    fn g() {}
    ; // useless, but legal, semicolon
    let z = 10;
}
```

So I no longer think it's worth bothering with the `;` case.  Just handling `,` is fine.  Thinking more about closures, there are some interesting issues as well, because requires and ensures in closures are already pushing the limits of how we parse commas.  I was trying out a bunch of tests:

```
fn test()  {
    let c = (|| { true }, || { true });
    let c = || requires true { true };
    let c = || requires true, { true };
//    let c = || requires true, { true } { true };
    let c = || requires true, ({ true }) { true };
//    let c = || requires true, { true }, { true };
    let c = || requires true, ({ true }), { true };
//    let c = || requires true true;
    let t = (|| true, || true);
//    let t = (|| true, || requires true { true }, || true);
    let t = (|| true, (|| requires true { true }), || true);
//    let t = (|| true, || requires true, { true }, || true);
//    let t = (|| true, || requires true, ensures true { true }, || true);
    let t = (|| true, (|| requires true, ensures true { true }), || true);
//    let t = (|| true, || requires true, true { true }, || true);
    let t = (|| true, (|| requires true, true { true }), || true);
    let t = (|| true, (|| requires true, { true }), || true);

//    let t = (|| true, || requires true, { true }, { true });
    let t = (|| true, (|| requires true, { true }), { true });

//    let t = (|| true, || requires true { true }, { true });
    let t = (|| true, (|| requires true { true }), { true });

    let t = (true, true, (|| requires true {}));
    let t = (true, true, (|| requires true, true {}));
    let t = (true, true, || requires true {});
    let t = (true, true, || requires true, true {});
}
```

Right now, we enforce a policy that a closure with requires or ensures cannot be directly followed by a comma.  This rules out cases like `(|| true, || requires true, { true }, { true })` where it's very confusing which commas belong to the tuple and which commas belong to the requires clauses.  I think we should continue being at least this restrictive with closure requires/ensures in comma-separated lists like tuples.  So for closures, I think we should continue to disallow unparenthesized blocks in requires/ensures.  This means we need to parse specs in closures more restrictively than specs in function declarations.  Maybe the easiest way to do this would be to parse all specs the same way, but make an extra check afterwards in the closure parsing.  A fancier way might be to make Specification, Requires, and Ensures parameterized over a generic const bool that says whether to allow unparenthesized blocks as requires/ensures clauses (I haven't tried this; I don't know whether it would work).

mkovaxx:
The `smoke-test` CI job was consistently failing at this step:
```
curl -LO https://www.singular.uni-kl.de/ftp/pub/Math/Singular/UNIX/Singular-4-3-2_M1.dmg
```

Error:
```
curl: (28) Failed to connect to www.singular.uni-kl.de port 443 after 75006 ms: Couldn't connect to server
```

To get around this issue, I've uploaded `Singular-4-3-2_M1.dmg` as a [release artifact](https://github.com/verus-lang/verus/releases/tag/dependency%2Fsingular-4.3.2) and changed the CI job to download it from GitHub.

.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ name: ci
 on:
   push:
     branches:
-      - 'main'
+      - "main"
```
mkovaxx, line 6: My YAML formatter is unstoppable, I'm sorry...

.github/workflows/ci.yml
```diff
@@ -194,14 +171,14 @@ jobs:
...
-          toolchain: 1.93.0
+          toolchain: 1.92.0
 
       - name: setup nextest
         uses: taiki-e/install-action@nextest
 
       - name: download singular
         if: matrix.features == 'singular'
-        run: curl -LO https://www.singular.uni-kl.de/ftp/pub/Math/Singular/UNIX/Singular-4-3-2_M1.dmg
+        run: curl -fLO https://github.com/verus-lang/verus/releases/download/dependency/singular-4.3.2/Singular-4-3-2_M1.dmg
```
mkovaxx, line 204: Needed this to prevent the download from failing CI.

mkovaxx:
@Chris-Hawblitzel This is ready to review. The grammar is now context-sensitive such that in `fn` definitions, parens are no longer required around braced blocks.

source/rust_verify_test/tests/mut_refs.rs
```diff
@@ -1679,6 +1679,52 @@ test_verify_one_file_with_options! {
     } => Err(err) => assert_fails(err, 3)
 }
 
+test_verify_one_file_with_options! {
+    #[test] test_params_with_reborrow_no_parens ["new-mut-ref"] => verus_code! {
```
Chris-Hawblitzel: Rather than duplicating existing tests to add separate "..._no_parens" tests, I think we should just edit the existing tests in place.  If we want separate parens and no_parens tests for testing the syntax, I think we should make new tests for this purpose to avoid duplicating existing tests.

dependencies/syn/src/item.rs
```diff
@@ -2906,7 +2906,7 @@ pub(crate) mod parsing {
                     None
                 };
                 generics.where_clause = input.parse()?;
-                let ensures: Option<Ensures> = input.parse()?;
+                let ensures: Option<Ensures> = Ensures::parse_optional_in(Context::Item, input)?;
```
Chris-Hawblitzel, line 2909: Why does this use `Context::Item` but the other places in `item.rs` use `Context::Expr`?

dependencies/syn/src/verus.rs
```diff
@@ -707,39 +748,10 @@ pub mod parsing {
...
-                if input.peek2(Token![invariant_except_break]) {
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by an 'invariant_except_break' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-                if input.peek2(Token![invariant]) {
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by an 'invariant' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-                if input.peek2(Token![decreases]) {
-                    return Err(input.error("This block would be parsed as the function/loop body, but it is followed immediately by a 'decreases' (if you meant this block to be part of the specification, try parenthesizing it)"));
-                }
-            }
```
Chris-Hawblitzel, line 741: I think these errors are still useful.  For example, if a user writes:

```
fn test()
    requires {}
{
}
```

or

```
fn test()
    requires {}
    ensures true,
{
}
```

or

```
fn test() {
    loop
        invariant
            {}
    {
    }
}
```

then these errors catch the problem and give a useful suggestion.

(That said, I don't think it was necessary for the error cases to duplicate so much of the same code 7 times.  This may have just been due to too much copy-and-paste code evolution.)

source/rust_verify_test/tests/mut_refs.rs
tjhance: I don't think we have a dedicated file for syntax-related tests, though perhaps we should? Otherwise they could go in basic.rs or maybe the examples/syntax.rs file

mkovaxx: Good point. Tests are no longer duplicated.

dependencies/syn/src/item.rs
mkovaxx: I think this is parsing the `ensures` clause that may appear between the type and the opening `{` of the "body" in a Verus-style `const` definition. Because the body of that is known to never be followed by a comma, it is safe to use the relaxed grammar i.e. `Context::Item`.

dependencies/syn/src/verus.rs
mkovaxx: These common mistakes in clauses are now detected:
- missing parentheses around block
- missing comma before block
- missing comma after block

PR #2099, opened 2026-01-21
added flag in settings.json.template to prevent error in rust analyzer

ahuoguo:
Rust analyzer will have an error in `source/cargo-verus/tests/src/utils.rs` without the added feature flag


mkovaxx:
@ahuoguo Apologies for the delay here!

The `integration-tests` feature has since been completely removed by #2383 as part of refactoring the `cargo-verus` tests, so this should no longer be an issue.

PR #2100, opened 2026-01-21
more precise prophecy-dependency-checking

tjhance:
(Depends on #2070)

This implements more precise prophecy-checking. At present, prophetic values are extremely restrictive. With this PR, you can save prophetic values in local ghost variables, and use them to call (non-tracked) lemmas.

### Design

Only ghost-mode variables may be prophetic. For any ghost-mode local, propheticness is automatically determined based on the propheticness of the initializer expression.

OK:
```rust
#[verifier::prophetic]
uninterp spec fn proph() -> bool;

fn test() {
    let ghost mut x = false;
    let r = Ghost(x);
}
```

Bad:
```rust
#[verifier::prophetic]
uninterp spec fn proph() -> bool;

fn test() {
    let ghost mut x = proph();
    let r = Ghost(x);
}
```

```
error: prophetic value not allowed for 'Ghost' wrapper
  --> tproph.rs:12:19
   |
 8 | uninterp spec fn proph() -> bool;
   |               ------------------ function is declared prophetic at this definition
...
11 |     let ghost mut x = proph();
   |                       ------- the result of this call is prophetic
12 |     let r = Ghost(x);
   |                   ^
   |                   |
   |                   the variable `x` is inferred as prophetic
   |                   this operand of this wrapper is expected to be non-prophetic
```

You can also explicitly mark a variable as prophetic (in case you want to assign a prophetic value to it later):

```rust
fn test() {
    #[verifier::prophetic]
    let ghost mut a = 0;
    // ...
    a = proph();
}
``` 

Though this is probably rare.

### Implemenation

We treat this like another dimension of the "mode". Mode-checking now computes both the mode and the propheticness of each expression. We also use an "outer_proph" (similar to "outer_mode") to track when we're inside a conditional dependent on a prophecy variable.

To handle the macro-expansion of `let ghost x = ...` we need an "infer_proph" concept similar to "infer_mode".

A slight technical snag was in the handling of StmtX::Decl. I had to move the call to `add_pattern` to after the recursive call to `check_place_has_mode`. This shouldn't really make a difference, but the previous configuration was covering up some bugs in the encoding of `mut` parameters. But I need to revamp that soon for new-mut-ref so I'll fix it then.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -214,6 +214,75 @@ impl Proph {
     }
 }
 
+/// For any expression that can't be allowed in prophecy-conditional context
+/// i.e., if it should be an error to do this:
+/// ```
+/// if proph_value() {
+///    e;
+/// }
+/// ```
+/// Then either this function should return a reason, OR it needs to be checked as part
+/// of the main `check_expr` function.
+///
+/// When in doubt, make it an error.
+///
+/// The main purpose of splitting this function out is to make it easy to add a check when
+/// someone adds a new variant to ExprX
+fn outer_reason_by_expr_kind(e: &Expr) -> Option<OuterProphReason> {
+    match &e.x {
+        ExprX::Const(_)
+            | ExprX::Var(_)
+            | ExprX::VarLoc(_)
+            | ExprX::VarAt(..)
+            | ExprX::ConstVar(..)
+            | ExprX::StaticVar(..)
+            | ExprX::Loc(_)
+            | ExprX::Call(..) // requires more complex checks
+            | ExprX::Ctor(..)
+            | ExprX::NullaryOpr(_)
+            | ExprX::Unary(..)
+            | ExprX::UnaryOpr(..)
+            | ExprX::Binary(..)
+            | ExprX::BinaryOpr(..)
+            | ExprX::Multi(..)
+            | ExprX::Quant(..)
+            | ExprX::Closure(..)
+            | ExprX::ArrayLiteral(_)
+            | ExprX::ExecFnByName(_)
+            | ExprX::Choose { .. }
+            | ExprX::WithTriggers { .. }
+            | ExprX::Assign { .. } // requires more complex checks
+            | ExprX::AssignToPlace { .. } // requires more complex checks
+            | ExprX::Fuel(..)
+            | ExprX::RevealString(..)
+            | ExprX::Header(..)
+            | ExprX::AssertAssume { .. }
+            | ExprX::AssertAssumeUserDefinedTypeInvariant { .. }
+            | ExprX::AssertBy { .. }
+            | ExprX::AssertQuery { .. }
+            | ExprX::AssertCompute { .. }
+            | ExprX::If(..)
+            | ExprX::Match(..)
+            | ExprX::Ghost { .. }
+            | ExprX::ProofInSpec(..)
+            | ExprX::Block(..)
+            | ExprX::AirStmt(..)
+            | ExprX::NeverToAny(..)
+            | ExprX::Nondeterministic
+            | ExprX::UseLeftWhereRightCanHaveNoAssignments(..)
+            | ExprX::ReadPlace(..)
+        => None,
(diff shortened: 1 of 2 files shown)


source/vir/src/ast_simplify.rs
```diff
@@ -153,7 +149,7 @@ fn pattern_to_exprs(
         // Mode doesn't matter at this stage; arbitrarily set it to 'exec'
```
Chris-Hawblitzel: Should this comment be changed to `... arbitrarily set it to None`?

source/vir/src/modes.rs
```diff
@@ -47,6 +49,263 @@ pub fn mode_meet(m1: Mode, m2: Mode) -> Mode {
...
+            OuterProphReason::Break => "break",
+            OuterProphReason::Continue => "continue",
+            OuterProphReason::OpenInvariant => "opening an invariant",
+            OuterProphReason::NonSpecClosure => "closure",
+            OuterProphReason::LetElse => "let-else statement",
+        };
+        let mut err = error_with_label(
+            span,
+            format!("{:} cannot occur in prophecy-conditional context", reason_str),
+            "this operation can be performed conditionally when the condition depends on a prophecy",
```
Chris-Hawblitzel: Is this supposed to say "can be" or "cannot be"?

source/vir/src/modes.rs
```diff
@@ -1020,19 +1329,21 @@ fn check_expr_handle_mut_arg(
     outer_mode: Mode,
     expect: Expect,
     expr: &Expr,
-) -> Result<(Mode, Option<Mode>), VirErr> {
-    let mode = match &expr.x {
-        ExprX::Const(_) => Ok(Mode::Exec),
+    outer_proph: &Proph,
+) -> Result<(Mode, Option<Mode>, Proph), VirErr> {
```
Chris-Hawblitzel, line 1395: For calling `outer_proph.outer_check`, I'm wondering if we should use a more defensive coding strategy, where by default, an expression is disallowed from running in a prophecy-conditional context, and you have to opt in to allow it (perhaps by having all the `outer_proph.outer_check`s in a separate match on `&expr.x`, or having the existing match return in indicator of whether `outer_proph.outer_check` needs to be called).  Specifically, I'm worried that if someone adds a new ExprX case, they might forget to think about whether they have to call `outer_proph.outer_check`.

source/vir/src/modes.rs
```diff
@@ -1511,43 +1939,78 @@ fn check_expr_handle_mut_arg(
...
+            let p = check_expr_has_mode(
+                ctxt,
+                record,
+                &mut typing,
+                Mode::Spec,
+                body,
+                Mode::Spec,
+                outer_proph,
+            )?;
+            Ok((Mode::Spec, p))
```
Chris-Hawblitzel, line 2079: Right now, we don't allow function names to be used directly as values of type spec_fn(...) -> ..., but in the future we might allow it.  Should we put in a test case to make sure the `#[verifier::prophetic]` doesn't get ignored in the process?

```
#[verifier::prophetic]
spec fn f() -> int {
    5
}

proof fn test() -> int {
    let g = f; // may be supported in the future, meaning the same as "let g = || f();"
    g() // returning g() should fail, because the result of calling g() is a prophetic int, not an unrestricted int
}
```

source/vir/src/modes.rs
```diff
@@ -1847,32 +2502,59 @@ fn check_expr_handle_mut_arg(
...
                     ctxt,
                     record,
                     &mut typing,
                     arm_outer_mode,
                     guard_condition_expect,
                     &arm.x.guard,
+                    &arm_outer_proph,
                 )?;
+                arm_outer_proph = arm_outer_proph.join(guard_proph.clone());
+                // REVIEW: does this need to accumulate modes from previous guards?
```
Chris-Hawblitzel, line 2601: In the original Verus OOPSLA design, yes.  In the syntax-macro-based design, I think it's an error to use a non-exec guard in an exec match at all, so the `(arm_outer_mode, guard_mode)` computation may be dead code.

Chris-Hawblitzel (approved):

PR #2102, opened 2026-01-21
cargo-verus: add verbose flag

bsdinis:
This allows us to see what commands cargo-verus issues

### DIFF
--- a/source/cargo-verus/src/cli.rs
+++ b/source/cargo-verus/src/cli.rs
@@ -58,6 +58,10 @@ pub struct VerifyCommand {
         help = "Arguments passed to 'verus' after `--`"
     )]
     pub verus_args: Vec<String>,
+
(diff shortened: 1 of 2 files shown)


source/cargo-verus/src/main.rs
```diff
@@ -44,27 +44,31 @@ pub fn main() -> Result<ExitCode> {
             compile_primary: false,
             verify_deps: true,
             warn_if_nothing_verified: true,
+            verbose: parsed_cli.verbose,
```
mkovaxx: By moving the `verbose` field from `CargoVerusCli` to `VerifyCommand`, it'll be included in the `options` above, making these lines unnecessary.

bsdinis: good catch!

bsdinis:
Good catch! thank you

PR #2104, opened 2026-01-21
cfg_features: add `verus_in_verification` feature and documentation on how to use it.

### LINKED ISSUE #2063: Document cfg verus_keep_ghost and perhaps add it to cargo-verus new template
## Background

Since Rust 1.80, `unexpected_cfgs` checking is opt-out - see [blog](https://blog.rust-lang.org/2024/05/06/check-cfg/). this makes verus-specific cfgs generate warning by default when using normal cargo check/build, including

* `cfg(verus_keep_ghost)`
* `cfg(verus_keep_ghost_body)`
* `cfg(verus_verify_core)`

In `vstd` itself, these are already declared in `Cargo.toml` under `[lints.rust]`:

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = [
  'cfg(verus_keep_ghost)',
  'cfg(verus_verify_core)',
  'cfg(verus_keep_ghost_body)',
] }
```

However, for a *new* Verus project created by `cargo verus new`, the generated `Cargo.toml` currently don't have these lints configured. This can be very confusing for new users when they write the following code

```rust
use vstd::prelude::*;

verus! {
    pub mod ghost_mod {
        pub closed spec fn ghost_fn() -> bool { true }
    }

    pub mod test_mod {
        #[cfg(verus_keep_ghost)]
        use crate::ghost_mod::ghost_fn;

        pub fn exec_fn() {
        }
    }
}
```

## Proposal

some template like this:

```toml
[package]
name = "{name}"
version = "0.1.0"
edition = "2021"

[dependencies]
vstd = "=0.0.0-2025-12-28-0056"

[package.metadata.verus]
verify = true

[lints.rust]
# Allow Verus-specific cfgs without triggering `unexpected_cfgs`,
# while still catching typos in cfg names.
unexpected_cfgs = { level = "warn", check-cfg = [
  'cfg(verus_keep_ghost)',
  'cfg(verus_keep_ghost_body)',
  'cfg(verus_verify_core)',
] }
```

bsdinis:
From the meeting yesterday, this implements the first step of the solution regarding advice to users on how to guard use statements of erased elements. Also adds the appropriate lint suppression to Cargo.toml generated by `cargo verus new` and inline documentation of the feature flag in the Cargo.toml and in the book.

Also started a [discussion](https://github.com/verus-lang/verus/discussions/2101) to track this more deeply.

Fixes #2063 

### DIFF
--- a/source/cargo-verus/src/subcommands.rs
+++ b/source/cargo-verus/src/subcommands.rs
@@ -64,7 +64,22 @@ vstd = "=0.0.0-2026-01-25-0057"
 
 [package.metadata.verus]
 verify = true
-"#
+
+[lints.rust]
+# Verus supports ghost code, code that is used for proofs but erased during compilation.
+# This means that ghost items that are imported via `use` will not exist during a normal
+# `cargo build`, leading to compilation errors. These errors can be prevented by guarding the
+# use statements with the feature flag `verus_only`, which Verus turns on during
+# verification.
+#
+# WARNING: this flag should only be used on import statements and setting config attributes,
+# see the documentation (https://verus-lang.github.io/verus/guide/erasure.html) for more details.
+#
+# This lint suppression prevents cargo from complaining about the
+# `verus_only` feature flag being undeclared.
+unexpected_cfgs = {{ level = "warn", check-cfg = [
+  'cfg(verus_only)',
+] }}"#
     );
 
     let project_dir = PathBuf::from(name);


source/cargo-verus/src/subcommands.rs
```diff
@@ -64,7 +64,22 @@ vstd = "=0.0.0-2026-01-25-0057"
 
 [package.metadata.verus]
 verify = true
-"#
+
+[lints.rust]
+# Because of ghost erasure, ghost items imported via `use` will not
```
parno: I think this paragraph might be slightly clearer as:

Verus supports ghost code, code that is used for proofs but erased
during compilation.  This means that ghost items that are imported
via `use` will not exist during a normal `cargo build`, leading to
compilation errors.  These errors can be prevented by guarding the
`use` statements with the feature flag `verus_in_verification`,
which Verus turns on during verification.

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
+# Ghost Erasure
+
+Verus performs ghost erasure -- code that exists for verification purposes is removed when building
```
parno: I'd use a colon instead of `--` here.  And change "code" to "ghost code"

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+the executable artifacts, ensuring they are minimally disturbed.
+
+One byproduct of this is that certain identifiers that exist at verification time will not exist at
+compile time. This means that code like the one below would fail to _compile_:
```
parno: I would change the last sentence to:

> This means that the code shown below would fail to _compile_:

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+use vstd::prelude::*;
+
+verus! {
+
+pub mod ghost_mod {
+    pub closed spec fn ghost_fn() -> bool { true }
+}
+
+pub mod test_mod {
+    use crate::ghost_mod::ghost_fn; // FAILS: during compilation ghost_fn is erased
```
parno: I think this would be clearer as: 

> FAILS: During compilation, ghost_fn is erased, causing rustc to complain about a missing definition

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+
+    pub fn exec_fn() -> u64 {
+        1
+    }
+}
+```
+
+To remedy this, Verus has the `verus_in_verification` flag, which is turned on during
```
parno: "Verus provides the..."

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+verification, but is otherwise off. This allows us to guard use statements like the one above:
```
parno: "guard `use` statements"

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+
+```rust
+#[cfg(verus_in_verification)]
+use crate::ghost_mod::ghost_fn; // OK: during compilation use statement is removed
```
parno: I'd  use:
> OK: During compilation, the `use` statement is removed

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+> [!CAUTION]
+> This should only be used to guard use statements. Using this feature flag
```
parno: `use`

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,60 @@
...
+> for conditional compilation of code can introduce **unsoundness**, breaking
+> the verification guarantees.
+
+> [!NOTE]
+> Verus is under _active development_ meaning this option may change or be removed
+> (in particular if a better solution to this problem is solved). See the [Github discussion](https://github.com/verus-lang/verus/discussions/2101)
```
parno: "is solved" -> "is devised"

source/rust_verify/src/cargo_verus.rs
```diff
@@ -54,12 +54,26 @@ pub fn extend_args_and_check_is_direct_rustc_call(
...
+    // For verifying client crates, we allow verus_in_verification
+    //
+    // NOTE: right now this code is unused while verifying verus because we verify via vargo,
+    // which calls rust_verify directly. If we support building verus with cargo-verus, then
+    // this would be "solved"
     if verus_crate {
+        const VERUS_CFGS: [&str; 4] =
+            ["verus_keep_ghost", "verus_keep_ghost_body", "verus_verify_core", "verus_no_vstd"];
+        // verus_crate ==> package_id is Some
+        // TODO: is there a better way of identifying vstd?
```
parno: @tjhance Do you know of something better here?

tjhance: i think the syntax macro checks if the package name is "vstd" based on the CARGO_PKG_NAME environment variable. I don't know what the package_id is or how that relates to the package name. But the starts_with check seems dubious to me.

bsdinis: `starts_with` is definitely dubious -- if someone has a crate called `vstd-extensions` it would be picked up here to. I can try to check if `CARGO_PKG_NAME` works here too, would be really nice

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,67 @@
...
+    use crate::ghost_mod::ghost_fn;
+//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//  OK: During compilation, the `use` statement is removed
+}
+```
+
+> [!CAUTION]
+> This should only be used to guard `use` statements. Using this feature flag
+> for conditional compilation of code can introduce **unsoundness**, breaking
+> the verification guarantees.
```
Chris-Hawblitzel: Can you add an example of what is unsound? It's definitely not subtle -- using cfgs, whether it's our cfg or some other cfgs, you can arrange to verify code that is completely different from the code you actually compile. For example (using one of the internal verus cfgs):

```
fn f() -> (u: u32)
    ensures u != 0
{
    #[cfg(verus_keep_ghost_body)]
    {
        return 3;
    }

    #[cfg(not(verus_keep_ghost_body))]
    {
        return 0;
    }
}

#[verifier::external]
fn main() {
    // when compiled, divides by zero:
    println!("{}", 30 / f());
}
```

bsdinis: Can definitely add an example. It is definitely not subtle but people might be tempted to use it to get away with stuff where Verus support is lacking. 

The difference with using our `cfg` is that we are the ones that control when that cfg is turned on. So users would have a way to hook into it when running code vs. verifying

source/docs/guide/src/erasure.md
```diff
@@ -0,0 +1,67 @@
...
+pub mod test_mod {
+    #[cfg(verus_in_verification)]
+    use crate::ghost_mod::ghost_fn;
+//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//  OK: During compilation, the `use` statement is removed
+}
+```
+
+> [!CAUTION]
+> This should only be used to guard `use` statements. Using this feature flag
```
Chris-Hawblitzel: I think this is not the only use.  You can also use it for other configuration, such as placing `#![cfg_attr(verus_keep_ghost, verus::trusted)]` or #![cfg_attr(verus_keep_ghost, verifier::loop_isolation(false))] at the root of a crate or beginning of a file.  There are also some more exotic examples in https://github.com/microsoft/verified-storage/blob/main/capybaraKV/capybarakv/src/pmem/pmem_async_equiv_t.rs .

source/rust_verify/src/cargo_verus.rs
bsdinis: it very much does

source/docs/guide/src/erasure.md
bsdinis: Added an example of this and changed the node to reflect that

source/docs/guide/src/erasure.md
bsdinis: Added the example

source/cargo-verus/src/subcommands.rs
```diff
@@ -64,7 +64,22 @@ vstd = "=0.0.0-2026-01-25-0057"
...
-"#
+
+[lints.rust]
+# Verus supports ghost code, code that is used for proofs but erased during compilation.
+# This means that ghost items that are imported via `use` will not exist during a normal
+# `cargo build`, leading to compilation errors. These errors can be prevented by guarding the
+# use statements with the feature flag `verus_only`, which Verus turns on during
+# verification.
+#
+# WARNING: this flag should only be used on import statements, see the documentation
```
Chris-Hawblitzel: This still says it should only be used on import statements.  Otherwise, I'm fine with this PR.

Chris-Hawblitzel (approved):

source/cargo-verus/src/subcommands.rs
bsdinis: fixed

PR #2107, opened 2026-01-21
real_to_int cast

ahuoguo:
- added real_to_int cast, which works as a `floor` function. This cast is needed for some theorems between real and int
- also added SpecNeg for reals

### DIFF
--- a/source/rust_verify/src/fn_call_to_vir.rs
+++ b/source/rust_verify/src/fn_call_to_vir.rs
@@ -1251,10 +1251,8 @@ fn verus_item_to_vir<'tcx, 'a>(
             let source_vir0 = expr_to_vir(bctx, &args[0], ExprModifier::REGULAR)?;
             let source_vir = source_vir0.consume(bctx, bctx.types.expr_ty_adjusted(&args[0]));
             let source_ty = undecorate_typ(&source_vir.typ);
-            match &*source_ty {
-                TypX::Real => mk_expr(ExprX::Unary(UnaryOp::RealToInt, source_vir)),
-                _ => err_span(expr.span, "floor expected real type"),
-            }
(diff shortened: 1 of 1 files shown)


tjhance:
I don't know if as-casting is a good choice for this. Would an explicit `floor` function be more appropriate?

Note that (in normal Rust) as-casting a float to an integer is round-towards-0, not floor.

ahuoguo:
> I don't know if as-casting is a good choice for this. Would an explicit `floor` function be more appropriate?
> 
> Note that (in normal Rust) as-casting a float to an integer is round-towards-0, not floor.

FWIW dafny uses `Floor` explicitly. I will switch to floor then

source/rust_verify/src/fn_call_to_vir.rs
```diff
@@ -1245,6 +1245,17 @@ fn verus_item_to_vir<'tcx, 'a>(
...
         }
+        VerusItem::UnaryOp(UnaryOpItem::RealFloor) => {
+            record_spec_fn_allow_proof_args(bctx, expr);
+            unsupported_err_unless!(args.len() == 1, expr.span, "expected 1 argument", &args);
+            let source_vir0 = expr_to_vir(bctx, &args[0], ExprModifier::REGULAR)?;
+            let source_vir = source_vir0.consume(bctx, bctx.types.expr_ty_adjusted(&args[0]));
+            let source_ty = undecorate_typ(&source_vir.typ);
+            match &*source_ty {
+                TypX::Real => mk_expr(ExprX::Unary(UnaryOp::RealToInt, source_vir)),
+                _ => err_span(expr.span, "floor expected real type"),
```
Chris-Hawblitzel: Since the `floor` function at the Rust level has a self argument type `real`, I think this should be guaranteed to be `real` here.  You could just make this an assertion.

Issue #2110, opened 2026-01-22
A bug in the interpreter


SSSPigeon:
```use verus_builtin::*;
use verus_builtin_macros::*;

verus! {

fn test (u : i8) {
    assert(u < 100 ==> true) by (compute);
    
}

fn main() {
    test(98);
    return
}


}
```

When running this, Veurs reports 

```
error: expression simplifies to false
 --> /playground/src/main.rs:7:12
  |
7 |     assert(u < 100 ==> true) by (compute);
  |            ^^^^^^^^^^^^^^^^

error: aborting due to 1 previous error
```
See the example here: [https://play.verus-lang.org/?version=stable&mode=basic&edition=2021](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021).

I think the bug is here: [https://github.com/verus-lang/verus/blob/main/source/vir/src/interpreter.rs#L1424](https://github.com/verus-lang/verus/blob/main/source/vir/src/interpreter.rs#L1424).
Instead of `Const(Bool(true)) => bool_new(false),`, it should be `Const(Bool(true)) => bool_new(true),`.

parno:
Good catch -- I think you've correctly diagnosed the bug and the necessary fix.

parno:
If you're able to open a PR with your proposed fix and a test case for it, that would be wonderful.  Otherwise, I'll add it to my queue of things to do.

SSSPigeon:
@parno Sure, I can do this.

PR #2114, opened 2026-01-26
Fixed a bug in the interpreter and added a test case

SSSPigeon:
This commit fixed the bug mention in #2110 , and added a test case.

### DIFF
--- a/source/rust_verify_test/tests/assert_by_compute.rs
+++ b/source/rust_verify_test/tests/assert_by_compute.rs
@@ -826,10 +826,5 @@ test_verify_one_file! {
         fn test_implies (u : i8) {
             assert(u < 100 ==> true) by (compute);
         }
-
-        fn main() {
-            test_implies(98);
-            return
-        }
     } => Ok(())
 }


source/rust_verify_test/tests/compute.rs
```diff
@@ -0,0 +1,17 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
```
tjhance: there's an existing file, `assert_by_compute.rs`, you can use

SSSPigeon:
@tjhance Thank you! I have moved the test case to that file.

parno:
Thanks for opening the PR!  For the test, you can remove the `main()` function.  The test harness shouldn't need it.

SSSPigeon:
> Thanks for opening the PR! For the test, you can remove the `main()` function. The test harness shouldn't need it.

Just removed it!

parno:
Perfect, thanks!

PR #2118, opened 2026-01-27
Verus internals updates

elanortang:
Port updates to Verus internals that Travis made on `verify-rustlib` to `main`. 
- Fixes for `ImportedViaCore`.
- Erase assume_specification if `erase_ghost` is true.
- Fixes an issue with stability attributes which comes up when verifying core. 
- Add `VstdItem::SharedReference` for the Rust stdlib verification effort.

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -154,6 +154,8 @@ struct CompilerCallbacksEraseMacro {
 }
 
 impl rustc_driver::Callbacks for CompilerCallbacksEraseMacro {
+    // Adding `override_stability` and `stable_attr` functions is a hacky solution specifically for verifying core,
+    // to fix an issue with stability attributes.
     fn config(&mut self, config: &mut rustc_interface::interface::Config) {
         if self.override_stability {
             config.override_queries = Some(|_session, providers| {
@@ -434,17 +436,6 @@ fn add_stable_attr<'tcx>(
 
     m.map.insert(hir_id.local_id, Box::leak(attrs.into_boxed_slice()));
     Box::leak(Box::new(m))
-
-    /*
-    let max_id = 0;
-    for (id, _) in attrmap.map.iter() {
-        let max_id = std::cmp::max(max_id, id.as_u32());
-    }
-    let local_id = ItemLocalId::from_u32(max_id + 1);
-    let span = tcx.hir_span(owner_id.hir_id);
-    m.map.insert(local_id, stable_attr(span));
-    Box::leak(Box::new(m))
-    */
 }
 
 fn needs_stable_attr<'tcx>(tcx: TyCtxt<'tcx>, owner_id: OwnerId) -> bool {


elanortang:
I added some `#[cfg(verus_verify_core)]` attributes in `source/rust_verify/src/driver.rs` - let me know if I should add more in other places or remove some of them.

source/rust_verify/src/driver.rs
```diff
@@ -147,9 +147,26 @@ for all functions (it would only be needed for functions with tracked data in pr
 */
 struct CompilerCallbacksEraseMacro {
     pub do_compile: bool,
+    pub override_stability: bool,
 }
 
 impl rustc_driver::Callbacks for CompilerCallbacksEraseMacro {
+    #[cfg(verus_verify_core)]
+    fn config(&mut self, config: &mut rustc_interface::interface::Config) {
+        if self.override_stability {
```
elanortang, line 160: The CI is failing because it thinks that override_stability is never read - is that because I have a cfg attribute on this function when I shouldn't have one?

elanortang: Never mind, the solution was to add more cfg attributes

elanortang:
It looks like the failure of `test_opaque_type_assume_spec_ok` can be traced to the changes in `source/builtin_macros/src/syntax.rs`: if I undo them (locally), this test succeeds. Do you have an idea of why those changes would cause type inference to fail?

parno:
> It looks like the failure of `test_opaque_type_assume_spec_ok` can be traced to the changes in `source/builtin_macros/src/syntax.rs`: if I undo them (locally), this test succeeds. Do you have an idea of why those changes would cause type inference to fail?

I'm just guessing, but this bit in the syntax macro seems like a plausible culprit:
```
if self.erase_ghost.erase() {
            Expr::Verbatim(quote! {
                ::core::unimplemented!()
            })
```
Previously this invoked a function on the arguments, which would add additional type constraints that presumably help type inference do a better job.

tjhance:
if the body is unimplemented, it won't be able to find a type to instantiate the opaque type.

tjhance:
possibly better solution: if `self.erase_ghost.erase()` is true, just erase the entire assume_specification function

elanortang:
> possibly better solution: if `self.erase_ghost.erase()` is true, just erase the entire assume_specification function

Would that mean that if `self.erase_ghost.erase()` is true, don't set `e` to be anything, and don't call `stmts.push(Stmt::Expr(e, None))` afterward?

source/rust_verify/src/driver.rs
```diff
@@ -147,9 +147,27 @@ for all functions (it would only be needed for functions with tracked data in pr
 */
 struct CompilerCallbacksEraseMacro {
     pub do_compile: bool,
+    #[cfg(verus_verify_core)]
+    pub override_stability: bool,
 }
 
 impl rustc_driver::Callbacks for CompilerCallbacksEraseMacro {
+    #[cfg(verus_verify_core)]
```
tjhance: Why are these cfg options necessary?

This is already a runtime option, we don't need it to be a compile-time option as well. I'd prefer to avoid unnecessary compile-time options.

tjhance (changes requested):

source/rust_verify/src/driver.rs
elanortang: The last time we talked about this code, you had said that for merging it into main, it was a kind of hacky way to get around errors about stable attributes, so you felt more comfortable putting it behind verify_core. 

But if upon further inspection you think it already has the necessary setup, I can take these cfgs out. (And to clarify, is it correct that you would want either all of them in or none of them?)

tjhance: As long as it's behind the runtime check, that's good enough.

elanortang: They should all be removed now

elanortang:
Regarding the changes to `source/builtin_macros/src/syntax.rs`: according to the [commit](https://github.com/verus-lang/verus/commit/dee1e8e2e2e85d69c1ffdd71fa660f1fa0f3b05a) where @tjhance added them originally, they exist to fix an issue with --compile. 

However, it seems like compilation of the Rust stdlib now works fine if I revert these changes - if I run `./compile.sh` in the branch [`rustlib/add-compiler-builtins`](https://github.com/secure-foundations/rustlib/pull/16) (which fixes errors in compiling `core`, among other things), it still goes through. 

@tjhance are there any other places I should test compilation before determining that the `--compile` issue is no longer a problem?

tjhance:
> @tjhance are there any other places I should test compilation before determining that the --compile issue is no longer a problem?

If it works on your rustlib branch, then it's all good.

elanortang:
> > @tjhance are there any other places I should test compilation before determining that the --compile issue is no longer a problem?
> 
> If it works on your rustlib branch, then it's all good.

It appears that although basic compilation works on `rustlib/add-compiler-builtins`, this is still a problem on `rustlib/vec` due to some additional assume_specifications there on `core::intrinsics::unchecked_add::<T>`. I get this error message: 

> thread 'rustc' panicked at compiler/rustc_mir_transform/src/validate.rs:80:25:
broken MIR in Item(DefId(0:27316 ~ core[856b]::specs::intrinsics::\_verus_external_fn_specification_656_core_32__58__58__32_intrinsics_32__58__58__32_unchecked__add_32__58__58__32__60__32_T_32__62_)) (after phase change to runtime-optimized) at bb0[0]:
Cannot AddUnchecked non-integer type T/#0

I will look into figuring out how to erase the entire assume_specification function.

source/rust_verify/src/driver.rs
```diff
@@ -361,3 +383,88 @@ pub fn run(
...
+        Some(attrs) => attrs.to_vec(),
+    };
+
+    let span = tcx.hir_span(hir_id);
+    attrs.push(stable_attr(span));
+
+    m.map.insert(hir_id.local_id, Box::leak(attrs.into_boxed_slice()));
+    Box::leak(Box::new(m))
+
+    /*
```
Chris-Hawblitzel: What's the commented-out code for?

elanortang: I copied it verbatim from what @tjhance did on the `verify-rustlib` branch - @tjhance, can that be deleted?

PR #2119, opened 2026-01-27
Add layout axioms

elanortang:
Add layout axioms, and move `is_power_2` to `arithmetic/power2.rs`. 

### DIFF
--- a/source/vstd/arithmetic/power2.rs
+++ b/source/vstd/arithmetic/power2.rs
@@ -56,7 +56,9 @@ pub open spec fn is_pow2_exists(n: int) -> bool {
 /// Proof that the recursive and existential specifications for `is_pow2` are equivalent.
 pub broadcast proof fn is_pow2_equiv(n: int)
     ensures
-        #[trigger] is_pow2(n) <==> #[trigger] is_pow2_exists(n),
+        #![trigger is_pow2(n)]
+        #![trigger is_pow2_exists(n)]
+        is_pow2(n) <==> is_pow2_exists(n),
 {
     if is_pow2(n) {
         assert(is_pow2_exists(n)) by {
--- a/source/vstd/layout.rs
+++ b/source/vstd/layout.rs
@@ -245,7 +245,7 @@ pub broadcast axiom fn align_properties<T>()
     ensures
         #![trigger align_of::<T>()]
         size_of::<T>() % align_of::<T>() == 0,
-        is_pow2_exists(align_of::<T>() as int),
+        is_pow2(align_of::<T>() as int),
 ;
 
 /// The alignment is at least 1
@@ -256,6 +256,7 @@ pub broadcast proof fn align_nonzero<T>()
         align_of::<T>() > 0,
 {
     broadcast use crate::vstd::arithmetic::power::lemma_pow_positive, align_properties;
+    broadcast use crate::vstd::arithmetic::power2::is_pow2_equiv;
 
 }
 


source/vstd/layout.rs
```diff
@@ -158,11 +209,167 @@ pub broadcast axiom fn layout_of_references_and_pointers_for_sized_types<T: Size
...
+
+/// `str` has the same layout as `[u8]`, which has the same layout as `u8`.
+/// ([Reference](https://doc.rust-lang.org/reference/type-layout.html#str-layout)).
+pub broadcast axiom fn layout_of_str(x: &str)
+    ensures
+        #![trigger spec_align_of_val::<str>(x)]
+        #![trigger spec_size_of_val::<str>(x)]
+        spec_align_of_val::<str>(x) == align_of::<
+            u8,
+        >(),
```
tjhance: ick, can you disable verusfmt for the file and fix this?

source/vstd/layout.rs
```diff
@@ -158,11 +209,167 @@ pub broadcast axiom fn layout_of_references_and_pointers_for_sized_types<T: Size
         align_of::<*mut T>() == align_of::<usize>(),
 ;
 
+/// Pointers to unsized types have the at least the size and alignment as pointers to sized types
```
tjhance: "have the at least"

source/vstd/layout.rs
```diff
@@ -158,11 +209,167 @@ pub broadcast axiom fn layout_of_references_and_pointers_for_sized_types<T: Size
...
+
+;
+
+/// The size is a multiple of alignment and alignment is always a power of 2
+/// ([reference](https://doc.rust-lang.org/reference/type-layout.html#r-layout.properties.size)).
+pub broadcast axiom fn align_properties<T>()
+    ensures
+        #![trigger align_of::<T>()]
+        size_of::<T>() % align_of::<T>() == 0,
+        is_power_2_exists(align_of::<T>() as int),
```
tjhance: be consistent and use `is_power_2`

source/vstd/layout.rs
```diff
@@ -158,11 +209,167 @@ pub broadcast axiom fn layout_of_references_and_pointers_for_sized_types<T: Size
...
+pub broadcast proof fn align_nonzero<T>()
+    ensures
+        #![trigger align_of::<T>()]
+        align_of::<T>() > 0,
+{
+    broadcast use crate::vstd::arithmetic::power::lemma_pow_positive, align_properties;
+
+}
+
+/// The size of a usize will always be a power of 2.
```
tjhance: I don't see a guarantee about this, but it seems to be true anyway (it seems Rust doesn't support 24-bit). This proof only follows because of Verus's extra assumption that the arch is either 32-bit or 64-bit.

That's fine, but can you make a note of it in the docstring?

source/vstd/layout.rs
```diff
@@ -158,11 +209,167 @@ pub broadcast axiom fn layout_of_references_and_pointers_for_sized_types<T: Size
...
+    reveal(is_power_2);
+
+    assert(is_power_2(4)) by (compute);
+    assert(is_power_2(8)) by (compute);
+}
+
+/// Bounds on the unsigned integer types, with respect to their sizes and bits.
+pub proof fn unsigned_int_max_bounds()
+    ensures
+        (usize::MAX as nat) < pow2(usize::BITS as nat),
```
tjhance: I think you can be more specific with a lot of these:

`usize::MAX == pow2(usize::BITS) - 1`

Since most of the constants are known, this lemma is really just saying

`pow2(8) == 256`, `pow2(16) == 65536` etc.

source/vstd/arithmetic/power2.rs
```diff
@@ -42,6 +43,88 @@ pub open spec fn pow2(e: nat) -> nat
     pow(2, e) as nat
 }
 
+/// Determines recursively if the given integer `n` is a power of 2.
+// TODO add some means for Verus to calculate the size & alignment of types
```
tjhance: remove this TODO

source/vstd/arithmetic/power2.rs
```diff
@@ -42,6 +43,88 @@ pub open spec fn pow2(e: nat) -> nat
     pow(2, e) as nat
 }
 
+/// Determines recursively if the given integer `n` is a power of 2.
```
tjhance: "Is the given integer a power of 2, defined recursively."

source/vstd/arithmetic/power2.rs
```diff
@@ -42,6 +43,88 @@ pub open spec fn pow2(e: nat) -> nat
...
+    if n <= 0 {
+        false
+    } else if n == 1 {
+        true
+    } else {
+        n % 2 == 0 && is_power_2(n / 2)
+    }
+}
+
+/// Existential specification for if the given integer `m` is a power of 2.
```
tjhance: "Is the given integer a power of 2, defined by existential quantification in terms of `pow`."

source/vstd/arithmetic/power2.rs
tjhance: I know this isn't in your diff but can you remove this unneeded decreases clause?

source/vstd/arithmetic/power2.rs
```diff
@@ -42,6 +43,88 @@ pub open spec fn pow2(e: nat) -> nat
...
+pub open spec fn is_power_2_exists(m: int) -> bool {
```
tjhance: be consistent about the variable name; in this file we currently use any of `e`, `n` or `m` for the exponent.

source/vstd/arithmetic/power2.rs
```diff
@@ -42,6 +43,88 @@ pub open spec fn pow2(e: nat) -> nat
     pow(2, e) as nat
 }
 
+/// Determines recursively if the given integer `n` is a power of 2.
+// TODO add some means for Verus to calculate the size & alignment of types
+#[verifier::opaque]
+pub open spec fn is_power_2(n: int) -> bool
```
tjhance: The function at the top is just `pow2`, so to be consistent this should probably be `is_pow2`.

tjhance (changes requested):

source/vstd/layout.rs
elanortang: I used `is_power_2_exists` because I specifically wanted the `exists` formulation in my broadcast axiom - if I use the other version, that changes what proofs go through automatically. 

For example, currently `align_nonzero` goes through with only the broadcast axioms. If I change to `is_power_2` in `align_properties`, even if I bring in the axiom `is_power_2_equiv` in the body of `align_nonzero`, the proof no longer goes through automatically.

source/vstd/layout.rs
elanortang: Done. I disabled it for just this function with `#[verusfmt::skip]`.

source/vstd/layout.rs
elanortang: Done - let me know if you want any wording changes.

source/vstd/arithmetic/power2.rs
elanortang: What about: "Returns true if the given integer is a power of 2 (defined recursively)."

source/vstd/arithmetic/power2.rs
elanortang: What about: "Returns true if the given integer is a power of 2 (defined by existential quantification in terms of `pow`)."

source/vstd/arithmetic/power2.rs
elanortang: Done. Does it still need to be marked opaque?

parno: Probably not.  It's just a wrapper around `pow`, which is already marked opaque.

source/vstd/layout.rs
elanortang: Done. I also made `signed_int_min_max_bounds` more specific.

source/vstd/arithmetic/power2.rs
elanortang: Okay, I also took out the opaque and any `reveal(pow2)` calls.

source/vstd/layout.rs
jaybosamiya-ms: Fascinatingly, moving the `// todo ...` comment to the line _before_ the `spec_align_of_val` actually handles it just fine.  Seems like a variant of https://github.com/verus-lang/verusfmt/issues/159

Basically, comments right after an expr is actually a particular sore spot (I would like to fix this but it is Hard™️), if you tend to put comments _before_ the expr, it is generally more likely to do the right thing.

And yes, if things are not ideal, disabling for a single function is better than disabling for the whole file.  If you ever hit a point where you need to do a whole file disable (not the case here, but just for future reference), it is _definitely_ worth opening an issue in verusfmt.

elanortang: Thanks, good to know! Will keep that in mind if I ever find myself wanting to disable verusfmt for an entire file. 

Fixed the verusfmt issue as per this suggestion.

source/vstd/layout.rs
tjhance: the arithmetic library has a broadcast axiom that they're equivalent: `is_pow2_equiv` so it shouldn't matter

tjhance (approved):

source/vstd/layout.rs
elanortang: As in, it doesn't matter so I can keep it at `is_pow2_exists`? 

Or I should be able to switch it to `is_pow2` without a problem? For some reason changing it to `is_pow2` and bringing in `broadcast use is_pow2_equiv` means I have to add an assert where I didn't used to need an assert.

tjhance: Oh, the triggers on that aren't as eager as I thought. Is that intentional? It needs to see both `is_pow` and `is_pow2_exists` before it triggers automatically.

elanortang: Oh I see. That may have been an accident when trying to make it trigger on either `is_pow` or `is_pow2_exists`, when I didn't understand fully how multiple triggers work. 

I changed it so it actually triggers on either `is_pow` or `is_pow2_exists`, and the proof goes through without an assert.

PR #2121, opened 2026-01-28
Allow mut refs to tracked locations

tjhance:
I'm officially scrapping the plan to create a `&mut tracked T` type.

The new plan, as implemented here:

* Allow taking mutable references to tracked places (which just become normal mutable references)
  * This is sound because a mutable borrow to a tracked place is itself tracked, and can never be coerced back to an exec mode borrow. (If we ever need to model the address, we can say it's some fake address to a fake tracked area of the heap that can never actually be conjured in exec code)
* Allow assigning through a mutable reference in proof mode in the following special situations:
  * The type being assigned is `Tracked<_>` or `Ghost<_>`. (In principle, we could allow any ZST.)
    * This is sound because a ZST assignment is a no-op; thus erasing one is safe.
  * The type being assigned in a struct marked `tracked` or `ghost` or is derived from one.
    * This is sound because no value of such a type can be produced in exec-mode. Therefore, if the location has such a type, then it is proof that the place was tracked-mode to begin with.
  * This check is currently implemented as an ad hoc check in modes.rs rather than (say) a trait. I think an ad hoc check is preferred since the condition is so complicated.

This leaves the question about what to do with generics. I think for generics, we should just have people use `&mut Tracked<T>`. We just need to have an easy way to convert `&mut A` <--> `&mut Tracked<A>`. The `<--` direction already exists, and for the `-->` direction we could add a conversion function with the checked conditions.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -1431,6 +1431,10 @@ fn check_place_rec_inner(
 ///  2. The type is tracked (and thus couldn't exist in an exec-mode place to begin with)
 ///
 /// For (1), we're keeping our check a little weaker,
+///
+/// It would be really nice to support `Option<T>`, but the problem is that this is a non-ZST;
+/// even if T is tracked, it's possible to create an `Option<T>` in exec-mode via `None`
+/// and prohibiting this would be difficult to do.
 fn ok_to_assign_exec_place_in_erased_code(ctxt: &Ctxt, place: &Place) -> bool {
     // Always say no if this doesn't involve a mutable reference.
     // This isn't really necessary as a restriction, but it's only for mutable references


source/rust_verify_test/tests/mut_refs_modes.rs
```diff
@@ -338,17 +338,65 @@ test_verify_one_file_with_options! {
...
+
+                *mut_ref = x1.get();
+            }
+        }
+    } => Ok(())
+}
+
+test_verify_one_file_with_options! {
+    #[test] modify_tracked_wrapped_type_in_proof_code ["new-mut-ref"] => verus_code! {
+        tracked struct X { y: Ghost<(int, int)> }
```
Chris-Hawblitzel, line 372: Here, X doesn't have to be a tracked struct, right?

tjhance: right.

PR #2122, opened 2026-01-30
add PointsToUnaligned to `raw_ptr`

JakeGinesin:
- Added struct, view, and impl for `PointsToUnaligned<T>`, adopting most functions (modulo alignment axioms) from `PointsTo<T>`
- Added `ptr_bounds` axiom to `PointsTo<T>` and `PointsToUnaligned<T>` (adopted from `PointsTo<[T]>`)
- Added conversions between `PointsToUnaligned<T>` and `PointsTo<[T]>`

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -373,7 +373,7 @@ impl<T> PointsTo<T> {
         self.mem_contents().value()
     }
 
-    /// Guarantee that the `PointsTo` for any non-zero-sized type points to a non-null address.
+    /// Guarantee that the `PointsTo` points to a non-null address.
     ///
     /// See <https://doc.rust-lang.org/std/ptr/#safety>
     pub proof fn is_nonnull(tracked &self)
@@ -405,7 +405,7 @@ impl<T> PointsTo<T> {
     pub proof fn ptr_bounds(tracked &self)
         ensures
             self.ptr()@.addr as int >= self.ptr()@.provenance.start_addr(),
-            self.ptr()@.addr as int + size_of::<T>() as int <= self.ptr()@.provenance.start_addr()
+            self.ptr()@.addr + size_of::<T>() <= self.ptr()@.provenance.start_addr()
                 + self.ptr()@.provenance.alloc_len(),
     {
         self.as_unaligned().ptr_bounds()
@@ -429,7 +429,7 @@ impl<T> PointsTo<T> {
         self.inner.is_disjoint(&other.inner)
     }
 
-    /// Convert an aligned PointsTo to an unaligned PointsToUnaligned.
+    /// Convert an aligned `PointsTo` to an unaligned `PointsToUnaligned`.
     /// This is always safe since aligned is stricter than unaligned.
     ///
     /// Ensures pointer locations remain the same, and memory
@@ -442,7 +442,7 @@ impl<T> PointsTo<T> {
         self.inner
     }
 
-    /// Borrow an aligned PointsTo as an unaligned PointsToUnaligned.
+    /// Borrow an aligned `PointsTo` as an unaligned `PointsToUnaligned`.
     /// This is always safe since aligned is stricter than unaligned.
     ///
     /// Ensures pointer locations remain the same, and memory
@@ -482,7 +482,7 @@ impl<T> PointsToUnaligned<T> {
         self.mem_contents().value()
     }
 
-    /// Guarantee that the `PointsToUnaligned` for any non-zero-sized type points to a non-null address.
+    /// Guarantee that the `PointsToUnaligned` points to a non-null address.
     ///
     /// See <https://doc.rust-lang.org/std/ptr/#safety>
     pub axiom fn is_nonnull(tracked &self)
@@ -501,7 +501,7 @@ impl<T> PointsToUnaligned<T> {
     pub axiom fn ptr_bounds(tracked &self)
         ensures
             self.ptr()@.addr as int >= self.ptr()@.provenance.start_addr(),
-            self.ptr()@.addr as int + size_of::<T>() as int <= self.ptr()@.provenance.start_addr()
+            self.ptr()@.addr + size_of::<T>() <= self.ptr()@.provenance.start_addr()
                 + self.ptr()@.provenance.alloc_len(),
     ;
 
@@ -535,6 +535,8 @@ impl<T> PointsToUnaligned<T> {
     ///
     /// Ensures pointer locations remain the same, and memory
     /// initializations states remain the same.
+    ///
+    /// Note: Currently an axiom since we don't have support for coercing equivalent references.
     pub axiom fn as_aligned(tracked &self) -> (tracked perm: &PointsTo<T>)
         requires
             self.ptr()@.addr as int % align_of::<T>() as int == 0,
@@ -567,11 +569,13 @@ impl<T> PointsTo<[T]> {
     }
 
     /// Returns `true` if all of the permission's associated memory is initialized.
+    #[verifier::inline]
     pub open spec fn is_init(&self) -> bool {
         self.is_init_subrange(0, self.mem_contents_seq().len())
     }
 
     /// Returns `true` if all of the permission's associated memory in the given subrange is initialized.
+    #[verifier::inline]
     pub open spec fn is_init_subrange(&self, start_index: int, len: nat) -> bool
         recommends
             0 <= start_index <= start_index + len <= self.mem_contents_seq().len(),
@@ -581,11 +585,13 @@ impl<T> PointsTo<[T]> {
     }
 
     /// Returns `true` if any part of the permission's associated memory is uninitialized.
+    #[verifier::inline]
     pub open spec fn is_uninit(&self) -> bool {
         !self.is_init()
     }
 
     /// Returns `true` if all of the permission's associated memory is uninitialized.
+    #[verifier::inline]
     pub open spec fn is_fully_uninit(&self) -> bool {
         forall|i|
             0 <= i < self.mem_contents_seq().len() ==> self.mem_contents_seq().index(i).is_uninit()
@@ -595,6 +601,7 @@ impl<T> PointsTo<[T]> {
     /// if the permission's associated memory at that index is initialized,
     /// the corresponding index in the sequence holds that value.
     /// Otherwise, the value at that index is meaningless.
+    #[verifier::inline]
     pub open spec fn value(&self) -> Seq<T>
         recommends
             self.is_init(),
@@ -606,6 +613,7 @@ impl<T> PointsTo<[T]> {
     /// if the permission's associated memory at that index is initialized,
     /// the corresponding index in the sequence holds that value.
     /// Otherwise, the value at that index is meaningless.
+    #[verifier::inline]
     pub open spec fn value_subrange(&self, start_index: int, len: nat) -> Seq<T>
         recommends
             0 <= start_index <= start_index + len <= self.mem_contents_seq().len(),
@@ -614,15 +622,11 @@ impl<T> PointsTo<[T]> {
         Seq::new(len, |i| self.mem_contents_seq().index(start_index + i).value())
     }
 
-    /// Guarantee that the `PointsTo` for any non-zero-sized type points to a non-null address.
+    /// Guarantee that the `PointsTo` points to a non-null address.
     ///
     /// Note that the size of a slice is given by the length * `size_of::<\T\>()`.
     /// <https://doc.rust-lang.org/reference/type-layout.html#slice-layout>
-    /// ZST pointers *are* allowed to be null, so we need a precondition that size != 0.
-    /// See <https://doc.rust-lang.org/std/ptr/#safety>
     pub proof fn is_nonnull(tracked &self)
-        requires
-            self.mem_contents_seq().len() * size_of::<T>() != 0,
         ensures
             self.ptr()@.addr != 0,
     {
@@ -657,10 +661,12 @@ impl<T> PointsTo<[T]> {
                 start_index as int + len as int,
             ),
     {
-        broadcast use axiom_ptr_mut_from_data;
-        broadcast use crate::vstd::group_vstd_default;
-        broadcast use group_layout_axioms;
-        broadcast use alloc_bound;
+        broadcast use {
+            axiom_ptr_mut_from_data,
+            crate::vstd::group_vstd_default,
+            group_layout_axioms,
+            alloc_bound,
+        };
 
         let tracked unaligned_self_ref = self.as_unaligned();
 
@@ -701,9 +707,7 @@ impl<T> PointsTo<[T]> {
 
         use_type_invariant(&*self);
         assert((self.ptr()@.addr + start_index * size_of::<T>()) as nat % align_of::<T>() == 0) by {
-            broadcast use lemma_mul_mod_noop_right;
-            broadcast use lemma_add_mod_noop;
-            broadcast use layout_of_sized;
+            broadcast use {lemma_mul_mod_noop_right, lemma_add_mod_noop, layout_of_sized};
 
         };
 
@@ -768,18 +772,10 @@ impl<T> PointsTo<[T]> {
             points_to.is_init(),
             points_to.value() as int == to_big_from_digits::<V, T>(self.value()).index(0),
     {
-        broadcast use axiom_ptr_mut_from_data;
-        broadcast use crate::vstd::group_vstd_default;
+        broadcast use {axiom_ptr_mut_from_data, crate::vstd::group_vstd_default};
 
         let tracked ua = self.as_unaligned();
-
-        assert(ua.is_init());
-        assert(ua.value().len() * size_of::<T>() == size_of::<V>());
-
         let tracked pt_unaligned = ua.cast_points_to_unaligned::<V>();
-
-        assert(self.value() =~= ua.value());
-
         pt_unaligned.as_aligned()
     }
 
@@ -818,15 +814,7 @@ impl<T> PointsTo<[T]> {
         broadcast use crate::vstd::group_vstd_default;
 
         let tracked ua = self.as_unaligned();
-
-        assert(ua.is_init());
-        assert(ua.value().len() * size_of::<T>() == size_of::<V>());
-
-        let tracked pt = ua.cast_points_to_unaligned::<V>();
-
-        assert(self.value() =~= ua.value());
-
-        pt
+        ua.cast_points_to_unaligned::<V>()
     }
 
     /// Guarantees that the memory ranges associated with two permissions will not overlap,
@@ -847,7 +835,7 @@ impl<T> PointsTo<[T]> {
         self.inner.is_disjoint(&other.inner)
     }
 
-    /// Convert an aligned PointsTo<[\T\]> to an unaligned PointsToUnaligned<[\T\]>.
+    /// Convert an aligned `PointsTo<[\T\]>` to an unaligned `PointsToUnaligned<[\T\]>`.
     /// This is always safe since aligned is stricter than unaligned.
     ///
     /// De-axiomitized: simply returns the inner [`PointsToUnaligned<[T]>`](PointsToUnaligned).
@@ -859,7 +847,7 @@ impl<T> PointsTo<[T]> {
         self.inner
     }
 
-    /// Borrow an aligned PointsTo<[\T\]> as an unaligned PointsToUnaligned<[\T\]>.
+    /// Borrow an aligned `PointsTo<[\T\]>` as an unaligned `PointsToUnaligned<[\T\]>`.
     /// This is always safe since aligned is stricter than unaligned.
     ///
     /// De-axiomitized: simply borrows the inner `PointsToUnaligned<[\T\]>`.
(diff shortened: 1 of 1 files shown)


source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
     ;
+
+    /// Convert an aligned PointsTo to an unaligned PointsToUnaligned.
+    /// This is always safe since aligned is stricter than unaligned.
+    pub axiom fn into_unaligned(tracked self) -> (tracked perm: PointsToUnaligned<T>)
+        ensures
+    // pointer locations remain the same
+
+            perm.ptr() == self.ptr(),
+            // the status of memory initialization also remains the same
```
elanortang: Can you move the information in these comments to the function documentation?

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+            perm.opt_value() == self.opt_value(),
+    ;
+
+    /// Borrow as unaligned (non-consuming).
+    pub axiom fn as_unaligned(tracked &self) -> (tracked perm: &PointsToUnaligned<T>)
+        ensures
+    // pointer locations remain the same
+
+            perm.ptr() == self.ptr(),
+            // the status of memory initialization also remains the same
```
elanortang: same here

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+    pub axiom fn into_unaligned(tracked self) -> (tracked perm: PointsToUnaligned<T>)
+        ensures
+    // pointer locations remain the same
+
+            perm.ptr() == self.ptr(),
+            // the status of memory initialization also remains the same
+            perm.opt_value() == self.opt_value(),
+    ;
+
+    /// Borrow as unaligned (non-consuming).
```
elanortang: Can you make this more detailed, closer to the documentation for `into_unaligned`?

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+    pub axiom fn into_aligned(tracked self) -> (tracked perm: PointsTo<T>)
+        requires
+    // require alignment as a precondition for conversion
+
+            self.ptr()@.addr as int % align_of::<T>() as int == 0,
+        ensures
+    // ensure the pointer values are the same
+
+            perm.ptr() == self.ptr(),
+            // and the memory initialization status
```
elanortang: Move comments into function documentation

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+            perm.opt_value() == self.opt_value(),
+    ;
+
+    pub axiom fn as_aligned(tracked &self) -> (tracked perm: &PointsTo<T>)
```
elanortang, line 540: Add documentation

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+    // pointer locations remain the same
+
+            perm.ptr() == self.ptr(),
+            // the status of memory initialization also remains the same
+            perm.opt_value() == self.opt_value(),
+    ;
+
+    /// Cast a `PointsTo<T>` to a `PointsTo<V>`, provided memory is initialized,
+    /// the pointer is aligned to `V`, and sizes match.
+    pub axiom fn cast_points_to<V>(tracked &self) -> (tracked points_to: &PointsTo<V>)
```
elanortang: You can't cast an arbitrary `T` to an arbitrary `V` - you have no idea if you are violating some type invariant of `V`. (This was what Natalie spent a long time working out with transmutes.) The reason that the `SharedReference<[T]>` cast was okay was because we restricted `T` and `V` to be `Integer` types, so there the cast was about interpreting an array of say `u8` as a `usize` (accounting for endianness). The documentation for that function should probably make that more clear. 

More generally, although you can always cast one kind of pointer to another kind of pointer, since pointers are just about their addresses, you cannot always cast one kind of `PointsTo` to another kind of `PointsTo`, because `PointsTo` fundamentally also represents the value pointed-to by the pointer. Any cast function between `PointsTo` also needs to account for how the value would change. (E.g., `points_to.value() as int == to_big_ne::<V, T>(self.value()).index(0)` below.)

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+        requires
+            self.ptr()@.addr as int % align_of::<T>() as int == 0,
+        ensures
+            perm.ptr() == self.ptr(),
+            perm.opt_value() == self.opt_value(),
+    ;
+
+    /// Cast a `PointsToUnaligned<T>` to a `PointsToUnaligned<V>`, provided memory
+    /// is initialized and sizes match. No alignment requirement.
+    pub axiom fn cast_points_to<V>(tracked &self) -> (tracked points_to: &PointsToUnaligned<V>)
```
elanortang: The same point applies

source/vstd/raw_ptr.rs
```diff
@@ -419,8 +576,8 @@ impl<T> PointsTo<[T]> {
     /// and `self.value().len() * size_of::<T>() == size_of::<V>()`,
     /// we can always cast a `[T]` permission to a `V` permission.
```
elanortang, line 1083: Can you update this documentation to say

```
/// We can cast a `[T]` permission to a `V` permission under the following conditions: 
/// 
/// (1) `T` and `V` are integer types where `V` is a power of 2
/// and the bit encoding of a `V` can be viewed as
/// the bit encoding for multiple `T`s 
/// (as defined precisely in the trait `CompatibleSmallerBaseFor<V>`). 
/// 
/// (2) Memory is initialized. 
/// 
/// (3) The pointer's address is aligned to `V`. 
/// 
/// (4) `self.value().len() * size_of::<T>() == size_of::<V>()`. 
```

source/vstd/raw_ptr.rs
```diff
@@ -285,6 +311,137 @@ impl<T> PointsTo<T> {
...
+            points_to.is_init(),
+    ;
+}
+
+impl<T> PointsToUnaligned<T> {
+    /// The pointer that this permission is associated with.
+    pub uninterp spec fn ptr(&self) -> *mut T;
+
+    /// The (possibly uninitialized) memory that this permission gives access to.
+    pub uninterp spec fn opt_value(&self) -> MemContents<T>;
```
elanortang: I currently have a PR under review which changes all `opt_value` to `mem_contents`. It might make more sense to merge this one in afterward, with the update to use `mem_contents` instead?

source/vstd/raw_ptr.rs
JakeGinesin: thanks for the clarification!

source/vstd/raw_ptr.rs
JakeGinesin: updated!

source/vstd/raw_ptr.rs
```diff
@@ -166,11 +167,10 @@ pub ghost struct PtrData<T: ?Sized> {
 //   no knowledge about what's in memory
 //   (to be pedantic, the bytes might be initialized in rust's abstract machine,
 //   but we don't know so we have to pretend they're uninitialized)
-#[verifier::external_body]
+//
```
elanortang: can delete this line

source/vstd/raw_ptr.rs
```diff
@@ -277,8 +285,18 @@ impl<T> View for PointsToUnaligned<T> {
 }
 
 impl<T> PointsTo<T> {
+
+    /// Alignment enforced as a type invariant.
+    // #[verifier::type_invariant]
+    // spec fn inv(self) -> bool {
```
elanortang: how come this is commented out at the moment? is there some issue with the type invariant?

source/vstd/raw_ptr.rs
```diff
@@ -332,14 +348,16 @@ impl<T> PointsTo<T> {
...
         ensures
             self.ptr() == old(self).ptr(),
             self.is_uninit(),
-    ;
+    {
+        self.inner.leak_contents();
+    }
 
-    /// The memory associated with a pointer should always be within bounds of its spatial provenance.
+       /// The memory associated with a pointer should always be within bounds of its spatial provenance.
```
elanortang: remove extra spaces

source/vstd/raw_ptr.rs
```diff
@@ -357,39 +375,45 @@ impl<T> PointsTo<T> {
...
+    }
 }
 
 impl<T> PointsToUnaligned<T> {
     /// The pointer that this permission is associated with.
-    pub uninterp spec fn ptr(&self) -> *mut T;
+    // pub uninterp spec fn ptr(&self) -> *mut T;
 
     /// The (possibly uninitialized) memory that this permission gives access to.
     pub uninterp spec fn mem_contents(&self) -> MemContents<T>;
```
elanortang, line 461: can this also be implemented as the `mem_contents` of the underlying `PointsToUnaligned`?

source/vstd/raw_ptr.rs
elanortang, line 540: It might be good to move this out of the `PointsTo` impl so that the `PointsTo` impl is entirely verified, and define it as a separate axiom (similar to how I did conversions from a `MapPointsTo` to something else).

source/vstd/raw_ptr.rs
```diff
@@ -492,12 +519,20 @@ pub broadcast axiom fn pt_slice_len<T>(pt: PointsTo<[T]>)
 ;
 
 impl<T> PointsTo<[T]> {
+
+    /// Alignment enforced as a type invariant for slices.
+    // #[verifier::type_invariant]
+    // spec fn inv(self) -> bool {
```
elanortang: same question

source/vstd/raw_ptr.rs
```diff
@@ -549,59 +581,122 @@ impl<T> PointsTo<[T]> {
...
-    ///
-    // Note that even for ZSTs, pointers need to be aligned.
-    pub proof fn is_aligned(tracked &self)
-        ensures
-            self.ptr()@.addr as nat % align_of::<T>() == 0,
-    {
-        let tracked m = self.as_map();
-        use_type_invariant(m);
-    }
+    pub axiom fn is_aligned(tracked &self)
```
elanortang: same question

source/vstd/raw_ptr.rs
```diff
@@ -549,59 +581,122 @@ impl<T> PointsTo<[T]> {
...
+    ensures
+        self.ptr()@.addr as nat % align_of::<T>() == 0,
+    ;
 
     /// The memory associated with a pointer should always be within bounds of its spatial provenance.
     pub proof fn ptr_bounds(
         tracked &self,
     )
-    // Q: do I need this requires? When the memory is zero-sized, is it true that we don't expect it to be "in bounds"?
```
elanortang, line 482: I wrote this question down as a note-to-self to investigate - before we delete it, I'd like to figure out the answer.

source/vstd/raw_ptr.rs
```diff
@@ -549,59 +581,122 @@ impl<T> PointsTo<[T]> {
...
-    pub axiom fn subrange(tracked &self, start_index: nat, len: nat) -> (tracked sub_points_to:
+    pub proof fn subrange(tracked &self, start_index: nat, len: nat) -> (tracked sub_points_to:
         &Self)
         requires
             start_index + len <= self.mem_contents_seq().len(),
         ensures
             sub_points_to.ptr() == ptr_mut_from_data::<[T]>(
                 PtrData {
-                    addr: (self.ptr()@.addr + start_index * size_of::<T>()) as usize,
+                    addr: #[verifier::truncate] ((self.ptr()@.addr + start_index * size_of::<T>()) as usize),
```
elanortang: we don't expect truncation behavior here, right? This address should be within usize bounds

source/vstd/raw_ptr.rs
```diff
@@ -549,59 +581,122 @@ impl<T> PointsTo<[T]> {
...
                     provenance: self.ptr()@.provenance,
-                    metadata: len as usize,
+                    metadata: #[verifier::truncate] (len as usize),
```
elanortang: same here

source/vstd/raw_ptr.rs
```diff
@@ -733,6 +842,151 @@ impl<T> PointsTo<[T]> {
...
+        ensures
+            perm.ptr() == self.ptr(),
+            perm.mem_contents_seq() == self.mem_contents_seq(),
+    {
+        PointsTo { inner: self }
+    }
+
+    /// Borrow an unaligned PointsToUnaligned<[T]> as an aligned PointsTo<[T]>.
+    /// Requires the pointer address to be properly aligned.
+    pub axiom fn as_aligned(tracked &self) -> (tracked perm: &PointsTo<[T]>)
```
elanortang, line 1023: similarly perhaps this can be outside of the impl block

JakeGinesin: I'm unsure what the utility of moving this outside the impl block is, especially since we cannot condense this function since they deal with the different PointsToUnaligned<[T]> and PointsToUnaligned<T> structs. Could you elaborate?

source/vstd/raw_ptr.rs
JakeGinesin: I'm unsure what you mean here. Were you thinking this is the `PointsTo` impl?

elanortang: oh wait yeah, I misread that

source/vstd/raw_ptr.rs
elanortang: Oh sorry I got mixed up again here with whether this was in a `PointsTo` vs a `PointsToUnaligned` impl, ignore that!

source/vstd/raw_ptr.rs
JakeGinesin: yup, truncate was ultimately not needed. I was fortunately able to remove every instance of it without consequence^

source/vstd/raw_ptr.rs
```diff
@@ -306,46 +338,179 @@ impl<T> PointsTo<T> {
...
+        broadcast use layout_of_sized;
+
+        use_type_invariant(&*self);
+        self.inner.leak_contents();
+    }
+
+    /// The memory associated with a pointer should always be within bounds of its spatial provenance.
+    pub proof fn ptr_bounds(tracked &self)
+        requires
+            size_of::<T>() != 0,
```
elanortang: can remove this sized-ness requirement also

source/vstd/raw_ptr.rs
```diff
@@ -306,46 +338,179 @@ impl<T> PointsTo<T> {
...
+        recommends
+            self.is_init(),
+    {
+        self.mem_contents().value()
+    }
+
+    /// Guarantee that the `PointsToUnaligned` for any non-zero-sized type points to a non-null address.
+    pub axiom fn is_nonnull(tracked &self)
+        requires
+            size_of::<T>() != 0,
```
elanortang: can you also bring over the comment about how "ZST pointers *are* allowed to be null" to explain the precondition?

source/vstd/raw_ptr.rs
```diff
@@ -306,46 +338,179 @@ impl<T> PointsTo<T> {
...
+    pub axiom fn leak_contents(tracked &mut self)
+        ensures
+            self.ptr() == old(self).ptr(),
+            self.is_uninit(),
+    ;
+
+    /// The memory associated with a pointer should always be within bounds of its spatial provenance.
+    pub axiom fn ptr_bounds(tracked &self)
+        requires
+            size_of::<T>() != 0,
```
elanortang: can remove this per discussion at our meeting

source/vstd/raw_ptr.rs
```diff
@@ -426,37 +589,437 @@ impl<T> PointsTo<[T]> {
...
+
+// PointsToUnaligned<[T]>: the unaligned slice permission that PointsTo<[T]> delegates to.
+impl<T> PointsToUnaligned<[T]> {
+    /// The sequence of (possibly uninitialized) memory that this permission gives access to.
+    pub uninterp spec fn mem_contents_seq(&self) -> Seq<MemContents<T>>;
+
+    /// Returns `true` if all of the permission's associated memory is initialized.
+    pub open spec fn is_init(&self) -> bool {
+        forall|i|
+            0 <= i < self.mem_contents_seq().len() ==> self.mem_contents_seq().index(i).is_init()
```
elanortang: can you change this to match the `is_init` definition for `PointsTo`?

source/vstd/raw_ptr.rs
```diff
@@ -426,37 +589,437 @@ impl<T> PointsTo<[T]> {
...
+
+    /// Returns a sequence where for each index,
+    /// if the permission's associated memory at that index is initialized,
+    /// the corresponding index in the sequence holds that value.
+    /// Otherwise, the value at that index is meaningless.
+    pub open spec fn value(&self) -> Seq<T>
+        recommends
+            self.is_init(),
+    {
+        Seq::new(self.mem_contents_seq().len(), |i| self.mem_contents_seq().index(i).value())
```
elanortang: can you change this to match the `value` definition for `PointsTo`?

JakeGinesin:
A summary of changes:

- Split `PointsTo` into `PointsTo` (enforces alignment via a type invariant) and `PointsToUnaligned` (no alignment constraint) for both T and [T].

- Ported specs and proofs (e.g., `is_nonnull`, `ptr_bounds`, `is_disjoint`) to `PointsToUnaligned` and made `PointsTo` wrap/delegate to it.

- Added `as_unaligned`, `into_unaligned`, `as_aligned`, and `into_aligned` to cast between the two pointer types.

- Added the `layout_of_sized` axiom and Base/BasePow2 trait implementations for `u32`.

source/vstd/raw_ptr.rs
```diff
@@ -341,43 +373,182 @@ impl<T> PointsTo<T> {
...
-    pub axiom fn is_nonnull(tracked &self)
+    /// Guarantee that the `PointsTo` for any non-zero-sized type points to a non-null address.
+    ///
+    /// ZST pointers *are* allowed to be null, so we need a precondition that size != 0.
+    /// See https://doc.rust-lang.org/std/ptr/#safety
+    pub proof fn is_nonnull(tracked &self)
+        requires
+            size_of::<T>() != 0,
         ensures
             self.ptr()@.addr != 0,
```
tjhance, line 381: Why is this precondition being added?

JakeGinesin: I believe this is a remnant of an old version of verify-rustlib -- I see this commit (https://github.com/verus-lang/verus/commit/53ee29f044879ec6ee889ee5b41df01137903afe) removes this precondition. Just updated my branch to match.

source/vstd/raw_ptr.rs
```diff
@@ -454,41 +616,452 @@ impl<T> PointsTo<[T]> {
...
+        ensures
+            self.ptr()@.provenance.start_addr() <= self.ptr()@.addr,
+            self.ptr()@.addr + self.mem_contents_seq().len() * size_of::<T>()
+                <= self.ptr()@.provenance.start_addr() + self.ptr()@.provenance.alloc_len(),
+    {
+        self.inner.ptr_bounds();
+    }
+
+    /// Given that the subrange is within bounds, it is always possible to get a permission to just that subrange.
+    pub proof fn subrange(tracked &self, start_index: nat, len: nat) -> (tracked sub_points_to:
```
elanortang, line 647: can you check to make sure all of the asserts in this proof are necessary?

source/vstd/raw_ptr.rs
```diff
@@ -452,43 +622,443 @@ impl<T> PointsTo<[T]> {
...
+    /// Guarantee that the `PointsTo` points to a non-null address.
     ///
-    /// Note that the size of a slice is given by the length * `size_of::<T>()`.
-    // https://doc.rust-lang.org/reference/type-layout.html#slice-layout
-    // ZST pointers *are* allowed to be null, so we need a precondition that size != 0.
-    // See https://doc.rust-lang.org/std/ptr/#safety
-    pub axiom fn is_nonnull(tracked &self)
-        requires
-            self.mem_contents_seq().len() * size_of::<T>() != 0,
+    /// Note that the size of a slice is given by the length * `size_of::<\T\>()`.
```
elanortang, line 627: do you still need \T\ if it's wrapped in the `` for code?

source/vstd/raw_ptr.rs
elanortang, line 1084: Does this have to be an axiom? It seems like it would follow from `cast_points_to_unaligned` and `as_aligned`.

JakeGinesin: No! I was able to make this a proof.

source/vstd/raw_ptr.rs
JakeGinesin: Fixed!

source/vstd/raw_ptr.rs
JakeGinesin: Removed some asserts! (Though less than I would have hoped)

PR #2125, opened 2026-01-30
Add EndianNat for reasoning about representations in different bases and endianness

zero-to-nat:
`EndianNat` provides utilities for reasoning about the per-digit representation of a non-negative number in a given base:
- `to_nat` and `from_nat` for converting to/from a natural number
- `to_big` and `from_big` for converting to/from a different base

These changes are part of the Rust standard library verification effort.

### DIFF
--- a/source/vstd/endian.rs
+++ b/source/vstd/endian.rs
@@ -1084,6 +1084,26 @@ impl<B: Base> EndianNat<B> {
         assert(Self::to_big(Self::from_big(n1)) == n1);
         assert(Self::to_big(Self::from_big(n2)) == n2);
     }
+
+    pub proof fn to_big_single<BIG>(x: EndianNat<B>) where
+        BIG: BasePow2,
+        B: CompatibleSmallerBaseFor<BIG>,
+
+        requires
+            x.len() == Self::exp(),
+            x.len() > 0,
+        ensures
+            Self::to_big(x).len() == 1,
+            Self::to_big(x).index(0) == x.to_nat() as int,
+    {
+        broadcast use EndianNat::exp_properties;
+
+        reveal(EndianNat::to_big);
+        assert(x =~= x.take_least(Self::exp()));
+        assert(Self::to_big(x) =~= Self::to_big(x.skip_least(Self::exp())).append_least(
+            EndianNat::new(x.endian, seq![x.take_least(Self::exp()).to_nat() as int]),
+        ));
+    }
 }
 
 /***** Functions involving both little and big endian *****/


source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+pub trait Base {
+    spec fn base() -> nat;
+
+    proof fn base_min()
+        ensures
+            Self::base() > 1,
+    ;
+}
+
+/// The maximum value that can be stored with `len` digits in base [`B::base()`].
```
elanortang: It seems like in other places in the standard library, "max" means that that value is allowed, but any more than that value cannot be represented (i.e., for `u8`, `u8::MAX` is 255 rather than 256). Here, this seems like an exclusive upper bound instead?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+pub trait CompatibleSmallerBaseFor<BIG: BasePow2>: BasePow2 {
+    proof fn compatible()
+        ensures
+            BIG::bits() > Self::bits() && BIG::bits() % Self::bits() == 0,
+    ;
+}
+
+/***** Endian *****/
+
+pub enum Endian {
```
elanortang, line 54: Add a line of documentation?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                        rest.len(),
+                    )) + rest.to_nat(),
+                    endian.to_nat() * base_max_val::<B>(rest.len()) * B::base() == endian.to_nat()
+                        * base_max_val::<B>(rest.len() + 1),
+                    other.len() == rest.len() + 1,
+            ;
+        }
+    }
+
+    /// Converts a natural number to an `EndianNat` representation with the specified number of digits (`len`).
```
elanortang: and the default endianness `endianness()`.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        reveal(EndianNat::to_nat);
+
+        if self.len() == 0 {
+        } else {
+            self.drop_least().to_nat_big_left_little_right_eq();
+        }
+    }
+
+    // TODO: How many of the triggering issues that arose with the broadcast pow properties are due to having mul_recursive and pow not opaque?
+    #[verifier::spinoff_prover]
```
elanortang: does the proof time out without `spinoff_prover`?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        reveal(EndianNat::to_nat_left);
+        reveal(EndianNat::to_nat);
+
+        if self.len() == 0 {
+        } else {
+            self.drop_least().to_nat_big_left_little_right_eq();
+        }
+    }
+
+    // TODO: How many of the triggering issues that arose with the broadcast pow properties are due to having mul_recursive and pow not opaque?
```
elanortang: I think this TODO can be deleted now? `pow` is now marked `opaque`. I don't remember if there was anything else to do with `mul_recursive` - @parno do you have any ideas?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    (self.drop_first().to_nat_left() * B::base() + self.first()) as int; {
+                        self.drop_first().nat_left_eq_nat_right_little();
+                    }
+                    self.to_nat_right() as int;
+                }
+            }
+        }
+    }
+
+    // TODO: How many of the triggering issues that arose with the broadcast pow properties are due to having mul_recursive and pow not opaque?
```
elanortang: same here

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    ); {
+                        self.drop_last().drop_first().nat_left_eq_nat_right_little();
+                    }
+                    self.drop_last().drop_first().to_nat_left() * B::base()
+                        + self.drop_last().first() + self.last() * pow(
+                        B::base() as int,
+                        (self.len() - 1) as nat,
+                    ); {
+                        assert(self.drop_last().drop_first() == self.drop_first().drop_last());
+                        //broadcast use group_mul_properties;
```
elanortang: can delete

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    ); {
+                        self.drop_first().drop_last().nat_left_eq_nat_right_big();
+                    }
+                    self.drop_first().drop_last().to_nat_right() * B::base()
+                        + self.drop_first().last() + self.first() * pow(
+                        B::base() as int,
+                        (self.len() - 1) as nat,
+                    ); {
+                        assert(self.drop_first().drop_last() == self.drop_last().drop_first());
+                        //broadcast use group_mul_properties;
```
elanortang: can delete

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+            let rest = n.drop_least();
+            Self::from_big_properties(rest);
+            Self::from_nat_properties(least, Self::exp());
+            assert(Self::from_big(n).len() == n.len() * Self::exp()) by (nonlinear_arith)
+                requires
+                    Self::from_big(n).len() == (n.len() - 1) * Self::exp() + Self::exp(),
+            ;
+            let small_least = EndianNat::from_nat(least, Self::exp());
+            let small_rest = Self::from_big(rest);
+            // small_rest.append_least(small_least).to_nat() == (small_rest.to_nat() * base_max_val::<B>(small_least.len())) + small_least.to_nat()
```
elanortang: delete?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+            Self::to_nat_append_least(small_rest, small_least);
+            // least == small_least.to_nat()
```
elanortang: delete?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    n.len() % Self::exp() == 0,
+                    Self::exp() > 0,
+                    n.len() > 0,
+            ;
+            assert(rest.len() % Self::exp() == 0) by (nonlinear_arith)
+                requires
+                    rest.len() == n.len() - Self::exp(),
+                    n.len() % Self::exp() == 0,
+            ;
+            // Self::from_big(Self::to_big(rest)) == rest
```
elanortang: Is this a comment to help structure and read the proof, or leftover from the process of writing it?

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    small.len() == 0,
+                    Self::exp() > 0,
+            ;
+            Self::to_big_properties(small);
+            assert(Self::to_big(Self::from_big(n)).digits == n.digits);
+            assert(Self::to_big(Self::from_big(n)).endian == n.endian);
+        } else {
+            let least = n.least();
+            let rest = n.drop_least();
+            // Self::to_big(Self::from_big(rest)) == rest
```
elanortang: same here

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+            ;
+
+            assert(Self::to_big(small) == n);
+        }
+    }
+}
+
+/***** Functions involving both little and big endian *****/
+
+pub open spec fn to_big_from_digits<BIG, B>(n: Seq<B>) -> EndianNat<BIG> where
```
elanortang, line 1112: add documentation

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        EndianNat { endian: endianness(), digits, phantom: PhantomData }
+    }
+
+    /// Number of digits in this `EndianNat`.
+    pub open spec fn len(self) -> nat {
+        self.digits.len()
+    }
+
+    /// The `i`th digit in this `EndianNat`.
+    pub open spec fn index(self, i: int) -> int {
```
parno: Perhaps we should add a `recommends 0 <= i < self.len()` clause

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+    /// The most significant digit in this `EndianNat`.
+    pub open spec fn most(self) -> nat {
+        match self.endian {
+            Endian::Little => self.last(),
+            Endian::Big => self.first(),
+        }
+    }
+
+    /// Constructs an `EndianNat` by skipping the first `n` digits of the original `EndianNat`. Ignores endianness.
+    pub open spec fn skip(self, n: nat) -> Self {
```
parno: As additional assistance, we could consider adding recommends clauses to these functions as well

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        self.digits.len()
+    }
+
+    /// The `i`th digit in this `EndianNat`.
+    pub open spec fn index(self, i: int) -> int {
+        self.digits[i]
+    }
+
+    /// The first digit in this `EndianNat`. Ignores endianness.
+    pub open spec fn first(self) -> nat {
```
parno: We could add a recommends here and on `last` that `self.len() > 0`

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                    )) + rest.to_nat(),
+                    endian.to_nat() * base_max_val::<B>(rest.len()) * B::base() == endian.to_nat()
+                        * base_max_val::<B>(rest.len() + 1),
+                    other.len() == rest.len() + 1,
+            ;
+        }
+    }
+
+    /// Converts a natural number to an `EndianNat` representation with the specified number of digits (`len`).
+    /// `n` should be less than the maximum number that can be represented in `len` digits in this base.
```
parno, line 424: It might be worth mentioning that it uses the default endianess of the machine

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+            #![trigger Self::from_nat(n, len)]
+            Self::from_nat(n, len).len() == len,
+            Self::from_nat(n, len).endian == endianness(),
+            Self::from_nat(n, len).wf(),
+        decreases len,
+    {
+        if len == 0 {
+        } else {
+            Self::base_max_val_len(n, len);
+            let endian = Self::from_nat(n, len);
```
parno: Can we choose a different name here?  I initially misread it as representing the endianness, which didn't match the RHS.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                B::base() > 0,
+        ;
+    }
+
+    pub broadcast proof fn from_nat_properties(n: nat, len: nat)
+        requires
+            n < base_max_val::<B>(len),
+        ensures
+            #![trigger Self::from_nat(n, len)]
+            Self::from_nat(n, len).len() == len,
```
parno, line 460: This suggests that we don't maintain an invariant that an EndianNat is always represented in the most compact form, right?  I.e., if we call `from_nat(1, 100)`, we might end up with a 100-digit sequence with lots of 0s for padding.  We should probably mention this when introducing `EndianNat` above.

source/vstd/endian.rs
parno: I think this comment is out of date -- `pow` is now marked `opaque`, and `mul_recursive` calls into an opaque function.  Let's just cut it.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        reveal(EndianNat::to_nat_right);
+        reveal(pow);
+        if self.len() == 0 {
+        } else {
+            if self.drop_last().len() == 0 {
+                // This proof is similar to Dafny's in that it uses the same number of steps and hints
+                // However, the first hint below isn't needed for Dafny, since Dafny
+                // provides 2 fuel by default, while Verus provides 1.
+                // The Dafny proof requires a reveal(pow) instead,
+                // because they hide pow, which we should probably do as well.
```
parno: Let's cut this last line, since we now mark `pow` as opaque.

source/vstd/endian.rs
parno: Let's cut this too.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+        reveal(EndianNat::to_nat_right);
+        reveal(pow);
+        if self.len() == 0 {
+        } else {
+            if self.drop_first().len() == 0 {
+                // This proof is similar to Dafny's in that it uses the same number of steps and hints
+                // However, the first hint below isn't needed for Dafny, since Dafny
+                // provides 2 fuel by default, while Verus provides 1.
+                // The Dafny proof requires a reveal(pow) instead,
+                // because they hide pow, which we should probably do as well.
```
parno: same as above.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+                        self.drop_first().nat_left_eq_nat_right_little();
+                    }
+                    self.to_nat_right() as int;
+                }
+            }
+        }
+    }
+
+    // TODO: How many of the triggering issues that arose with the broadcast pow properties are due to having mul_recursive and pow not opaque?
+    proof fn nat_left_eq_nat_right_big(self)
```
parno: In theory we ought to be able to do these proofs in an endian-agnostic way, so that we don't have to duplicate the code like this.  Might be more trouble than it's worth, unless an LLM can do it.

source/vstd/endian.rs
```diff
@@ -0,0 +1,1252 @@
...
+    /// [`B::base()`] to the power of `exp()` is [`BIG::base()`].
+    /// In other words, `exp()` is the number of digits in base `B` that correspond to a single digit in base `BIG`.
+    ///
+    /// [`B::base()`]: Base::base
+    /// [`BIG::base()`]: Base::base
+    pub open spec fn exp<BIG>() -> nat where BIG: BasePow2, B: CompatibleSmallerBaseFor<BIG> {
+        BIG::bits() / B::bits()
+    }
+
+    pub proof fn exp_properties<BIG>() where BIG: BasePow2, B: CompatibleSmallerBaseFor<BIG>
```
parno: Should we make this a broadcast lemma?

source/vstd/endian.rs
parno, line 1112: Can we add a doc string for this?

parno (approved):
Thanks for putting this together.  I added various minor comments and suggestions.

source/vstd/endian.rs
elanortang: What are times when we might care about whether we are calculating `to_nat` left-to-right or right-to-left? I remember that `to_nat_right` and `to_nat_left` were defined before defining the simpler `to_nat`, which uses `least` and `drop_least` to avoid messy definitions.

parno: Good question.  An important use case I can imagine for this library is verifying a big-integer library, e.g., for computing RSA values.  There you'll have functions that take in two BigIntegers `x` and `y`, and returns `z`, where you want to prove something like `z.to_nat() == x.to_nat() + y.to_nat()`.  Your implementation is likely to be a loop over the elements of the BigInteger digits, so you'll want a loop invariant about the partial construction of the `nat` value.  If we can do that from the cleaner `least` and `drop_least`, then we can omit the left-to-right and right-to-left definitions.

zero-to-nat: @elanortang @parno For `to_nat_left` and `to_nat_right`, these are not currently used anywhere in the rustlib project, so it would be fine for us to remove them. I kept the specs and proofs since they are related to `to_nat` and seemed like they could be useful.

To the question of whether `least`/`drop_least` is cleaner than `first`/`drop_first`: FWIW, I did rewrite the original `to_big` and `from_big` specs to use `take_least`/`skip_least` instead of `take`/`skip` because I found that it made the proofs simpler -- the arithmetic was the same for both little endian and big endian. 

However, I could see how the `first`/`drop_first` approach would be more helpful if that aligned with how the implementation processed the data. With that in mind, I would be fine to either keep the spec or remove it.

source/vstd/endian.rs
zero-to-nat: thanks, these were leftover from proof development. I just deleted these comments since they aren't that helpful for understanding the proof IMO

source/vstd/endian.rs
zero-to-nat: It does! I had to remove `reveal(pow)` from the top of the proof and move it to the one place it was needed

source/vstd/endian.rs
parno: I think it's the case that `to_nat` is identical to `to_nat_right` for little endian, and `to_nat_left` for big endian, right?  If so, then I think an implementation that wanted to use `to_nat_right` on a little endian machine, for example, would be able to just use `to_nat`.  If that's the case, then perhaps we want to replace `to_nat_right` and `to_nat_left` with `to_nat_msb`, which would be the dual of `to_nat`.  Then we'd only have two definitions for how to convert an EndianNat to a nat, and we should be able to eliminate the duplication of `nat_left_eq_nat_right_little` with `nat_left_eq_nat_right_big`

zero-to-nat: Following up on this: I added `to_nat_most` and the proof of equivalence to `to_nat`, removing `to_nat_right` and `to_nat_left`.

PR #2132, opened 2026-02-02
feat: support `verus_spec` in verusdoc

### LINKED ISSUE #2082: verusdoc: cannot display `ensures` and `requires`for `verus_spec`
I'm using verusdoc to generate documentation for our verification project. However, the verusdoc does not support the display of the `ensures` and `requires`for `verus_spec`.

Here is an example:

The method `BoxRef::deref_target` is
```rust
    /// Dereferences `self` to get a reference to `T` with the lifetime `'a`.
    #[verus_spec(ret => ensures *ret == self.value())]
    pub fn deref_target(&self) -> &'a T {
```

and the documentation only shows
```
impl<'a, T> BoxRef<'a, T>
  pub exec fn deref_target(&self) -> &'a T
    Dereferences self to get a reference to T with the lifetime 'a.
```
The `ensures *ret == self.value()` specification is not shown anywhere in the generated docs.

~https://asterinas.github.io/vostd/ostd/sync/non_null/struct.BoxRef.html#method.deref_target~

Marsman1996:
Fix #2082

Now the verusdoc  will display the doc for
```rust
#[verus_spec(ret =>
    with
        ghost_in1: Tracked<int>,
        ghost_in2: Tracked<int>,
        ->
        ghost_out: Ghost<(int,int)>,
    requires
        ghost_in1@ >= 0,
    ensures
        ghost_out@.0 == ghost_in1@,
        ret == 0,
)]
pub fn spec_with_example(&self) -> u32 {
```
as
```
pub exec fn spec_with_example(&self) -> ret : u32
with
  ghost_in1: Tracked <int>,
  ghost_in2: Tracked <int>,
  ->
  ghost_out: Ghost <(int,int)>,
requires
  ghost_in1@ >= 0,
ensures
  ghost_out@.0 == ghost_in1@,
  ret == 0,
```

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -36,6 +36,7 @@
 /// - Refer to `examples/syntax_attr.rs`.
 use proc_macro2::TokenStream;
 use quote::{ToTokens, quote, quote_spanned};
+use syn::parse::Parser;
 use syn::visit_mut::VisitMut;
 use syn::{Expr, Item, ItemConst, parse2, spanned::Spanned};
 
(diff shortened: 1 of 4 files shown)


rikosellic:
The generated doc of the example can be found [here](https://asterinas.github.io/vostd/ostd/sync/struct.SpinLock.html#method.spec_with_example).

source/builtin_macros/src/rustdoc.rs
```diff
@@ -330,6 +335,89 @@ fn encoded_str(kind: &str, data: &str) -> String {
...
+        verus_syn::FnArgKind::Typed(pt) => {
+            let s = format_pat_type(pt);
+            format!("{tracked}{s}")
+        }
+    }
+}
+
+fn format_pat_type(pt: &verus_syn::PatType) -> String {
+    let pat = normalize_ws(&pt.pat.to_token_stream().to_string());
+    let ty = tighten_type_spacing(&normalize_ws(&pt.ty.to_token_stream().to_string()));
```
tjhance: We have a printer library that can print verus expressions is a pretty way. For example, we print expressions by doing `verus_prettyplease::unparse_expr(&code);` (see above). You can do the same for types and then you shouldn't have to all this whitespace tightening.

`unparse_expr` is defined here: https://github.com/verus-lang/verus/blob/233fd26e98d209a10e0f43bc85735fcdbb619244/dependencies/prettyplease/src/lib.rs#L388

You should be able to similarly add an `unparse_ty` and `unparse_pat` without much trouble.

tjhance (changes requested):

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -497,9 +498,88 @@ pub(crate) fn rewrite_verus_spec_on_fun_or_loop(
             // To avoid misuse of the unverified function,
             // we add `requires false` and thus prevent verified function to use it.
             // Allow unverified code to use the function without changing in/output.
```
ziqiaozhou: NIT: The comment should be close to the code related to it.

Marsman1996: Updated

PR #2134, opened 2026-02-02
feat: complex return value patterns

mkovaxx:
- Extend the lowering path of `ensures` to support complex patterns for binding the return value.
- Add tests to exercise some complex patterns.
- Extend identifier collision checks to gather all identifiers from a pattern.
- Refactor and simplify some related code.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -536,7 +536,7 @@ impl Visitor {
         // some closures also use this function to handle
         is_closure: bool,
         // function name
-        ident: &Ident,
+        fn_ident: &Ident,
         generics: Option<impl ToTokens>,
         inputs: (Option<impl ToTokens>, impl ToTokens), // optional self and args
     ) -> Vec<Stmt> {
@@ -553,11 +553,17 @@ impl Visitor {
 
         // Either the single identifier in the return pattern, or a fresh identifier.
         let ret_val_ident: &Ident = match ret_pat {
-            Some(Pat::Ident(pat)) if &pat.ident != ident => &pat.ident,
+            Some(Pat::Ident(pat)) if pat.ident != *fn_ident => &pat.ident,
             _ => &Ident::new("_VERUS_ret_ident", Span::call_site()),
         };
 
         fn wrap_with_ret_binding_pat(expr: &mut Expr, ret_val_ident: &Ident, ret_pat: &Pat) {
+            if let Pat::Ident(pat) = ret_pat {
+                if pat.ident == *ret_val_ident {
+                    // The binding would be unnecessary.
+                    return;
+                }
+            }
             let expr_span = expr.span();
             let attrs = expr.replace_attrs(Vec::new());
             let inner = take_expr(expr);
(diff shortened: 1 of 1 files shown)


source/builtin_macros/src/syntax.rs
```diff
@@ -546,37 +548,35 @@ impl Visitor {
...
-        fn wrap_with_ret_binding_pat(expr: &mut Expr, ret_pat: &Pat, final_ret_pat: &Option<Pat>) {
-            if let Some(final_ret_pat) = final_ret_pat {
-                let expr_span = expr.span();
-                let attrs = expr.replace_attrs(Vec::new());
-                let inner = take_expr(expr);
-                let wrapped: Expr =
-                    parse_quote_spanned!(expr_span => { let #final_ret_pat = #ret_pat; #inner });
-                *expr = wrap_expr_with_attrs(wrapped, attrs);
-            }
+        let ret_val_ident: Ident = Ident::new("_VERUS_ret_ident", Span::call_site());
```
mkovaxx: @Chris-Hawblitzel Unconditionally introducing this "internal" identifier allows completely removing identifier collision checks (e.g. against the function's name).

source/vir/src/poly.rs
```diff
@@ -719,8 +713,11 @@ fn visit_exp(ctx: &Ctx, state: &mut State, exp: &Exp) -> Exp {
                 state.types.push_scope(true);
                 let mut new_bs: Vec<VarBinder<Exp>> = Vec::new();
                 for b in bs.iter() {
-                    // TODO: this might be better as just visit_exp:
-                    let a = visit_exp_native(ctx, state, &b.a);
+                    let a = if typ_is_poly(ctx, &b.a.typ) {
+                        visit_exp_poly(ctx, state, &b.a)
+                    } else {
+                        visit_exp_native(ctx, state, &b.a)
+                    };
```
Chris-Hawblitzel: I just fixed this in a different way in https://github.com/verus-lang/verus/commit/51d6c7aa33e949116621129f4bc5632e6b89c96b , so changing poly.rs should no longer be needed.

source/builtin_macros/src/syntax.rs
```diff
@@ -546,37 +548,35 @@ impl Visitor {
...
+
+        fn wrap_with_ret_binding_pat(expr: &mut Expr, ret_val_ident: &Ident, ret_pat: &Pat) {
+            let expr_span = expr.span();
+            let attrs = expr.replace_attrs(Vec::new());
+            let inner = take_expr(expr);
+            let wrapped: Expr =
+                parse_quote_spanned!(expr_span => { let #ret_pat = #ret_val_ident; #inner });
+            *expr = wrap_expr_with_attrs(wrapped, attrs);
```
Chris-Hawblitzel, line 572: As an optimization, can we omit this wrapping in the case where ret_pat is just a single variable that doesn't collide with the function name?  Most functions will fall into this case.

source/builtin_macros/src/syntax.rs
Chris-Hawblitzel: I think this is good, because it's better to handle the case where the return value and the function name are the same than it is to reject the code with an error.

source/vir/src/poly.rs
mkovaxx: Thanks! :)

source/builtin_macros/src/syntax.rs
```diff
@@ -546,37 +551,45 @@ impl Visitor {
...
+            _ => &Ident::new("_VERUS_ret_ident", Span::call_site()),
+        };
+
+        fn wrap_with_ret_binding_pat(expr: &mut Expr, ret_val_ident: &Ident, ret_pat: &Pat) {
+            if let Pat::Ident(pat) = ret_pat {
+                if pat.ident == *ret_val_ident {
+                    // The binding would be unnecessary.
+                    return;
+                }
             }
```
mkovaxx, line 566: @Chris-Hawblitzel The optimization you suggested is implemented here.

source/builtin_macros/src/syntax.rs
mkovaxx: Yes, thanks for the reminder! Done. :)

PR #2137, opened 2026-02-03
fix: skip proof code in doc

### LINKED ISSUE #2135: verusdoc: fail when generating doc for spec with proof code
Verusdoc will panic when generating doc for spec with proof code, which is caused by the code below:

https://github.com/verus-lang/verus/blob/53ee29f044879ec6ee889ee5b41df01137903afe/dependencies/prettyplease/src/expr.rs#L92-L94

Previous there is no proof code in spec function, but now things are different, we can use proof blocks inside spec code is currently for recursion now.
```rust
    pub open spec fn view_rec(self) -> Seq<<T as View>::V>
        decreases L - self.level,
        when self.level < L && self.inv()
    {
        if self.level == L - 1 {
            Seq::empty().push(self.value.view())
        } else {
            proof {
                assert(self.inv_children());
```

Marsman1996:
Fix #2135 

I think skip the proof code in doc is reasonable. We shall focus on the spec code.

With this PR, for code
```rust
pub open spec fn view_rec(self) -> Seq<<T as View>::V>
        decreases L - self.level,
        when self.level < L && self.inv()
    {
        if self.level == L - 1 {
            Seq::empty().push(self.value.view())
        } else {
            proof {
                assert(self.inv_children());
                assert(forall|i: int|
                    0 <= i < Self::size() ==> match #[trigger] self.children[i] {
                        Some(child) => child.level < L,
                        None => true,
                    });
                assert(forall|i: int|
                    0 <= i < Self::size() ==> match #[trigger] self.children[i] {
                        Some(child) => L - child.level < L - self.level,
                        None => true,
                    });
                assert(forall|i: int|
                    0 <= i < Self::size() ==> match #[trigger] self.children[i] {
                        Some(child) => child.inv(),
                        None => true,
                    });
            }

            self.value.view_rec_step(
                Seq::flat_map(
                    Seq::new(self.children.len(), |i: int| i),
                    |i: int|
                        if 0 <= i < Self::size() {
                            match self.children[i] {
                                Some(child) => {
                                    proof {
                                        self.remaining_level_decreases();
                                        assert(self.children[i].is_some());
                                        assert(L - child.level < L - self.level);
                                    }
                                    child.view_rec()
                                },
                                None => Seq::empty(),
                            }
                        } else {
                            Seq::empty()
                        },
                ),
            )
        }
    }
```

The doc is
```
pub open spec fn view_rec(self) -> Seq<<T as View>::V>
{
    if self.level == L - 1 {
        Seq::empty().push(self.value.view())
    } else {
        self.value
            .view_rec_step(
                Seq::flat_map(
                    Seq::new(self.children.len(), |i: int| i),
                    |i: int| {
                        if 0 <= i < Self::size() {
                            match self.children[i] {
                                Some(child) => {
                                    child.view_rec()
                                }
                                None => Seq::empty(),
                            }
                        } else {
                            Seq::empty()
                        }
                    },
                ),
            )
    }
}
```

### DIFF
--- a/dependencies/prettyplease/src/expr.rs
+++ b/dependencies/prettyplease/src/expr.rs
@@ -89,9 +89,7 @@ impl Printer {
             Expr::GetField(expr) => self.expr_get_field(expr),
             Expr::Matches(m) => self.expr_matches(m),
 
-            Expr::Assume(_) | Expr::Assert(_) | Expr::AssertForall(_) | Expr::RevealHide(_) => {
-                unimplemented!("unknown Expr {:?}", expr)
-            }
+            Expr::Assume(_) | Expr::Assert(_) | Expr::AssertForall(_) | Expr::RevealHide(_) => {}
         }
 
         if needs_paren {
--- a/dependencies/prettyplease/src/stmt.rs
+++ b/dependencies/prettyplease/src/stmt.rs
@@ -4,7 +4,7 @@ use crate::expr;
 use crate::fixup::FixupContext;
 use crate::mac;
 use crate::INDENT;
-use verus_syn::{BinOp, Expr, Stmt};
+use verus_syn::{BinOp, Expr, Stmt, UnOp};
 
 impl Printer {
     pub fn stmt(&mut self, stmt: &Stmt, is_last: bool) {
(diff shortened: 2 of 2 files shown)


Marsman1996:
See 
https://asterinas.github.io/vostd/vstd_extra/ghost_tree/struct.Node.html#method.view_rec
for generated doc of the example `view_rec`.

PR #2138, opened 2026-02-04
add support for opaque types defined by assume specification

FeizaiYiHao:
This PR replaces the opaque types in `assume_specification` with the actual opaque types defined by the external function in the Rust to VIR stage, specifically in `mid_ty_to_vir`. All opaque types defined by `assume_specification` are completely ignored and not included in the VIR `Krate`. 

### DIFF
--- a/source/rust_verify/src/rust_to_vir.rs
+++ b/source/rust_verify/src/rust_to_vir.rs
@@ -162,7 +162,8 @@ fn check_item<'tcx>(
         }
 
         let mid_ty = ctxt.tcx.type_of(def_id).skip_binder();
-        let vir_ty = ctxt.mid_ty_to_vir(def_id, item.span, &mid_ty, false).map_err(|e| vec![e], None)?;
+        let vir_ty =
+            ctxt.mid_ty_to_vir(def_id, item.span, &mid_ty, false, None).map_err(|e| vec![e])?;
 
         crate::rust_to_vir_func::check_item_const_or_static(
             ctxt,
@@ -184,8 +185,8 @@ fn check_item<'tcx>(
 
     match &item.kind {
         ItemKind::Fn { sig, generics, body: body_id, .. } => {
-            let _ =
-                check_fn_opaque_ty(ctxt, vir, &item.owner_id.to_def_id()).map_err(|e| vec![e])?;
+            // let _ =
+            //     check_fn_opaque_ty(ctxt, &mut vir.opaque_types, &item.owner_id.to_def_id()).map_err(|e| vec![e])?;
             check_item_fn(
                 ctxt,
                 &mut vir.functions,
--- a/source/rust_verify/src/rust_to_vir_impl.rs
+++ b/source/rust_verify/src/rust_to_vir_impl.rs
@@ -457,12 +457,6 @@ pub(crate) fn translate_impl_item<'tcx>(
                 ImplItemKind::Fn(sig, body_id) => {
                     let kind = mk_trait_function_kind();
 
-                    let _ = crate::rust_to_vir_base::check_fn_opaque_ty(
-                        ctxt,
-                        vir,
-                        &impl_item.owner_id.to_def_id(),
-                    )?;
-
                     check_item_fn(
                         ctxt,
                         &mut vir.functions,
@@ -522,8 +516,7 @@ pub(crate) fn translate_impl_item<'tcx>(
             if let ImplItemKind::Const(_ty, ConstItemRhs::Body(body_id)) = &impl_item.kind {
                 let def_id = body_id.hir_id.owner.to_def_id();
                 let mid_ty = ctxt.tcx.type_of(def_id).skip_binder();
-                let vir_ty =
-                        ctxt.mid_ty_to_vir(def_id, impl_item.span, &mid_ty, false, None)?;
+                let vir_ty = ctxt.mid_ty_to_vir(def_id, impl_item.span, &mid_ty, false, None)?;
                 if trait_path_typ_args.is_none() {
                     crate::rust_to_vir_func::check_item_const_or_static(
                         ctxt,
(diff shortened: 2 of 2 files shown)


FeizaiYiHao:
@tjhance Here's the new PR. I'll close the old one.

examples/opaque_type.rs
```diff
@@ -0,0 +1,40 @@
+use vstd::prelude::*;
+verus!{
+trait DummyTrait{
```
tjhance: a lot of weird formatting in this file

tjhance: also I'm not sure we really need an example file for this in the first place

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -469,11 +470,44 @@ fn compare_external_ty<'tcx>(
...
+    {
+        compare_external_ty(tcx, verus_items, &ty1, &ty2, external_trait_from_to)
+    } else if let (rustc_middle::ty::TyKind::Pat(ty1, _), rustc_middle::ty::TyKind::Pat(ty2, _)) =
+        (ty1.kind(), ty2.kind())
+    {
+        compare_external_ty(tcx, verus_items, &ty1, &ty2, external_trait_from_to)
+    } else if let (rustc_middle::ty::TyKind::Slice(ty1), rustc_middle::ty::TyKind::Slice(ty2)) =
+        (ty1.kind(), ty2.kind())
+    {
+        compare_external_ty(tcx, verus_items, &ty1, &ty2, external_trait_from_to)
```
tjhance: Why is this casework restricted to a small subset of TyKind variants?

examples/opaque_type.rs
FeizaiYiHao: We don't... Committed by accident

source/rust_verify/src/rust_to_vir_func.rs
FeizaiYiHao: If we have:
```
        fn return_opaque_variable<T>(x:T) -> (impl DummyTrait<Output = T>, impl DummyTrait<Output = T>)
        {
            x
        }
        assume_specification<T> [ return_opaque_variable::<T> ](x:T) -> (ret: (impl DummyTrait<Output = T>, impl DummyTrait<Output = T>))
```
We need to allow this to pass `compare_external_ty`. 

Are you saying that we are not recursive on enough variants of TyKind? Actually, I did miss two cases, I think:
```
RawPtr(Ty<'tcx>, Mutability)

Ref(Region<'tcx>, Ty<'tcx>, Mutability)
```

Do you think there are more cases? Should we add `Param(ParamTy)`?

tjhance: as far as I know, there are no restrictions on what kind of positions an opaque type can appear, so it seems to me we'll need to check any TyKind that could recursively contain a Ty ([reference](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/type.TyKind.html)). This includes every kind that has a GenericsArgs field.

FeizaiYiHao:
@tjhance fixed. But I'm not so sure about my fix on `compare_external_ty()`, most types containing GenericsArgs are not supported yet with assume_specification if I understand correctly? So I left them commented. `opaque_def_to_vir` should be good tho. Let me know!

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -3134,6 +3070,65 @@ fn expr_assign_to_vir_innermost<'tcx>(
...
+            ExprKind::Field(lhs, _) => {
+                let deref_ghost = mid_ty_to_vir_ghost(
+                    bctx.ctxt.tcx,
+                    &bctx.ctxt.verus_items,
+                    None,
+                    bctx.fun_id,
+                    lhs.span,
+                    &bctx.types.expr_ty_adjusted(lhs),
+                    true,
+                    None,
```
tjhance: Should this be `self.external_opaque_type_map.as_ref()`?

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -453,29 +460,111 @@ fn compare_external_ty<'tcx>(
...
+                rustc_middle::ty::TyKind::Alias(rustc_middle::ty::AliasTyKind::Opaque, al_ty2),
+            ) => {
+                // two opaque types. We compare their trait bounds
+                let ty1_bounds = tcx.item_bounds(al_ty1.def_id).instantiate(tcx, al_ty1.args);
+                let ty2_bounds = tcx.item_bounds(al_ty2.def_id).instantiate(tcx, al_ty2.args);
+                if ty1_bounds.len() != ty2_bounds.len() {
+                    return false;
+                }
+                for (bound1, bound2) in ty1_bounds.iter().zip(ty2_bounds.iter()) {
+                    if !compare_clasue_kind(
```
tjhance: typo: clasue -> clause

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: Should this be `self.external_opaque_type_map.as_ref()`?

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -403,8 +398,11 @@ fn compare_external_ty_or_true<'tcx>(
     }
 }
 fn compare_clasue_kind<'tcx>(
```
tjhance: typo clasue -> clause

tjhance:
github doubled up all my comments for some reason ....

FeizaiYiHao:
@tjhance Pushed more fixes. Also rebased.

FeizaiYiHao:
> @tjhance Pushed more fixes. Also rebased.

@tjhance ping in case github notification didn't work

FeizaiYiHao:
OK. Rebased. Ready to be merged. @tjhance

parno:
@FeizaiYiHao It looks like some additional conflicts emerged before this could be merged in.  Would you mind resolving them, and then we can merge it in?  Sorry for the additional work.

FeizaiYiHao:
@parno No problem. Just rebased it again. Let me know.

parno:
Thanks!  Everything looks green, so I'll merge it in.

PR #2147, opened 2026-02-07
Rename exec_spec! to exec_spec_verified!, add features, and introduce exec_spec_unverified!

zero-to-nat:
### Renaming

Rename `exec_spec!` to `exec_spec_verified!`.

### Additions to `exec_spec_verified!` (previously `exec_spec!`)

- quantifiers
  - guard exprs can now have `<` and `<=` for both lower and upper bound
- Compilation of structs and enums
  - add impl of `ExecSpecEq`
  - add `#[derive(Eq, Hash, PartialEq)]` in order to ensure compatibility with `HashMap`, `HashSet`, `ExecMultiset`
- More `Seq` functions*
  - `add`, `push`, `update`, `subrange`, `empty`, `to_multiset`, `drop_first`, `drop_last`, `take`, `skip`, `last`, `first`, `is_prefix_of`, `is_suffix_of`, `contains`, `index_of`, `index_of_first`, `index_of_last`
- `Map<K, V>` compiles to `HashMap<K, V>`, with support for these functions:
  - equality, `len`, indexing**, `empty`, `dom`, `insert`, `remove`, `get`
- `Set<T>` compiles to `HashSet<T>`, with support these functions:
    - equality, `len`, `empty`, `contains`, `insert`, `remove`, `union`, `intersect`, `difference`
- `Multiset<T>` compiles to (new type) `ExecMultiset<T>`, with support for these functions:
  - equality, `len`, `count`, `empty`, `singleton`, `add`, `sub`
- More `Option` functions
  - `unwrap`
- More informative errors: supported types, expected quantifier format

### Modifications for `exec_spec_unverified!` mode only
- all function signatures are marked with `#[verifier::external_body]`
- multivariable quantifiers are supported in this mode only. The current code does not attempt to generate loop invariants or triggers for quantifiers with multiple variables
- quantifiers over `char`. My implementation uses `char::from_u32`, for which Verus does not currently have a spec

***Function support:** many of the new and existing spec-to-exec translations for supported functions on vstd types are marked external_body. Some cases are due to the use of iterators, and other cases due to a nontrivial proof (e.g. see my note for Map). The documentation has been updated to state that many translations are trusted.

****Map indexing**: indexing is only supported on `Map<K, V>` where `K` is a primitive type (e.g. `usize`). I couldn't get the conversion between ref and owned types to work correctly for more complex types.

**View and DeepView for HashMap and HashSet**: The impls for these traits were moved into vstd/view.rs to resolve issues with building without Verus. This follows analogous changes in [https://github.com/verus-lang/verus/pull/1140](https://github.com/verus-lang/verus/pull/1140) and [https://github.com/verus-lang/verus/pull/1827](https://github.com/verus-lang/verus/pull/1827).


source/vstd/view.rs
```diff
@@ -134,6 +134,38 @@ impl<T: DeepView> DeepView for alloc::vec::Vec<T> {
     }
 }
 
+#[cfg(all(feature = "alloc", feature = "std"))]
```
zero-to-nat: Hi @parno or @Chris-Hawblitzel, would you be able to double-check that these configs are correct? I copied them from the config annotations on the module in std_specs where these impls used to live: [https://github.com/verus-lang/verus/blob/main/source/vstd/std_specs/mod.rs#L17](https://github.com/verus-lang/verus/blob/main/source/vstd/std_specs/mod.rs#L17)

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -136,6 +167,20 @@ fn compile_type(typ: &Type, ctx: TypeKind) -> Result<TokenStream2, Error> {
...
+                    || type_path.path.segments[0].ident.to_string() == "HashSet"
+                    || type_path.path.segments[0].ident.to_string() == "ExecMultiset"
+                    || type_path.path.segments[0].ident.to_string() == "String"
+                    || type_path.path.segments[0].ident.to_string() == "str"
+                    || type_path.path.segments[0].ident.to_string() == "nat"
+                    || type_path.path.segments[0].ident.to_string() == "int"
+                {
+                    return Err(Error::new_spanned(
+                        &typ,
+                        "Type cannot be compiled from spec code to exec code. Hint: supported types are primitive integers (uN, usize, iN, isize), bool, char, Seq<char> (for strings), Seq, Multiset, Map, Set.",
```
zhengyao-lin: Can we replace `Seq<char>` in the error message to `SpecString`? `Seq<char>` currently gets lowered to `Vec<char>`.

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -637,13 +806,25 @@ fn compile_sig(ctx: &mut LocalCtx, item_fn: &ItemFn) -> Result<TokenStream2, Err
...
-            #decreases
+    let sig = if trusted {
+        quote_spanned! { span =>
+            #[verifier::external_body]
+            #vis fn #exec_name(
+                #(#params,)*
+            ) -> (res: #ret_type)
+                requires #requires
+                ensures res.deep_view() #ext_eq #spec_name(#(#args_deep_view),*)
+                #decreases
```
zhengyao-lin: Can we factor this part out from both branches, since they are identical?

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -940,46 +1187,124 @@ fn compile_match_arm(ctx: &LocalCtx, arm: &Arm) -> Result<TokenStream2, Error> {
...
 /// or
-///   `|x| <lower> <= x < <upper> && <body>`
-fn get_guarded_range_quant(closure: &ExprClosure) -> Result<GuardedQuantifier, Error> {
+///   `|x| <guard> && <body>`
+/// where <guard> is one of:
+///   `<lower> <= x < <upper>`
+///   `<lower> <= x <= <upper>`
+///   `<lower> < x < <upper>`
+///   `<lower> < x <= <upper>`
+fn get_guarded_range_quant_untrusted(
```
zhengyao-lin: Is it easy to merge `GuardedQuantifier` and `GuardedQuantifierUntrusted` (and the `*_quant_untrusted` and `*_quant` functions)? It seems like their only difference is that the untrusted/verified version only supports a single quantified variable.

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -1783,22 +3021,22 @@ fn compile_spec_fn(item_fn: &ItemFn) -> Result<TokenStream2, Error> {
...
+        Item::Fn(item_fn) => compile_spec_fn(&item_fn, trusted),
         Item::Struct(item_struct) => compile_struct(&item_struct),
         Item::Enum(item_enum) => compile_enum(&item_enum),
         _ => Err(Error::new_spanned(item, "unsupported item")),
     }
 }
 
 /// Parses and compiles a list of items.
-pub fn exec_spec(input: TokenStream) -> TokenStream {
+pub fn exec_spec(input: TokenStream, trusted: bool) -> TokenStream {
```
zhengyao-lin: I wonder if we should toggle trusted/verified using an attribute on the function itself (instead of having two macros), since they behave the same on structs/enums.

zero-to-nat: This is a good point. Since the trusted mode currently supports more Verus features than the untrusted mode, I would be somewhat inclined towards keeping them separate macros, in order to emphasize the differences between them.

FWIW, it appears that users can add some definitions within `exec_spec!` and use them within `exec_spec_trusted!`. I added a unit test for this scenario.

@parno, would you have any opinions on this?

parno: I think my top-level concern is that it should be clear at the source level how much trust these macros require.  For example, I don't like the idea that, say, `exec_spec(true)` would mean trusted and `exec_spec(false)` would mean untrusted, since a reader/reviewer won't immediately know what that means.  I could switching to a derive macro style, where there's a `derive(ExecSpec)` for datatypes, and a `derive(ExecSpecTrusted)` and `derive(ExecSpecUntrusted)` for functions, though even there, it might be confusing as to whether `derive(ExecSpec)` is trusted or untrusted.

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -127,6 +125,39 @@ fn compile_type(typ: &Type, ctx: TypeKind) -> Result<TokenStream2, Error> {
                         TypeKind::Owned => Ok(quote_spanned! { span => Vec<#param> }),
                         TypeKind::Ref => Ok(quote_spanned! { span => &[#param] }),
                     };
+                } else if type_path.path.segments[0].ident.to_string() == "Multiset" {
+                    // todo(nneamtu):
```
parno: Ideally we should try to resolve (or remove) todo items before merging into `main`

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -1140,7 +1519,399 @@ fn compile_guarded_quant(ctx: &LocalCtx, op: &UnOp, expr: &Expr) -> Result<Token
...
+    bounds: GuardedQuantifierBounds,
+}
+
+struct GuardedQuantifier {
+    guard_op: BinOp,
+    body: Box<Expr>,
+    guarded_vars: Vec<GuardedQuantifierVar>,
+}
+
+const UNSUPPORTED_QUANTIFIER_ERROR_MSG: &str = "Within the exec_spec! macro, quantifier expressions must match one of these forms:
```
parno: > Within the exec_spec! macro

This should be:
> Within the exec_spec_trusted! macro

right?

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -1140,7 +1519,399 @@ fn compile_guarded_quant(ctx: &LocalCtx, op: &UnOp, expr: &Expr) -> Result<Token
...
+Where <guardI> is one of:
+- <lowerI> <= xI < <upperI>
+- <lowerI> <= xI <= <upperI>
+- <lowerI> < xI < <upperI>
+- <lowerI> < xI <= <upperI>
+And <lowerI> and <upperI> may mention xJ for all J < I.";
+
+const UNSUPPORTED_QUANTIFIED_TYPE_ERROR_MSG: &str = "Unsupported quantified type.
+
+Within the exec_spec! macro, quantified variables must have one of the following Rust types: u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize, char. Note: int and nat are not allowed.";
```
parno: And possibly the same change here?

source/docs/guide/src/exec_spec.md
```diff
@@ -51,14 +63,149 @@ Currently, `exec_spec!` supports these basic features:
...
+```
+pub open spec fn my_arith(x: u64, y: u64) -> u64 
+{
+    (x + y) as u64
+}
+```
+
+### Running `exec_spec_trusted!` code
+
+When using `exec_spec_trusted!`, the macro does not generate Verus proofs that the executable code produced by the macro does not perform arithmetic overflow or violate any preconditions of the functions it invokes (i.e., `recommends` clauses are not checked). This differs from `exec_spec!`. Instead, the code generated by the macro will have a runtime error if such a precondition is violated (e.g. unwrapping a `None`) or if overflow occurs.
```
parno: The double negative in the first sentence is a bit confusing.  Can we rephrase it?

Also, this bit "Instead, the code generated by the macro" is ambiguous w.r.t. which macro it is referring to.

source/builtin_macros/src/contrib/exec_spec.rs
parno: Or perhaps generalizing them by removing the `nneamtu` so others can potentially pick them up

source/vstd/view.rs
parno: To refer to HashMap, I do think that we need `std` here.  I think `std` re-exports the contents of `alloc`, so I'm less certain we need it here.

parno (approved):
Thanks for all of the hard work you've put into this!  This is a very nice expansion of both the functionality and the usability of `exec_spec`.

I think we should also update the documentation for `exec_spec` to clarify what parts are actually verified.  Right now, I think the docs make it seem like all of the generated code is verified, but some of it depends on `external_body` functions, which isn't as strong of a guarantee.

source/builtin_macros/src/contrib/exec_spec.rs
zero-to-nat: I took a stab at reducing the repeated code here, including merging the types that you pointed out and factoring out some common helper functions. I wasn't able to find a clean way to merge them completely -- there are a lot of differences that arise from the fact that the trusted version supports `char` and multiple quantified variables.

source/builtin_macros/src/contrib/exec_spec.rs
zero-to-nat: In that case, perhaps it is cleaner to stick with the current macro approach. Would you prefer that we rename `exec_spec!` to `exec_spec_untrusted!`?

source/builtin_macros/src/contrib/exec_spec.rs
zero-to-nat: thanks for catching this!

source/docs/guide/src/exec_spec.md
zero-to-nat: Thanks for the feedback! I tried to make the wording clearer.

source/builtin_macros/src/contrib/exec_spec.rs
zhengyao-lin: Sounds good. Regarding naming, `exec_spec_unverified` might be more intuitive?

zero-to-nat:
> Thanks for all of the hard work you've put into this! This is a very nice expansion of both the functionality and the usability of `exec_spec`.
> 
> I think we should also update the documentation for `exec_spec` to clarify what parts are actually verified. Right now, I think the docs make it seem like all of the generated code is verified, but some of it depends on `external_body` functions, which isn't as strong of a guarantee.

Thank you for the comments @parno!

I have updated the guide documentation to make explicit exactly which `vstd` functions have trusted translations to exec code.

Also, upon reflection, I think that it does not make sense to add pre- and postconditions to functions compiled by `exec_spec_trusted!`. I believe the intention is that such code should not be used in verified projects, and so it should not have specs. I removed this part of the macro, but let me know if you have any further thoughts on this.

source/docs/guide/src/exec_spec.md
```diff
@@ -9,18 +9,29 @@ The `exec_spec!` macro simplifies this process: you only need
...
+## Trusted `vstd` functions within by the `exec_spec!` and `exec_spec_trusted!` macros
+
+To enable greater expressivity, the `exec_spec!` and `exec_spec_trusted!` macros support the use of
+several spec functions on types in `vstd` (such as `Seq`) within the macros. This means that exec 
+translations of supported spec functions are already provided by the macro. These translations 
+are used for compiling spec code which invokes such functions to exec code. The full list of 
+functions supported by each macro can be found below.
+
+For many `vstd` functions, however, the translation from spec code to the exec equivalent is trusted. 
+Each case is marked clearly below. We hope to reduce the number of trusted translations as Verus expands
```
parno: Perhaps "is marked with a `*` below"?

parno (approved):

source/docs/guide/src/exec_spec.md
parno: Thanks, it looks better now.

parno:
> I believe the intention is that such code should not be used in verified projects

I think there may be a spectrum of projects.  Some projects want everything, even macro-generated code to be verified.  Others don't care about verification; they just want to write specifications and test them.  I think it's possible to imagine a middle ground, where a project is willing to trust the `exec_spec_trusted` macro but still wants to verify interactions with the generated code.  I also don't think it costs us anything to add the specs, right?

source/builtin_macros/src/contrib/exec_spec.rs
parno: I guess the cleanest might be to have the pair `exec_spec_verified` and `exec_spec_unverified`.  Renaming the existing `exec_spec` macro would be a breaking change though.  Do we know if anyone other than Verdict and VerusGym are using it?

zero-to-nat: I am good with that naming. I can make a Zulip post to ask if anyone else is using it / notify them about the proposed change.

zero-to-nat:
> > I believe the intention is that such code should not be used in verified projects
> 
> I think there may be a spectrum of projects. Some projects want everything, even macro-generated code to be verified. Others don't care about verification; they just want to write specifications and test them. I think it's possible to imagine a middle ground, where a project is willing to trust the `exec_spec_trusted` macro but still wants to verify interactions with the generated code. I also don't think it costs us anything to add the specs, right?

Okay, that sounds fair. I have added the specs back for `exec_spec_unverified`.

source/builtin_macros/src/contrib/exec_spec.rs
zero-to-nat: I have finished applying the renaming to the code and documentation. I will wait until Monday to see if anyone responds to the [Zulip post](https://verus-lang.zulipchat.com/#narrow/channel/394395-general/topic/Projects.20using.20exec_spec!.20macro.3F/with/574741870) before proceeding.

parno: Great, thanks for wrapping this up.  Maybe we should give it until ~Wednesday?  That will be after the Verus weekly meeting

PR #2163, opened 2026-02-10
New approach to specifying iterators via a prophetic sequence encoding

parno:
So far, this only updates the iterator definitions for `Vec` and `Range`.  If we decide we like this style, we'll need to update the other containers, as well as `Slice` for `&Vec`.


source/rust_verify_test/tests/loops_no_spinoff.rs
```diff
@@ -1234,131 +1236,131 @@ test_verify_one_file_with_options! {
...
-                invariant
-                    b <==> (forall|i: int| 0 <= i < iter.cur ==> v[i] > 0),
-            {
-                b = b && x > 0;
-            }
-            b
-        }
-    } => Ok(())
-}
+//test_verify_one_file_with_options! {
```
parno: I wasn't sure if we still wanted a test along these lines, but I can cook one up if we do.

source/vir/src/prelude.rs
parno, line 1: I initially couldn't remember why this file changed, but it was due to 612d2092029b17 to make the encoding of the decreases clause more flexible.

parno:
It looks like the most recent test run passed all of the regression tests, but [failed on Linux](https://github.com/verus-lang/verus/actions/runs/22955958618/job/66632708315?pr=2163) due bad UTF-8 coming back from Z3 when verifying vstd.  I suspect it's a variant of [this Z3 bug](https://github.com/z3prover/z3/issues/8057).  Hopefully additional edits to this PR will randomly bump us away from that bug :-/

examples/guide/iterators.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+// ANCHOR_END: double_iter_spec
+
+fn test_basic() {
+    let v: Vec<u8> = vec![1, 2, 3, 4, 5, 6];
+    let mut w: Vec<u8> = Vec::new();
+
+    for x in iter: vec_iter(&v)
+        invariant
+            w@ == iter.seq().take(iter.index@).map_values(|v: &u8| *v),
```
Chris-Hawblitzel: In loop invariants, I would rather encourage people to use forall rather than == on sequences.  For example, prior to commit https://github.com/verus-lang/verus/commit/e484bd7e942483d579ae559e93db080dc3dd3b7b ( Add lemma_seq_two_subranges_index and group_seq_lemmas_expensive ), the following failed:

```
fn test(a: &Vec<u8>) {
    let mut b = Vec::new();
    for n in 0..a.len()
        invariant
            b.len() == n,
            b@ =~= a@.take(n as int),
    {
        b.push(a[n]);
    }
}
```

whereas the forall version succeeds:

```
fn test(a: &Vec<u8>) {
    let mut b = Vec::new();
    for n in 0..a.len()
        invariant
            b.len() == n,
            forall|i: int| 0 <= i < n ==> b[i] == a[i],
    {
        b.push(a[n]);
    }
}
```

Commit https://github.com/verus-lang/verus/commit/e484bd7e942483d579ae559e93db080dc3dd3b7b added  `lemma_seq_two_subranges_index` as a special case to help invariants like `b@ =~= a@.take(n as int)`, but it general, more complex `=~=` loop invariants would need the expensive `group_seq_lemmas_expensive`.

source/rust_verify_test/tests/btree.rs
```diff
@@ -265,43 +265,46 @@ test_verify_one_file_with_options! {
...
-
-            for k in iter: m_keys
+            for k in iter: m.keys()
                 invariant
-                    iter.keys == g_keys,
-                    g_keys == seq![3u32, 6u32],
-                    increasing_seq(g_keys),
-                    items@ == iter@,
+                    g_keys == iter.seq(),
+                    items@ == iter.seq().take(iter.index@).map_values(|v: &u32| *v),
```
Chris-Hawblitzel: In my comment above, I said I preferred forall over == on sequences in loop invariants, but I realize that there are many existing loops that already use ==, and I'm not proposing to change them all to forall.  Can we add a couple helpers to seq_lib to avoid having to write out `map_values(|v: &u32| *v)`?

```
impl<A> Seq<A> {
    pub open spec fn as_ref(&self) -> Seq<&A> {
        Seq::new(self.len(), |i: int| &self[i])
    }
}

impl<A> Seq<&A> {
    pub open spec fn unref(self) -> Seq<A> {
        Seq::new(self.len(), |i: int| *self[i])
    }
}
```

source/rust_verify_test/tests/btree.rs
```diff
@@ -399,11 +400,12 @@ test_verify_one_file_with_options! {
             m.insert(6, -8);
 
             let mut idx = 0;
-            let m_iter = m.iter();
-            for (k, v) in iter: m_iter
+            for (k, v) in iter: m.iter()
                 invariant
-                    iter.kv_pairs.to_set() =~= set![(3u32, 4i8), (6u32, -8i8)],
+                    iter.seq().map_values(|v: (&u32, &i8)| (*v.0, *v.1)).to_set() =~= set![(3u32, 4i8), (6u32, -8i8)],
```
Chris-Hawblitzel: I think the invariant `m@.kv_pairs() == set![(3u32, 4i8), (6u32, -8i8)]` is simpler.

Chris-Hawblitzel (approved):
I wrote some individual comments about map_values, but one way or another, I think map_values can be removed from all the loop invariants.

source/vir/src/prelude.rs
Chris-Hawblitzel: I went ahead and merged https://github.com/verus-lang/verus/pull/2010 .

parno: Great, thanks!

source/rust_verify_test/tests/btree.rs
parno: Yes, that's a nice idea.  How do you feel about calling the second one `deref` instead of `unref`?

parno: Also, should we add one for this?
```
impl<A, B> Seq<(&A, &B)> {
    pub open spec fn unref(self) -> Seq<(A, B)> {
        Seq::new(self.len(), |i: int| (*self[i].0, *self[i].1))
    }
}
```
It's a bit of a niche case, but it comes up for vstd's hashmap and btree_map.

Chris-Hawblitzel: > Yes, that's a nice idea. How do you feel about calling the second one `deref` instead of `unref`?

Maybe, although `deref` in `Deref` has signature `fn deref(&self) -> &Self::Target` and `as_deref` in `Option` has signature `fn as_deref(&self) -> Option<&<T as Deref>::Target>`.  The closest thing I could find was `copied` in `Option`, which has signature `impl<T> Option<&T> fn copied(self) -> Option<T>`.

parno: Huh, that's counterintuitive, but if that's the Rust convention, then inventing our own name is probably better.  Thanks for those examples.

examples/guide/iterators.rs
parno: That's a good point.  I've updated the example to use the quantified version.

source/rust_verify_test/tests/btree.rs
parno: I've applied these changes.  They do lead to specs that are easier to read/write.  Thanks for the suggestion!

source/rust_verify_test/tests/btree.rs
parno: Good point -- done!

parno:
@Chris-Hawblitzel Thanks for the initial round of comments.  I think they're all addressed now.  If you get a chance, I'd be interested in your thoughts on the various `// REVIEW` items, or any other structural feedback you might have.

Chris-Hawblitzel:
Would you mind doing a merge or rebase to bring the branch up to date with main?

parno:
> Would you mind doing a merge or rebase to bring the branch up to date with main?

Done.

source/builtin_macros/src/syntax.rs
```diff
@@ -3481,144 +3505,200 @@ impl Visitor {
...
         } else {
             None
         };
-        self.add_loop_specs(&mut stmts, None, invariant_for, None, ensure, decreases);
-        let body_exec = Expr::Verbatim(quote_spanned!(span => {
+        self.add_loop_specs(&mut stmts, inv_except_break, invariant_for, None, ensure, decreases);
+        let body_exec = Expr::Verbatim(quote_spanned_vstd!(vstd, span => {
+            #[verus::internal(spec)]
+            #[verus::internal(unwrapped_binding)]
+            let mut #x_iter_body_old; // REVIEW: We appear to need this to be mut so we can initialize it in a separate proof block
```
Chris-Hawblitzel: Could you try it without the `mut`?

source/builtin_macros/src/syntax.rs
```diff
@@ -3481,144 +3505,200 @@ impl Visitor {
...
-            match ::core::iter::Iterator::next(&mut #x_exec_iter) {
+            match #vstd::std_specs::iter::VerusForLoopWrapper::next(&mut #x_iter_name) {
                 ::core::option::Option::Some(VERUS_loop_val) => {
                     VERUS_loop_next = VERUS_loop_val;
                 }
                 ::core::option::Option::None => break,
             };
             let #pat = VERUS_loop_next;
+            #[verus::internal(spec)]
+            #[verus::internal(unwrapped_binding)] // REVIEW: We appear to need this to be mut so we can initialize it in a separate proof block
```
Chris-Hawblitzel: Could you try it without the `mut`?

source/builtin_macros/src/syntax.rs
```diff
@@ -3481,144 +3505,200 @@ impl Visitor {
...
+                // Spec-level iterator (relies on `when_used_as_spec` on into_iter)
+                #[verifier::ghost_wrapper]
+                ::vstd::prelude::ghost_exec(#[verifier::ghost_block_wrapped] Some(&#x_exec_iter)),
+            );
+            // Hold on to the initial snapshot value so that after the loop, we know it didn't change
             #[allow(non_snake_case)]
-            let VERUS_iter = #expr;
+            #[verus::internal(spec)]
+            #[verus::internal(unwrapped_binding)]
+            let mut #x_snapshot;    // REVIEW: We appear to need this to be mut so we can initialize it in a separate proof block
```
Chris-Hawblitzel: Could you try it without the `mut`?

source/builtin_macros/src/syntax.rs
parno: Huh, you're right.  They don't seem to be necessary any more.  I'm fairly confident that it produced errors before.

parno: I've removed all three `mut`s.

source/builtin_macros/src/syntax.rs
```diff
@@ -3481,144 +3505,200 @@ impl Visitor {
...
-            ))
+            let mut auto_ensures: Ensures = parse_quote_spanned_vstd!(vstd, span =>
+                ensures
+                    #[verus::internal(auto_decreases)]
+                    #vstd::std_specs::iter::IteratorSpec::completes(&#x_iter_name.snapshot.view()),
+                    #[verus::internal(auto_decreases)]
+                    #vstd::prelude::spec_eq(#x_iter_name.index.view(), #x_iter_name.seq().len()),
+                    true,
+            );
+            // REVIEW: Do we need to combine user_ensures.attrs?
```
Chris-Hawblitzel: Yes, that would be a good idea.

parno: Done.

source/rust_verify_test/tests/btree.rs
```diff
@@ -265,43 +265,46 @@ test_verify_one_file_with_options! {
...
-            let ghost g_keys = m_keys@.1;
+            let ghost m_keys = m.keys();
+            let ghost g_keys = m_keys.remaining();
             assert(increasing_seq(g_keys));
             assert(g_keys.no_duplicates());
-            assert(g_keys == seq![3u32, 6u32]) by {
-                assert forall|i, j| 0 <= i < j < g_keys.len() implies g_keys[i] < g_keys[j] by {
-                    assert(<u32 as vstd::std_specs::cmp::OrdSpec>::cmp_spec(&g_keys[i], &g_keys[j]) is Less);
+
+            // REVIEW: Needed because vstd lacks OrdSpec impl for &T
```
Chris-Hawblitzel: I just added OrdSpec for `&`

source/vstd/prelude.rs
```diff
@@ -72,6 +72,8 @@ pub use super::string::StringExecFns;
 #[cfg(all(feature = "alloc", not(verus_verify_core)))]
 pub use super::string::StringExecFnsIsAscii;
 
+// REVIEW: Should we re-export IteratorSpec in the prelude?
```
Chris-Hawblitzel: I'm leaning against this, just because it could create name conflicts and because most Verus code doesn't need to import IteratorSpec.

source/vstd/std_specs/iter.rs
```diff
@@ -0,0 +1,364 @@
...
+}
+
+
+/********************************************************************************
+ * Collect our broadcast definitions
+ ********************************************************************************/
+
+// REVIEW: Can we automatically pull these in?
+pub broadcast group group_iter_axioms {
+    rev_postcondition,
```
Chris-Hawblitzel, line 371: It sounds good to me to add it to the vstd default broadcast.  The trigger on rev_postcondition is restrictive enough that this will be filtered out for most code.

source/vstd/prelude.rs
parno: Sounds reasonable to me.

source/vstd/std_specs/iter.rs
parno: Okay, thanks, I've added it.

source/rust_verify_test/tests/btree.rs
parno: Thanks!  That plus the second fix enabled me to remove these assumptions.

tjhance:
something's wrong with the branch here, there seem to be a lot of unrelated commits

parno:
> something's wrong with the branch here, there seem to be a lot of unrelated commits

It's due to my most recent merges of `main` into the branch.  I've been experimenting with [jj](https://github.com/jj-vcs/jj), and it apparently doesn't default to a squash merge when pulling in another branch.

source/vstd/std_specs/iter.rs
```diff
@@ -0,0 +1,362 @@
...
+    spec fn peek(&self, index: int) -> Option<Self::Item>;
+
+    // Provided methods
+
+    // TODO: Once we can add when_used_as_spec to provided trait methods, this would be a simpler encoding:
+    //#[verifier::when_used_as_spec(into_rev_spec)]
+    fn rev(self) -> (r: Rev<Self>)
+        where Self: Sized,
+        requires
+            self.obeys_prophetic_iter_laws(),   // REVIEW: Should this be moved to an implication on the ensures clauses?
```
Chris-Hawblitzel: In general, for Rust std traits, we've been putting things in ensures clauses except for requirements needed to prevent panics, like arithmetic overflow checks and index bounds checks.

parno: Makes sense.  I updated it accordingly.

parno:
@Chris-Hawblitzel Is this now waiting on @tjhance, or are you still reading through it?

Chris-Hawblitzel:
> @Chris-Hawblitzel Is this now waiting on @tjhance, or are you still reading through it?

I'm finished reviewing it.

tjhance:
there's still something mucked up with the git branch here, the github diff view shows a whole bunch of unrelated commits that are already on main

tjhance:
The fundamentals look fine. However, I will need an accurate diff to do a fine-grained code review.

parno:
> The fundamentals look fine. However, I will need an accurate diff to do a fine-grained code review.

Hmm, I'm not sure what the best way is to fix the GitHub diff, but I'm open to suggestions.  FWIW, if you check the `iter-clean` branch out locally, then `git diff main...HEAD` gives an accurate diff.

source/builtin_macros/src/syntax.rs
```diff
@@ -3365,49 +3365,66 @@ impl Visitor {
...
-        //                              &::vstd::pervasive::ForLoopGhostIteratorNew::ghost_iter(
-        //                                  &::core::iter::IntoIterator::into_iter(VERUS_iter)),
-        //                              &::vstd::pervasive::ForLoopGhostIteratorNew::ghost_iter(
-        //                                  &::core::iter::IntoIterator::into_iter(e)),
-        //                          )),
-        //                      { let x =
-        //                          ::vstd::pervasive::ForLoopGhostIterator::ghost_peek_next(&y)
-        //                          .unwrap_or(vstd::pervasive::arbitrary());
-        //                        inv },
+        //                     // We track the continuitiy of the snapshot and the initial iterator-creation expression
```
tjhance: typo - continuity

source/vir/src/sst_to_air.rs
```diff
@@ -1229,6 +1229,11 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<
                 // A `proof_note` label is metadata and has no effect otherwise.
                 return exp_to_expr(ctx, e, expr_ctxt);
             }
+            UnaryOpr::AutoDecreases => {
```
tjhance: could we filter these out during ast_to_sst instead? if not, why not?

source/vir/src/ast_to_sst.rs
```diff
@@ -2492,12 +2492,35 @@ pub(crate) fn expr_to_stm_opt(
             } else {
                 invs.clone()
             };
-            let has_break = loop_body_has_break(label, body);
+            let num_breaks = loop_body_has_breaks(label, body);
+            let has_user_break = num_breaks > if is_for_loop { 1 } else { 0 };
+            let invs = if is_for_loop && has_user_break {
+                // If the user added a break statement, then we need to remove the auto-generated ensures
+                // clauses (since they typically don't apply any more)
```
tjhance, line 2401: Am I right that `auto_decreases` seems to be doing 2 different things? One, to filter out decreases-related invariants if we aren't doing decreases checking. And two, to filter out these extra ensures clauses?

should we use two different attributes for these 2 purposes? They seem unrelated.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2914,6 +2915,12 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
             } else {
                 header.decrease.clone()
             };
+            if expr_vattrs.auto_decreases && allow_no_decreases {
+                // Filter out auto_decreases invariants if we're not checking for termination
+                // (at present, we only add them to invariant_except_break)
+                Arc::make_mut(&mut header.invariant_except_break)
+                    .retain(|e| !matches!(&e.x, ExprX::UnaryOpr(UnaryOpr::AutoDecreases, _)));
```
tjhance: When you drop a VIR node, you need to call `crate::erase::mark_tree_for_erasure`

source/vir/src/sst_to_air.rs
parno: Yes, I think that would be feasible.  I think we'd still need to case on it in places like this though, since SST recycles the AST's UnaryOpr.  In that case, should we just assert that we shouldn't hit the case?

source/vir/src/ast_to_sst.rs
parno: Yes, that's true, the extra ensures are tagging along on the auto-decreases mechanism.  I can split that out into another marker.

source/rust_verify/src/rust_to_vir_expr.rs
parno: Done.

source/builtin_macros/src/syntax.rs
parno: Thanks; fixed.

tjhance (approved):

source/vir/src/sst_to_air.rs
parno: Should we defer this change until the larger change that gives SST it's on UnaryOpr?  Or should we move ahead with filtering during ast_to_sst and leaving stubs here?

source/vir/src/ast_to_sst.rs
parno: Done.

source/vir/src/sst_to_air.rs
tjhance: I don't think it matters.

source/docs/migration-iterators.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Change 5: Removing the `ForLoopGhostIterator` implementation
+
+*Relevant to anyone who implemented a custom iterator for use in verified `for` loops.*
+
+The old system required two types (one exec, one ghost) and multiple trait implementations
+to construct your own iterator type.
+
+The new system requires only one type to define the iterator type and an implementation
+of the `IteratorSpecImpl` trait.  See the new docs for details.
```
jaylorch: The first sentence of this paragraph seems ambiguous. Which of the following does it mean?

* The new system requires only two things: (1) a type to define the iterator type, and (2) an implementation of the `IteratorSpecImpl` trait.
* The new system requires only one type. That type is used to both (1) define the iterator type, and (2) provide an implementation of the `IteratorSpecImpl` trait.

source/docs/guide/src/iterators.md
```diff
@@ -0,0 +1,68 @@
...
+        // can refer to iter.index@ and iter.seq()
+{
+    // x is the current element; iter tracks ghost state
+}
+```
+You can use this name (`my_iter` above) to refer to the number of iterations
+thus far: `my_iter.index@`.  You can also use `my_iter.seq()` to refer to the
+complete sequence of items the iterator will yield, from its starting position.
+This is a *prophetic* value: Verus knows it up front even though the iterator
+hasn't run yet.  In `for` loops without a `break`, after the loop completes, we
```
jaylorch, line 26: Why doesn't this hold when there's a `break`? Can't the prophet foresee that a `break` will occur?

source/docs/guide/src/iterator-specs.md
```diff
@@ -0,0 +1,91 @@
...
+so we're going to implement one for it.
+
+### 1. The iterator struct
+
+`VecIterator` holds a reference to the underlying `Vec` plus indices `i` and `j`
+marking the current and end positions. The type invariant enforces that `i <= j <=
+v.len()` at all times.
+
+```rust
+{{#include ../../../../examples/guide/iterators.rs:iter_def}}
```
parno, line 25: Ah, sorry about that.  They were in an earlier commit.  You can see that file here:
https://github.com/verus-lang/verus/blob/iter-clean/examples/guide/iterators.rs

source/docs/guide/src/iterators.md
parno: The prophet is roughly saying, "Here are the values you will receive each time you call `next()` until you receive a `None`".  In the case of a user-supplied `break`, you're exiting the loop before you've done all of the `next` calls you could have done.

source/docs/guide/src/iterator-specs.md
```diff
@@ -0,0 +1,91 @@
...
+- **`obeys_prophetic_iter_laws`** — return `true` to assert that this iterator satisfies
+  the Verus specification for `next`.  Most iterator implementations should return `true` here,
+  and most iterator adaptors will return their inner iterator's value.
+- **`remaining`** — a prophetic spec function returning the sequence of items that the iterator will
+  eventually produce. For `VecIterator`, this is the subrange `v[i..j]`.
+- **`completes`** — return `true` if the iterator will eventually return `None`.
+  Infinite iterators or iterators driven by a non-terminating closure may return `false`.
+- **`decrease`** — a termination metric for Verus's decreases checker.  By default,
+  `for` loops expect this to return `Some(n)` where `n` decreases on every call to `next`. Here `j - i` works.
+- **`initial_value_inv`** — a prophetic invariant relating the current iterator state to
```
jaylorch: This doesn't seem consistent with the example in `examples/guide/iterators.rs::iter_spec`. There, the first conjunct in `initial_value_inv` says `IteratorSpec::remaining(init) == IteratorSpec::remaining(self)`. But if it were relating the current state to the initial state, they shouldn't have the same sequence of remaining items. There should be, in general, fewer remaining items in the current state.

source/docs/guide/src/iterator-specs.md
```diff
@@ -0,0 +1,91 @@
...
+
+If your iterator supports backward traversal, implement the standard Rust
+`DoubleEndedIterator` trait, which adds a `next_back` method:
+
+```rust
+{{#include ../../../../examples/guide/iterators.rs:double_iter_spec}}
+```
+
+For full verification support with `.rev()`, also implement `DoubleEndedIteratorSpecImpl`
+(analogous to `IteratorSpecImpl`), providing a `peek_back` function that returns the item
```
jaylorch, line 103: The example doesn't show `peek_back`.

source/docs/guide/src/iterator-specs.md
```diff
@@ -0,0 +1,91 @@
+# Iterator Specifications for a Custom Type
+
+To reason about iteration over your own custom type,
+you need to:
+1. Define the iterator struct, similar to how various Rust types have a corresponding iterator
+type; e.g., a Rust slice relies on a [`std::slice::Iter`](https://doc.rust-lang.org/std/slice/struct.Iter.html) struct.
+2. Implement the standard Rust `Iterator` trait (`next`).
+3. Implement the `IteratorSpecImpl` trait to provide Verus specification.
```
jaylorch, line 8: The examples use `IteratorSpec::` a lot, but this guide never talks about that. I suggest the guide discuss what `IteratorSpec` is and what it provides.

source/docs/migration-iterators.md
parno: Thanks, I updated this as discussed on Slack.

source/docs/guide/src/iterator-specs.md
parno: Good point; I fixed it to say:
> a prophetic invariant relating the iterator's initial state to the exec expression used to initialize it.

source/docs/guide/src/iterator-specs.md
parno: Good point; I revamped the opening paragraph of step 3 to say:

> In `vstd`, Verus provides `IteratorSpec`, an extension of the Rust [`Iterator`](https://doc.rust-lang.org/std/iter/trait.Iterator.html) trait that defines a variety of specification functions, as well as the Verus specs for the `next()` function.  To enable us to reason about our custom iterator, we need to implement the Verus-provided `IteratorSpecImpl` trait (not the `IteratorSpec` trait that defines the specs -- see ["External trait specifications"](external_trait_specifications.md) for more details).  Here's a brief summary of the specification functions, with a focus on how we define them for our custom iterator.

source/docs/guide/src/iterator-specs.md
parno: That's true.  I've added a code snippet to accompany that text as well.

PR #2167, opened 2026-02-11
[rust_verify] feat: `proof_note` labels

mkovaxx:
Include `proof_note` labels in errors about `requires`/`ensures` clauses and `assert`/`assume` statements.

Caveats for the `proof_note` attribute to attach correctly:
- A `requires`/`ensures` clause must be in parentheses (or braces).
- To label an `assert`/`assume` statement, the attribute must be placed on the expression _inside_. The expression also must be in parentheses (or braces), same as a clause.
- See `source/rust_verify_test/tests/basic.rs` for examples.

### DIFF
--- a/source/rust_verify_test/tests/basic.rs
+++ b/source/rust_verify_test/tests/basic.rs
@@ -134,7 +134,7 @@ test_verify_one_file! {
             test_requires1(a + a, b + b, c + c);
             test_requires1(a + a, b + b, a + c); // FAILS
         }
-    } => Err(err) => assert_one_fails(err)
+    } => Err(err) => assert_help_error_msg(err, "note: Test label #1")
 }
 
 test_verify_one_file! {


mkovaxx:
@parno The `proof_note` attribute on `requires` and `ensures` clauses works. A minor annoyance at the moment is that the clause must be in either `()` or `{}`, otherwise the `Attribute` doesn't always attach to the whole `Expr` but sometimes to a sub-`Expr`.

I think it's higher priority to get the `assume`/`assert` reporting done, but please let me know if that's not the case.

mkovaxx:
The `assert`/`assume` cases are working as well. Similar caveat to the `requires`/`ensures` clauses: the `proof_note` attribute must be on the argument `Expr` for now.

Next step will be to address those issues, likely in the macro layer.

source/vir/src/ast_to_sst.rs
```diff
@@ -548,6 +549,16 @@ pub(crate) fn init_var(span: &Span, x: &UniqueIdent, exp: &Exp) -> Stm {
     )
 }
 
+pub(crate) fn expr_get_proof_note(expr: &Expr) -> Option<Arc<String>> {
+    match &expr.x {
+        ExprX::UnaryOpr(UnaryOpr::Box(_), e) => expr_get_proof_note(e),
+        ExprX::UnaryOpr(UnaryOpr::Unbox(_), e) => expr_get_proof_note(e),
```
Chris-Hawblitzel: I think Box and Unbox should not exist at this stage -- they only get introduced after ast_to_sst.

source/vir/src/ast_to_sst.rs
```diff
@@ -548,6 +549,16 @@ pub(crate) fn init_var(span: &Span, x: &UniqueIdent, exp: &Exp) -> Stm {
     )
 }
 
+pub(crate) fn expr_get_proof_note(expr: &Expr) -> Option<Arc<String>> {
```
Chris-Hawblitzel: If expr_get_proof_note isn't used by ast_to_sst, I think it would be better placed in either ast_util or well_formed.

Chris-Hawblitzel (approved):

mkovaxx:
Going to land this as soon as the tests pass.

Remaining work to eliminate syntax quirks is tracked here: https://github.com/verus-lang/verus/issues/2181

PR #2173, opened 2026-02-13
Add documentation for trait extensions

parno:
We've had a number of questions about this feature, so it seems worthwhile to document it properly.  This PR puts the existing trait specification discussion plus the new documentation into a new file that appears in the Reference section, and then adds suitable pointers to it.

### DIFF
--- a/source/docs/guide/src/external_trait_specifications.md
+++ b/source/docs/guide/src/external_trait_specifications.md
@@ -116,5 +116,6 @@ to say that `obeys_eq_spec()` is true.
  * The specification trait should not have a body for any method.
  * Generic parameters and associated types must match the external trait exactly.
  * When using `external_trait_extension`, the two names (`SpecTrait` and `SpecImplTrait`)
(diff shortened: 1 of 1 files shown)


source/docs/guide/src/external_trait_specifications.md
```diff
@@ -0,0 +1,120 @@
...
+to say that `obeys_eq_spec()` is true.
+
+## Rules and restrictions
+
+ * The `ExternalTraitSpecificationFor` associated type is required and must name the
+   external trait.
+ * The specification trait should not have a body for any method.
+ * Generic parameters and associated types must match the external trait exactly.
+ * When using `external_trait_extension`, the two names (`SpecTrait` and `SpecImplTrait`)
+   become real trait names that can be used in bounds and `impl` blocks.
```
Chris-Hawblitzel: `SpecImplTrait` should not be used in bounds.

parno: Good point -- I pushed a fix to clarify.

PR #2182, opened 2026-02-14
[rust_verify] feat: JSON output has failed obligations (`proof_note`)

mkovaxx:
- Extend `MessageLabel` with an `is_proof_note: bool` field to distinguish proof notes from plain labels.
- Render proof note labels as `note: {label}` in diagnostics.
- Extend verifier to record proof note labels of failing obligations per function.
- Emit those under `func-details` in JSON output.
- Extend test harness to handle JSON output and add `tests/output_json.rs`.

### DIFF
--- a/source/air/src/messages.rs
+++ b/source/air/src/messages.rs
@@ -59,7 +59,6 @@ pub struct AirSpan {
 pub struct AirMessageLabel {
     pub span: AirSpan,
     pub note: String,
-    pub is_proof_note: bool,
 }
 
 /// Very simple implementation of Diagnostics for use in AIR
@@ -132,7 +131,6 @@ impl MessageInterface for AirMessageInterface {
         Arc::new(AirMessageLabel {
             span: AirSpan { as_string: air_span.to_owned() },
             note: note.to_owned(),
-            is_proof_note: false,
         })
     }
 
@@ -145,7 +143,7 @@ impl MessageInterface for AirMessageInterface {
                 span: None,
             })
         } else {
-            let AirMessageLabel { span, note, is_proof_note: _ } =
+            let AirMessageLabel { span, note } =
                 labels[0].downcast_ref::<AirMessageLabel>().unwrap().clone();
             Arc::new(AirMessage {
                 span: Some(span),
--- a/source/air/src/main.rs
+++ b/source/air/src/main.rs
@@ -173,9 +173,7 @@ pub fn main() {
                 let err: &AirMessage =
                     err.downcast_ref().expect("unexpected value in Any -> Message conversion");
                 println!("Error at {}", err.note);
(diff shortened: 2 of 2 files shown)


mkovaxx:
@parno Here's some progress on collecting the failed proof obligation labels per function.

I'm not sure whether some details are correct, e.g. precondition failures are assigned to the caller, whereas postcondition failures are assigned to the target (callee).

Please take a look at the tests.

parno:
> I'm not sure whether some details are correct, e.g. precondition failures are assigned to the caller, whereas postcondition failures are assigned to the target (callee).

Yes, that makes sense to me.  A function always assumes its preconditions, so a function should never "see" an error in its own preconditions (absent a recursive call).

source/rust_verify/src/verifier.rs
```diff
@@ -871,6 +900,15 @@ impl Verifier {
                         }
                     }
 
+                    // Collect `proof_note` labels related to this failure.
+                    let mut proof_notes = vec![];
+                    for label in &error.labels {
+                        if label.note.starts_with("note: ") {
```
parno: Is `label.note` the newly added field for proof notes?  If so, is the prefix check for "note: " needed?

mkovaxx: That was half-baked code to get things working fast. `MessageLabel` now has `is_proof_note: bool` to internally distinguish proof note and message note labels.

source/rust_verify_test/tests/output_json.rs
```diff
@@ -32,17 +37,21 @@ test_verify_one_file_with_options! {
...
                 (z == x + y),
+                #[verifier::proof_note("Label 451")]
+                (z == x - y),
         {
             x
         }
 
         fn caller() {
-            let _ = example(1, 2); // postcondition fails
+            let _ = example(1, 2); // postconditions fail
```
parno: I would shift this comment to be on `example`, since the caller doesn't know the postcondition fails; that's a property of the callee (and is independent of the callsite).

source/air/src/messages.rs
```diff
@@ -59,6 +59,7 @@ pub struct AirSpan {
 pub struct AirMessageLabel {
     pub span: AirSpan,
     pub note: String,
+    pub is_proof_note: bool,
```
Chris-Hawblitzel: You can omit `is_proof_note` here.  AirMessageLabel is only for air's internal use when air is run as a standalone command-line tool.  From Verus's perspective, air is an independent tool and it doesn't have to mimic Verus's features.

mkovaxx: Thanks for lagging this. I've reverted the changes to AIR.

source/rust_verify_test/tests/output_json.rs
mkovaxx: Good point. I moved the comment.

PR #2183, opened 2026-02-14
Jonh/hashmap remove returns removed value

jonhnet:
Rust `HashMap::remove()` returns the element, if found. This change adds that behavior to our `external_body` spec.


jialin-li (approved):
looks good to me!

jaylorch (approved):
Looks good to me. Thanks for correcting this oversight!

PR #2188, opened 2026-02-16
vstd: add warning supression for layout_for_val_is_valid

bsdinis:
The lack of this suppression creates a warning on client crates when building.
 

elanortang:
This makes sense to me since from an `exec` point of view, `val` is never used in the body of `layout_for_val_is_valid`. However, I don't know if I have final say on anything that's merged into `main` - I've always gotten an approval from Bryan, Chris, or Travis.

bsdinis:
Requesting @tjhance for a review, but in my view I think you'd be within your purview to accept by yourself (especially given it's such a small change -- only worth it because this will warn on client `cargo build`s). Maybe Travis can clarify

PR #2189, opened 2026-02-16
vstd: add specs for BTreeMap and BTreeSet

bsdinis:

### DIFF
--- a/source/rust_verify_test/tests/btree.rs
+++ b/source/rust_verify_test/tests/btree.rs
@@ -1,0 +1,480 @@
+#![feature(rustc_private)]
+#[macro_use]
+mod common;
+use common::*;
+
+test_verify_one_file! {
+    #[test] test_btree_map verus_code! {
+        use std::collections::BTreeMap;
+        use vstd::prelude::*;
+        fn test()
+        {
+            let mut m = BTreeMap::<u32, i8>::new();
+            assert(m@ == Map::<u32, i8>::empty());
+
+            let b = m.is_empty();
+            assert(b);
+
+            m.insert(3, 4);
+
+            let b = m.is_empty();
+            assert(!b);
+
+            m.insert(6, -8);
+            assert(m@[3] == 4);
+
+            let b = m.contains_key(&3);
+            assert(b);
+
+            let n = m.len();
+            assert(n == 2);
+
+            let v = m.get(&6);
+            match v {
+                Some(v) => assert(*v == -8),
+                None => assert(false),
+            };
+
+            m.remove(&6);
+            assert(!m@.contains_key(6));
+            assert(m@.contains_key(3));
+
+            m.clear();
+            assert(!m@.contains_key(3));
+            let b = m.contains_key(&3);
+            assert(!b);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_btree_set verus_code! {
+        use std::collections::BTreeSet;
+        use vstd::prelude::*;
+        fn test()
+        {
+            let mut m = BTreeSet::<u32>::new();
+            assert(m@ == Set::<u32>::empty());
+
+            let b = m.is_empty();
+            assert(b);
+
+            let res = m.insert(3);
+            assert(res);
+            m.insert(6);
+
+            let b = m.is_empty();
+            assert(!b);
+
+            let res = m.insert(3);
+            assert(!res);
+
+            let b = m.contains(&3);
+            assert(b);
+
+            let n = m.len();
+            assert(n == 2);
+
+            let v = m.get(&6);
+            match v {
+                Some(v) => assert(*v == 6),
+                None => assert(false),
+            };
+
+            let res = m.remove(&6);
+            assert(res);
+            let res = m.remove(&6);
+            assert(!res);
+            assert(!m@.contains(6));
+            assert(m@.contains(3));
+
+            m.clear();
+            assert(!m@.contains(3));
+            let b = m.contains(&3);
+            assert(!b);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_btree_map_struct verus_code! {
+        use std::collections::BTreeMap;
+        use vstd::prelude::*;
+
+        #[derive(PartialEq, Eq, PartialOrd, Ord)]
+        struct MyStruct
+        {
+            pub i: u16,
+            pub j: i32,
+        }
+
+        fn test()
+        {
+            assume(vstd::laws_cmp::obeys_cmp_spec::<MyStruct>());
+
+            let mut m = BTreeMap::<MyStruct, u32>::new();
+            assert(m@ == Map::<MyStruct, u32>::empty());
+            let s1 = MyStruct{ i: 3, j: 7 };
+            m.insert(s1, 4);
+
+            let s2 = MyStruct{ i: 3, j: 7 };
+            assert(m@[s2] == 4);
+            assert(m@.contains_key(s2));
+
+            let b = m.contains_key(&s2);
+            assert(b);
+
+            let v = m.get(&s2);
+            match v {
+                Some(v) => assert(*v == 4),
+                None => assert(false),
+            }
+
+            m.clear();
+            assert(!m@.contains_key(s2));
+            let b = m.contains_key(&s2);
+            assert(!b);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_btree_set_struct verus_code! {
+        use std::collections::BTreeSet;
+        use vstd::prelude::*;
+
+        #[derive(PartialEq, Eq, PartialOrd, Ord)]
+        struct MyStruct
+        {
+            pub i: u16,
+            pub j: i32,
+        }
+
+        fn test()
+        {
+            assume(vstd::laws_cmp::obeys_cmp_spec::<MyStruct>());
+
+            let mut m = BTreeSet::<MyStruct>::new();
+            assert(m@ == Set::<MyStruct>::empty());
+            let s1 = MyStruct{ i: 3, j: 7 };
+            let res = m.insert(s1);
+            assert(res);
+            let res = m.insert(MyStruct{ i: 3, j: 7 });
+            assert(!res);
+
+            let s2 = MyStruct{ i: 3, j: 7 };
+            assert(m@.contains(s2));
+
+            let b = m.contains(&s2);
+            assert(b);
+
+            let v = m.get(&s2);
+            match v {
+                Some(v) => assert(*v == s2),
+                None => assert(false),
+            }
+
+            let s3 = MyStruct { i: 9, j: 9 };
+
+            m.insert(MyStruct { i: 9, j: 9 });
+            let res = m.remove(&s3);
+            assert(res);
+            let res = m.remove(&s3);
+            assert(!res);
+
+            m.clear();
+            assert(!m@.contains(s2));
+            let b = m.contains(&s2);
+            assert(!b);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_btree_map_struct_fails verus_code! {
+        use std::collections::BTreeMap;
+        use vstd::prelude::*;
+
+        #[derive(PartialEq, Eq, PartialOrd, Ord)]
+        struct MyStruct
+        {
+            pub i: u16,
+            pub j: i32,
+        }
+
+        fn test()
+        {
+            // Missing `assume(vstd::laws_cmp::obeys_cmp_spec::<MyStruct>());`
+
+            let mut m = BTreeMap::<MyStruct, u32>::new();
+            let s1 = MyStruct{ i: 3, j: 7 };
+            m.insert(s1, 4);
+
+            let s2 = MyStruct{ i: 3, j: 7 };
+            assert(m@[s2] == 4); // FAILS
+        }
+    } => Err(err) => assert_one_fails(err)
+}
+
+test_verify_one_file! {
+    #[test] test_btree_set_struct_fails verus_code! {
+        use std::collections::BTreeSet;
+        use vstd::prelude::*;
+
+        #[derive(PartialEq, Eq, PartialOrd, Ord)]
+        struct MyStruct
+        {
+            pub i: u16,
+            pub j: i32,
+        }
+
+        fn test()
+        {
+            // Missing `assume(vstd::laws_cmp::obeys_cmp_spec::<MyStruct>());`
+
+            let mut m = BTreeSet::<MyStruct>::new();
+            let s1 = MyStruct{ i: 3, j: 7 };
+            m.insert(s1);
+
+            let s2 = MyStruct{ i: 3, j: 7 };
+            assert(m@.contains(s2)); // FAILS
+        }
+    } => Err(err) => assert_one_fails(err)
+}
+
+test_verify_one_file! {
+    #[test] test_btree_map_deep_view verus_code! {
+        use std::collections::BTreeMap;
+        use vstd::prelude::*;
+        use vstd::std_specs::btree::*;
+
+        fn test(m: BTreeMap<u64, Vec<bool>>, k: u64)
+            requires
+                m@.contains_key(k),
+                m[k]@ == seq![true],
+        {
+            broadcast use lemma_btree_map_deepview_properties;
+            assert(m.deep_view()[k] == seq![true]);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file_with_options! {
+    #[test] test_btree_map_keys ["exec_allows_no_decreases_clause"] => verus_code! {
+        use std::collections::BTreeMap;
+        use std::collections::btree_map::Keys;
+        use vstd::prelude::*;
+        use vstd::std_specs::btree::*;
+        fn test()
+        {
+            let mut m = BTreeMap::<u32, i8>::new();
+            assert(m@ == Map::<u32, i8>::empty());
+
+            m.insert(3, 4);
+            m.insert(6, -8);
+            let m_keys = m.keys();
+            assert(m_keys@.0 == 0);
+            assert(m_keys@.1.to_set() =~= set![3u32, 6u32]);
+            let ghost g_keys = m_keys@.1;
+            assert(increasing_seq(g_keys));
+            assert(g_keys.no_duplicates());
+            assert(g_keys == seq![3u32, 6u32]) by {
+                assert forall|i, j| 0 <= i < j < g_keys.len() implies g_keys[i] < g_keys[j] by {
+                    assert(<u32 as vstd::std_specs::cmp::OrdSpec>::cmp_spec(&g_keys[i], &g_keys[j]) is Less);
+                }
+                assert(g_keys.len() == 2) by {
+                    g_keys.unique_seq_to_set();
+                    assert(set![3u32, 6u32].len() == 2);
+                }
+                assert(g_keys.to_set().contains(g_keys[0]));
+                assert(g_keys.to_set().contains(g_keys[1]));
+            }
+
+
+            let mut items = Vec::<u32>::new();
+            assert(items@ =~= g_keys.take(0));
+
+            for k in iter: m_keys
+                invariant
+                    iter.keys == g_keys,
+                    g_keys == seq![3u32, 6u32],
+                    increasing_seq(g_keys),
+                    items@ == iter@,
+            {
+                assert(iter.keys.take(iter.pos).push(*k) =~= iter.keys.take(iter.pos + 1));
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel:
Rather than making `obeys_key_model` uninterp, can we use `obeys_cmp_spec` from `vstd/laws_cmp.rs`?

bsdinis:
Good point, will look into it

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1296 @@
...
+    open spec fn view(&self) -> Seq<(Key, Value)> {
+        self.kv_pairs.take(self.pos)
+    }
+}
+
+// To allow reasoning about the ghost iterator when the executable
+// function `iter()` is invoked in a `for` loop header (e.g., in
+// `for x in it: v.iter() { ... }`), we need to specify the behavior of
+// the iterator in spec mode. To do that, we add
+// `#[verifier::when_used_as_spec(spec_iter)` to the specification for
```
jaylorch: Missing ] after #[verifier::when_used_as_spec(spec_iter). (I guess this is a comment bug in `hash.rs` as well!)

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1296 @@
...
+
+#[verifier::when_used_as_spec(spec_btree_map_iter)]
+pub assume_specification<'a, Key, Value, A: Allocator + Clone>[ BTreeMap::<Key, Value, A>::iter ](
+    m: &'a BTreeMap<Key, Value, A>,
+) -> (iter: btree_map::Iter<'a, Key, Value>)
+    ensures
+        obeys_key_model::<Key>() ==> {
+            let (index, s) = iter@;
+            &&& index == 0
+            &&& s.to_set() == m@.kv_pairs()
```
jaylorch, line 420: I was expecting to see a conjunct here saying that the keys are in order. After all, that's the interesting part of the specification for a `BTreeMap` and `BTreeSet`: it iterates through the keys in order.

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1296 @@
...
+;
+
+pub assume_specification<'a, Key, Value, A: Allocator + Clone>[ BTreeMap::<Key, Value, A>::keys ](
+    m: &'a BTreeMap<Key, Value, A>,
+) -> (keys: Keys<'a, Key, Value>)
+    ensures
+        obeys_key_model::<Key>() ==> {
+            let (index, s) = keys@;
+            &&& index == 0
+            &&& s.to_set() == m@.dom()
```
jaylorch, line 809: I think there should be another conjunct here saying that the sequence is in order. After all, that's one of the important ways in which `BTreeMap` differs from `HashMap`.

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1296 @@
...
+;
+
+pub assume_specification<'a, Key, Value, A: Allocator + Clone>[ BTreeMap::<Key, Value, A>::values ](
+    m: &'a BTreeMap<Key, Value, A>,
+) -> (values: Values<'a, Key, Value>)
+    ensures
+        obeys_key_model::<Key>() ==> {
+            let (index, s) = values@;
+            &&& index == 0
+            &&& s.to_set() == m@.values()
```
jaylorch, line 822: I think there should be another conjunct here saying that the values are in order by key. The documentation at `https://doc.rust-lang.org/std/collections/struct.BTreeMap.html` says they are.

source/vstd/std_specs/btree.rs
bsdinis: Yeah, this was a very lazy copy-over from HashMap -- Will try to make the `obeys_cmp_spec` work

source/vstd/std_specs/btree.rs
jaylorch: I would suggest doing the following. Write a function:

```
pub open spec fn set_corresponds_to_increasing_seq<K: Ord>(m: Set<K>, s: Seq<K>) -> bool
{
    &&& s.to_set() == m
    &&& forall|i, j| 0 <= i < j < s.len() ==> s[i].cmp_spec(s[j]) is Less // using `Less` enforces no duplication
}
```

Then add a conjunct here:
```
exists|ks: Seq<Key>| set_corresponds_to_increasing_seq(m@.keys(), ks) && s == ks.map(|k| m@[k])
```

The function `set_corresponds_to_increasing_seq` will also be useful in the spec for `keys` and `iter`.

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1296 @@
...
+;
+
+pub assume_specification<'a, Key, A: Allocator + Clone>[ BTreeSet::<Key, A>::iter ](
+    m: &'a BTreeSet<Key, A>,
+) -> (r: btree_set::Iter<'a, Key>)
+    ensures
+        obeys_key_model::<Key>() ==> {
+            let (index, s) = r@;
+            &&& index == 0
+            &&& s.to_set() == m@
```
jaylorch, line 1158: There should be a conjunct here saying that the keys are in order. Perhaps that should be `set_corresponds_to_increasing_seq(m@, s)` using the function definition I suggested in an earlier comment.

jaylorch:
Another suggestion I have is to add tests, like the ones for `HashMap` and `HashSet`. I find that Claude Opus is great at writing such tests. :-)

bsdinis:
One quite annoying thing here is that `obeys_cmp_spec` requires `K: Ord`, while (for instance) `BTreeMap::iter` does not (the ordering is ensured because the only way values are inserted is if `K: Ord` because `insert` does require it). This means we can't change the spec in a way that allows us to use `obeys_cmp_spec`. @jaylorch do you have any idea around this? I'm also going to think more about it and try to do the correct things where I can.

jaylorch:
> One quite annoying thing here is that `obeys_cmp_spec` requires `K: Ord`, while (for instance) `BTreeMap::iter` does not (the ordering is ensured because the only way values are inserted is if `K: Ord` because `insert` does require it). This means we can't change the spec in a way that allows us to use `obeys_cmp_spec`. @jaylorch do you have any idea around this? I'm also going to think more about it and try to do the correct things where I can.

Ah, that is annoying! But I think I have a solution to this problem:

```
pub uninterp spec fn key_obeys_cmp_spec<K>() -> bool;

pub broadcast proof fn axiom_key_obeys_cmp_spec_meaning<K: Ord>()
    ensures
        #[trigger] key_obeys_cmp_spec::<K>() <==> K::obeys_cmp_spec(),
{
    admit();
}

pub uninterp spec fn set_corresponds_to_increasing_seq<K>(m: Set<K>, s: Seq<K>) -> bool;

pub broadcast proof fn axiom_set_corresponds_to_increasing_seq_meaning<K: Ord>(m: Set<K>, s: Seq<K>)
    ensures
        #[trigger] set_corresponds_to_increasing_seq(m, s) <==> {
            &&& s.to_set() == m
            &&& forall|i, j| 0 <= i < j < s.len() ==> s[i].cmp_spec(&s[j]) is Less
        }
{
    admit();
}
```

bsdinis:
Thank you so much @jaylorch , something like what you described works. I'm going to work on some tests, but it should be ok

source/rust_verify_test/tests/btree.rs
```diff
@@ -0,0 +1,436 @@
...
+        fn test()
+        {
+            let mut m = BTreeMap::<u32, i8>::new();
+            assert(m@ == Map::<u32, i8>::empty());
+
+            m.insert(3, 4);
+            m.insert(6, -8);
+            let m_keys = m.keys();
+            assert(m_keys@.0 == 0);
+            assert(m_keys@.1.to_set() =~= set![3u32, 6u32]);
```
jaylorch, line 277: This should probably be `assert(m_keys@.1 == [3u32, 6u32])` because they're in order.

source/rust_verify_test/tests/btree.rs
```diff
@@ -0,0 +1,436 @@
...
+        use vstd::std_specs::btree::*;
+        fn test()
+        {
+            let mut m = BTreeMap::<u32, i8>::new();
+            assert(m@ == Map::<u32, i8>::empty());
+            assert(m@.values() =~= Set::<i8>::empty());
+
+            m.insert(3, 4);
+            m.insert(6, -8);
+            assert(m@.values() == set![4i8, -8i8]) by {
```
jaylorch, line 330: This should probably be `assert(m@.values == [4i8, -8i8])` because they're in order by key.

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1210 @@
...
+        #[trigger] increasing_seq(s) <==> forall|i, j|
+            0 <= i < j < s.len() ==> s[i].cmp_spec(&s[j]) is Less,
+;
+
+/// Whether a sequence corresponds to the values in the set in an ordered fashion
+/// This only has meaning if `K: Ord` and [`obeys_cmp_spec::<K>`].
+///
+/// See [`axiom_set_corresponds_to_increasing_seq_meaning`] for an interpretation of this predicate.
+pub uninterp spec fn set_corresponds_to_increasing_seq<K>(set: Set<K>, seq: Seq<K>) -> bool;
+
```
jaylorch: Given, `increasing_seq`, we probably don't also need `set_corresponds_to_increasing_seq`, since we can say `set == seq.to_set()` whether `K` satisfies the `Ord` trait bound or not.

source/rust_verify_test/tests/btree.rs
bsdinis: It doesn't quite work directly, but we can say that `m_keys@.1.to_set() == set![3u32, 6u32] && increasing_seq(m_keys@.1)`

jaylorch (commented):
Thanks for addressing the issues! I have one major suggestion (in next paragraph) and some small individual code suggestions.

Major suggestion: Your first commit in this PR had, for various specs, requirements on keys like (1) any two `Key`s are identical if and only if the executable `==` operator considers them equal, and (2) `Key::clone` produces a result equal to its input. We suggested adding the additional requirement of `obeys_cmp_seq`. You seem to have *replaced* the old two requirements with this new requirement, but it should be in addition to them.

source/rust_verify_test/tests/btree.rs
jaylorch: OK. We can "leave the proof to the reader" that those two imply `m_keys@ == [3u32, 6u32]`. But it might be nice to have one case in these tests that give that proof, to serve as an example for people reading the tests to see how to use `increasing_seq`.

jaylorch:
In `test_btree_map_keys`, you can use this proof to pin down the sequence of keys:

```
use vstd::std_specs::btree::KeysAdditionalSpecFns;

assert(m_keys@.1 == seq![3u32, 6u32]) by {
    let s: Seq<u32> = m_keys@.1;
    assert forall|i, j| 0 <= i < j < s.len() implies s[i] < s[j] by {
        assert(<u32 as vstd::std_specs::cmp::OrdSpec>::cmp_spec(&s[i], &s[j]) is Less);
    }
    assert(s.len() == 2) by {
        s.unique_seq_to_set();
        assert(set![3u32, 6u32].len() == 2);
    }
    assert(s.to_set().contains(s[0]));
    assert(s.to_set().contains(s[1]));
}
```

jaylorch:
Here's the full test I was able to get to work:

```
    #[test] test_btree_map_keys ["exec_allows_no_decreases_clause"] => verus_code! {
        use std::collections::BTreeMap;
        use std::collections::btree_map::Keys;
        use vstd::prelude::*;
        use vstd::std_specs::btree::*;
        fn test()
        {
            let mut m = BTreeMap::<u32, i8>::new();
            assert(m@ == Map::<u32, i8>::empty());

            m.insert(3, 4);
            m.insert(6, -8);
            let m_keys = m.keys();
            assert(m_keys@.0 == 0);
            assert(m_keys@.1.to_set() =~= set![3u32, 6u32]);
            let ghost g_keys = m_keys@.1;
            assert(increasing_seq(g_keys));
            assert(g_keys.no_duplicates());

            assert(g_keys == seq![3u32, 6u32]) by {
                assert forall|i, j| 0 <= i < j < g_keys.len() implies g_keys[i] < g_keys[j] by {
                    assert(<u32 as vstd::std_specs::cmp::OrdSpec>::cmp_spec(&g_keys[i], &g_keys[j]) is Less);
                }
                assert(g_keys.len() == 2) by {
                    g_keys.unique_seq_to_set();
                    assert(set![3u32, 6u32].len() == 2);
                }
                assert(g_keys.to_set().contains(g_keys[0]));
                assert(g_keys.to_set().contains(g_keys[1]));
            }

            let mut items = Vec::<u32>::new();
            assert(items@ =~= g_keys.take(0));

            for k in iter: m_keys
                invariant
                    iter.keys == g_keys,
                    g_keys == seq![3u32, 6u32],
                    increasing_seq(g_keys),
                    items@ == iter@,
            {
                assert(iter.keys.take(iter.pos).push(*k) =~= iter.keys.take(iter.pos + 1));
                items.push(*k);
            }
            assert(items@ == seq![3u32, 6u32]) by {
                assert(g_keys.take(g_keys.len() as int) =~= g_keys);
            }
            assert(increasing_seq(items@));
            assert(items@.no_duplicates());
        }
    } => Ok(())
}
```

bsdinis:
Thanks for the help with these test proofs!

I still want to give the example of the values being in order. It's slightly hairier, but shouldn't be too bad.

re:
> Major suggestion: Your first commit in this PR had, for various specs, requirements on keys like (1) any two Keys are identical if and only if the executable == operator considers them equal, and (2) Key::clone produces a result equal to its input. We suggested adding the additional requirement of obeys_cmp_seq. You seem to have replaced the old two requirements with this new requirement, but it should be in addition to them.

I had interpreted Chris's comments to be a replacement when possible. And I believe it makes sense, because:
1. `obeys_cmp_spec ==> obeys_eq_spec`, which is what (1) says
2. (2) just means that it obeys the clone spec. Interestingly, we have no `obeys_clone_spec` (which arguably we should have). The only situation where this would apply would be in `BTreeMap::clone` (and BTreeSet). I could introduce this and make that work.

(This reminds me that whatever the disposition, the doc comments should be updated)

Let me know if I'm missing something

jaylorch:
@Chris-Hawblitzel , what do you think? Should the precondition just be `obeys_cmp_spec`, or should it be stronger than that?

Chris-Hawblitzel:
> @Chris-Hawblitzel , what do you think? Should the precondition just be `obeys_cmp_spec`, or should it be stronger than that?

I think it should just be `obeys_cmp_spec`.  I agree with what @bsdinis wrote here:

> I had interpreted Chris's comments to be a replacement when possible. And I believe it makes sense, because:
> obeys_cmp_spec ==> obeys_eq_spec, which is what (1) says
> (2) just means that it obeys the clone spec. Interestingly, we have no obeys_clone_spec (which arguably we should have). The only situation where this would apply would be in BTreeMap::clone (and BTreeSet). I could introduce this and make that work.

I haven't thought about what the clone spec should be here, but I agree that it's only needed on `BTreeMap::clone` and `BTreeSet::clone`.

briangmilnes:
I had to add the clone specs for my use of various collections in:
https://github.com/briangmilnes/APAS-VERUS/blob/main/src/vstdplus/clone_plus.rs
.
But I bet it can be cleaner than this in vstd.

jaylorch:
> > @Chris-Hawblitzel , what do you think? Should the precondition just be `obeys_cmp_spec`, or should it be stronger than that?
> I think it should just be `obeys_cmp_spec`. I agree with what @bsdinis wrote here:

OK, thanks. In that case I'll approve the PR.

jaylorch:
@bsdinis, do you have permission to merge, or do you need/want me to do that for you?

Chris-Hawblitzel:
I haven't finished my review.  I have more requests for changes coming.

source/vstd/std_specs/btree.rs
```diff
@@ -0,0 +1,1193 @@
...
+#[verifier::external_body]
+#[verifier::accept_recursive_types(Key)]
+#[verifier::accept_recursive_types(Value)]
+pub struct ExKeys<'a, Key, Value>(Keys<'a, Key, Value>);
+
+pub trait KeysAdditionalSpecFns<'a, Key, Value> {
+    spec fn view(self: &Self) -> (int, Seq<Key>);
+}
+
+impl<'a, Key, Value> KeysAdditionalSpecFns<'a, Key, Value> for Keys<'a, Key, Value> {
```
Chris-Hawblitzel, line 69: Can you say why this introduces a new trait `KeysAdditionalSpecFns` rather than using `View`?

bsdinis: That's a good point, I just followed the lead of the hashmap on this

bsdinis: Managed to fix it -- https://github.com/verus-lang/verus/pull/2205

Chris-Hawblitzel:
Right now, there are n * m axioms for n kinds of borrows vs. m kinds of operations.  I don't think we need to go for full generality in supporting borrows (that would probably involve something like `exists|k| k.borrows_spec() == key`), but how about generalizing the PR's approach to borrows, which works by introducing coercions in the reverse direction of the borrow (e.g. by using `Box::new`).  Here's an idea for generalizing over things like `Box::new`:

```
// probably in vstd/std_specs/borrow.rs

#[verifier::external_trait_extension(BorrowSpec via BorrowSpecImpl)]
pub trait ExBorrow<Borrowed> where Borrowed: ?Sized {
    type ExternalTraitSpecificationFor: core::borrow::Borrow<Borrowed>;

    fn borrow(&self) -> &Borrowed;

    spec fn borrow_spec(&self) -> &Borrowed;
}

pub open spec fn obeys_borrow_spec<Borrowed: ?Sized, T: core::borrow::Borrow<Borrowed> + ?Sized>() -> bool {
    forall|v: &T, ret: &Borrowed| call_ensures(T::borrow, (v,), ret) ==> ret == T::borrow_spec(v)
}

pub assume_specification<T: ?Sized>[ <T as core::borrow::Borrow<T>>::borrow ](t: &T) -> (r: &T)
    ensures
        r == t,
;

pub assume_specification<T: ?Sized, A: std::alloc::Allocator>[ <Box<T, A> as core::borrow::Borrow<T>>::borrow ](t: &Box<T, A>) -> (r: &T)
    ensures
        r == t,
;

impl<T> BorrowSpecImpl<T> for T {
    open spec fn borrow_spec(&self) -> &T {
        self
    }
}

impl<T: ?Sized, A: std::alloc::Allocator> BorrowSpecImpl<T> for Box<T, A> {
    open spec fn borrow_spec(&self) -> &T {
        &**self
    }
}

pub trait Unborrow<Borrowed: ?Sized>: core::borrow::Borrow<Borrowed> {
    spec fn unborrow(b: &Borrowed) -> &Self;

    proof fn obeys_borrow_spec()
        ensures
            obeys_borrow_spec::<Borrowed, Self>();

    proof fn borrow_unborrow(&self)
        ensures
            Self::unborrow(self.borrow_spec()) == self;

    proof fn unborrow_borrow(b: &Borrowed)
        ensures
            Self::unborrow(b).borrow_spec() == b;
}

impl<T> Unborrow<T> for T {
    open spec fn unborrow(b: &T) -> &Self {
        b
    }

    proof fn obeys_borrow_spec()
    {
    }

    proof fn borrow_unborrow(&self)
    {
    }

    proof fn unborrow_borrow(b: &T)
    {
    }
}

impl<T> Unborrow<T> for Box<T> {
    open spec fn unborrow(b: &T) -> &Self {
        &Box::new(*b)
    }

    proof fn obeys_borrow_spec()
    {
    }

    proof fn borrow_unborrow(&self)
    {
    }

    proof fn unborrow_borrow(b: &T)
    {
    }
}
```

This way, there would only need to be m axioms, each using `unborrow` rather than something more specific like `Box::new`.

jaylorch:
I like the idea of having a spec for `Borrow` in `std_specs/borrow.rs`, which we could use in `btree.rs` and `hash.rs`. @bsdinis, do you want to do that or would you like me to?

bsdinis:
I also like it!

bsdinis:
@jaylorch feel free to go for it, I could also give it a go some time tonight

jaylorch:
I've got a lot of other things on my plate right now, so if you want to go for it be my guest. :-)

Chris-Hawblitzel:
> I like the idea of having a spec for `Borrow` in `std_specs/borrow.rs`, which we could use in `btree.rs` and `hash.rs`. @bsdinis, do you want to do that or would you like me to?

And I guess the `Ord` definitions for `T: Borrow<Q> + Ord, Q: Ord` also need to be consistent between `T` and `Q`.

PR #2190, opened 2026-02-17
new-mut-ref: handle old function properly

tjhance:
* Using `old` on a parameter from the body of a function will produce a VarAt node for the pre-state (as it does when new-mut-ref is disabled)
  * In principle we could extend this to all `mut` params, not just `&mut` params, but that requires changing the type signature of `old`, which I don't want to do just yet
* Using `old` is optional in a requires clause and does nothing
* Using `old` or `fin` (to-be-renamed as `final`) is required in a postcondition

Example: suppose the user writes as they would today:

```rust
fn test(x: &mut u64)
    requires *old(x) == 0,
    ensures *x == *old(x) + 1 
{
    *x += 1;
}
```

They'll get this error:

```
error: to dereference a mutable reference parameter in a postcondition, disambiguate by wrapping it either `old` or `final`
  --> tx.rs:51:14
   |
51 |     ensures *x == *old(x) + 1
   |              ^
```

They can update the function to:

```rust
fn test(x: &mut u64)
    requires *old(x) == 0,
    ensures *fin(x) == *old(x) + 1 
{
    *x += 1;
}
```

or:

```rust
fn test(x: &mut u64)
    requires *x == 0,
    ensures *fin(x) == *old(x) + 1 
{
    *x += 1;
}
```

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -3941,6 +3941,21 @@ pub(crate) fn expr_to_loc_coerce_modes(expr: &vir::ast::Expr) -> Result<vir::ast
 }
 
 pub(crate) fn deref_mut(bctx: &BodyCtxt, span: Span, place: &Place) -> Result<Place, VirErr> {
+    // Error if `*x` is used in the postcondition where `x` is a mutable reference
+    //
+    // Ok:
+    //    *old(x)
+    //    *old(x.field)
+    //    old(*x)
+    //    *fin(x)
+    // Bad:
+    //    *x
+    //    *x.field
+    //    fin(*x)
+    //
+    // Note that in order to display an error we need both:
+    //  - dereference is not inside an `old` node
+    //  - there is no `old` node between the dereference and the variable
     if bctx.in_postcondition && !bctx.in_old {
         if let Some(local_place) = vir::ast_util::place_get_local(place) {
             let PlaceX::Local(name) = &local_place.x else { unreachable!() };
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -1141,7 +1141,11 @@ pub enum ExprX {
     /// The right side should only contain `assume(has_resolved(...))` statements
     /// emitted by the resolution analysis.
     EvalAndResolve(Expr, Expr),
-    /// The `old` node
+    /// The `old` node. (new-mut-ref only)
+    /// Note: to explicitly refer to the pre-state of a variable, the ExprX::VarAt(e, Pre)
+    /// node should be used. The 'old' node itself is used for some bookkeeping purposes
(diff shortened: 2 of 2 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -3884,7 +3937,20 @@ pub(crate) fn expr_to_loc_coerce_modes(expr: &vir::ast::Expr) -> Result<vir::ast
...
-pub(crate) fn deref_mut(bctx: &BodyCtxt, span: Span, place: &Place) -> Place {
+pub(crate) fn deref_mut(bctx: &BodyCtxt, span: Span, place: &Place) -> Result<Place, VirErr> {
+    if bctx.in_postcondition && !bctx.in_old {
+        if let Some(local_place) = vir::ast_util::place_get_local(place) {
+            let PlaceX::Local(name) = &local_place.x else { unreachable!() };
+            if bctx.is_param_for_innermost_fn_or_non_spec_closure(name) {
+                // TODO(new_mut_ref): link to documentation or something
+                return Err(vir::messages::error(
+                    &place.span,
+                    "to dereference a mutable reference parameter in a postcondition, disambiguate by wrapping it either `old` or `final`",
```
parno: "wrapping it *in* either"

source/vir/src/ast.rs
```diff
@@ -1138,6 +1138,8 @@ pub enum ExprX {
     /// The right side should only contain `assume(has_resolved(...))` statements
     /// emitted by the resolution analysis.
     EvalAndResolve(Expr, Expr),
+    /// The `old` node
```
Chris-Hawblitzel: It looks like the semantics of `old` are handled in the HIR->VIR translation rather than by processing `ExprX::Old`.  Right now it looks like the only thing that uses `ExprX::Old` is the error message when `old` is used in a spec function.  Is `ExprX::Old` intended to be used for more in the future?

tjhance: I included it mainly for these minor bookkeeping roles, yeah. Another functionality of the node which is perhaps nonobvious is that it prevents ~~[this check](https://github.com/verus-lang/verus/blob/1bfaba28d0622f88e58e07b11a669304c51cf3d7/source/rust_verify/src/rust_to_vir_expr.rs#L3933)~~ from firing when you write `*old(x)`

tjhance: or more accurately [this check](https://github.com/verus-lang/verus/blob/1bfaba28d0622f88e58e07b11a669304c51cf3d7/source/rust_verify/src/rust_to_vir_expr.rs#L3934), rather

Chris-Hawblitzel: Could we have the comment say this, pointing to ExprX::VarAt(x, VarAt::Pre) as the main place where the semantics of `old` are expressed?

tjhance: done

PR #2192, opened 2026-02-18
Suppress warnings with `#[allow(clone_without_copy)]`

jaylorch:
This lets you suppress all the annoying warnings that are emitted when you derive `Clone` but not `Copy`. You can do this with `#[allow(clone_without_copy)]`.


jaylorch:
Not needed yet

PR #2193, opened 2026-02-18
implement 'final' operator

tjhance:
Replaces 'fin'.

### DIFF
--- a/source/vstd/std_specs/option.rs
+++ b/source/vstd/std_specs/option.rs
@@ -360,9 +360,8 @@ pub assume_specification<T>[ Option::as_mut ](option: &mut Option<T>) -> (res: O
     ensures
         (match *old(option) {
             None => final(option).is_none() && res.is_none(),
-            Some(r) => final(option).is_some() && res.is_some() && *res.unwrap() === r && *final(
-                res.unwrap()
-            ) === final(option).unwrap(),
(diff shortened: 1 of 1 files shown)


tjhance:
@jaybosamiya this needs updates from the verusfmt side. `final` is a reserved word in rust so the `final(x)` syntax is rejected.

bsdinis:
small question -- have we figured out a way to prevent this from breaking syntax highlighting?

jaybosamiya:
Head's up: `final` may become a used keyword in Rust, see https://github.com/rust-lang/rfcs/pull/3678

De-restricting it as a keyword inside verusfmt would unblock the formatter, but I just wanted to make sure that there has been a discussion of the implications of using `final` here.  Is it only in expr position in a function-call-like way, and thus we're assuming it is fine to do?  If so, the better way to fix it on the verusfmt side would be to actually have it as a special case for call-like exprs that way.  Is there some situation where it might be used that I'm missing?

tjhance:
It's like a unary operator, it's only ever used as `final(expr)`

jaybosamiya-ms:
verusfmt [v0.6.3](https://github.com/verus-lang/verusfmt/releases/tag/v0.6.3) (just released) supports the `final` unary operator in `expr` position.

tjhance:
thank you!

PR #2196, opened 2026-02-19
[rust_verify] feat: JSON output has obligation proof notes

mkovaxx:
- Extend JSON `FuncDetails`with `obligation_proof_notes` and merge it alongside `failed_proof_notes` across verifier runs.
- Extend `vir::sst_util` to collect obligation proof notes from a `fn`'s own `ensures`, `assert`s in its body, and
  callee `requires` clauses.
- Extend `Verifier` to populate `obligation_proof_notes` for every function before query generation.
- Update `tests/output_json.rs` to assert exact `obligation_proof_notes` and `failed_proof_notes` sets for `requires`, `ensures`, and `assert` cases, noting the current `--no-cheating` JSON limitation.

### DIFF
--- a/source/vir/src/sst_util.rs
+++ b/source/vir/src/sst_util.rs
@@ -767,12 +767,6 @@ pub fn func_collect_obligation_proof_notes(
     if let Some(func_check) = func.x.exec_proof_check.as_ref() {
         collector.collect_func_check(func_check);
     }
-    if let Some(func_check) = func.x.recommends_check.as_ref() {
-        collector.collect_func_check(func_check);
-    }
-    if let Some(func_check) = func.x.safe_api_check.as_ref() {
-        collector.collect_func_check(func_check);
-    }
     if let Some(spec_axioms) = func.x.axioms.spec_axioms.as_ref()
         && let Some(func_check) = spec_axioms.termination_check.as_ref()
     {


source/vir/src/sst_util.rs
```diff
@@ -746,6 +748,89 @@ pub(crate) fn sst_exp_get_proof_note(exp: &Exp) -> Option<Arc<String>> {
     }
 }
 
+/// Collect proof notes from a function's requires clauses.
+pub fn func_collect_requires_proof_notes(function: &FunctionSst) -> HashSet<String> {
+    function.x.decl.reqs.iter().filter_map(sst_exp_get_proof_note).map(|s| s.to_string()).collect()
+}
+
+/// Collect proof notes from a function's proof obligations.
+pub fn func_collect_obligation_proof_notes(
```
mkovaxx, line 762: I'm not sure which passes are actually relevant here. Please check if this is correct.

mkovaxx:
@parno @Chris-Hawblitzel Please look at the tests and the function I commented on to see if this direction is correct.

source/vir/src/sst_util.rs
parno: Chris can correct me, but I think we only need the first one below, i.e., `exec_proof_check`.  The `recommends` are only used to help diagnose proof failures.  The `safe_api_check` is a check that the public portion of your API doesn't have proof obligations (as a safety precaution against unverified code calling with inputs that violate those preconditions), and the `axioms` are trusted, so they don't generate proof obligations.

Chris-Hawblitzel: We now allow assertions and calls to proof functions in proof blocks for spec fn termination checking, and these show up in `spec_axioms.termination_check`.

```
proof fn p() {
}

spec fn f() -> int
    decreases 0int
{
    proof {
        p();
        assert(false);
    }
    f()
}
```

mkovaxx: Thanks for clarifying! I've updated this `fn` to only consider `exec_proof_check` and `termination_check`.

PR #2200, opened 2026-02-20
vstd: strengthen postcondition on len0 <==> empty lemma

bsdinis:
Before, we could only prove that a `set.len() == 0 ==> set.is_empty()`. This way, we can also derive the reverse `set.len() > 0 ==> !set.is_empty()`.


Chris-Hawblitzel:
Don't we already get this from `set.rs`, without even having to use `set_lib`?

```
// Trusted axioms about len
// Note: we could add more axioms about len, but they would be incomplete.
// The following, with axiom_set_ext_equal, are enough to build libraries about len.
/// The empty set has length 0.
pub broadcast proof fn axiom_set_empty_len<A>()
    ensures
        #[trigger] Set::<A>::empty().len() == 0,
{
    fold::lemma_fold_empty(0, |b: nat, a: A| b + 1);
}
```

So this already works automatically:

```
proof fn f<A>(s: Set<A>)
    requires s.len() > 0
    ensures !s.is_empty()
{
}
```

bsdinis:
I've been working around some examples, and indeed there is always a way to make it prove automatically (although not all constructions are automatically proved).

Happy to leave as is if you think there are any downsides.

parno:
I don't see any harm to strengthening a manually invoked lemma.  One downside to this specific change is that it makes the failure mode trickier to track down if you call `lemma_len0_is_empty` on a non-empty set.  Before, you'd get a precondition failure, alerting you to the issue.  Now, you'll just fail to get something useful from the postcondition, so you'll have to debug the failed proof starting from later step of the proof.

bsdinis:
@parno that is correct (FWIW my experience with sets is that they are pretty hard to debug in any scenario)

PR #2201, opened 2026-02-20
Pass through non-Verus cfg settings

### LINKED ISSUE #2194: Passing cfgs through the verus macro
Currently, patterns like this
```
use vstd::prelude::*;

verus! {

#[cfg(target_pointer_width = "32")]
global layout usize is size == 4;

#[cfg(target_pointer_width = "64")]
global layout usize is size == 8;

}
```
are not supported in Verus, because the `target_pointer_width` cfg is not being passed through the Verus macro. (This produces an error about duplicate `global layout` declarations, even when running verus with a `--target` specified.)

It would be good to add support for passing these kinds of cfg attributes through the `verus` macro.

parno:
Fixes #2194.

Note that of the two new tests, each only runs when on the given platform.  I'm not sure if there's an easy way to run both of them regardless of the underlying platform.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -1501,7 +1501,19 @@ impl Visitor {
         for item in items.iter_mut() {
             match &item {
                 Item::Global(global) => {
-                    let Global { attrs: _, global_token: _, inner, semi: _ } = global;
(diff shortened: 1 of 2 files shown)


parno:
Merging this based on our discussion at the Verus meeting.

PR #2204, opened 2026-02-23
Wrapping specs

elanortang:
This PR 
1. Pulls the wrapping arithmetic specifications from `std_specs/num.rs` to its own file `wrapping.rs`, and defines the `num` wrapping specs in terms of these. 
2. Adds `#[cfg(not(verus_verify_core))]` to the arithmetic assume_specifications in `num`, as well as to the `overflow` module in `arithmetic` (since I don't believe the Rust stdlib verification effort will need to use the Checked structs found there). 
3. Updates `atomics` to use the `wrapping_add` and `wrapping_sub` specs defined in `wrapping.rs`. 

### DIFF
--- a/source/vstd/atomic.rs
+++ b/source/vstd/atomic.rs
@@ -11,25 +11,26 @@ use core::sync::atomic::{AtomicI64, AtomicU64};
 use super::modes::*;
 use super::pervasive::*;
 use super::prelude::*;
+use super::wrapping::*;
 
 macro_rules! make_unsigned_integer_atomic {
-    ($at_ident:ident, $p_ident:ident, $p_data_ident:ident, $rust_ty: ty, $value_ty: ty) => {
+    ($at_ident:ident, $p_ident:ident, $p_data_ident:ident, $rust_ty: ty, $value_ty: ty, $modname:ident) => {
         atomic_types!($at_ident, $p_ident, $p_data_ident, $rust_ty, $value_ty);
         #[cfg_attr(verus_keep_ghost, verus::internal(verus_macro))]
         impl $at_ident {
             atomic_common_methods!($at_ident, $p_ident, $p_data_ident, $rust_ty, $value_ty, []);
-            atomic_integer_methods!($at_ident, $p_ident, $rust_ty, $value_ty);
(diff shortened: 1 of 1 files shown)


source/vstd/std_specs/num.rs
```diff
@@ -76,42 +77,27 @@ macro_rules! num_specs {
...
 
             #[verifier::allow_in_spec]
+            #[cfg(not(verus_verify_core))]
             pub assume_specification[<$uN>::wrapping_mul](x: $uN, y: $uN) -> $uN
-                returns ((x as nat * y as nat) % $range as nat) as $uN;
+                returns $mod_u::wrapping_mul(x, y);
 
             #[verifier::allow_in_spec]
+            #[cfg(not(verus_verify_core))]
             pub assume_specification[<$uN>::checked_add](x: $uN, y: $uN) -> Option<$uN>
```
parno, line 101: Should `wrapping.rs` define all of these checked specs too?  Or do you think that only `num.rs` will care about this version of wrapping definitions?

elanortang: Reasons to not define checked specs: 
- We can think of `wrapping.rs` as axiomatizing the wrap-around arithmetic, as a complement to the way that spec arithmetic is unbounded. If you have wrap-around and unbounded arithmetic as the two primitives, you should be able to define everything else in terms of that. 
- At the moment it seems like `arithmetic/overflow.rs` is the only place that uses checked arithmetic, and there you want the exec fns rather than the specs. 

Reasons to define checked specs:
- If we anticipate other `vstd` modules needing them. 
- Avoiding duplication between `num` assume_specifications and the verified specifications in Rustlib. 

Since these specs are not very complicated, I'm not sure whether it would be more clear to write them out explicitly, e.g.
```
        pub const fn checked_add(self, rhs: Self) -> Option<Self>
            returns
                (if self + rhs <= Self::MAX {
                    Some((self + rhs) as Self)
                } else {
                    None
                }),
```
or to say `returns spec_checked_add(self, rhs)` and go look up the definition of `spec_checked_add`.

parno: I don't feel strongly about it, so if there aren't currently other `vstd` modules that need them, or modules that we expect might need them, then we can probably leave things as is here.

PR #2205, opened 2026-02-24
vstd: impl View for HashMap/BTreeMap iterators

bsdinis:
Before, we were using one-off traits to extend the type. But View works just as well


jaylorch (approved):
Great, thanks!

PR #2208, opened 2026-02-24
Avoid spurious warnings generated by the combination of a `prophetic` attribute and trait extension.

### LINKED ISSUE #2080: Trait extension plus Verus attributes produces spurious warnings
In this code, we want to add a prophetic spec function to an existing trait `T`.
```
use vstd::prelude::*;

verus! {

#[verifier::external]
trait T {
}

#[verifier::external_trait_extension(TSpec via TSpecImpl)]
trait Ex {
    
    #[verifier::prophetic]
    spec fn good_outcome(&self) -> bool;
}

} // verus!
```
However, including the `#[verifier::prophetic]` attribute causes Verus to output the identical warning twice:
```
warning: verus-related attribute has no effect because item is already marked external
  --> main.rs:23:10
   |
23 |     spec fn good_outcome(&self) -> bool;
   |          ^^
```
Looking at the macro-expanded version of the code, the offending code is:
```
#[verifier::external]
#[allow(non_camel_case_types)]
#[verus::internal(verus_macro)]
trait TSpecImpl: T {
    #[verifier::prophetic]
    #[verus::internal(verus_macro)]
    fn good_outcome(&self)
    -> bool;
    #[verifier::prophetic]
    #[doc(hidden)]
    #[verus::internal(verus_macro)]
    fn VERUS_SPEC__good_outcome(&self) -> bool {
        ::vstd::prelude::no_method_body()
    }
}
```
which is why we see two warnings (one from `good_outcome` and one from `VERUS_SPEC__good_outcome`).

I'm not sure what the right direction for a fix is.  Should we just strip off the `#[verifier::prophetic]` attribute when creating `TSpecImpl`?

parno:
Fixes #2080.

### DIFF
--- a/source/builtin_macros/src/syntax_trait.rs
+++ b/source/builtin_macros/src/syntax_trait.rs
@@ -126,7 +126,7 @@ fn expand_extension_trait<'tcx>(
                 f_tspec_impl.default = None;
                 f_tspec_impl.sig.mode = FnMode::Default;
                 // Remove #[verifier::prophetic] from the TSpecImpl methods.
-                // The TSpecImpl trait is marked #[verifier::external], so any Verus-specific 
+                // The TSpecImpl trait is marked #[verifier::external], so any Verus-specific
                 // attributes cause a spurious warning about having no effect.
                 f_tspec_impl.attrs.retain(|attr| {
                     !(attr.path().segments.len() == 2


parno:
Merging based on discussion in today's Verus meeting.

PR #2215, opened 2026-02-27
borrow check for prophetic uses

tjhance:
The error looks like this:

```rust
fn test() {
    let mut a = 0;
    let a_ref = &mut a;
    assert(a == 0);
    *a_ref = 10;
}
```

```rust
error[E0502]: cannot borrow `(Verus spec a)` as immutable because it is also borrowed as mutable
 --> example.rs:8:12
  |
7 |     let a_ref = &mut a;
  |                 ------ mutable borrow occurs here
8 |     assert(a == 0);
  |            ^ immutable borrow occurs here
9 |     *a_ref = 10;
  |     ----------- mutable borrow later used here
```

The error message isn't great — the ideal message would be something like "cannot take spec-snapshot of `a` because it is borrowed as mutable" or something. It also kind of sucks that these error messages show up redundantly even if there are already "real" borrow errors. Maybe we can figure out some kind of post-processing to improve the messages somehow. 

There are some minor bugs which are noted in the tests that I'll fix later. For example: right now the way loops are encoded, the invariants don't appear in the right place in the source-encoding that corresponds to their actual evaluation point in the VIR semantics. 

### Implementation

See the documentation in the new file for details.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -1971,6 +1971,8 @@ pub fn array_index<T, const N: usize>(_a: [T; N], _i: int) -> T {
     unimplemented!()
 }
 
+/// Needed for the THIR-based erasure (rustc_mir_build_additional_files)
+
 #[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::verus_builtin::erased_ghost_value"]
 pub fn erased_ghost_value<S, T>(_: S) -> T {
@@ -1995,6 +1997,14 @@ pub fn dummy_capture_new<'a>() -> DummyCapture<'a> {
 pub fn dummy_capture_consume<'a>(_dc: DummyCapture<'a>) {
     unimplemented!()
 }
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::mutable_reference_tie"]
+pub fn mutable_reference_tie<'a, T: ?Sized, U: ?Sized>(_a: &'a mut T, _b: &'a mut U) -> &'a mut T {
+    unimplemented!()
+}
+
+/// Directives and spec functions related to &mut references
 
 #[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::verus_builtin::has_resolved"]
--- a/source/rust_verify/src/config.rs
+++ b/source/rust_verify/src/config.rs
@@ -83,6 +83,7 @@ pub struct ArgsX {
     pub no_external_by_default: bool,
     pub no_verify: bool,
     pub no_lifetime: bool,
+    pub no_erasure_check: bool,
     pub no_auto_recommends_check: bool,
     pub no_cheating: bool,
     pub time: bool,
@@ -130,6 +131,7 @@ impl ArgsX {
             no_external_by_default: Default::default(),
             no_verify: Default::default(),
             no_lifetime: Default::default(),
+            no_erasure_check: Default::default(),
             no_auto_recommends_check: Default::default(),
             no_cheating: Default::default(),
             time: Default::default(),
@@ -194,9 +196,13 @@ pub fn enable_default_features_and_verus_attr(
 ) {
     if syntax_macro {
         // REVIEW: syntax macro adds superfluous parentheses and braces
-        for allow in
-            &["unused_parens", "unused_braces", "unconditional_panic", "arithmetic_overflow"]
-        {
+        for allow in &[
+            "unused_parens",
+            "unused_braces",
+            "unconditional_panic",
+            "arithmetic_overflow",
+            "irrefutable_let_patterns",
+        ] {
             rustc_args.push("-A".to_string());
             rustc_args.push(allow.to_string());
         }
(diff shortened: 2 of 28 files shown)


tjhance:
Here's the documentation in markdown, for your convenience:

---

The main point of transformation in this file is to prevent prophetic "paradoxes" that arise
from taking a spec snapshot of a variable while it is mutably borrowed.

```rust
let mut x = 0;
let x_ref = &mut x;

let ghost x_snapshot = x;

*x_ref = 20;
```

Taking such a snapshot would be prophetic, but Verus mode-checking (modes.rs) assumes by default
that uses of a local variable are non-prophetic. Thus, for soundness reasons, we need to disallow
code like the above, unless `x` is specifically used in a way that indicates it is prophetic.

Besides soundness issues, it's also a usability issue. It's perfectly sound to use a prophetic
variable in an assert (in general, it's fine for safety conditions), for example,

```rust
let mut x = 0;
let x_ref = &mut x;

assert(x == 0); // this check would fail; that might be confusing

*x_ref = 20;
```

But the behavior would be pretty confusing, so this is also disallowed (again, unless it's
used in a way that indicates propheticness is intentional).

To implement this restriction, we essentially need to "partially borrow check" all spec code.
Note that we still to opt-out of other borrowck restrictions (e.g. we want to be able to
take spec snapshots of moved variables).
Specifically, we want to check that certain spec uses do not reference data which is currently
_mutably_ borrowed.

# The transformation

At a high-level, the key idea is that for any (non-ghost) variable `x`, we introduce a secondary
variable, "the shadow variable", `x_shadow` which represents the ability to take a non-prophetic
spec snapshot.  The variable `x_shadow` will always be initialized, and never moved-from,
only borrowed-from.

When `x` is declared, we insert a declaration of `x_shadow`:

```rust
let x: T = ...;                               // Existing line
let x_shadow = arbitrary_ghost_value::<T>();  // Added by our transformation
```

Next, whenever `x` appears in spec code, if it requires this special checking, we replace
the usage with a borrow of x_shadow: `&x_shadow`.
For the sake of this file, we can assume that it's already determined _which_ spec variables
need this extra checking, specifically, any variable usage whose erasure mode is set to
`VarErasure::Shadow` needs this special checking.

Finally, whenever we take a mutable reference from `x`, we also "tie" the lifetime together:
`&mut x` becomes `mutable_reference_tie(&mut x, &mut x_shadow)` where
`mutable_reference_tie<'a, T, U>(&'a mut T, &'a mut U) -> &'a mut T`.
This has the effect of forcing `x_shadow` to be mutably-borrowed as long as `x` is borrowed,
but won't have any other effect on lifetime checking.

More generally,
`&mut place` becomes `mutable_reference_tie(&mut place, &mut place_shadow)` where
`place` is any place expression (like `x.field`) and `place_shadow` is the corresponding
place in the shadow variable (like `x_shadow.field`).

# Details

There are a lot of different cases to consider, accounting for all the ways mutable references
can actually be declared.

### Two-phase borrows

For two-phase borrows, we can't use the `mutable_references_tie` solution like we would
for normal mutable borrows. Imagine we expanded:

```
f(&mut[two_phase] x, y)
```

to:

```
f(mutable_reference_tie(&mut[two_phase] x, &mut x_shadow), y)
```

The problem is that the scope of the two-phase borrow only extends through the
artificial `mutable_reference_tie` call, not through the entire `f` call.

We do this instead:

```
fake_call(
    f(&mut[two_phase] x, y),
    &mut x_shadow,
    ...
)
```

where the `fake_call` wires up any lifetime variables appropriately to the output of `f`.
Note that the mutable borrow of `x_shadow` doesn't actually occur until `f` returns, but
this is fine because the two-phase borrow doesn't properly start until the end of the args
to `f`.

### Patterns

It's possible to create mutable references via patterns, e.g.:

```rust
let (ref mut x, y) = place;
```

Again, we need to do a mutable borrow from the corresponding shadow place. We translate
the above like this:

```rust
let (ref mut x_half1, y) = place;
let (ref mut x_half2, _) = shadow_place;
// For each binder with a `ref mut` binding in the pattern:
let x = mutable_reference_tie(x_half1, x_half2);
// Binders for shadow vars of the newly declared vars:
let x_shadow = arbitrary_ghost_value();
let y_shadow = arbitrary_ghost_value();
```

We call these the "half patterns".

And of course we need to apply this principle to anywhere a pattern can appear:
match, let, let-else, let expression (like in an if-let).

### Assignments

When assigning to `x`, we also need to assign to `x_shadow`:

```rust
x = ...;                                  // Existing line
x_shadow = arbitrary_ghost_value::<T>();  // Added by our transformation
```

This is needed to make Rust always considers `x` and `x_shadow` to have the exact same
set of outstanding borrows.
This is relevant in a situation that reborrows from a mutable
reference and then overwrites that reference:

```rust
let mut x_ref = &mut a;
let mut reborrow = &mut *x_ref;
x_ref = &mut b;
let mut reborrow2 = &mut *x_ref;

*reborrow = 0;   // extends lifetime from `&mut a`
*reborrow2 = 0;  // extends lifetime from `&mut b`
```

Without the assignment, we would have `*x_ref` borrowed twice at the same time;
but _with_ the assignment, the two mutable references `&mut *x_ref` don't intersect.
Thus, the assignment line is load-bearing for Rust to accept this code, so it needs
to be replicated for the shadow var.

### Closures

Inside a closure, we don't emit shadow vars for captured vars. Instead, those are
emitted by the enclosing fn/closure at the point the closure is created.
This is handled with the rest of closure-handling, in verus.rs.

PR #2216, opened 2026-02-27
Set build

Chris-Hawblitzel:
Many months ago, I promised to add `set_build!` as an independent macro from https://github.com/verus-lang/verus/pull/1766 as part of making it easier to migrate to https://github.com/verus-lang/verus/pull/1766 .  This macro provides a syntax for building various kinds of sets that are finite by construction and therefore don't need a separate proof of finiteness.  It also provides a friendlier interface to `map_by` and `map_flatten_by` for sets, which are usually easier to work with in proofs than `map` and `flatten`.

## Documentation:

This macro takes an expresion of the form:
`set_build!{ elem_expr: optional_typ | x1: typ1 in ..., ..., xn: typn in ..., cond1, ..., condm }`
or just:
`set_build!{ x: typ in ..., cond1, ..., condm }`
where each `xk: typk in ...` has one of the following forms:
- `xk: typk` for finite types (implementing FiniteFull)
- `xk: typk in expr`, where expr has type Set<typk>
- `xk: typk in lo..hi`, where lo and hi have type typk, where typk implements FiniteRange
- `xk: typk in lo..=hi`, where lo and hi have type typk, where typk implements FiniteRange
and each condk is a boolean expression.

From this, the setbuild macro uses map_by, map_flatten_by, filter, etc. to build a set of
elements specified by elem_expr.
Important restriction: by default, the elem_expr must be a variable, tuple, or datatype such
that all of the variables x1, ..., xn can be easily found with nothing more that tuple/datatype
field accesses.  In exchange for this restriction, set_build guarantees not to introduce
any extra existential quantifiers into to constructed set.  This makes it easy for proofs
to use sets constructed with set_build, when compared to other forms of Set construction
(like Set::map or Set::flatten) that do introduce existential quantifiers.
To override this default and remove this restriction, you can mark one or more variables as
`exists x: typ` rather than just `x: typ`, and set_build will use map/flatten for these
variables. This will, however, make proofs about the constructed set more difficult.

Example:
`set_build!{ (x, y, x - y): (int, int, int) | x: int in 10..20, y: int in x..20, x + y != 25 }`
From this, set_build generates:
```
Set::<int>::range(10, 20)
    .map_flatten_by(
        |__VERUS_x: int| {
            let x = __VERUS_x;
            Set::<int>::range(x, 20)
                .filter(|y: int| (x + y != 25))
                .map_by(
                    |y: int| ((x, y, x - y)),
                    |__VERUS_x: (int, int, int)| (__VERUS_x.1),
                )
        },
        |__VERUS_x: (int, int, int)| __VERUS_x.0,
    )
```

(Note: set_build_debug is like set_build, but also prints the generated builder to stderr)

### DIFF
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -2678,11 +2678,11 @@ dependencies = [
 
 [[package]]
 name = "verus_builtin"
-version = "0.0.0-2026-02-24-1505"
+version = "0.0.0-2026-03-01-0109"
 
 [[package]]
 name = "verus_builtin_macros"
-version = "0.0.0-2026-02-24-1505"
+version = "0.0.0-2026-03-01-0109"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2702,7 +2702,7 @@ dependencies = [
 
 [[package]]
 name = "verus_state_machines_macros"
-version = "0.0.0-2026-02-24-1505"
+version = "0.0.0-2026-03-01-0109"
 dependencies = [
  "indexmap 1.9.3",
  "proc-macro2",
--- a/source/builtin/Cargo.toml
+++ b/source/builtin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "verus_builtin"
-version = "0.0.0-2026-02-24-1505"
+version = "0.0.0-2026-03-01-0109"
 edition = "2018"
 license = "MIT"
 description = "Verus Builtins: Rust definitions for core constructs in Verus"
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -1971,6 +1971,8 @@ pub fn array_index<T, const N: usize>(_a: [T; N], _i: int) -> T {
     unimplemented!()
 }
 
+/// Needed for the THIR-based erasure (rustc_mir_build_additional_files)
+
 #[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::verus_builtin::erased_ghost_value"]
 pub fn erased_ghost_value<S, T>(_: S) -> T {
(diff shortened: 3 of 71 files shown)


source/vstd/set_lib.rs
```diff
@@ -30,6 +30,32 @@ impl<A> Set<A> {
...
+    /// Set::map_by is like Set::map, but map only takes a forward function fwd: spec_fn(A) -> B,
+    /// while map_by also takes a reverse function rev: spec_fn(B) -> A.
+    /// This reverse function can make proofs easier
+    /// by avoiding the "exists" that appears in lemmas about Set::map.
+    /// Example: for a set s: Set<int>, to map each i in s to (i, 10 * i),
+    /// we can write either s.map(|i: int| (i, 10 * i))
+    /// or s.map_by(|i: int| (i, 10 * i), |p: (int, int)| p.0);
+    /// the version with map_by is usually easier to use in proofs.
+    /// Also see the `set_build!` macro for a convenient interface to map_by.
+    pub open spec fn map_by<B>(self, fwd: spec_fn(A) -> B, rev: spec_fn(B) -> A) -> Set<B> {
```
parno: Would it be useful to add this
```
recommends forall |b| b == fwd(rev(b)))
```
to help catch places where someone accidentally provides a pair of functions that lack this property?

jonhnet (approved):
High order bit: hooray! Thank you.

Low order feedback: The interface is very powerful; so powerful that I had a difficult time guessing what set gets constructed by the example definition. May I suggest:
* add an example of something that belongs to the set: `assert(s.contains((3, 6, 9)))` and maybe `assert(!s.contains(10, 15, 25))` (because x+y!=25).
* adding a couple of simpler examples that, rather than showing the full power, show ways you might see set_build applied using only parts of its glory:
  -  set_build!(Pointer(page, offset): Pointer<Page, nat>, valid_page(page), offset < PAGE_SIZE) `assert(s.contains(Pointer(Page::new(0), 13)))`
  - set_build!(x + y, smallprimes.contains(x), smallprimes.contains(y)) `assert(s.contains(12))` because 5 and 7 are in smallprimes.

source/vstd/set_lib.rs
```diff
@@ -30,6 +30,32 @@ impl<A> Set<A> {
...
 
+    /// Set::map_by is like Set::map, but map only takes a forward function fwd: spec_fn(A) -> B,
+    /// while map_by also takes a reverse function rev: spec_fn(B) -> A.
+    /// This reverse function can make proofs easier
+    /// by avoiding the "exists" that appears in lemmas about Set::map.
+    /// Example: for a set s: Set<int>, to map each i in s to (i, 10 * i),
+    /// we can write either s.map(|i: int| (i, 10 * i))
+    /// or s.map_by(|i: int| (i, 10 * i), |p: (int, int)| p.0);
+    /// the version with map_by is usually easier to use in proofs.
+    /// Also see the `set_build!` macro for a convenient interface to map_by.
```
tjhance: can you use markdown in the docstring (e.g. backticks around the code) so it's more suitable to appear in the verusdoc?

PR #2225, opened 2026-03-05
Add bit lemmas for `usize` type

luojia65:
Those lemmas are useful in proving `usize` typed platform specific CPU control and state registers, like x86 CRs, MSRs, Arm system registers and RISC-V CSRs. Software adapters for those registers usually take `usize` as register types (for example: [riscv::register::mepc::read()](https://docs.rs/riscv/latest/riscv/register/mepc/fn.read.html))

By extending Verus `vstd` bit lemmas to `usize` we can now formally model those platform specific registers in Verus.

r? @jaylorch


jaylorch (approved):
Looks good to me. Thanks!

PR #2226, opened 2026-03-05
feat: streamline `proof_note` syntax on `requires`/`ensures` clauses

mkovaxx:
This is part 1 of eliminating the `proof_note` syntax quirks:
- Proof notes on `requires`/`ensures` clauses are now inner attributes, i.e. `#![proof_note("label")]`.
- Extend `verus_syn` to parse inner attributes on each clause in a `Specification`.
- Special-case inner attributes `#![trigger ...]`, `#![all_triggers]`, and `#![auto]` on the first clause of an `ensures` group, so that they apply to the entire group.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -128,6 +128,15 @@ fn path_is_ident(path: &Path, s: &str) -> bool {
     segments.len() == 1 && segments.first().unwrap().ident == s
 }
 
+fn path_matches_idents(path: &Path, expected: &[&str]) -> bool {
+    let segments = &path.segments;
+    segments.len() == expected.len()
+        && segments
+            .iter()
+            .zip(expected.iter())
+            .all(|(segment, expected)| segment.ident == *expected)
+}
+
 pub(crate) fn into_spans(span: Span) -> proc_macro2::extra::DelimSpan {
     let mut group = proc_macro2::Group::new(proc_macro2::Delimiter::None, TokenStream::new());
     group.set_span(span);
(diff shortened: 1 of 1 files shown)


mkovaxx:
@Chris-Hawblitzel This implements your idea of making `proof_note` an inner attribute.

parno:
> Special-case inner attributes #![trigger ...], #![all_triggers], and #![auto] on the first clause of an ensures group, so that they apply to the entire group.

I'm not sure I understand this part.  Those attributes should never appear at the top level of a requires/ensures; they're only allowed inside a quantifier (e.g., as `forall |x| ![trigger ...] P(x)`).  Given that, do we still need special casing?

mkovaxx:
@parno The inner attributes `#![trigger]` and `#![all_triggers]` sometimes appear in `ensures` groups.

Here are two examples:

https://github.com/verus-lang/verus/blob/e97969ef173a7165673ffbb48738c4a1b975daeb/source/vstd/array.rs#L49-L57 (There are many more cases in `vstd` similar to this one above.)

https://github.com/verus-lang/verus/blob/e97969ef173a7165673ffbb48738c4a1b975daeb/source/rust_verify_test/tests/triggers.rs#L512-L520

I couldn't find any examples of `#![auto]` in an `ensures` group, but included it in the special-casing for symmetry.

parno:
> @parno The inner attributes `#![trigger]` and `#![all_triggers]` sometimes appear in `ensures` groups.

Oh, I see.  Thanks for including those examples.  In a `broadcast` proof function, we can indeed potentially see those attributes outside of a quantifier.  I didn't think about that case.

dependencies/syn/src/verus.rs
```diff
@@ -891,13 +933,14 @@ pub mod parsing {
     impl Ensures {
         /// Parse an `ensures` clause group in a given context.
         pub fn parse_in(ctx: Context, input: ParseStream) -> Result<Self> {
-            let mut attrs = Vec::new();
             let token = input.parse()?;
-            attr::parsing::parse_inner(input, &mut attrs)?;
+            let mut exprs = Specification::parse_in(ctx, input)?;
+            // Hoist attributes that semantically belong to the entire group.
+            let top_level_attrs = exprs.remove_top_level_attrs();
```
mkovaxx, line 939: @parno @Chris-Hawblitzel To clarify: not hoisting inner trigger-related attributes (i.e. setting `attrs: Vec::new()`), causes widespread failures, among other things, preventing `vstd` from compiling.

PR #2229, opened 2026-03-06
Support IEEE floating point SMT theory

Chris-Hawblitzel:
This adds support for the SMT solver's IEEE floating point theory ( https://smt-lib.org/theories-FloatingPoint.shtml ).  This completes the last major item from the plan for floating point support in https://github.com/verus-lang/verus/discussions/1853 , building on https://github.com/verus-lang/verus/pull/1859 .  As with bit vectors, the IEEE operations are uninterpreted by default but are interpreted in assert-by-bit_vector, so that assert-by-bit_vector can be used for floating point reasoning.  Specifications can refer to the IEEE operators by name (`ieee_add`, `ieee_le`, etc.), or by using overloaded operators (`+`, `<=`, etc.).

In general, we do not assume that Rust floating point semantics match IEEE floating point semantics (see https://github.com/rust-lang/rfcs/blob/master/text/3514-float-semantics.md for details).  However, if users are working on a platform where they are willing to make this assumption, they can use the module `assume_ieee_float`, which contains broadcast axioms that express this assumption.  Here is a small example that uses `assume_ieee_float` in combination with assert-by-bit_vector to prove a simple property of a Rust f32 operation:

```
fn test(x: f32, y: f32) -> (z: f32)
    requires
        1.0f32 <= x <= 2.0f32,
        4.0f32 <= y <= 8.0f32,
    ensures
        5.0f32 <= z <= 10f32,
{
    broadcast use vstd::contrib::assume_ieee_float::assume_ieee_float;

    let z = x + y;
    assert(5.0f32 <= x + y <= 10.0f32) by(bit_vector)
        requires
            1.0f32 <= x <= 2.0f32,
            4.0f32 <= y <= 8.0f32;
    z
}
```


source/air/src/parser.rs
```diff
@@ -234,10 +266,82 @@ impl Parser {
...
+                        let exp_sig_bits = match (&nodes[2], &nodes[3]) {
+                            (Node::Atom(s2), Node::Atom(s3)) => {
+                                match (s2.parse::<u32>(), s3.parse::<u32>()) {
+                                    (Ok(exp_bits), Ok(sig_bits)) => Some((exp_bits, sig_bits)),
+                                    _ => None,
+                                }
+                            }
+                            _ => None,
+                        };
+                        if let Some((exp_bits, sig_bits)) = exp_sig_bits {
```
jaylorch, line 315: It seems overly verbose to do a match that sets `exp_sig_bits`, then match on `exp_sig_bits` to extract the values. It seems easier to combine the previous match with the subsequent action into a single match.

source/vir/src/bitvector_to_air.rs
```diff
@@ -865,17 +977,21 @@ fn bv_typ_for_vir_typ(state: &mut State, span: &Span, typ: &Typ) -> Result<BvTyp
...
+            TypX::Float(32) => Ok(f32_typ()),
+            TypX::Float(64) => Ok(f64_typ()),
+            TypX::Real => Ok(BvTyp::Real),
+            _ => unreachable!(),
         }
+    } else {
+        Err(error(
+            span,
+            format!(
+                "error: bit_vector prover cannot handle this type (bit_vector can only handle variables of type `bool` or of fixed-width integers)"
```
jaylorch: This error message should be updated to include "or of floating point".

source/air/src/parser.rs
```diff
@@ -234,10 +266,82 @@ impl Parser {
                     }
                     _ => {}
                 }
-                let args = self.nodes_to_exprs(&nodes[1..])?;
+                let round = |args: &mut Exprs| {
```
jaylorch, line 269: This is a closure, but it doesn't seem to need to capture anything. It seems like it should just be a separate function. Also, it doesn't seem to need to take a mutable expression, since it can just do `nodes.get(2)` instead of `remove(0)` and then `nodes.get(1)`.

source/air/src/typecheck.rs
```diff
@@ -235,6 +245,69 @@ fn check_bv_exprs(
...
+    typing: &mut Typing,
+    op: UnaryOp,
+    f_name: &str,
+    expr: &Expr,
+) -> Result<Typ, TypeError> {
+    let t0 = check_expr(typing, expr)?;
+    match &*t0 {
+        TypX::Float { .. } => match op {
+            UnaryOp::FloatNeg => Ok(t0.clone()),
+            UnaryOp::FloatRoundToInt(_) => Ok(t0.clone()),
```
jaylorch, line 259: Doesn't `FloatRoundToInt` produce a value with an integer type, not a floating-point value?

source/builtin/src/lib.rs
```diff
@@ -1414,6 +1416,8 @@ impl_ord!([
 ]);
 
 impl_ord_self_rhs!([real]);
+impl_ord_self_rhs!([f32]);
```
jaylorch, line 1419: I don't think `f32` and `f64` implement `Ord`, because of weirdness with `NaN`. It implements `PartialOrd`. See https://doc.rust-lang.org/std/primitive.f32.html

source/rust_verify/src/verus_items.rs
```diff
@@ -217,13 +217,45 @@ pub(crate) enum SpecBitwiseItem {
...
+    IsPositive,
+}
+
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Hash)]
+pub(crate) enum IeeeFloatBinaryItem {
+    Add,
+    Sub,
+    Mul,
+    Div,
+    Eq,
```
jaylorch, line 244: No `Ne`?

source/rust_verify_test/tests/float.rs
```diff
@@ -49,3 +49,92 @@ test_verify_one_file! {
...
 }
+
+test_verify_one_file! {
+    #[test] f32_ieee verus_code! {
+        fn test1(x: f32, y: f32) {
+            assert(2.0f32 <= x <= 5.0f32 ==> x + x <= 10.0f32) by(bit_vector);
+            assert(2.0f32 <= x <= 5.0f32 && 2.0f32 <= y <= 5.0f32 ==> x + y == y + x) by(bit_vector);
+            assert(0.5f32.ieee_cast() === 0.5real) by(bit_vector);
+            assert(0.5f32 === 0.5real.ieee_cast()) by(bit_vector);
+            assert(4f32.ieee_cast() === 4u8) by(bit_vector);
```
jaylorch, line 62: I thought we were avoiding `===` in favor of `==` now.

Chris-Hawblitzel:
> It's unfortunate that the new axioms aren't meaningfully distinct from the other axioms in vstd. There's no indication that this is a platform specific assumption other than via the name. Is there a way we could unify this, design-wise, with our global directive for specifying platform assumptions?

Ok, I split this aspect into a separate pull request https://github.com/verus-lang/verus/pull/2232 so we can discuss this separately.

source/vir/src/sst_to_air.rs
```diff
@@ -1068,6 +1067,50 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<
...
                 Arc::new(ExprX::Unary(air::ast::UnaryOp::RealToInt, expr))
             }
+            UnaryOp::FloatToBits => exp_to_expr(ctx, e, expr_ctxt)?,
+            UnaryOp::IeeeFloat(crate::ast::IeeeFloatUnaryOp::Cast) => {
+                let t_from = undecorate_typ(&e.typ);
+                let t_to = undecorate_typ(&exp.typ);
+                let mut args: Vec<Expr> = Vec::new();
+                let fname = match (&*t_from, &*t_to) {
+                    (TypX::Real, _) => {
+                        args.push(typ_to_id(ctx, &t_to));
```
jaylorch, line 1081: Why doesn't this path need `t_from`?

source/vir/src/sst_to_air.rs
```diff
@@ -1068,6 +1067,50 @@ pub(crate) fn exp_to_expr(ctx: &Ctx, exp: &Exp, expr_ctxt: &ExprCtxt) -> Result<
...
+                        crate::def::IEEE_FLOAT_CAST_FROM_REAL
+                    }
+                    (_, TypX::Real) => {
+                        args.push(typ_to_id(ctx, &t_from));
```
jaylorch, line 1085: Why doesn't this path need `t_to`?

source/vir/src/bitvector_to_air.rs
Chris-Hawblitzel: Thanks for spotting this.  It's fixed now.

source/air/src/typecheck.rs
Chris-Hawblitzel: The SMTLIB spec defines the types, and "round" produces a float.

source/builtin/src/lib.rs
Chris-Hawblitzel: This isn't implementing `Ord` or `PartialOrd`, but rather Verus's internal spec-level syntax handling (`SpecOrd`).

source/rust_verify/src/verus_items.rs
Chris-Hawblitzel: The SMTLIB spec ( https://smt-lib.org/theories-FloatingPoint.shtml ) omits Ne.

source/vir/src/sst_to_air.rs
Chris-Hawblitzel: I added a comment about this and a case to make explicit that (real, real) is unreachable.

source/vir/src/sst_to_air.rs
Chris-Hawblitzel: I added a comment about this and a case to make explicit that (real, real) is unreachable.

Issue #2219, opened 2026-03-01
ghost uses of never type invalidates borrowchecking


tjhance:
Verus wrongly accepts this:

```rust
proof fn test(nvr: !) { }
proof fn consume<T>(tracked t: T) { }

proof fn test2<T>(tracked t: T) {
    test(arbitrary());

    consume(t);
    consume(t);
}
```

because the ghost use of the `!` type causes Rust to think the `consume` statements are unreachable.

PR #2230, opened 2026-03-06
keep erased_ghost_value calls from influencing the CFG

tjhance:
partially addresses #2219


tjhance:
this doesn't actually totally fix the problem, there's still proof functions that return spec values.

PR #2233, opened 2026-03-07
feat: streamline `proof_note` syntax on `assert`/`assume` statements

mkovaxx:
This is part 2 of eliminating the `proof_note` syntax quirks:
- A proof note on an `assert`/`assume` is now simply an outer attribute on the statement itself.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -138,10 +138,9 @@ fn path_matches_idents(path: &Path, expected: &[&str]) -> bool {
 }
 
 fn split_off_proof_note_attrs(attrs: Vec<Attribute>) -> (Vec<Attribute>, Vec<Attribute>) {
-    let (mut proof_note_attrs, other_attrs): (Vec<Attribute>, Vec<Attribute>) =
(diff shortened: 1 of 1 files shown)


mkovaxx:
@parno @Chris-Hawblitzel There's nothing complex in here; just a simple follow-up to #2226.

Chris-Hawblitzel:
Is it possible this could be done in rust_verify/src/attributes.rs and rust_verify/src/rust_to_vir_expr.rs rather than in the syntax macro?

mkovaxx:
@Chris-Hawblitzel

> Is it possible this could be done in rust_verify/src/attributes.rs and rust_verify/src/rust_to_vir_expr.rs rather than in the syntax macro?

It's technically possible, but seems both more complex and more brittle. The core issue is that `builtin_macros` expands `assume`/`assert` into forms wrapped by `#[verifier::proof_block]`, which severs the connection between the `proof_note` and the statement. Reconstructing that connection later in `rust_verify` must then take into account multiple lowering paths and expression forms. Because of that, the macro rewrite point seems like the cleaner and more robust place to handle this.

PR #2238, opened 2026-03-09
Support for unicode str, specs for std::str fns, specs for UTF-8

zero-to-nat:
Moving many definitions for UTF-8 over from the Rust standard library verification project. Made a couple simplifications to vstd/string.rs and added support for reasoning about the byte-level representation of a `str`.

Simplifications/modifications:
- Implement definition for `str::is_ascii` and `String::is_ascii`, which were previously uninterp
- Remove builtin `strslice_is_ascii` for string literals
- cfg for items in vstd/string.rs module now includes `not(verus_verify_core)`
- Fix bugs in a few specs for exec fns implemented on `str` in vstd/string.rs

Additions to vstd/string.rs:
- `spec_bytes()` for reasoning about the byte-level representation of a `str`
- specs for these `std::str` functions: `as_bytes`, `len`, `is_empty`, `is_char_boundary`, `split_at`, `from_utf8_unchecked`

New file: vstd/utf8.rs
- specs: `encode_utf8`, `decode_utf8`, `valid_utf8`, `partial_valid_utf8`, `is_char_boundary`
- proofs that `encode_utf8` and `decode_utf8` are inverses, and that the result of `encode_utf8` satisfies `valid_utf8`
- lemmas for other useful facts about the above specs

### DIFF
--- a/source/vstd/utf8.rs
+++ b/source/vstd/utf8.rs
@@ -260,7 +260,7 @@ proof fn lemma_pop_first_scalar_decreases(bytes: Seq<u8>)
     };
 }
 
-/// Takes the bytes corresponding to the the first scalar encoded in UTF-8 in the given byte sequence, assuming that the sequence begins with a well-formed encoding of a single scalar.
+/// Takes the bytes corresponding to the first scalar encoded in UTF-8 in the given byte sequence, assuming that the sequence begins with a well-formed encoding of a single scalar.
 pub open spec fn take_first_scalar(bytes: Seq<u8>) -> Seq<u8>
     recommends
         valid_first_scalar(bytes),
@@ -307,7 +307,7 @@ pub open spec fn length_of_last_scalar(bytes: Seq<u8>) -> int
     }
 }
 
-/// Takes the bytes corresponding to the the last scalar encoded in UTF-8 in the given byte sequence, assuming that the bytes form a valid UTF-8 encoding.
+/// Takes the bytes corresponding to the last scalar encoded in UTF-8 in the given byte sequence, assuming that the bytes form a valid UTF-8 encoding.
 pub open spec fn take_last_scalar(bytes: Seq<u8>) -> Seq<u8>
     recommends
         valid_utf8(bytes),
@@ -964,7 +964,7 @@ pub broadcast proof fn is_char_boundary_start_end_of_seq(bytes: Seq<u8>)
     }
 }
 
-/// Ensures that any byte in a valid UTF-8 byte sequence falls on a character boundary if and only if it does not have the form of a UTF-8 continuation byte.
+/// Ensures that any byte in a valid UTF-8 byte sequence falls on a character boundary (i.e. the first byte in a codepoint's encoding) if and only if it does not have the form of a UTF-8 continuation byte.
 pub broadcast proof fn is_char_boundary_iff_not_is_continuation_byte(bytes: Seq<u8>, index: int)
     requires
         valid_utf8(bytes),
@@ -985,7 +985,7 @@ pub broadcast proof fn is_char_boundary_iff_not_is_continuation_byte(bytes: Seq<
     }
 }
 
-/// Ensures that any byte in a valid UTF-8 byte sequence falls on a character boundary if and only if it has the form of a UTF-8 leading byte.
+/// Ensures that any byte in a valid UTF-8 byte sequence falls on a character boundary (i.e. the first byte in a codepoint's encoding) if and only if it has the form of a UTF-8 leading byte.
 pub broadcast proof fn is_char_boundary_iff_is_leading_byte(bytes: Seq<u8>, index: int)
     requires
         valid_utf8(bytes),


source/vstd/string.rs
```diff
@@ -158,76 +241,70 @@ impl StrSliceExecFns for str {
...
-            self.is_ascii(),
         ensures
-            ret.view() == Seq::new(self.view().len(), |i| self.view().index(i) as u8),
+            ret@ == self.spec_bytes(),
     {
-        let mut v = alloc::vec::Vec::new();
-        for c in self.as_bytes().iter() {
-            v.push(*c);
+        proof {
+            is_ascii_spec_bytes(self);
```
parno: Do we still need this call?

source/vstd/string.rs
```diff
@@ -106,10 +192,9 @@ impl StrSliceExecFns for str {
     #[verifier::external_body]
     fn get_char(&self, i: usize) -> (c: char)
         requires
-            i < self@.len(),
+            0 <= i < self@.len(),
```
parno: Very minor, but I think the new `0 <=` is redundant with the type of `i` being `usize`.

source/vstd/string.rs
```diff
@@ -158,76 +241,70 @@ impl StrSliceExecFns for str {
...
         let byte_end = byte_end.unwrap();
+        // Range::index panics if from > to or from > self@.len()
         &self[byte_start..byte_end]
     }
 
-    #[verifier::external_body]
     fn get_ascii(&self, i: usize) -> (b: u8)
         requires
             self.is_ascii(),
+            0 <= i < self@.len(),
```
parno: We could also consider removing the `0 <=` here.  And the lack of the upperbound in the previous version was definitely a problem!

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
+//! Definitions for UTF-8 encoding and decoding of character sequences.
+//!
+//! [UTF-8](https://en.wikipedia.org/wiki/UTF-8) is a variable-width character encoding scheme.
+//! Each character is encoded with between 1 and 4 bytes.
+//! Specifications for encoding and decoding characters to their UTF-8 byte sequences are given by [`encode_utf8`] and [`decode_utf8`], respectively.
+//! Characters in the ASCII character set are encoded in UTF-8 with 1-byte encodings identical to those used by ASCII.
+//! Thus, some UTF-8 byte sequences can also be considered ASCII byte sequences, as defined in [`is_ascii_chars`].
+//!
+//! UTF-8 encodes numerical values called Unicode _scalars_ (see below), which assign a unique value to each Unicode character.
+//! A scalar value is encoded in UTF-8 using a leading byte and between 0 and 2 continuation bytes, where larger scalar values require more continuation bytes.
```
parno: Shouldn't this be `0 and 3` continuation bytes?

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
...
+//! The first part of the bit pattern in the leading byte is reserved for describing the number of bytes in the scalar's encoding (e.g., [`is_leading_byte_width_1`]).
+//! The rest of the leading byte contains data bits corresponding to the scalar's value (e.g., [`leading_bits_width_1`]).
+//! The continuation bytes also follow a specific bit pattern ([`is_continuation_byte`]) and contain the remainder of the data bits ([`continuation_bits`]).
+//!
+//! This module makes use of terminology from the [Unicode standard](https://www.unicode.org/glossary/).
+//! A Unicode _scalar_ is a numerical value (represented in this module as a `u32`) corresponding to a character that can be encoded in UTF-8.
+//! All Rust `char`s correspond to Unicode scalars ([`char_is_scalar`]),
+//! and every numerical value encoded in a UTF-8 byte sequence must fall within the range defined for Unicode scalars ([`is_scalar`]).
+//! The Unicode standard also defines a _codepoint_ to be numerical value which falls in the range available for encoding characters in UTF-8.
```
parno: "to be **a** numerical"

source/vstd/utf8.rs
parno, line 1: The documentation and terminology at the top of the file is fantastic.  Thanks for assembling that!

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
...
+pub open spec fn valid_leading_and_continuation_bytes_first_codepoint(bytes: Seq<u8>) -> bool {
+    ||| (bytes.len() >= 1 && is_leading_byte_width_1(bytes[0]))
+    ||| (bytes.len() >= 2 && is_leading_byte_width_2(bytes[0]) && is_continuation_byte(bytes[1]))
+    ||| (bytes.len() >= 3 && is_leading_byte_width_3(bytes[0]) && is_continuation_byte(bytes[1])
+        && is_continuation_byte(bytes[2]))
+    ||| (bytes.len() >= 4 && is_leading_byte_width_4(bytes[0]) && is_continuation_byte(bytes[1])
+        && is_continuation_byte(bytes[2]) && is_continuation_byte(bytes[3]))
+}
+
+/// The first codepoint encoded in UTF-8 in the given byte sequence, assuming that the sequence begins with a well-formed leading byte and an appropriate number of well-formed continuation bytes.
```
parno: Might be clearer if the comment started with "Returns the first ..."

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
...
+    } else if is_leading_byte_width_2(bytes[0]) {
+        codepoint_width_2(bytes[0], bytes[1])
+    } else if is_leading_byte_width_3(bytes[0]) {
+        codepoint_width_3(bytes[0], bytes[1], bytes[2])
+    } else {
+        codepoint_width_4(bytes[0], bytes[1], bytes[2], bytes[3])
+    }
+}
+
+/// The length in bytes of first codepoint encoded in UTF-8 in the given byte sequence, assuming that the sequence begins with a well-formed leading byte and an appropriate number of well-formed continuation bytes.
```
parno: "The length in bytes of **the** first"

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+/// The length in bytes of first scalar encoded in UTF-8 in the given byte sequence, assuming that the sequence begins with a well-formed encoding of a single scalar.
+pub open spec fn length_of_first_scalar(bytes: Seq<u8>) -> int
+    recommends
+        valid_first_scalar(bytes),
+{
+    length_of_first_codepoint(bytes)
+}
+
+proof fn lemma_pop_first_scalar_decreases(bytes: Seq<u8>)
```
parno, line 251: Can we move this below the definition of `pop_first_scalar`?

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+/// True when the given value is a Unicode scalar with a 4-byte UTF-8 encoding.
+pub open spec fn has_width_4_encoding(v: u32) -> bool {
+    0x10000 <= v <= 0x10FFFF
+}
+
+/// True when the given `u32` represents a Unicode scalar, i.e., a value that can be encoded in UTF-8.
+/// This definition is equivalent to: `0 <= v <= 0x10ffff && !(0xD800 <= v <= 0xDFFF)`.
+pub open spec fn is_scalar(v: u32) -> bool {
+    has_width_1_encoding(v) || has_width_2_encoding(v) || has_width_3_encoding(v)
```
parno: I think verusfmt would provide a nicer result here if you use:
```
||| has_width_1_encoding(v)
||| has_width_2_encoding(v)
...
```

zero-to-nat:
> I think this looks great -- it's very nicely organized and the new specs and definitions are cleanly stated and documented. I also tested it with the (verifying) projects in `verita` and I didn't see any changes in proof success or timing.
> 
> I'd still like to get @dschoepe's thoughts, however, as I believe he and his team use strings pretty heavily.

Thank you for the review, and for running verita on the changes! And that sounds good, it would be great to get Daniel's thoughts.

source/vstd/string.rs
```diff
@@ -158,76 +241,66 @@ impl StrSliceExecFns for str {
...
+        broadcast use is_ascii_spec_bytes;
+        // panics if i is not a valid index
+
         self.as_bytes()[i]
     }
 
-    // TODO:This should be the as_bytes function after
-    // slice support is added
-    // pub fn as_bytes<'a>(&'a [u8]) -> (ret: &'a [u8])
-    #[cfg(feature = "alloc")]
```
elanortang: Should this still have the "alloc" cfg, since it returns a `Vec`?

source/vstd/string.rs
```diff
@@ -85,10 +171,10 @@ pub trait StrSliceExecFns {
 
     fn get_ascii(&self, i: usize) -> u8;
 
-    #[cfg(feature = "alloc")]
     fn as_bytes_vec(&self) -> alloc::vec::Vec<u8>;
 }
 
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
```
elanortang: It definitely makes sense to apply `not(verus_verify_core)` to the entire impl, but maybe double-check the guidelines for when `feature = "alloc"` is to be applied? I'm not sure if that's supposed to only be on functions which refer to things in `alloc` (i.e. `as_bytes_vec`) or if it also makes sense to do it over the whole impl.

source/vstd/string.rs
```diff
@@ -28,51 +34,131 @@ impl DeepView for str {
...
+pub assume_specification[ str::as_bytes ](s: &str) -> (b: &[u8])
+    ensures
+        b@ == s.spec_bytes(),
+;
+
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ str::len ](s: &str) -> (res: usize)
+    ensures
+        res == s.spec_bytes().len(),
```
elanortang: extremely minor, but personally I like to use `returns` whenever possible - there's less to parse, and you don't have to bother with naming the output variable. up to you though!

source/vstd/string.rs
```diff
@@ -28,51 +34,131 @@ impl DeepView for str {
...
+;
+
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ str::is_empty ](s: &str) -> (res: bool)
+    ensures
+        res == (s@.len() == 0),
```
elanortang: same here

source/vstd/string.rs
```diff
@@ -28,51 +34,131 @@ impl DeepView for str {
...
+;
+
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ str::is_char_boundary ](s: &str, index: usize) -> (res: bool)
+    ensures
+        res == (is_char_boundary(s.spec_bytes(), index as int)),
```
elanortang: same here

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1037 @@
...
+    }
+}
+
+/// Properties of the first scalar from the result of [`decode_utf8`].
+pub proof fn decode_utf8_first_scalar(bytes: Seq<u8>)
+    requires
+        valid_utf8(bytes),
+        bytes.len() > 0,
+    ensures
+        encode_scalar((decode_first_scalar(bytes) as char) as u32) == take_first_scalar(bytes),
```
elanortang, line 684: how come you cast from u32 to char to u32 here?

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1037 @@
...
+
+            assert(pop_first_scalar(bytes.subrange(j, i)) == bytes.subrange(j + k, i));
+
+            partial_valid_utf8_invalid_subrange_helper(bytes, i, j + k);
+
+            assert(bytes_j.subrange(k, bytes_j.len() as int) == bytes.subrange(
+                j + k,
+                bytes.len() as int,
+            ));
+        } else {
```
elanortang: I know this is copied directly from my proof but looking at it now I think line 723 can be removed

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1037 @@
...
+{
+    if b1.len() == 0 {
+        assert(b1 + b2 == b2) by { Seq::add_empty_left(b1, b2) };
+        assert(valid_utf8(b1 + b2));
+    } else {
+        let rest = pop_first_scalar(b1);
+        // so valid_utf8(b1) implies valid_utf8(rest)?
+        assert(pop_first_scalar(b1).len() < b1.len()) by { lemma_pop_first_scalar_decreases(b1) };
+        valid_utf8_concat(rest, b2);
+        // now know valid_utf8(rest+b2)
```
elanortang: can probably delete the comments - I think I was just figuring things out

source/vstd/utf8.rs
```diff
@@ -0,0 +1,1037 @@
...
+    valid_utf8_concat(bytes.subrange(0, i), scalar);
+    assert(bytes.subrange(0, i) + scalar =~= bytes.subrange(
+        0,
+        i + length_of_first_scalar(bytes.subrange(i, bytes.len() as int)),
+    ));
+}
+
+/* Reasoning about character boundaries */
+
+/// True when the given index into the byte sequence falls on a character boundary, assuming that the sequence is valid UTF-8.
```
elanortang: by "on" a character boundary, you mean that the boundary falls just before the byte being indexed?

source/vstd/string.rs
dschoepe: Another advantage is that you can then call `s.len()` in a `spec` context when using `returns`.

dschoepe:
> > I think this looks great -- it's very nicely organized and the new specs and definitions are cleanly stated and documented. I also tested it with the (verifying) projects in `verita` and I didn't see any changes in proof success or timing.
> > I'd still like to get @dschoepe's thoughts, however, as I believe he and his team use strings pretty heavily.
> 
> Thank you for the review, and for running verita on the changes! And that sounds good, it would be great to get Daniel's thoughts.

No concerns from my side, thanks for putting this together! I like @elanortang's suggestion of using `returns` where possible since it allows using those functions directly in `spec` contexts.

parno:
> No concerns from my side, thanks for putting this together! I like @elanortang's suggestion of using `returns` where possible since it allows using those functions directly in `spec` contexts.

Indeed, that can add to the convenience, but I think for that, you'll also need to add an [![verifier::allow_in_spec]](https://verus-lang.github.io/verus/guide/reference-attributes.html?highlight=allow_in#verifierallow_in_spec).  I'd be in favor of making that automatic whenever you use `returns`, but I think it probably needs discussion at a Verus meeting.

dschoepe:
> Indeed, that can add to the convenience, but I think for that, you'll also need to add an [![verifier::allow_in_spec]](https://verus-lang.github.io/verus/guide/reference-attributes.html?highlight=allow_in#verifierallow_in_spec). I'd be in favor of making that automatic whenever you use returns, but I think it probably needs discussion at a Verus meeting.

You're right, I forgot that that's needed.

source/vstd/string.rs
zero-to-nat: Thanks for the suggestion! That is a nice improvement

source/vstd/utf8.rs
zero-to-nat: This is brought over from the definitions of `decode_utf8` and `encode_utf8`. I decided to keep it the exact definition, because this lemma is about roundtrip properties

source/vstd/utf8.rs
zero-to-nat: Good point, thanks for the clarification! I borrowed the phrasing from the Rust documentation which says that the index is the first byte in a codepoint

source/vstd/utf8.rs
```diff
@@ -289,6 +289,52 @@ pub open spec fn decode_utf8(bytes: Seq<u8>) -> Seq<char>
...
+    } else if !is_continuation_byte(bytes[n - 2]) {
+        2
+    } else if !is_continuation_byte(bytes[n - 3]) {
+        3
+    } else {
+        4
+    }
+}
+
+/// Takes the bytes corresponding to the the last scalar encoded in UTF-8 in the given byte sequence, assuming that the bytes form a valid UTF-8 encoding.
```
elanortang: typo: "the the last"

source/vstd/string.rs
zero-to-nat: thank you for catching this! indeed, the configs I added differed from the original behavior. I changed it back so that `as_bytes_vec` is the only fn cfg-ed on `alloc`

zero-to-nat:
> thanks so much for putting this together and reorganizing everything!! I only had very minor comments.
> 
> As far as I could tell from just looking at things, this seemed to be compatible with str targets - we won't know for sure until we try to verify them, of course, but there was nothing glaring.
> 
> Do you have a sense of what changes `rustlib` will need for when this is merged into `verify-rustlib`? Does it break anything in the existing str functions?

Thank you for the review and the helpful comments! 

I have a version of `rustlib`'s main branch which verifies against this library. It might need some more clean up, but at least we know it works. Overall, the structure stayed the same, and I just needed to add a few more lemmas/triggers. Most of the things that broke had to do with the new definitions for `is_leading_byte_width_1` etc. -- the code in the standard library tends to use bitwise operators like `&` rather than ranges of values

source/vstd/utf8.rs
elanortang: the updated documentation here looks good! The documentation for  `is_char_boundary_iff_not_is_continuation_byte` and `is_char_boundary_iff_is_leading_byte` also talks about being "on" a character boundary - could you update that correspondingly as well?

zero-to-nat:
> > > I think this looks great -- it's very nicely organized and the new specs and definitions are cleanly stated and documented. I also tested it with the (verifying) projects in `verita` and I didn't see any changes in proof success or timing.
> > > I'd still like to get @dschoepe's thoughts, however, as I believe he and his team use strings pretty heavily.
> > 
> > 
> > Thank you for the review, and for running verita on the changes! And that sounds good, it would be great to get Daniel's thoughts.
> 
> No concerns from my side, thanks for putting this together! I like @elanortang's suggestion of using `returns` where possible since it allows using those functions directly in `spec` contexts.

Thank you for the review! Agreed, that was a good suggestion.

PR #2243, opened 2026-03-12
Allow `#[verus_verify]` on `const` definitions

jaylorch:
Written by Github Copilot + Claude Opus 4.6 (after I rejected its first suggestion, which was way too elaborate, and suggested it just do what `verus!` does).

### DIFF
--- a/source/rust_verify_test/tests/syntax_attr.rs
+++ b/source/rust_verify_test/tests/syntax_attr.rs
@@ -1281,3 +1281,16 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] test_verus_verify_on_const code! {
+        #[verus_verify]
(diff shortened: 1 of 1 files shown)


ziqiaozhou:
Another way to fix  (https://github.com/verus-lang/verus/commit/ae5efd49a54ed71178f2572b0c421d348db36c97) without changing the verus internal behavior is to add #[verus_spec] when applying #[verus_verify] to properly desugar const item, similar to how I handle ImplItemConst.

jaylorch:
> Another way to fix ([ae5efd4](https://github.com/verus-lang/verus/commit/ae5efd49a54ed71178f2572b0c421d348db36c97)) without changing the verus internal behavior is to add #[verus_spec] when applying #[verus_verify] to properly desugar const item, similar to how I handle ImplItemConst.

Oh, great, thanks. I didn't see that example. I'll close the PR.

PR #2246, opened 2026-03-13
[builtin_macros] fix: handle `default_ensures` with return value pattern

mkovaxx:
- Test that `default_ensures` works when the return value's name collides with the `fn`'s.
- Update the code to correctly handle `default_ensures` clauses when such a collision happens.
- Inline a single-use function into its call site.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -557,6 +557,8 @@ impl Visitor {
             }
         }
 
+        // Rewrite the `ensures` clauses to protect against edge cases e.g.
+        //     a name collision between the return-value and the function
         let ensures = ensures.map(|mut ensures| {
             if let Some((ret_pat, _)) = &ret_pat {
                 for expr in &mut ensures.exprs.exprs {
@@ -566,6 +568,8 @@ impl Visitor {
             ensures
         });
 
+        // Rewrite the `default_ensures` clauses to protect against edge cases e.g.
+        //     a name collision between the return-value and the function
         let ensures = {
             let mut ensures = ensures;
             if let Some(DefaultEnsures { token, mut exprs }) = default_ensures {


source/builtin_macros/src/syntax.rs
```diff
@@ -483,29 +483,6 @@ fn rewrite_args_unwrap_ghost_tracked(erase_ghost: &EraseGhost, arg: &mut FnArg)
...
-    ensures: Option<Ensures>,
-    default_ensures: Option<DefaultEnsures>,
-) -> Option<Ensures> {
-    let Some(default_ensures) = default_ensures else {
-        return ensures;
-    };
-    let DefaultEnsures { token, mut exprs } = default_ensures;
-    for expr in exprs.exprs.iter_mut() {
-        let span = expr.span();
-        *expr = parse_quote_spanned_builtin!(verus_builtin, span => #verus_builtin::default_ensures(#expr));
```
mkovaxx, line 496: Here each `default_ensures` clause is rewritten to `default_ensures(#expr)`.

When the return value name collides with the function name, each `ensures` clause must also be rewritten to`{ let #final_ret_pat = #p; #expr }`.

If the `default_ensures` rewrite happens first, the result is `{ let #final_ret_pat = #p; default_ensures(#expr) }`. Downstream transformations expect builtins such as `default_ensures` to be the outermost layer, so that causes a failure.

After this fix, the result of the rewrite is `default_ensures({ let #final_ret_pat = #p; #expr })`.

source/builtin_macros/src/syntax.rs
```diff
@@ -569,7 +546,45 @@ impl Visitor {
...
+                let expr_span = expr.span();
+                let attrs = expr.replace_attrs(Vec::new());
+                let inner = take_expr(expr);
+                let wrapped: Expr =
+                    parse_quote_spanned!(expr_span => { let #final_ret_pat = #ret_pat; #inner });
+                *expr = wrap_expr_with_attrs(wrapped, attrs);
+            }
+        }
+
+        let ensures = ensures.map(|mut ensures| {
```
parno, line 562: Could we add a comment here and above the block below to summarize what's happening?  E.g., here, it might be "Protect against name collisions between the return-value name and the function name"

mkovaxx, line 562:
```suggestion
        // Rewrite the `ensures` clauses to protect against edge cases e.g.
        //     a name collision between the return-value and the function
        let ensures = ensures.map(|mut ensures| {
```

source/builtin_macros/src/syntax.rs
```diff
@@ -569,7 +546,45 @@ impl Visitor {
...
+            if let Some((ret_pat, _)) = &ret_pat {
+                for expr in &mut ensures.exprs.exprs {
+                    wrap_with_ret_binding_pat(expr, ret_pat, &final_ret_pat);
+                }
+            }
+            ensures
+        });
+
+        let ensures = {
```
mkovaxx, line 573:
```suggestion
        // Rewrite the `default_ensures` clauses to protect against edge cases e.g.
        //     a name collision between the return-value and the function
        let ensures = {
```

PR #2247, opened 2026-03-14
Fixed mistake in "Putting it all together" page in docs

jsenn:

### DIFF
--- a/examples/guide/recursion.rs
+++ b/examples/guide/recursion.rs
@@ -105,7 +105,7 @@ fn rec_triangle(n: u32) -> (sum: u32)
 // ANCHOR: rec
 fn rec_triangle(n: u32) -> (sum: u32)
     requires
-        triangle(n as nat) < 0x1_0000_0000,
+        triangle(n as nat) < u32::MAX,
     ensures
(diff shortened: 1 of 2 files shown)


tjhance:
I agree we should use `u32::MAX`, it's helpful for users to know that they can use those

jsenn:
Updated the text per @parno's suggestion and changed the recursion example to use u32::MAX throughout instead of the magic number.

This probably should've been an issue not a PR so you could word things how you want, sorry about that.

parno:
Thanks for making the updates!  I think it looks good now.

> This probably should've been an issue not a PR so you could word things how you want, sorry about that.

No worries -- we appreciate contributions in all forms :)

tjhance:
yes, PRs are a perfectly fine way to iterate on minor changes

jaylorch:
Should this use `<= u32::MAX` instead of `< u32::MAX`?

jsenn:
> Should this use `<= u32::MAX` instead of `< u32::MAX`?

yes it should, good catch: https://github.com/verus-lang/verus/pull/2256

Issue #1739, opened 2025-06-12
Using verus in a workspace breaks cargo


bsdinis:
I'm trying to use verus in a workspace (i.e., separate but dependent crates).
Verus (or verus' invokation of cargo) fails to find the "library" crate:

```
error: extern location for lib does not exist: /path/to/verus-workspace/target/debug/deps/liblib-8475093cacd66a51.rlib
 --> lib-user/src/main.rs:6:12
  |
6 |     assert(lib::is_even(2));
  |            ^^^

error: extern location for lib does not exist: /path/to/verus-workspace/target/debug/deps/liblib-8475093cacd66a51.rlib
 --> lib-user/src/main.rs:7:13
  |
7 |     assert(!lib::is_even(3));
  |             ^^^
```

# Steps to reproduce

I've created a minimal example here: https://github.com/bsdinis/verus-workspace

Running `cargo verus` yields the above error.

mkovaxx:
I briefly looked into this.

Brought @bsdinis's example back in sync, PR here: https://github.com/bsdinis/verus-workspace/pull/1

Can confirm that the issue persists.

Running `cargo-verus` from the `verus/source` dir as follows:
```
./target-verus/debug/cargo-verus check -v --manifest-path $HOME/prog/verus-workspace/Cargo.toml
```

Output:
```
running cargo command:
[...]

PR #2251, opened 2026-03-18
[cargo-verus] fix: make it possible to use verified lib from same package

### LINKED ISSUE #1905: Code in `src/bin` recognizes traits in lib as external, not verified
Hi,

I'm brand new to this ecosystem, I just discovered Verus and Anvil a few days ago. It seems Verus is in the middle of pushing towards making things feel more out-of-the-box and cargo-native with the addition of `cargo verus`, which is great!

I [opened an issue in the Anvil repo](https://github.com/anvil-verifier/anvil/issues/697) about supporting `cargo verus` and have been having a go at that, which involves moving some of the code into cargo binaries. However, I discovered a peculiar issue. I have a minimal repro, so let me explain -- here is my project layout:

```sh
test
├── Cargo.lock
├── Cargo.toml
├── src
│   ├── lib.rs
│   └── main.rs
└── test2
    ├── Cargo.lock
    ├── Cargo.toml
    └── src
        └── lib.rs
```

`test2` has `test` as a dependency. `test/src/lib.rs` defines a trait like this:
```rust
verus! {

pub trait Foo {
    open spec fn hello() -> int {
        42
    }
}

}
```

And then in both `src/main.rs` and `test2/src/lib.rs` I have a trait impl like this:

```rust
verus! {

struct Bar {}
    
impl test::Foo for Bar {
    open spec fn hello() -> int {
        43
    }
}

}
```

As it turns out, the one in `test2/src/lib.rs` compiles, but the one in `src/main.rs` fails with the following error:

```sh
❯ cargo verus build
   Compiling test v0.1.0 (/home/suremarc/anvil/test)
error: function for external trait must have mode 'exec'
 --> src/main.rs:8:15
  |
8 |     open spec fn hello() -> int {
  |               ^^^^^^^^^^^^^^^^^

error: could not compile `test` (bin "test") due to 1 previous error
```

Being new to the ecosystem, I'm not 100% what's happening here, but from what I can gather, traits are either marked as 'external' (not proof or spec-carrying) or verified. Since `Foo` comes from a `verus!` invocation it should be verified, and indeed, again the code in `test2/src/lib.rs` compiles fine even though `Foo` is external to `test2`. It seems like maybe this is a case overlooked by Verus? 

I've attached the repro as a .zip file below. Thanks for reading this far!

[test.zip](https://github.com/user-attachments/files/22670094/test.zip)

jsenn:
Fixes #1905. When verifying an example/test/bin/bench target in a package that has a lib target, the .vir files are being generated correctly. However, the driver doesn't see the lib target's .vir file, as it only considers what is passed in with `import-dep-if-present`.

The fix here is to add the lib itself to `import-dep-if-present`. One weird consequence is the lib will be passed as a dependency to itself. This is harmless as the driver only uses deps that are in its externs, but it is a little awkward.

Also added a test that exercises this case with a mock example target.

### DIFF
--- a/source/cargo-verus/tests/test_build.rs
+++ b/source/cargo-verus/tests/test_build.rs
@@ -8,11 +8,8 @@ fn lib_with_example_imports_own_lib() {
     let package_name = "mylib";
     let args_prefix = format!(" __VERUS_DRIVER_ARGS_FOR_{package_name}-0.1.0-");
 
-    let project_dir = MockPackage::new(package_name)
-        .lib()
-        .example("foo")
-        .verify(true)
-        .materialize();
+    let project_dir =
+        MockPackage::new(package_name).lib().example("foo").verify(true).materialize();
 
     let (status, data) = run_cargo_verus(|cmd| {
         cmd.current_dir(&project_dir).arg("build");
@@ -33,10 +30,7 @@ fn bin_only_no_own_lib_import() {
     let package_name = "mybin";
     let args_prefix = format!(" __VERUS_DRIVER_ARGS_FOR_{package_name}-0.1.0-");
 
-    let project_dir = MockPackage::new(package_name)
-        .bin("main")
-        .verify(true)
-        .materialize();
+    let project_dir = MockPackage::new(package_name).bin("main").verify(true).materialize();
 
     let (status, data) = run_cargo_verus(|cmd| {
         cmd.current_dir(&project_dir).arg("build");


mkovaxx:
This looks good to me, and I've confirmed that the `cargo-verus` tests succeed.
```
cargo +nightly test -p cargo-verus --features integration-tests
```

I don't feel knowledgeable enough about the sanity of treating a crate as its own dep in Verus via `import-dep-if-present`, so I'll defer confirming that to @Chris-Hawblitzel.

Chris-Hawblitzel:
@jsenn It looks like this change leads to an error in the verified-storage project in https://github.com/verus-lang/verus/tree/main/tools/verita :

```
error: duplicate specification for this trait implementation
  --> src/main_t.rs:44:5
   |
44 | /     impl CheckPermission<Seq<u8>> for TrustedPermission {
45 | |         closed spec fn check_permission(&self, state: Seq<u8>) -> bool {
46 | |             (self.is_state_allowable)(state)
47 | |         }
48 | |     }
   | |_____^
   |
  ::: verified-storage/pmemlog/src/main_t.rs:44:5
   |
44 | /     impl CheckPermission<Seq<u8>> for TrustedPermission {
45 | |         closed spec fn check_permission(&self, state: Seq<u8>) -> bool {
46 | |             (self.is_state_allowable)(state)
47 | |         }
48 | |     }
   | |_____^

error: could not compile `pmemlog` (bin "pmemlog") due to 1 previous error
```

From the error message, I'm guessing that the same `trait_impl.x.impl_path` ends up in the `Krate` twice.  Do you have any idea why the change might lead to this error?

parno:
FWIW, the pmemlog code is here:
https://github.com/microsoft/verified-storage/tree/main/pmemlog
I suspect this issue might be related to the fact that the crate includes both a `src/lib.rs` and a `src/main.rs`.

jsenn:
Yes the problem is that `pmemlog/src/main.rs` is recompiling the same modules as the lib rather than importing them. To be honest I didn't know you could "overlap" a lib and a bin this way.

This fix works by adding the lib as a dependency to all targets (including the lib itself). I thought this was ok because there's a downstream check where it only uses deps that are in the externs coming from the build system. But in this case, the lib is building some of those modules, then it's getting added as a dep to the bin, which itself is building the modules, which triggers the duplication error.

A workaround for pmemlog would be to `use` the modules from the lib rather than rebuilding them in the bin.

It might be tricky to handle this case properly. I don't think the driver has enough information to know when it needs to keep a dependency or not. Maybe cargo verus needs to pass an extra arg to help the driver deduplicate.

jsenn:
fwiw I confirmed locally that the following patch in verified-storage builds and verifies cleanly against verus main:

EDIT: actually no, it builds but it doesn't verify.

```diff
diff --git a/pmemlog/src/main.rs b/pmemlog/src/main.rs
index 7f9d8d2d..48b4d5ad 100644
--- a/pmemlog/src/main.rs
+++ b/pmemlog/src/main.rs
@@ -3,17 +3,9 @@
 use vstd::prelude::*;
 use vstd::slice::*;
 
-pub mod infinitelog_t;
-pub mod logimpl_v;
-pub mod main_t;
-pub mod math_v;
-pub mod pmemmock_t;
-pub mod pmemspec_t;
-pub mod sccf;
-
-use crate::main_t::*;
-use crate::pmemspec_t::*;
-use crate::pmemmock_t::*;
+use pmemlog::main_t::*;
+use pmemlog::pmemspec_t::*;
+use pmemlog::pmemmock_t::*;
 
 verus! {
```

parno:
Ah, interesting, thanks for the analysis!  @jaylorch What do you think about applying the patch above to the pmemlog repo?

jsenn:
> Ah, interesting, thanks for the analysis! @jaylorch What do you think about applying the patch above to the pmemlog repo?

sorry, edited the comment but you probably saw the original--the patch doesn't actually work: `cargo verus build` works, and a subsequent `cargo verus verify` works, but `cargo clean && cargo verus verify` fails, not sure why.

jaylorch:
I'm confused. Is there a patch that works?

jsenn:
> I'm confused. Is there a patch that works?

No. There are 2 problems:
1. The original deduplication issue. That one is fixed by the patch in my comment.
2. Fixing that reveals a deeper issue with this fix. A clean `cargo verus verify` *never* worked when you have a lib and a non-lib target in the same package.

I think the second issue is this. `cargo verus verify`, unlike `cargo verus build`, avoids compiling the primary package (`compile_when_primary_package`). But that means that even with my original fix, if you have a cargo example, bin, etc. that depends on the lib, if you run a clean `cargo verus verify`, the lib is not actually compiled, so its symbols can't be seen by the other target.

When I tested my original change I must have only been doing `build` or `build` then `verify`, never a clean `verify`, so didn't catch it.

The trick is to allow the non-lib targets to see the lib's symbols when doing a verify.

jsenn:
ok I think I know what's going on with problem 2.

## TLDR
`cargo verus verify` explicitly short-circuits rustc before it can generate metadata, which means there's no way for internal dependencies to work. Proposed fix is to instead use `cargo check` and allow it to do its thing, including generating metadata. But not 100% sure.

## Background
This PR was trying to fix a limitation of Verus where a non-lib target in a lib package (say an example or bin) was using the lib. Building or verifying the non-lib target would always fail, even if the code was correct. This is what #1905 was about, and it's also probably why pmemlog was written the way it is, with the bin rebuilding the lib modules instead of just consuming the lib (i.e. it's a workaround for #1905).

## Problem
As written, my change only works with `cargo verus build`, or `cargo verus verify` after a previous build. In particular, `cargo clean && cargo verus verify` will still always fail when you try to build the non-lib target.

## Details
The reason for this is as follows. [`cargo verus verify`](https://github.com/verus-lang/verus/blob/060fe640b605b67a9494b72bdd35e352f15e07ff/source/cargo-verus/src/main.rs#L45) works by invoking `cargo build`, but setting a `do_compile` flag to `false`. In the [`after_expansion`](https://github.com/verus-lang/verus/blob/060fe640b605b67a9494b72bdd35e352f15e07ff/source/rust_verify/src/driver.rs#L175) callback in the driver, when `do_compile` is false, the [`check()`](https://github.com/verus-lang/verus/blob/060fe640b605b67a9494b72bdd35e352f15e07ff/source/rust_verify/src/driver.rs#L42) function is invoked, which does some limited borrow-checking, then `Compilation::Stop` is returned, which tells `rustc` not to generate code. Unfortunately, **it also won't emit metadata**. This means that the library's structs, functions, etc. are not visible to the non-lib targets, which is why clean verification fails.

## Possible solutions
To make this work for non-lib targets, somehow **verus needs to let rustc emit the metadata even when only verifying.** One way to do this might be to use `cargo check` as the subcommand for `verify`, instead of doing `cargo build` but stopping before compilation happens. Then the `after_expansion` just always lets `Continue`.

I hacked that in locally and it seems to work, but the code was probably written this way for a reason. In particular, the `check()` function only does limited checking of functions and trait impls (as these are the only places you could find spec or proof code I guess), skipping everything else. So I guess doing a normal `cargo check` would be doing extra work.

Another way would be to have verus invoke whatever rust internals are needed to generate the metadata, in the same way that it manually does a limited borrow check pass, but I don't know how/if that would work.

jsenn:
Here is a PR that implements the proposed fix: https://github.com/verus-lang/verus/pull/2280. Instead of implementing `cargo verus verify` by short-circuiting `cargo build` and running a custom borrow-checking pass, it just uses `cargo check`, and allows it to run to completion, generating metadata. It is now doing more borrow checking than it was previously, so there could be a performance tradeoff.

This, in combination with the patch above, fixes the error in pmemlog.

mkovaxx:
Related issue: #1739.

jsenn:
@Chris-Hawblitzel @jaylorch FYI the above PR has been merged into main. The issue in verified-storage with latest Verus will be fixed if you apply [the patch mentioned above](https://github.com/verus-lang/verus/pull/2251#issuecomment-4145104337) to avoid building the modules twice (once in lib and once in bin).

jaylorch:
OK, we've applied the patch to `pmemlog` so it should work now. Is that going to be a permanent requirement or will Verus eventually be fixed to support `pmemlog` without that change?

jsenn:
> OK, we've applied the patch to `pmemlog` so it should work now. Is that going to be a permanent requirement or will Verus eventually be fixed to support `pmemlog` without that change?

@jaylorch created #2324 which explains what I think the problem is and at a high level how to fix it, but it could be an involved fix.

PR #2257, opened 2026-03-20
restrictions for `tracked_swap`

tjhance:
tracked_swap needs additional restrictions, similar to assignment.

Special-casing tracked_swap kind of sucks, but this is probably the best low-effort, decently backwards compatible solution.

If the user wanted to write their own "tracked_swap", they would have to do it like this:

```
proof fn tracked_swap<A>(tracked a: &mut Tracked<A>, tracked b: &mut Tracked<A>) { ... }
```

The special casing allows us to keep the current type signature, which is:

```
proof fn tracked_swap<A>(tracked a: &mut A, tracked b: &mut A) { ... }
```

~~Option::tracked_take is unfortunately going to be a casualty.~~ EDIT: ok, I decided to special-case this one as well.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -1227,7 +1227,8 @@ fn check_place(
             place_mode
         }
         PlaceAccess::MutBorrow(Some(mut_ref_tracked_typ)) => {
-            let found = record.mut_bor_place_modes.insert(place.span.id, (place_mode, note));
+            let found =
+                record.mut_bor_place_modes.insert(place.span.id, (place_mode, note.clone()));
             if found.is_some() {
                 return Err(error(
                     &place.span,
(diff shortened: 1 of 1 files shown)


source/rust_verify_test/tests/mut_refs_modes.rs
```diff
@@ -1051,7 +1050,169 @@ test_verify_one_file_with_options! {
...
+    } => Err(err) => assert_vir_error_msg(err, "cannot mutate exec-mode place in ghost code")
+}
+
+test_verify_one_file_with_options! {
+    #[test] tracked_swap_requires_non_exec_place3 ["new-mut-ref"] => verus_code! {
+        use vstd::prelude::*;
+        use vstd::modes::*;
+        fn test(x: &mut u32, y: &mut u32) {
+            let tracked a: u32 = 0;
+            let tracked b: u32 = 0;
```
Chris-Hawblitzel: Shouldn't a and b be declared `tracked mut` to be mutated?  (Also, x and y aren't used.)

source/rust_verify_test/tests/mut_refs_modes.rs
```diff
@@ -1051,7 +1050,169 @@ test_verify_one_file_with_options! {
...
+        use vstd::modes::*;
+        tracked struct X { a: u64 }
+        // The type argument being a ZST isn't sufficient
+        fn test(Tracked(x): Tracked<&mut Option<X>>, Tracked(y): Tracked<&mut Option<X>>) {
+            proof {
+                tracked_swap(x, y);
+            }
+        }
+    } => Err(err) => assert_vir_error_msg(err, "cannot mutate exec-mode place in ghost code")
 }
```
Chris-Hawblitzel, line 1234: We don't currently support creating a proof_fn directly from a function value, but for future-proofing, could we add a negative test for:

```
proof fn q<V>(tracked f: proof_fn(tracked &mut V, tracked &mut V) -> ()) {
}
proof fn r() {
    q(tracked_swap);
}
```

source/rust_verify_test/tests/mut_refs_modes.rs
tjhance: the mut check doesn't work for some of these declarations that get de-sugared. long-standing bug

Issue #2236, opened 2026-03-08
False positive when checking for type parameter constraints


parno:
In the code below, Verus [reports an error](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=4a7bd3eec50c53bc88271272839d41d3):
```
error: the type parameter `A7_B` is not constrained by the impl trait, self type, or predicates
  --> scratch/map.rs:15:6
   |
15 | impl <B, I, F> T for core::iter::Map<I, F>
   |      ^^^^
```
even though `B` appears in the bound for `F`.  I think we need to fix this if we want to provide a specification for the [`std::iter::Map`](https://doc.rust-lang.org/std/iter/struct.Map.html) iterator adaptor.

```
use vstd::prelude::*;

verus! {
#[verifier::external_body]
#[verifier::external_type_specification]
#[verifier::reject_recursive_types(I)]
#[verifier::reject_recursive_types(F)]
pub struct ExMap<I, F>(core::iter::Map<I, F>);

trait T: core::iter::Iterator {
    spec fn test() -> bool;
}

impl <B, I, F> T for core::iter::Map<I, F>
 where
     I: Iterator,
     F: FnMut(I::Item) -> B,
{
    spec fn test() -> bool {
        true
    }
}
}
```

Chris-Hawblitzel:
The repro listed above should be fixed by https://github.com/verus-lang/verus/pull/2259 , https://github.com/verus-lang/verus/commit/7a0c75f901bb5501be03bcf05a01288e08b0ebc7 , https://github.com/verus-lang/verus/commit/b9d49b15402379e4b15750d50552e84394da82d2 , and https://github.com/verus-lang/verus/commit/1a0d1ffdfb01148ed0e64afad99461405c4986ab .

parno:
Amazing, thanks for working on this!

PR #2259, opened 2026-03-21
Treat Copy, FnOnce, FnMut, Fn, and Tuple as first-class traits rather than ignoring them in rust_to_vir_base

Chris-Hawblitzel:
This pull request supports Copy, FnOnce, FnMut, Fn, and Tuple as first-class traits (no longer ignored by rust_to_vir_base).  It also fixes Clone for tuples as part of supporting Copy, Clone, and Tuple for tuples.

The only remaining traits being ignored by rust_to_vir_base are the auto traits Send, Sync, and Unpin.  These will need special computation, particularly since their computation can depend on other modules' private fields (see https://lang-team.rust-lang.org/design_notes/auto_traits.html ), which Verus tries to prune away, so I prefer to wait to make a separate pull request for auto traits.

These changes pass the verita tests, given a couple rlimit/spinoff_prover tweaks, except for an intermittent failure in one vest-dsl-tls function that can be mitigated with an arbitrary lemma call like "vstd::slice::axiom_slice_ext_equal(&[true], &[true]);".  The SMT time changes by at most a few percent for performance tests (mimalloc, capybara, vest-dsl-tls).

These changes are part of a long term plan, but in the short term they are also needed as part of fixing the blocking issue https://github.com/verus-lang/verus/issues/2236 .

### DIFF
--- a/tools/verita/run_configuration_all.toml
+++ b/tools/verita/run_configuration_all.toml
@@ -78,7 +78,7 @@ name = "human-eval"
 #git_url = "https://github.com/secure-foundations/human-eval-verus.git"
 #refspec = "main"
 git_url = "https://github.com/Chris-Hawblitzel/human-eval-verus.git"
-refspec = "rlimit"
+refspec = "origin/rlimit"
 crate_roots = ["tasks/lib.rs"]
 extra_verus_args = ["--crate-type=lib"]
 


tjhance (approved):
nice timing, this helps me fix an issue I was having with FnMut.

PR #2262, opened 2026-03-23
vstd: add some documentation to tracked methods on map

bsdinis:

### DIFF
--- a/source/vstd/map.rs
+++ b/source/vstd/map.rs
@@ -108,17 +108,26 @@ impl<K, V> Map<K, V> {
         self.dom().len()
     }
 
+    /// Create an empty tracked map.
+    ///
+    /// This allows us to create a map, which we know is empty, that is _tracked_.
     pub axiom fn tracked_empty() -> (tracked out_v: Self)
         ensures
             out_v == Map::<K, V>::empty(),
     ;
 
+    /// Inserts the given `(key, tracked value)` pair into the map.
+    ///
+    /// If the key is already present from the map, then its existing value is overwritten
+    /// by the new value.
     pub axiom fn tracked_insert(tracked &mut self, key: K, tracked value: V)
         ensures
             *self == Map::insert(*old(self), key, value),
     ;
 
-    /// todo fill in documentation
+    /// Removes the given key and its associated _tracked_ value from the map.
+    ///
+    /// The key must exist in the map
     pub axiom fn tracked_remove(tracked &mut self, key: K) -> (tracked v: V)
         requires
             old(self).dom().contains(key),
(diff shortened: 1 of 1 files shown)


parno (commented):
Thanks for expanding the documentation.  I'd suggest that we try to mirror the documentation for the untracked varieties, with modifications to account for the `tracked` aspect.  E.g, the existing docs for `insert` say:
```
    /// Inserts the given (key, value) pair into the map.
    ///
    /// If the key is already present in the map, then its existing value is overwritten
    /// by the new value.
```

For a complex operation like `tracked_map_keys`, I think we could use a bit more text to explain (at a high-level) what it's doing, perhaps even with an example.  Same for `tracked_remove_keys`.

bsdinis:
that sounds good! done

source/vstd/map.rs
```diff
@@ -127,13 +136,18 @@ impl<K, V> Map<K, V> {
             v == old(self)[key],
     ;
 
+    /// Index into a tracked map, getting a tracked borrow to the value
```
parno: Very minor, but I think "borrow **of** the value" sounds a bit better

source/vstd/map.rs
```diff
@@ -127,13 +136,18 @@ impl<K, V> Map<K, V> {
...
         requires
             self.dom().contains(key),
         ensures
             *v === self.index(key),
     ;
 
+    /// Change the keys of a map, by lookup in a different map.
+    ///
+    /// For all the `(key, value)` pairs in the old map,
+    /// the new map will have `(key, key_map[key])`
```
parno: I don't think this last bit is quite right.  I think we want something like this, no?

> For each `(old_key, new_key)` pair in `key_map`, the new map will have `(new_key, old_map[old_key])`.  Note the new map may be smaller than the old map if the `key_map` omits mappings for some of the old keys.

source/vstd/map.rs
```diff
@@ -154,6 +168,9 @@ impl<K, V> Map<K, V> {
                 ) == old_map.index(key_map.index(j)),
     ;
 
+    /// Extract a set of keys out of the map.
```
parno: How about this?

> Extract a set of keys (and their corresponding values) from the map.

parno (approved):
Thanks for the updates.  I added a few more suggestions, and then I think it'll be good to go.

source/vstd/map.rs
bsdinis: This sounds great! Just updated it

source/vstd/map.rs
bsdinis: Sounds great!

bsdinis:
Just noticed I left a bunch of these out. Fixed!

parno (approved):
Looks good to me.

PR #2264, opened 2026-03-23
vstd: add tracked Set methods

bsdinis:


tjhance:
What do you want this for? I'm somewhat dubious it will be easier to use than Map

tjhance:
I was skeptical about tracked seq too, though we eventually added tracked seq functions because so many people clamored for them. But I'm even more skeptical about tracked set.

bsdinis:
I had some use-case that I think I moved on from. Happy to drop this, thought it might be useful somehow.
Out of curiosity, why would this not be useful?

tjhance:
When manipulating maps or seqs of ghost objects, being able to index into the collection with some elementary datatype is crucial to being able to use them well, in my experience. I think it would be very difficult to write meaningful predicates about sets of PointsTo objects or other ghost objects. If you have an idea on how you'd like to use them, I'd be happy to see it.

PR #2266, opened 2026-03-23
vstd: add set equality lemma based on injective_on

bsdinis:
Also add a subset lemma for injective_on.

This allows more restrictive uses. With adequate broadcast proofs (e.g., `injective ==> injective_on(Set::full())`) we could remove direct references to the full injectivity from `vstd`'s set functions.

### DIFF
--- a/source/vstd/relations.rs
+++ b/source/vstd/relations.rs
@@ -143,4 +143,21 @@ pub proof fn lemma_new_first_element_still_sorted_by<T>(
     }
 }
 
+/// If a function is injective on a set `s2`, then it is also injective on any subset `s1` of `s2`.
+pub proof fn lemma_injective_on_subset<X, Y>(r: spec_fn(X) -> Y, s1: Set<X>, s2: Set<X>)
+    requires
+        s1 <= s2,
+        injective_on(r, s2),
+    ensures
+        injective_on(r, s1),
+{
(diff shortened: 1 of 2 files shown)


source/vstd/relations.rs
```diff
@@ -143,4 +143,21 @@ pub proof fn lemma_new_first_element_still_sorted_by<T>(
     }
 }
 
+/// If a function is injective on some superset, then it is also injective in the subset
```
parno: What about tweaking the comment to be the following?

> If a function is injective on a set `s2`, then it is also injective on any subset `s1` of `s2`

source/vstd/set_lib.rs
```diff
@@ -958,6 +958,28 @@ pub proof fn lemma_sets_eq_iff_injective_map_eq<T, S>(s1: Set<T>, s2: Set<T>, f:
     }
 }
 
+/// Two sets are equal iff mapping `f` results in equal sets, if `f` is injective on their union.
```
parno: What about

> Two sets are equal iff applying an injective function `f` to each set produces equal sets.

?

bsdinis: Changed to this with a minor tweak: the function is known to be injective on `s1 + s2`, so added a parenthetical claiming it's "injective (in the union of the sets)"

PR #2268, opened 2026-03-24
refactor: replace `custom_req_err` and `custom_err` with `proof_note_custom_err`

mkovaxx:
- Replace internal uses of `#[verifier::custom_req_err("msg")]` and `#[verifier::custom_err("msg")]`.
- Make a small adjustment to ensure output remains unaltered.
- Update an example.

### DIFF
--- a/source/vir/src/sst_to_air_func.rs
+++ b/source/vir/src/sst_to_air_func.rs
@@ -480,7 +480,12 @@ fn req_ens_to_air(
                 }
             }
             let mut labels: Vec<ArcDynMessageLabel> = Vec::new();
-            if let Some(msg) = msg {
+            let proof_note = sst_exp_get_proof_note(exp);
+            let has_custom_err_label =
+                proof_note.as_ref().map_or(false, |label| label.is_custom_err);
+            if let Some(msg) = msg
+                && !has_custom_err_label
+            {
                 labels.push(Arc::new(MessageLabel {
                     span: exp.span.clone(),
                     note: msg.clone(),
@@ -488,7 +493,7 @@ fn req_ens_to_air(
                     is_custom_err: false,
                 }));
             }
-            if let Some(label) = sst_exp_get_proof_note(exp) {
+            if let Some(label) = proof_note {
                 labels.push(Arc::new(MessageLabel {
                     span: exp.span.clone(),
                     note: label.text.to_string(),


mkovaxx:
@Chris-Hawblitzel @parno Please sanity-check that the intent of the original use sites is preserved.

source/state_machines_macros/src/to_relation.rs
```diff
@@ -466,11 +466,11 @@ fn prepend_let_stmt(
 fn prepend_conjunct(e: &Expr, p: Option<TokenStream>, msg: &str) -> Option<TokenStream> {
     let msg_lit = Lit::Str(LitStr::new(msg, e.span()));
     let err_attr = quote_spanned! { e.span() =>
-        #[verifier(custom_err(#msg_lit))] /* vattr */
+        #[verifier::proof_note(#msg_lit)] /* vattr */
     };
     match p {
         None => Some(quote_spanned! { e.span() => #err_attr (#e) }),
-        Some(r) => Some(quote_spanned! { e.span() => (#err_attr (#e) && #r) }),
+        Some(r) => Some(quote_spanned! { e.span() => #err_attr ((#e) && #r) }),
```
tjhance: it looks like the meaning is being changed here (applying the attribute to the whole expression vs just the first conjunct)

parno:
I agree with Travis: I think our goal should be to match the current behavior/output.  I think this means that `proof_note` will need some extra syntax for cases like the ones above where we want to modify the entire error report, not just add a `note: ...`

mkovaxx:
@tjhance @parno Thanks for the review, I agree with the points you raised. I'll put this PR on hold and will come back to it later, after making the necessary extensions to `proof_note` in a different PR.

mkovaxx:
> I'll put this PR on hold and will come back to it later, after making the necessary extensions to `proof_note` in a different PR.

PR to implement those extensions: https://github.com/verus-lang/verus/pull/2286

source/state_machines_macros/src/to_relation.rs
mkovaxx: Opened #2311 to carefully confirm the behavior here.

source/vstd/pervasive.rs
```diff
@@ -18,9 +18,9 @@ pub proof fn assume(b: bool)
 }
 
 // TODO: remove this
-#[verifier(custom_req_err("assertion failure"))]
 pub proof fn assert(b: bool)
     requires
+        #![verifier::proof_note_custom_err("assertion failure")]
```
sunmy2019, line 23: A question, why we need a `!` here?

mkovaxx: @sunmy2019 Good question. The `!` makes that an inner (as opposed to an outer) attribute.

> *Inner attributes*, written with a bang (`!`) after the hash (`#`), apply to the form that the attribute is declared within.

You can find more detail here: https://doc.rust-lang.org/reference/attributes.html

On a `requires`/`ensures` clause, we want to attach the `proof_note`/`proof_note_custom_err` to the entire expression (as opposed to a subexpression), so we make it an inner attribute.

In the following example, the attribute attaches to the expression `x < y`:
```rust
requires
    #![verifier::proof_note("label")]
    x < y,
```

On the other hand, in the following the attribute attaches to only the subexpression `x`:
```rust
requires
    #[verifier::proof_note("label")]
    x < y,
```

Even though the outer attribute would also work on most use sites in this change, we always use the inner attribute for consistency.

mkovaxx:
@parno @Chris-Hawblitzel @tjhance #2311 was merged so we can now continue with this migration.

Quick summary: the migration has 4 steps:
1. [[done]](https://github.com/verus-lang/verus/pull/2286) Introduce the `proof_note_custom_err` attribute.
2. **[this PR] Replace `custom_req_err` and `custom_err` with `proof_note_custom_err`.**
4. Remove code paths that serve the `custom_err` and `custom_req_err` attributes.
5. Rename `proof_note_custom_err` to simply `custom_err`, finalizing the migration.

parno:
This generally looks reasonable to me.  Did you confirm that the output on the state machines example Travis showed above now matches the original output, both in terms of the error message and the reported span?

mkovaxx:
> Did you confirm that the output on the state machines example Travis showed above now matches the original output, both in terms of the error message and the reported span?

@parno @tjhance I couldn't find `tclos.rs`, but injected an error into `examples/state_machines/tutorial/ref_cell.rs` as shown in Travis's example (commenting out `assert_multisets_equal!(...)`), and that does produce the expected custom error message:

```
error: could not show invariant `flag_inv` on the `post` state
   --> ../examples/state_machines/tutorial/ref_cell.rs:113:5
    |
 26 | /  tokenized_state_machine!(RefCounter<S> {
 27 | |      fields {
 28 | |          #[sharding(constant)]
 29 | |          pub pcell_loc: CellId,
...   |
113 | |/     #[inductive(do_withdraw)]
114 | ||     fn do_withdraw_inductive(pre: Self, post: Self) {
115 | ||         //assert_multisets_equal!(post.reader, Multiset::<Perm<S>>::empty());
116 | ||     }
    | ||_____^
...   |
154 | |  });
    | |___- failed precondition

verification results:: 17 verified, 1 errors
```

The error still mentions a precondition. I'll look into changing that as well.

**UPDATE**

Made a small change and now the output is identical:

```
error: could not show invariant `flag_inv` on the `post` state
   --> ../examples/state_machines/tutorial/ref_cell.rs:113:5
    |
113 | /     #[inductive(do_withdraw)]
114 | |     fn do_withdraw_inductive(pre: Self, post: Self) {
115 | |         //assert_multisets_equal!(post.reader, Multiset::<Perm<S>>::empty());
116 | |     }
    | |_____^

verification results:: 17 verified, 1 errors
```

source/vir/src/sst_to_air_func.rs
```diff
@@ -480,15 +480,20 @@ fn req_ens_to_air(
...
+            {
                 labels.push(Arc::new(MessageLabel {
                     span: exp.span.clone(),
                     note: msg.clone(),
                     is_proof_note: false,
                     is_custom_err: false,
                 }));
             }
-            if let Some(label) = sst_exp_get_proof_note(exp) {
+            if let Some(label) = proof_note {
```
mkovaxx, line 496: This small change ensures that the `proof_note_custom_err` is reported the same way as `custom_err` and `custom_req_err` were, without any additional notes ("failed precondition", etc.).

parno: Great, thanks for making that adjustment!

PR #2271, opened 2026-03-25
[guide] docs: document proof_note attributes

mkovaxx:

### DIFF
--- a/source/docs/guide/src/reference-attributes.md
+++ b/source/docs/guide/src/reference-attributes.md
@@ -143,7 +143,7 @@ These attributes attach a string note to a `requires`/`ensures` clause or `assum
 - The outer attribute `#[verifier::proof_note]` must attach to an `assume`/`assert` statement.
 - The inner attribute `#![verifier::proof_note]` (note the `!`) must attach to a `requires`/`ensures` clause.
 
-The note is included in errors for failed proof obligations, as well as in the JSON output under the key `func-details`.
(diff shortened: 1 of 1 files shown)


source/docs/guide/src/reference-attributes.md
```diff
@@ -135,6 +136,15 @@ should "memoize" the results of this function.
...
 The definition can then be revealed locally via the [`reveal` and `reveal_with_fuel` directives](./reference-reveal-hide.md).
 
+## `#[verifier::proof_note("text")]` and `#![verifier::proof_note("text")]`
+
+These attributes attach a string note to a `requires`/`ensures` clause or `assume`/`assert` statement.
+
+- The outer attribute `#[verifier::proof_note]` must attach to an `assume`/`assert` statement.
+- The inner attribute `#![verifier::proof_note]` (note the `!`) must attach to a `requires`/`ensures` clause.
+
+The note is included in errors for failed proof obligations, as well as in the JSON output under the key `func-details`.
```
parno: I think we could expand this to something like:
> If a proof obligation (requires/ensures/assert) fails, then the `"text"` of the note is included in the error message, as well as in the JSON output under the key `func-details`.  An assume flagged by the `--no-cheating` mode is treated similarly.  This can be useful for connecting informal spec requirements (say from a text description of desired properties) to obligations in the verified code.

mkovaxx: Thanks for the suggestion. It's now part of the section.

PR #2272, opened 2026-03-25
verus_verify attribute: Properly desugar forloop

### LINKED ISSUE #2270: Compiler crashes with `for` loop inside `#[verus_verify]` function
If you have a function that adds verification via attributes rather than `verus!`, then a `for` loop without a `verus_spec` annotation will cause the compiler to crash when running verification. For instance, consider the following code (note the absence of `verus!`):

```
use vstd::prelude::*;

#[verus_verify]
fn for_loop_crashes_compiler()
{
    for i in 0..10 {
    }
}
```

Verifying this will cause an error message like:
```
thread 'rustc' (37588) panicked at /rustc-dev/4a4ef493e3a1488c6e321570238084b38948f6db/compiler\rustc_middle\src\ty\generic_args.rs:54:14:
index out of bounds: the len is 0 but the index is 0
```

It would be nice if this generated a nicer error message, since Verus novices who are trying to add verification to their code piecemeal may run into it and not understand what's going on. Though, if they ask Claude Opus it may tell them; that's how I figured out the root cause of this! Claude says:

> This is a Verus compiler bug — the for loop desugars into IntoIterator::into_iter() + Iterator::next() on [Range<usize>](vscode-file://vscode-app/c:/Programs/Microsoft%20VS%20Code/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and Verus's modified rustc crashes when resolving the generic arguments for those trait impls (the 0-length generic args being indexed at 0).

ziqiaozhou:
It fixes #2270 

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -181,9 +181,8 @@ pub(crate) fn rewrite_verus_attribute(
     }
 
     // Special handling for impl blocks, add marker attribute to each method for `#[verus_spec]`.
-    let mut impl_item_replacer = ImplItemReplacer { verify_const: true };
-    impl_item_replacer.visit_item_mut(&mut item);
-
+    let mut visitor = VerusVerifyVisitor { verify_const: true };
+    visitor.visit_item_mut(&mut item);
     let mut new_stream = quote_spanned! {item.span()=>
         #(#attributes)*
         #item
@@ -303,7 +302,7 @@ impl VisitMut for ExecReplacer {
     }
 }
 
-struct ImplItemReplacer {
+struct VerusVerifyVisitor {
     verify_const: bool,
 }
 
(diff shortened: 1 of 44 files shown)


source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -282,6 +282,25 @@ impl VisitMut for ExecReplacer {
...
+
+    fn visit_expr_for_loop_mut(&mut self, for_loop: &mut syn::ExprForLoop) {
+        syn::visit_mut::visit_expr_for_loop_mut(self, for_loop);
+
+        if !self.erase.keep() {
+            return;
+        }
+
+        // In verification mode, even without verus spec on the loop, we still
+        // need to desugar the forloop.
```
jaylorch: I suggest a space between `for` and `loop`.

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -282,6 +282,25 @@ impl VisitMut for ExecReplacer {
...
+        // Thus, add empty verus_spec if not exists
```
jaylorch: For greater clarity, I suggest rewording this line to:
```
// So, if there's no `verus_spec` attribute, we need to add an empty one.
```

Issue #1938, opened 2025-11-01
`cargo verus` passes arguments like `--verify-module` to dependencies too


jaybosamiya-ms:
`cargo verus`'s implementation seems to pass along the parameters after the `--` to all dependencies (including vstd), so something like `cargo verus verify -- --verify-module bar` for the following files will fail on the vstd build.

Cargo.toml
```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"
[dependencies]
vstd = "=0.0.0-2025-08-12-1837"
[package.metadata.verus]
verify = true
```

src/lib.rs
```rs
use vstd::prelude::*;

verus! {

mod bar {
    use vstd::prelude::*;

    fn baz() {
        assert(1 == 0 + 1);
    }

}

} // verus!
```

The workaround for this is fairly straightforward, just run `cargo verus verify` first (so that vstd finishes; although you'll have to delete `target/` first to clear out the cache), followed by `cargo verus verify -- --verify-module bar`, which then works as expected.

However, this workaround works fine only for single-crate situations.  If we are in a multi-crate workspace (e.g., https://github.com/verus-lang/verus-mode.el/tree/main/verus-examples/cv-workspace), then the there is no way to restrict the `--verify-module` to just a single one of the crates in the workspace.  Thus, we cannot benefit from partial verification which can be a performance bottleneck _especially_ when working on a file (you can prob see why I hit this issue when adding support for `cargo verus` on workspaces within `verus-mode.el`).

parno:
This came up w.r.t. verus-analyzer as well.  Chris [proposed](https://verus-lang.zulipchat.com/#narrow/channel/399252-dev/topic/cargo-verus.20vs.2E.20crate-specific.20options/with/544189479) adding a `--verify-crate C` option, which would skip verification of all crates except `C`.  We could also have `cargo verus` complain if you tried to use `--verify-module` without using `--verify-crate` (with a helpful suggestion, of course).  One subtlety is that we'd have to remember that we skipped verification, so that subsequent uses of `cargo verus` without `--verify-crate` would go back and verify any unverified crates.

jaybosamiya-ms:
I like `--verify-crate`.

[...]

PR #2277, opened 2026-03-26
[cargo-verus] feat: option to control Verus arg forwarding

mkovaxx:
- Handle unfinished business from #1938.
- Add `--fwd-verus-args-to` option to more precisely control which crates receive forwarded Verus arguments.
- Three possible values: `all`, `roots`, and `deps`.
- The default is `all`, except for the `focus` subcommand where it's `roots`.
- Add tests to cover basic usage and defaults for all subcommands.
- Small refactor: consolidate all CLI-related logic in `cli.rs`.

### DIFF
--- a/source/cargo-verus/src/cli.rs
+++ b/source/cargo-verus/src/cli.rs
@@ -60,11 +60,12 @@ pub struct VerifyCommand {
         long,
         value_name = "SELECTOR",
         help = "\
-Crates to receive forwarded Verus args. Defaults to `roots` in `focus`, `all` otherwise.\n",
+Crates to receive forwarded Verus args. Defaults to `all`, except in `focus` mode where it defaults to `roots`. Use `deps` to pass args ONLY to dependencies and NOT to roots.\n",
         long_help = "\
 Crates to receive forwarded Verus args.
 
-Defaults to `all`, except in `focus` mode where it defaults to `roots`."
+Defaults to `all`, except in `focus` mode where it defaults to `roots`.
+Use `deps` to pass args ONLY to dependencies and NOT to roots."
     )]
     pub fwd_verus_args_to: Option<VerusArgFwdSelector>,
 


source/cargo-verus/tests/test_focus.rs
```diff
@@ -171,3 +171,54 @@ fn workspace_package_hasdeps() {
     data.assert_env_has_no_key_prefix(&verify_optout_prefix);
     data.assert_env_has_no_key_prefix(&verify_unset_prefix);
 }
+
+#[test]
+fn workspace_package_hasdeps_forwards_verus_args_only_to_roots() {
```
mkovaxx, line 176: This is the test to confirm that the issue originally raised in #1938 is solved by this change.

source/cargo-verus/src/cli.rs
```diff
@@ -153,7 +174,40 @@ fn has_late_verus_arg(opts: &CargoOptions) -> bool {
...
 
 impl CargoVerusCli {
-    pub fn clap_trailing_args_hotfix(mut self) -> Self {
+    pub fn from_args(args: impl Iterator<Item = String>) -> Result<Self> {
+        let normalized_args = normalize_args(args);
+        let mut parsed_cli = CargoVerusCli::parse_from(normalized_args).clap_trailing_args_hotfix();
+
+        if parsed_cli.has_inadvisable_verus_arg() {
+            eprintln!("Args forwarded to Cargo must precede args forwarded to Verus");
+            // TODO: Consider replacing this with `return Err(anyhow!("message above^"))`.
```
parno, line 184: What's the downside to using `return Err...` here?

source/cargo-verus/src/cli.rs
```diff
@@ -58,10 +76,13 @@ pub struct VerifyCommand {
...
+}
 
-    /// Make cargo-verus verbose
-    #[arg(short, long)]
-    pub verbose: bool,
+#[derive(Clone, Copy, Debug, ValueEnum)]
+pub enum VerusArgFwdSelector {
+    All,
+    Roots,
+    Deps,
```
parno, line 86: Is there a use-case where we want to pass args to the dependencies but not the root?  If so, should we mention the deps option in the help text?  If not, we could just cut the `Deps` constructor.

source/cargo-verus/src/cli.rs
mkovaxx: It breaks all tests in `test_late_args.rs` because they expect an exit code of 2. I'll refactor the Cargo arg forwarding next, in another PR.
https://github.com/verus-lang/verus/blob/44c4848497c38139e067673c53849c225ae2e5a7/source/cargo-verus/tests/test_late_args.rs#L40

source/cargo-verus/src/cli.rs
mkovaxx: One use-case I thought of is passing `--no-cheating` to dependencies while allowing cheating for the root crate(s) under development. Is that something that might be useful?

mkovaxx: Even though the help text overrides don't contain it, `clap` automatically injects a message about possible values.

In the long help output, i.e. `cargo-verus verify --help`:
```
      --fwd-verus-args-to <SELECTOR>
          Crates to receive forwarded Verus args.
          
          Defaults to `all`, except in `focus` mode where it defaults to `roots`.
          
          [possible values: all, roots, deps]
```

In the short help output, i.e. `cargo-verus verify -h`:
```
      --fwd-verus-args-to <SELECTOR>  Crates to receive forwarded Verus args. Defaults to `roots` in `focus`, `all` otherwise.
                                       [possible values: all, roots, deps]
```

parno: Thanks, that's helpful to see.  Just to be fully explicit, could we say something like this?
> Defaults to `all`, except in `focus` mode where it defaults to `roots`.  Use `deps` to pass args only to the dependencies and *not* the roots.

source/cargo-verus/src/cli.rs
parno: Okay, deferring to that PR is fine with me.

source/cargo-verus/src/cli.rs
mkovaxx: Yep, added that clarification to the help descriptions. :)

PR #2280, opened 2026-03-28
[cargo-verus] fix: verify bug with in-package dependencies

jsenn:
In https://github.com/verus-lang/verus/pull/2251, I attempted to fix https://github.com/verus-lang/verus/issues/1905, which is a bug where verus was unable to verify code in a lib package that used the lib. For example, if you wanted to add a test, bin, or example to your package that used the lib, verus would fail to verify the non-lib target, even if the code was correct.

The attempted fix was to pass the lib as a dependency to the rustc driver. This works if you use `cargo verus build`, or if you do a build then `cargo verus verify`. However, `cargo clean && cargo verus verify` would still fail.

The problem is in the driver itself. `cargo verus verify` works by invoking `cargo build`, but setting some special flags that tell the driver to interrupt compilation before code generation. Unfortunately, it also prevents rustc from emitting metadata. The upshot is the non-lib targets have no way of seeing the lib target's definitions, which leads to compiler errors.

The fix here its to:
1. Use `cargo check` instead of `cargo build` as the subcommand for `cargo verus (verify|focus)`
2. Get rid of the previous "limited check then short-circuit compilation" approach to verification passes. Instead, rustc is always allowed to run to completion, emitting the necessary metadata to fix the bug.

Now, applying those changes makes the test harness unhappy. The main issue was some minor linter nits (naming conventions) that never showed up before as the short-circuit behaviour skipped linting altogether. I have fixed those.

~~However, there are 2 remaining test failures, and I *think* they reveal genuine bugs in Verus's erasure logic, though I don't fully understand. It would be easy to fix the errors by declaring the unstable features in the test code, but I have avoided doing so you can see the test failures to judge if this is a real bug.~~ EDIT: This was a misunderstanding on my part. These errors just need the feature switch for `sized_hierarchy` and `const_destruct`, as they are unstable features.

Now, technically this is doing a bit more checking than previously. In particular, the previous code was running a stripped-down custom `check()` routine that did type checking (`rustc_hir_analysis::check_crate()`), then only did borrow checking on function definitions. With this change everything is checked. This seems right to me, but there could be some reason it was like this.

Another implication is that `cargo verus check` and `cargo verus verify` are exactly the same. I'm not really sure what the difference between them was *supposed* to be. From a little spelunking in the git history it looks like this was a refactor of a previous API where these were independent flags, and the distinction isn't very meaningful anymore.

I have also added an end-to-end test that exercises the lib + non-lib target case. It would have caught this bug.

### DIFF
--- a/dependencies/prettyplease/Cargo.toml
+++ b/dependencies/prettyplease/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "verus_prettyplease"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 authors = ["David Tolnay <dtolnay@gmail.com>", "The Verus Contributors"]
 autoexamples = false
 categories = ["development-tools"]
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -2650,11 +2650,11 @@ dependencies = [
 
 [[package]]
 name = "verus_builtin"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 
 [[package]]
 name = "verus_builtin_macros"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2666,7 +2666,7 @@ dependencies = [
 
 [[package]]
 name = "verus_prettyplease"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 dependencies = [
  "proc-macro2",
  "verus_syn",
--- a/source/air/src/typecheck.rs
+++ b/source/air/src/typecheck.rs
@@ -14,7 +14,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 pub(crate) type Declared = Arc<DeclaredX>;
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub(crate) enum DeclaredX {
     Type,
     Var { typ: Typ, mutable: bool },
--- a/source/builtin/Cargo.toml
+++ b/source/builtin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "verus_builtin"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 edition = "2018"
 license = "MIT"
 description = "Verus Builtins: Rust definitions for core constructs in Verus"
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -12,6 +12,7 @@
     register_tool(verifier)
 )]
 
+use core::future::Future;
 use core::marker::PhantomData;
 
 #[cfg(verus_keep_ghost)]
@@ -244,6 +245,13 @@ pub fn with_triggers<A, B>(_triggers_tuples: A, body: B) -> B {
 #[verifier::spec]
 pub fn constrain_type<T>(_x: T, _y: T) -> bool {
     true
+}
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::get_future_output_type"]
+#[verifier::spec]
+pub fn get_future_output_type<T>(_x: impl Future<Output = T>) -> T {
+    unimplemented!()
 }
 
 // example: forall with three triggers [f(x), g(y)], [h(x, y)], [m(y, x)]:
--- a/source/builtin_macros/Cargo.toml
+++ b/source/builtin_macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "verus_builtin_macros"
-version = "0.0.0-2026-04-05-0114"
+version = "0.0.0-2026-04-12-0118"
 edition = "2018"
 license = "MIT"
 description = "Verus Builtin Macros: Various macros used by Verus"
@@ -19,7 +19,7 @@ quote = "1.0"
 synstructure = "0.13.2"
 syn = { version = "2.0", features = ["full", "visit", "visit-mut", "extra-traits"] }
 verus_syn = { version = "=0.0.0-2026-04-05-0114", path = "../../dependencies/syn", features = ["full", "visit", "visit-mut", "extra-traits"] }
-verus_prettyplease = { version = "=0.0.0-2026-04-05-0114", path = "../../dependencies/prettyplease" }
+verus_prettyplease = { version = "=0.0.0-2026-04-12-0118", path = "../../dependencies/prettyplease" }
 
 [package.metadata.verus]
 is-builtin-macros = true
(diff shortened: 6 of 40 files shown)


mkovaxx:
@jsenn Thanks for flagging the issue and proposing this fix! :)

I've only made a quick glance, but this change seems to resolve #1739 as well.
Even though it isn't yet clear whether this is a complete fix, it's already valuable. 👏

jsenn:
Might be worth taking a look at the first commit by itself as well: https://github.com/verus-lang/verus/pull/2280/changes/0c9ef56e55637223a836560cc3c6547c9389415d.

This seems cleaner to me, and I feel like it *should* work, but it broke the test harness in weird ways. The current iteration keeps the old limited-check-then-short-circuit behaviour for non-cargo invocations, but that seems arbitrary.

EDIT: you can see the test failures [here](https://github.com/verus-lang/verus/actions/runs/23675199976/job/68976775447?pr=2280). When I run locally it seems like at least some of the failing tests emit warnings about snake case and such, but not sure if that’s related or why

jsenn:
heh, ok I think I see why the test harness didn't like the simpler fix. The short-circuit behaviour *also* skips the normal linting pass. The tests were failing with minor linter warnings (e.g. upper case letter for module, stuff like that).

I think I'll go back to the original simpler fix, though I'll also have to tweak the test harness to run rustc in "check mode" (emit metadata not emit link).

This avoids the "special pleading" when running via cargo

jsenn:
still lots of test failures due to snake_case and other minor naming violations. I don't want to introduce too much churn here. Maybe a more pragmatic fix would be to just suppress these linter warnings in the test harness for now?

EDIT: also a couple of tests are using a deprecated API which triggers a warning when it didn't before
EDIT2: pushed a commit that silences warnings in examples. It woudl be a fair amount of churn to fix them properly. ~There are still the 2 genuine errors that I think reflect bugs in erasure~

source/rust_verify/src/verifier.rs
```diff
@@ -3330,7 +3330,6 @@ impl rustc_driver::Callbacks for VerifierCallbacksEraseMacro {
                         // because the error messages are likely to be better.
                         let compile_status = crate::driver::run_with_erase_macro_compile(
```
parno: Does this change mean that we're going to end up doing codegen even when we're just collecting errors for lifetime issues?

examples/guide/const.rs
parno, line 1: Instead of adding all of these `#![allow(non_snake_case)]` should we consider passing `--allow non_snake_case` in the example test runner?

source/rust_verify/src/verifier.rs
jsenn: good catch. I think you're right that when verus hits this code path, the current code will always short-circuit after the custom `check()` pass, but now its behaviour depends on the rustc args (i.e. `--emit`).

However, this particular callsite is only hit (as I understand it) when something triggered a lifetime error, in which case I don't think that either metadata or code actually *can* be generated?

Even if that's true though this is potentially brittle, and there could still be a performance penalty even if the actual codegen isn't happening--it's relying on rustc's internals more than it used to.

[Another way of solving this problem](https://github.com/verus-lang/verus/compare/4191be5169af90b21bfc95a1d8e616da0b261934) is to keep the current short-circuiting behaviour, but set the `do_compile` to true whenever we invoke this via cargo. The verifier callsite would stay the same and have the same behaviour.

This would also avoid having to handle/suppress all the warnings generated from the tests and examples, as the test harness invokes the driver directly, so the special pleading for cargo verus doesn't trigger. I moved away from this idea because it felt hacky, but maybe this is the more pragmatic fix.

tjhance: if it helps, i believe this entire code path can (and probably should) be removed. The original rationale for running rust in the event of a "lifetime error" was to get better lifetime errors. However, lifetime handling has since changed. So this code path is now (despite what the comment says) only invoked for _trait conflict errors_ and there's no reason that trait conflict errors should trigger an attempt to get better lifetime errors.

examples/guide/const.rs
jsenn: that would eliminate most of the churn here. One reason to prefer per-file suppression would be to avoid accidentally introducing minor naming convention violations in *new* code that will create warnings when users copy-paste an example into their editor, but I don't know if that's worth the noise in this diff.

source/rust_verify/src/verifier.rs
jsenn: @tjhance gonna try to say this back to see if I understand. So this invokes Verus's trait conflict checker, which resynthesizes Rust source from the VIR and runs `rustc` against that. The resynthesis step is needed because Verus has its own model of rust traits, so some impls that `rustc` considers distinct could be collapsed by Verus, causing contradictory axioms to be fed into the solver (soundness bug). Originally that code also did borrow checking, but it no longer does. Presumably the error messages from the resynthesized source were confusing for users, so the second pass was to get better errors out of `rustc` for the cases where it's a regular rust error rather than a verus-specific one.

This callsite says if there was a trait conflict, run a type check + borrow check (the `check()` function I deleted in this PR). If that *succeeds*, print the trait conflict, else end compilation. But we could only get here in the first place if `rustc` accepted the code. In other words, the only error this could be is a mismatch between Verus's trait model and `rustc`'s, so there's no way the call to `run_erase_macro_compile` could possibly fail.

So to @parno's point, yes this code will now be doing redundant code generation, but what you're saying is we can fix it by removing the callsite entirely because it no longer serves a purpose.

EDIT: a genuine trait impl conflict (i.e. one that `rustc` considers a conflict, as opposed to one that's due to Verus and `rustc` getting out of sync), would be caught in the call to `rustc_hir_analysis::check_crate(tcx)` in `construct_vir_crate()`, which would stop compilation before this Verus-specific trait conflict check is run.

examples/guide/const.rs
jsenn: added snake_case and deprecation suppression in the test runner itself, so these warnings are now allowed in examples and tests (snake_case was already allowed).

Eliminates the couple dozen changes in example files.

source/rust_verify/src/verifier.rs
jsenn: pushed a commit that removes this call to rustc per above discussion

parno:
Thanks for the updates!  I'd like to run a few performance tests to check that this doesn't slowdown the code-prove-debug cycle too much, and then I think we can probably land this.

parno:
I tried running this on a largish project, using `cargo verus` to focus on verifying a specific function in the project.  I measured the time when the function was expected to succeed and after I edited it to fail, with 5 runs in each condition.  The resulting differences seem safely within the noise:
```
                 success                 fail                                                                                                                     
  ----------------------------------------------------                                                                                                            
      main    2.353 ± 0.013         1.856 ± 0.005                                                                                                                 
        pr    2.330 ± 0.014         1.848 ± 0.024                                                                                                                 
  ----------------------------------------------------   
```
so I think it's reasonable to merge this.  Would you mind doing one more merge from `main`, and then we can land it?

jsenn:
> Would you mind doing one more merge from `main`, and then we can land it?

done! Thanks for reviewing and testing

parno:
Thanks again for the contribution!

jaylorch:
Thanks! We've updated `verified-storage` to use this so it now verifies.

PR #2282, opened 2026-03-28
new-mut-ref: fix issues related to FnMut

tjhance:
~~(Depends on https://github.com/verus-lang/verus/pull/2259)~~

This fixes a few tests that would otherwise break when new-mut-ref is enabled.

Until now we've been loose about the difference between `call_requires(f, ...)`, `call_requires(&f, ...)` and `call_requires(&mut f, ...)`. This has been ok so far because `call_requires` actually ignores the decoration on the first arg. But `&mut T` will soon become a non-decoration type. We need to specify the relationship between `call_requires(&mut f, ...)` and `call_requires(f, ...)`. This is easy to do since we still don't actually support closures that capture mutably.


tjhance:
I realized I needed to fix a few more things ...

In order for this to work everywhere, we need to emit an axiom for the bound `closure_type: FnMut`. So I added that (in ast_simplify, when creating the Datatype that represents the closure type).

Then to make _that_ work, I needed to make the closure type generic, rather than pretending the type parameters are bound outside the definition of the type.

PR #2284, opened 2026-03-29
Add #[verifier::allow(...)] to suppress warnings

Chris-Hawblitzel:
This adds a `#[verifier::allow(...)]` attribute to suppress warnings generated by Verus.  For now, the attribute is only implemented for items like functions and modules, not for expressions, so you have to suppress warnings across entire functions or modules.

Example:

```
#[derive(Clone)]
struct S;
```

This generates:

```
warning: Verus does not (yet) support autoderive Clone impl when the clone is not a copy; continuing, but without adding a specification for the derived Clone impl
  --> .\scratch.rs:55:10
   |
55 | #[derive(Clone)]
   |          ^^^^^
   |
   = help: to suppress this warning, use `#[verifier::allow(autoderive_clone_without_spec)]` on the surrounding function, datatype, or module or `#![verifier::allow(autoderive_clone_without_spec)]` in the module or crate
```

Rewriting the code as:

```
#[verifier::allow(autoderive_clone_without_spec)]
#[derive(Clone)]
struct S;
```

removes the warning.

The implementation works differently for rust_verify and vir.  In rust_verify, for each warning, the implementation walks up the DefId tree on demand to look for matching `verifier::allow` attributes.  For vir, the implementation preemptively walks up the DefId tree from each function during translation to vir, keeping all the `verifier::allow` attributes in a table for later reference, and then uses the table to handle any warnings generated inside vir.


source/rust_verify/src/external.rs
```diff
@@ -652,12 +652,14 @@ fn get_attributes_for_automatic_derive<'tcx>(
...
+        crate::attributes::warning_maybe(
+            ctxt.tcx,
+            general_item.id().owner_id().to_def_id(),
             span,
-            format!(
-                "Verus doesn't known how to handle this automatically derived item; ignoring it"
-            ),
-        )));
+            "unknown_automatic_derive",
+            || "Verus doesn't known how to handle this automatically derived item; ignoring it",
```
parno: "known" -> "know"

source/vir/src/interpreter.rs
```diff
@@ -2042,11 +2094,12 @@ fn eval_expr_launch(
                 let res = cleanup_exp(&res)?;
                 // Send partial result to Z3
                 if exp.definitely_eq(&res) {
-                    let msg = format!(
-                        "Failed to simplify expression <<{}>> before sending to Z3",
-                        exp.x.to_user_string(&ctx.global)
+                    ctx.warning(
+                        &exp.span,
+                        "assert_compute_unsimplified",
+                        || "Failed to simplify expression <<{}>> before sending to Z3",
```
parno: I think we might need a `format!(...)` here, so that the "<<{}>>" gets resolved properly

parno (approved):
Thanks, this looks like an elegant framework.  It'll be nice to silence some of the warnings that have persistently cluttered some of my projects for a while now.

source/vir/src/context.rs
```diff
@@ -35,6 +35,23 @@ pub struct ChosenTriggers {
     pub manual: bool,
 }
 
+#[derive(Debug, Clone)]
+pub struct WarningConfig(pub Vec<String>);
```
tjhance: should we use an enum instead of String for the warning names?

Chris-Hawblitzel: Ok, I switched it to an enum.

PR #2286, opened 2026-03-30
feat: `proof_note_custom_err` attribute

mkovaxx:
- Extend internal `proof_note` machinery with functionality to match `custom_err` and `custom_req_err`.
- Add support for a new attribute: `proof_note_custom_err`.
- Exclude new `proof_note_custom_err` labels from the JSON output.
- Add tests.

### DIFF
--- a/source/rust_verify_test/tests/basic.rs
+++ b/source/rust_verify_test/tests/basic.rs
@@ -188,6 +188,19 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] test_multiple_proof_notes_on_requires verus_code! {
+        fn example(x: u64, y: u64) -> (z: u64)
+            requires
+                #![verifier::proof_note("Property 732")]
+                #![verifier::proof_note("Property 451")]
+                x == y,
+        {
+            x + y
+        }
+    } => Err(err) => assert_vir_error_msg(err, "at most one `proof_note` attribute is allowed")
+}
+
+test_verify_one_file! {
     #[test] test_proof_note_on_ensures verus_code! {
         fn example(x: u64, y: u64) -> (z: u64)
             ensures
@@ -204,12 +217,35 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] test_multiple_proof_notes_on_ensures verus_code! {
+        fn example(x: u64, y: u64) -> (z: u64)
+            ensures
+                #![verifier::proof_note("Property 732")]
+                #![verifier::proof_note("Property 451")]
+                z == x + y,
+        {
+            x
+        }
+    } => Err(err) => assert_vir_error_msg(err, "at most one `proof_note` attribute is allowed")
+}
+
+test_verify_one_file! {
     #[test] test_proof_note_on_assert verus_code! {
         fn caller() {
             #[verifier::proof_note("Statement known to be false")]
             assert(1 > 2); // assertion fails
         }
     } => Err(err) => assert_help_error_msg(err, "note: Statement known to be false")
+}
+
+test_verify_one_file! {
+    #[test] test_multiple_proof_notes_on_assert verus_code! {
+        fn caller() {
+            #[verifier::proof_note("Statement known to be false")]
+            #[verifier::proof_note("This is another proof label")]
+            assert(1 > 2); // assertion fails
+        }
+    } => Err(err) => assert_vir_error_msg(err, "at most one `proof_note` attribute is allowed")
 }
 
 test_verify_one_file_with_options! {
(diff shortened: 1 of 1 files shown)


mkovaxx:
@parno @Chris-Hawblitzel This is step 1 of merging `custom_err` and `custom_req_err` functionality into `proof_note`.

The migration plan has 4 steps:
1. (this PR) Introduce the `proof_note_custom_err` attribute.
2. Resume #2268
4. Remove code paths that serve the `custom_err` and `custom_req_err` attributes.
5. Rename `proof_note_custom_err` to simply `custom_err`, finalizing the migration.

Please let me know if that plan makes sense, or if you think something is missing.

Also, please review the tests added in this change to confirm `proof_note_custom_err` semantics.

mkovaxx:
> What does the implementation do if you mark multiple requires clauses with `proof_note_custom_err`

@parno Good question, thanks for raising it. The general `proof_note` machinery rejects multiple attributes. That covers `proof_note` and `proof_note_custom_err` together, in any combination. At most one such attribute is allowed on each `requires`/`ensures` clause or `assert`/`assume` statement.

I've added a few tests to document that.

parno:
Looking at the tests, I think you answered a slightly different question.  I'm curious what happens in an example like this:
```
fn example(x: u64, y: u64) -> (z: u64)
    requires
        #![verifier::proof_note_custom_err("x req")]
        x < 10,
        #![verifier::proof_note_custom_err("y req")]
        y < 100,
{
    x + y
}

fn test() {
   example(500, 500);
}
```
In other words, each requires clause has its own custom error message and both are violated at the same time.

mkovaxx:
> Looking at the tests, I think you answered a slightly different question. I'm curious what happens in an example like this:

@parno You're right, I misunderstood the question. Thanks for the clarification.

In terms of which failed obligations are reported, the `proof_note_custom_err` and `proof_note` attributes should have no effect.

In the example you gave, only the first error is reported, regardless of whether a label attribute is present.

I.e. the following test fails:
```rust
test_verify_one_file! {
    #[test] test_proof_note_custom_err_on_multiple_requires verus_code! {
        fn example(x: u64, y: u64) -> (z: u64)
            requires
                #![verifier::proof_note_custom_err("x req")]
                x < 10,
                #![verifier::proof_note_custom_err("y req")]
                y < 100,
        {
            x + y
        }

        fn test() {
           example(500, 500);
        }
    } => Err(err) => assert_vir_error_msgs(err, &["x req", "y req"])
}
```

Relaxing the test to expect only the first error makes it pass:
```rust
test_verify_one_file! {
    #[test] test_proof_note_custom_err_on_multiple_requires verus_code! {
        // fn example same as above

        // fn test same as above
    } => Err(err) => assert_vir_error_msg(err, "x req")
}
```

Only one precondition failure is reported, despite the fact that the test harness uses `--multiple-errors 2` by default.

Another way to make the test pass is to have two call sites:
```rust
test_verify_one_file! {
    #[test] test_proof_note_custom_err_on_multiple_requires verus_code! {
        // fn example same as above

        fn test() {
           example(500, 0);
           example(0, 500);
        }
    } => Err(err) => assert_vir_error_msgs(err, &["x req", "y req"])
}
```

Having two call sites results in the precondition failures reported separately.

For postconditions, a single use-site is enough to report both failures:
```rust
test_verify_one_file! {
    #[test] test_proof_note_custom_err_on_multiple_ensures verus_code! {
        fn example(x: u64, y: u64) -> (z: u64)
            ensures
                #![verifier::proof_note_custom_err("x ens")]
                x < 10,
                #![verifier::proof_note_custom_err("y ens")]
                y < 100,
        {
            x + y
        }

        fn test() {
            example(500, 500);
        }
    } => Err(err) => assert_vir_error_msgs(err, &[
        "x ens",
        "y ens",
        "possible arithmetic underflow/overflow",
    ])
}
```

Strangely enough, the `ensures` and the two-callsite `requires` test also print the following, but not the single-site `requires` test:
```
note: function body check: not all errors may have been reported; rerun with a higher value for --multiple-errors to find other potential errors in this function
```

Do any of those results seem incorrect?

parno:
I'm a bit surprised that the `test_proof_note_custom_err_on_multiple_requires` test only reports the first precondition, even when the test passes `--multiple-errors 2`.  @Chris-Hawblitzel is that expected?

As a side note, in the `test_proof_note_custom_err_on_multiple_ensures` test, I think you should see those errors even without the `test()` function.

Chris-Hawblitzel:
> I'm a bit surprised that the `test_proof_note_custom_err_on_multiple_requires` test only reports the first precondition, even when the test passes `--multiple-errors 2`. @Chris-Hawblitzel is that expected?

It is expected.  You can use --expand-errors to get multiple precondition failures (at least if the second precondition can still fail even after the first one is assumed to have succeeded, which is not the case in this example).

PR #2287, opened 2026-03-30
add Send and Sync impl for Ghost

elanortang:
This PR
- Adds `Send` and `Sync` impls for any `Ghost` struct, since it's "spec-mode" and not checked by the borrow-checker.
- Adds `not(verus_verify_core)` condition on `no_std` attribute to fix an error in the Rust standard library verification.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -1,4 +1,4 @@
-#![no_std]
+#![cfg_attr(not(verus_verify_core), no_std)]
 #![allow(internal_features)]
 #![cfg_attr(
     verus_keep_ghost,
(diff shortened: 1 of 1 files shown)


source/builtin/src/lib.rs
```diff
@@ -601,6 +601,12 @@ impl<A: Copy> Clone for Tracked<A> {
 
 impl<A: Copy> Copy for Tracked<A> {}
 
+/// `Ghost` structs are always `Send`, since they are spec mode.
+unsafe impl<A> Send for Ghost<A> {}
+
+/// `Ghost` structs are always `Sync`, since they are spec mode.
+unsafe impl<A> Sync for Ghost<A> {}
```
elanortang, line 608: I noticed that some of the other trait impls for `Ghost` and `Tracked` also had a `verus_keep_ghost` cfg - do we need that here too?

Chris-Hawblitzel: You don't need a cfg here.  Ghost will always be Send and Sync no matter whether we're verifying code or just type checking and compiling it.

Issue #2275, opened 2026-03-25
Issue with calling `use_type_invariant` when defined in another crate


elanortang:
### Problem
If you try to call `use_type_invariant` on a `struct` defined in another crate, Verus cannot find the corresponding `inv` function in the context. (It has been mistakenly pruned away.) Adding in any call to the `inv` function resolves the error, since Verus now realizes that it shouldn't be pruned.

#### Files
foo.rs
```rust
use vstd::prelude::*;

verus! {

pub struct Foo {
    n: usize,
}

impl Foo {
    #[verifier::type_invariant]
    pub open spec fn inv(self) -> bool {
        self.value() == 0
    }

    pub closed spec fn value(self) -> usize {
        self.n
    }
}

}
```

bar.rs
```rust
use vstd::prelude::*;
use foo::Foo;

verus! {

fn test(f: Foo) {
    proof {
        // FAILS with "Verus Internal Error: missing type invariant function"
        use_type_invariant(f);
        // Uncommenting the following line fixes it
        // assert(f.inv());
    }
}

}
```

#### How to Run
To trigger the error (assuming foo.rs and bar.rs are in the same directory), run
```
verus --crate-type=lib foo.rs --export foo.vir --compile
verus --crate-type=lib bar.rs --import foo=foo.vir --extern foo=libfoo.rlib
```

Error produced:
```
error: Verus Internal Error: missing type invariant function
 --> bar.rs:9:28
  |
9 |         use_type_invariant(f);
  |                            ^

error: aborting due to 1 previous error
```

tjhance:
~~error message should be changed to say that `f` isn't visible at that program point. Issue is the function was pruned, so we can't look up the visibility in the first place, hence the internal error. We'll need to store the visibility separately, so we can do the visibility check without having to look up a possibly-pruned-away item.~~ edit: never mind, looks like the function is public in your example. your diagnosis is probably correct

elanortang:
(As a separate item) I did find the error message confusing, and I think it would be useful to make that more clear - before I dug into the Verus code, I thought it meant that the struct didn't have a type invariant function on it.

tjhance:
It seems there are 2 related issues. In this one, the problem is that the function is being removed in prune.rs. 

There's also the related https://github.com/verus-lang/verus/issues/2171, which I'm still pretty sure is the thing I said above, the function is private so not being exported at all.

PR #2289, opened 2026-03-30
make type inv error more verbose

elanortang:
Instead of the error "missing type invariant function" (which I incorrectly interpreted as meaning that the datatype itself doesn't have a type invariant function), make it clear that you hit this error when the type invariant fn is not in scope.

### DIFF
--- a/source/vir/src/user_defined_type_invariants.rs
+++ b/source/vir/src/user_defined_type_invariants.rs
@@ -140,7 +140,10 @@ pub(crate) fn annotate_one(
             {
                 let fun = typ_get_user_defined_type_invariant(datatypes, &expr.typ).unwrap();
                 let Some(function) = functions.get(&fun) else {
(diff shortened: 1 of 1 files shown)


tjhance:
to give a correct user-facing error message, we need to know *why* the function is not in the map. As I write in https://github.com/verus-lang/verus/issues/2275, there are multiple potential causes. This message is the result of a bug, which is why it's an internal_error.

elanortang:
> to give a correct user-facing error message, we need to know _why_ the function is not in the map. As I write in #2275, there are multiple potential causes. This message is the result of a bug, which is why it's an internal_error.

I was thinking it would just be more helpful to be more clear in the error message _that_ the function is not in the map, until we can figure out why. I'm not very attached to my current phrasing, so if you think there's something more accurate I'd be happy to change it.

PR #2291, opened 2026-03-31
Account for implicit pointer coercion by Rust

### LINKED ISSUE #2290: verus internal error: codegen_select_candidate failed
The following program causes the error `verus internal error: codegen_select_candidate failed`:

```
verus! {

fn compare_pointers(p1: *const u32, p2: *mut u32) -> bool
{
    p1 < p2
}

}
```

jaylorch:
If Rust adds an implicit pointer coercion, e.g., from `*mut u32` to `*const u32`, account for that coercion during verification.

Fixes #2290

This code was written in collaboration with Github Copilot + Claude Opus 4.6.


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1766,8 +1783,11 @@ fn operator_overload_to_vir<'tcx>(
         let (fun_sym, Some(trait_id)) = lang_item_for_op(tcx, op, span)? else {
             crate::internal_err!(span, "operator needs an accessible trait");
         };
-        let lhs_ty = bctx.types.node_type(lhs.hir_id);
-        let rhs_ty = bctx.types.node_type(rhs.hir_id);
+        // When constructing the substs for trait resolution, we need to account
+        // for any coercions applied to the operands (e.g., *mut T -> *const T).
+        // Apply non-Borrow adjustments to get the types used for trait resolution.
```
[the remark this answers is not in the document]
jaylorch: Funny. I was making that change as you were generating this review. Done!

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1631,6 +1631,23 @@ enum OpKind {
...
+) -> rustc_middle::ty::Ty<'tcx> {
+    let mut ty = types.node_type(operand.hir_id);
+    for adj in types.expr_adjustments(operand) {
+        match adj.kind {
+            rustc_middle::ty::adjustment::Adjust::Pointer(_) => ty = adj.target,
+            _ => {}
+        }
+    }
+    ty
+}
```
tjhance, line 1690: you don't need to reimplement `expr_ty_adjusted`

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: this one should also be `expr_ty_adjusted`, can you fix it as well?

source/rust_verify_test/tests/raw_ptrs.rs
```diff
@@ -330,6 +330,24 @@ test_verify_one_file! {
...
+        pub assume_specification<T: core::marker::PointeeSized>[ <*const T as core::cmp::PartialOrd>::lt ] (
+            p: &*const T,
+            q: &*const T,
+        ) -> (result: bool)
+        ensures
+            result == ((*p as usize) < (*q as usize)),
+        ;
+
+        fn compare_pointers(p1: *const u32, p2: *mut u32) -> bool {
+            p1 < p2
```
tjhance, line 348: can you test the semantics of the operation as well?

tjhance (changes requested):

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: Oh, didn't know that already existed. Thanks! Fixed.

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: Sure, fixed!

source/rust_verify_test/tests/raw_ptrs.rs
jaylorch: Added!

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: Hmm, fixing it created a problem. Apparently making *all* adjustments, including borrows, is problematic. For instance, it makes the `eq_cmp::eq_cmp1` test fail with:

```
error: `core::cmp::impls::impl&%9::eq` is not supported (note: you may be able to add a Verus specification to this function with `assume_specification`) (note: the vstd library provides some specification for the Rust std library, but it is currently limited)
   --> /home/lorch/verus-jaylorch/source/target/debug/test_inputs/eq_cmp-5b709e559c865d5c-eq_cmp1/test.rs:106:13
    |
106 |     let b = P(3, true) == P(3, false);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
            pub assume_specification<'a, 'b, A, B> [<&'b A as std::cmp::PartialEq<&B>>::eq] (_0: &&'b A, _1: &&B) -> bool
            where
            A: std::cmp::PartialEq<B> + ?Sized,
            B: ?Sized,;
```

So it looks like it would be best to go back to the way it was, with a custom function that only applies pointer adjustments.

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: I updated this one to use the custom function to avoid similar issues.

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: Note that the original function that Claude wrote applied all adjustments *except* Borrow. I thought that was too aggressive, and changed it to only do Pointer adjustments. Should I change it "back" to applying all non-Borrow adjustments?

tjhance: Ah! Now I see why you/claude tried to factor it out.

In general, adjustments can do pretty complicated transformations, so I don't think a good idea to try to cleverly skip over them. It's best practice to always use expr_ty_adjusted for getting the type of an argument to a function.

I presume that the extra reference is appearing in the type because that's the what the type signature of the function expects (in this case `<*const T as core::cmp::PartialOrd>::lt` expects the args to be `&*const T` and `&*const T`. So the most reliable thing to do is to get the type with expr_ty_adjusted, then explicitly strip off the Ref, and then you'll have the type you're looking for.

Note that the user could also write `&a < &b`, in which case there would be no adjustments at all, so attempting to strip off the reference by simply filtering it out of the adjustments wouldn't work. (And I'm sure this case is also broken in the old code that calls `node_type` as well.) (EDIT: I'd have to double check to be sure; it might be Rust adds a & and * adjustment in this case.)

tjhance (changes requested):
i gave my recommendation in my comment above. looks like this is still in my queue though, so i'm officially Requesting Changes

jaylorch:
> i gave my recommendation in my comment above. looks like this is still in my queue though, so i'm officially Requesting Changes

OK, great! Are the changes I just made what you had in mind?

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: OK, done.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1767,8 +1775,12 @@ fn operator_overload_to_vir<'tcx>(
         let (fun_sym, Some(trait_id)) = lang_item_for_op(tcx, op, span)? else {
             crate::internal_err!(span, "operator needs an accessible trait");
         };
-        let lhs_ty = bctx.types.node_type(lhs.hir_id);
-        let rhs_ty = bctx.types.node_type(rhs.hir_id);
+        // When constructing the substs for trait resolution, we use
+        // expr_ty_adjusted to account for all adjustments (e.g., pointer
+        // coercions like *mut T -> *const T), then strip off any Ref
+        // since auto-borrow adds a reference we don't want here.
```
tjhance: This is gonna sound pedantic, but the "auto-borrow" is really a red herring. The arguments are typed as borrow types because that's the type signature of the lt operation. The fact that the reference is usually added automatically is just an incidental detail.

tjhance (approved):

source/rust_verify/src/rust_to_vir_expr.rs
jaylorch: Thanks. I've updated the comment to reflect that.

PR #2297, opened 2026-04-01
new-mut-ref doc updates

tjhance:
* Add a draft guide page for mutable references. [rendered doc page](https://github.com/verus-lang/verus/blob/new-mut-ref-doc-updates/source/docs/new-mut-ref.md)

 * Prepare for making the experiment "official"

Once this (and my other open PRs) are merged in, I'll "officially" announce it (as experimental) (i.e., invite people to try out `-V new-mut-ref` and report any issues they encounter)

### DIFF
--- a/source/docs/new-mut-ref.md
+++ b/source/docs/new-mut-ref.md
@@ -331,7 +331,7 @@ fn loop_test() -> (r: u64)
 }
 ```
 
-## The `has_resolved` operator
+## The `has_resolved` operator: When mutable references are "done"
 
 How does Verus know when a mutable borrow is "complete"?
 As we've seen, the creation of a mutable borrow lets us immediately reason about the


source/docs/migration-mut-ref.md
```diff
@@ -74,8 +74,8 @@ around mutable references and no longer makes sense.
 We aim to treat mutable references uniformly with other types, and that means that referring to
 any input parameter should always refer to its pre-state value.
 
-In the new system, mentioning a mutable variable in the precondition (e.g., `*a`) always refers to the value at entry. You can still optionally use `*old(a)`, but this is redundant.
-In the post-condition, you either use `*old(a)` to refer to the value at entry, or use the newly introduced `final` operator to refer to the updated value (`*final(a)`) of the mutable variable.
+In the new system, mentioning a mutable variable in the **pre**-condition (e.g., `*a`) always refers to the value at entry. You can still optionally use `*old(a)`, but this is redundant.
```
parno: I think in most of the guide, we don't hyphenate precondition or postcondition.

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
+# Feature status
+
+The content in this document applies only to Verus's **experimental**
+new mutable reference support,
+which can be enabled with the Verus command line option `-V new-mut-ref`.
+
+If you're familiar with the old design, see [migration-mut-ref.md](./migration-mut-ref.md)
+for more information on the transition and migration issues.
```
parno: Perhaps "potential migration issues"?

source/docs/new-mut-ref.md
parno, line 1: Should this be linked from the table of contents?  Or are you holding off on that to allow smaller-scale experiments?  Either way, if it's in the main Tutorial section, I think it should be pretty late, as it's pretty technical material.

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+    assert(a == 20);
+}
+```
+
+For more complex examples, we often need to write specifications about mutable references.
+In the rest of the section, we'll see how to do that.
+
+## Function specifications
+
+One of the most common ways to work with mutable references is to have a function taking
```
parno: "taking" -> "that takes"

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+    let r = get_mut_fst(&mut p);
+    *r = 100;
+
+    assert(p == (100, 20));
+}
+```
+
+Think for a moment about how you might write a specification for `get_mut_fst`.
+This is a little more challenging because the "final" value of `pair.0` isn't known concretely at
+the end of the function. Instead, this value can additionally be mutated by the caller who can
```
parno: I would cut "additionally"

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+this _relation_ between `final(ret)` and `final(pair)` is known.
+
+To prove `get_mut_fst_test`, Verus reasons roughly as follows:
+
+```rust
+fn get_mut_fst_test() {
+    let mut p = (10, 20);
+
+    let r = get_mut_fst(&mut p);
+    //                  ^^^^^^ call this value `pair`
```
parno: I'd say "let's call this value `pair`", since "call" is a bit overloaded here at the point where you're actually making a call

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+    assert(x == 20);
+}
+```
+
+1. This is the relation between the final value of `x_ref` and the updated value of `x`,
+   briefly mentioned above.
+
+2. Here we have an operator, `has_resolved`, that we haven't introduced yet.
+   The expression `has_resolved(x_ref)` is equivalent to `*x == *final(x)`,
+   meaning `x_ref` has attained its final value, i.e., it won't be mutated subsequently.
```
parno: Need a semicolon before "i.e." here

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+}
+```
+
+## Building a list
+
+Finally, let's do a more complex example to really get deeper into the mindset of
+thinking about mutable references in terms of the relationships between the "final values".
+
+Specifically, we'll see how mutable references can be used to build up a cons-list
+"from the root down", and how to verify it.
```
parno, line 432: This might just be me, but I found the example code difficult to decode, even with the accompanying graphic.  Maybe it's because it's a counterintuitive way to build a Cons list.  Or because I tried to figure out what the code was doing before scrolling down to read the text and the graphic, which helped but still took some puzzling.

If we want to keep this example, it might help to move the explanation before the code.  I think adding something like:
> The code maintains the invariant that `cur` points at the `Nil` entry at the tail of the list, and each iteration replaces the `Nil` with a `Cons(0, Nil)`

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+        Some(inner_ref) => {
+            // Obtain a reference to the contents of the Option
+            *inner_ref = 20;
+        }
+        None => { }
+    }
+
+    assert(a == 20);
+}
+```
```
Chris-Hawblitzel, line 43: When I try this example, it says "type `{integer}` cannot be dereferenced" for `*inner_ref = 20`.

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+    &mut pair.0
+}
+
+fn get_mut_fst_test() {
+    let mut p = (10, 20);
+
+    let r = get_mut_fst(&mut p);
+    *r = 100;
+
+    assert(p == (100, 20));
```
Chris-Hawblitzel: I got a type inference error for this line, but `assert(p.0 == 100 && p.1 == 20);` worked.

source/docs/new-mut-ref.md
```diff
@@ -0,0 +1,513 @@
...
+    let a_ref = &mut a;
+
+    loop
+        invariant after_borrow(a) == *a_ref,
+        decreases 0nat
+    {
+        *a_ref = 30;
+        return a;
+    }
+}
```
Chris-Hawblitzel, line 331: I see an error on this example.

source/docs/new-mut-ref.md
tjhance: I'll add to the guide when the feature fully releases

PR #2298, opened 2026-04-01
Fix the followed var and ghost fields in proof_with! 

### LINKED ISSUE #2283: Conflict caused by `proof_with!` on struct
The new [`proof_with!`-on-struct](https://github.com/verus-lang/verus/pull/2279) feature, unfortunately, conflicts with an existing use case where `proof_with` is used to return a ghost value that is unrelated to the returned struct.

For example, the following code returns a ghost integer and worked in earlier versions. However, with the new behavior, it now attempts to construct an `S` with a ghost field, which results in panic:
```
struct S {}

#[verus_spec(
    with 
        -> res: Ghost<int>
)]
pub fn f() -> S {
    proof_with!(|= Ghost(0int));
    S {}
}
```
@jaylorch  @ziqiaozhou  Could you advise on how to address this or avoid this?

ziqiaozhou:
Supersedes #2285
Fixes #2283

* Move ghost fields into WithSpecOnExpr to unify the implementation of proof_with! so that other specs applied in proof_with! are not affected by the ghost fields. It also reverts some changes made in #2279
* Added more syntax error checks for ghost fields.

### DIFF
--- a/dependencies/syn/src/verus.rs
+++ b/dependencies/syn/src/verus.rs
@@ -266,7 +266,7 @@ ast_struct! {
         pub inputs: Punctuated<Expr, Token![,]>,
         pub outputs: Option<(Token![=>], Pat)>,
         pub follows: Option<(Token![|=], Pat)>,
-        pub ghost_fields: Punctuated<FieldValue, Token![,]>, // only supported if applied to a struct construction expression
+        pub erased_fields: Punctuated<FieldValue, Token![,]>, // only supported if applied to a struct construction expression
     }
 }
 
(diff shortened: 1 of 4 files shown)


dependencies/syn/src/verus.rs
```diff
@@ -266,6 +266,7 @@ ast_struct! {
         pub inputs: Punctuated<Expr, Token![,]>,
         pub outputs: Option<(Token![=>], Pat)>,
         pub follows: Option<(Token![|=], Pat)>,
+        pub ghost_fields: Punctuated<FieldValue, Token![,]>, // only supported if applied to a struct construction expression
```
jaylorch: This applies to not just ghost fields but also tracked fields, right? If so, we may want a name that reflects this, e.g., `ghost_or_tracked_fields` or `erased_fields`.

source/builtin_macros/src/attr_rewrite.rs
```diff
@@ -776,14 +734,80 @@ fn rewrite_verus_spec_on_expr_local(
...
+/// Apply ghost fields in `with` clause to a struct constructor expression.
+/// Return Err if the ghost fields are not valid.
+fn apply_ghost_fields<'a>(
+    erase: EraseGhost,
+    expr: &mut Expr,
+    ghost_fields: impl Iterator<Item = &'a verus_syn::FieldValue>,
+) -> Result<(), ()> {
+    let syn::Expr::Struct(expr_struct) = expr else {
+        // If there's no struct constructor, we cannot apply ghost fields.
+        if let Some(field) = ghost_fields.last() {
```
jaylorch: Perhaps this should just be a `match ghost_fields.last()` rather than an `if let`, since there are only two options?

PR #2299, opened 2026-04-01
account for dummy args when translating f.requires and f.ensures

### LINKED ISSUE #2296: Panic when using `call_ensures` in `ensures` clause
### POC
```
use vstd::prelude::*;
verus!{
pub fn f()->Option<()> {
    Some(())
}

pub fn call_f() -> (ret: Option<()>) 
    ensures
        f.ensures((), ret),
{
    f()
}

fn main() {}
}
```

### Panic Log
```
thread '<unnamed>' (46984) panicked at vir\src\poly.rs:212:43:
no entry found for key

thread '<unnamed>' (46984) panicked at rust_verify\src\verifier.rs:2217:29:
worker thread panicked
```
### Playground Link
https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=df4553312fa089be8f2fa551d77edb1a

tjhance:
fixes #2296

### DIFF
--- a/source/rust_verify_test/tests/fndef_types.rs
+++ b/source/rust_verify_test/tests/fndef_types.rs
@@ -1901,3 +1901,37 @@ test_verify_one_file! {
         }
     } => Err(err) => assert_fails(err, 3)
 }
+
+test_verify_one_file! {
+    #[test] zero_arg_fn_issue2296 verus_code! {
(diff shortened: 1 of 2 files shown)


tjhance:
just to check, the "dummy_param_name" is unambiguous, right? like, it can't be the result of a user param?

PR #2300, opened 2026-04-02
Translate general case of ExprKind::Index into calls to Index::index

Chris-Hawblitzel:
This supports exec-mode index expressions in general, including range index expressions like `vec[1..3]`.

I also ran the added tests with new-mut-ref and they passed (to avoid duplicating the tests, I didn't commit this).

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -1,4 +1,4 @@
-#![no_std]
+#![cfg_attr(not(verus_verify_core), no_std)]
 #![allow(internal_features)]
 #![cfg_attr(
     verus_keep_ghost,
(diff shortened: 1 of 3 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -3000,7 +3079,24 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
...
+            // For borrow checking and SMT efficiency, ((array | slice), usize) is special case
+            // For SMT efficiency, (Vec, usize) is special case
+            // All other cases go through general Index trait
+            let array_kind = match tgt_expr_ty.kind() {
+                TyKind::Array(..) => Some(vir::ast::ArrayKind::Array),
+                TyKind::Slice(..) => Some(vir::ast::ArrayKind::Slice),
+                _ => None,
+            };
+            if let (Some(array_kind), false, true) =
+                (array_kind, bctx.types.is_method_call(expr), idx_ty.is_usize())
```
tjhance: Can you keep the original condition, and just case on `is_method_call` at the top level? In general (for any overloadable operator) I think it is best to consider the two cases entirely separately. i.e. anything falling into the `!is_method_call` case should require special handling. (Even the form taken by the adjusted types of the arguments differ between the `!is_method_call`/`method_call` cases.)

If I'm not mistaken, then for Index in particular, anything other than (array, usize) or (slice, usize) should already fall in the `is_method_call` case, so checking the other fields should be redundant, anyway.

But if I _am_ mistaken — i.e., if there _is_ another case besides (array, usize) or (slice, usize) which falls in the `!is_method_call` case — it probably needs special handling as a Place node, and we should consider it explicitly.

PR #2322, opened 2026-04-13
doc: Add guide chapter on `async/await`

dschoepe:


source/docs/guide/src/async-await.md
```diff
@@ -0,0 +1,110 @@
+# Verifying asynchronous code
+
+Rust supports writing asynchronous code without manually managing which thread a
```
parno, line 3: I might rephrase as:
> To support writing asynchronous code without manually assigning each task to a specific thread, Rust provides the `async` and `await` keywords.

source/docs/guide/src/async-await.md
```diff
@@ -0,0 +1,110 @@
...
+
+Verus supports reasoning about code using `async`/`await`. Just
+like regular `fn`s, `async fn`s can be annotated with `requires`
+and `ensures` clauses:
+
+```rust
+{{#include ../../../../examples/guide/async_await.rs:basic_async_ensures}}
+```
+
+`result` in the `ensure`s clause refers to the value of the future,
```
parno, line 19: I think we want `ensures` rather than `ensure`s

examples/guide/async_await.rs
```diff
@@ -0,0 +1,88 @@
...
+// ANCHOR_END: basic_async_ensures
+
+// ANCHOR: basic_async_caller
+async fn async_call(x: u64)
+{
+    let result = add1(41).await;
+    assert(result == 42);
+    let future = add1(41);
+    // ^ This future has not yet executed, so the following
+    // assertion would fail; the value of a future can be obtained
```
parno, line 26: The first line of the comment says that the assertion will fail, but then it also says that you can use `@` to obtain the value, so it sounds a bit contradictory.  Maybe it would be clearer to say something like "once the future has executed, the value it computed can be obtained..."?

examples/guide/async_await.rs
```diff
@@ -0,0 +1,88 @@
...
+    let f2 = add1(41);
+    let result = branch_and_await(true, f1, f2).await;
+    assert(result == 23 && f1.awaited());
+}
+// ANCHOR_END: async_future_branch
+
+// ANCHOR: async_mut
+async fn mutating_arg(x: &mut usize) -> (result: ())
+    requires *old(x) < usize::MAX,
+    ensures *x == *old(x) + 1
```
parno, line 73: We'll need to wrap this in `final(x)` before it goes live.

source/docs/guide/src/async-await.md
```diff
@@ -0,0 +1,110 @@
...
+values. The mutations are only known to have taken effect once the future
+has been awaited.
+
+```rust
+{{#include ../../../../examples/guide/async_await.rs:async_mut}}
+```
+
+*Note*: Technically, Rust guarantees that after creating, but not `await`ing,
+`f`, `*x` will still have value `42`. However, to avoid complex tracking of when
+a future may have started executing, Verus treats this imprecisely and considers
```
parno, line 56: Perhaps:
> Verus treats this conservatively and considers...

source/docs/guide/src/async-await.md
```diff
@@ -0,0 +1,110 @@
...
+
+```rust
+select! {
+    add1(42) = r1 => r1,
+    add1(21) = r2 => r2,
+}
+```
+
+This code runs both futures `add1(42)` and `add1(21)` concurrently and returns
+the value of whichever completes first; the syntax `expr = var => ret` binds
```
parno, line 79: I think we want:

> the syntax `expr = var => ret` binds the final value of `expr` to the name `var` and then executes expression `ret`.

source/docs/guide/src/async-await.md
```diff
@@ -0,0 +1,110 @@
...
+the value `expr` completed with to `var` and executes expression `ret`.
+
+While `verus` does not natively understand how to reason about the code
+generated by the `select!` macro, we can use mechanisms for specifying
+assumptions on external code, such as `#[verifier::external_body]` for
```
parno, line 84: Can we update this to:
[`#[verifier::external_body]`(./calling-unverified-from-verified.md)
so the reader can find more info about that attribute?

source/docs/guide/src/SUMMARY.md
```diff
@@ -107,6 +107,8 @@
   - [Memory safety is conditional on verification](./memory-safety.md)
   - [Calling verified code from unverified code](./call-from-unverified-code.md)
 
+- [Verifying asynchronous code](./async-await.md)
```
parno, line 110: Maybe we could list this earlier, perhaps before "Unsafe code & complex ownership"?  That portion seems to focus more on specific Rust features.

parno (commented):
Thanks for putting this together.  I added a variety of mostly minor comments.

I think @Chris-Hawblitzel suggested we hold off on merging this, however, until we're able to try `async` on some larger examples and work out some of the completeness issues that emerge.

PR #2326, opened 2026-04-14
Extend Verus with Logical Atomicity

slerpyyy:
This PR extends Verus with a built-in notion of logical atomicity, heavily inspired by its implementation in Iris.


tools/vargo/.cargo/config.toml
```diff
@@ -3,3 +3,4 @@
 
 [build]
 target-dir = "target"
+build-dir = "target"
```
tjhance: extraneous change?

source/vstd/atomic.rs
```diff
@@ -87,6 +92,12 @@ macro_rules! atomic_types {
             pub open spec fn id(&self) -> AtomicCellId {
                 self.view().patomic
             }
+
+            /// Implies that `a@ == b@ ==> a == b`.
+            pub broadcast axiom fn view_bijective(self)
+                ensures
+                    Self::view_inverse_for_eq(#[trigger] self@) == self,
+            ;
```
tjhance: We ended up not needing this, right?

It might be best to split this out and discuss it separately ... I'm not sure we actually want these axioms for all the different PointsTo types in Verus, this can have some subtle soundness implications

source/.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+# Set target directory in case it was overridden by the environment
+# https://doc.rust-lang.org/cargo/reference/config.html
+
+[build]
+target-dir = "target"
+build-dir = "target"
```
tjhance: are these config changes meant to be included in the PR?

source/vir/src/ast.rs
```diff
@@ -1470,6 +1491,9 @@ pub struct FunctionAttrsX {
     pub bit_vector: bool,
     /// Is atomic (i.e., can be inside an invariant block)
     pub atomic: bool,
+    /// Instructs the `AtomicInstCollector` to ignore all calls to this function.
+    /// This allows the function to be called freely in invariant and atomic update blocks.
+    pub skip_inst_collector: bool,
```
tjhance: Is there anything preventing the user from adding these to their own functions?

source/vir/src/def.rs
```diff
@@ -1156,6 +1156,13 @@ pub fn fn_set_contains_name() -> Fun {
     })
 }
 
+pub fn result_type_path() -> Path {
+    Arc::new(PathX {
+        krate: CrateId::Core,
+        segments: Arc::new(vec![Arc::new("result".to_owned()), Arc::new("Result".to_owned())]),
+    })
```
tjhance: There's a macro, `path!`, you can use to write this more succinctly

tjhance (commented):
did a really cursory pass. I am going to spend some time looking carefully at the VC gen and then probably leave some more comments.

source/builtin_macros/src/syntax.rs
```diff
@@ -527,6 +558,345 @@ impl Visitor {
         }
     }
 
+    fn resolve_receiver(&self, receiver: &Receiver, name: &Ident) -> PatType {
+        match &receiver.colon_token {
+            None => {
+                let (_generics, mut ty) = self
+                    .inside_impl
+                    .as_deref()
+                    .expect("cannot resolve self type outside of impl block")
```
tjhance: should be a user error rather than a panic, e.g., this:

```
pub fn read(&self, key: Key) -> (value: Option<Value>)
    atomically (au) { ... }
```

outside of an 'impl' block should be a parse error, not a panic

tjhance:
Seems to be a bug in the following:

```
use vstd::prelude::*;
use vstd::atomic::*;

verus!{

type Key = u64;
type Value = u64;

impl MyHashMap {
}

pub tracked struct MPointsTo { }

struct MyHashMap { }

impl MPointsTo {
    pub uninterp spec fn key(&self) -> Key;
    pub uninterp spec fn value(&self) -> Option<Value>;
    pub uninterp spec fn instance_id(&self) -> int;
}

impl MyHashMap {
    pub uninterp spec fn wf(&self) -> bool;
    pub uninterp spec fn id(&self) -> int;

    pub fn read(&self, key: Key) -> (value: Option<Value>)
        atomically (au) {
            type ReadPred,
            (mpt: MPointsTo) -> (new_mpt: I<MPointsTo>),
            requires
                key == mpt.key(),
                self.id() == mpt.instance_id(),
            ensures
                new_mpt@ == mpt,

            outer_mask any,
            inner_mask none,
        }
        requires self.wf(),
        ensures value == mpt.value(),
    {
        return None;
    }
}

pub type ReadAU = AtomicUpdate<MPointsTo, I<MPointsTo>, ReadPred>;


}
```

I get "cannot use assert or assume in exec mode". For reasons that are unclear to me, the syntax-macro skips adding a proof block around the assume

source/vstd/atomic.rs
slerpyyy: I've removed it from the PR

source/vir/src/ast.rs
slerpyyy: Removed!

source/.cargo/config.toml
slerpyyy: I have opened a separate issue #2366 about this, and extracted out my current fix #2368

Can you make sure the right people see this?

source/builtin_macros/src/syntax.rs
slerpyyy: Fixed

source/vir/src/ast.rs
```diff
@@ -1046,9 +1057,18 @@ pub enum ExprX {
...
+        target: CallTarget,
+        args: Exprs,
+        /// To be executed *after* the arguments but *before* the call.
+        ///
+        /// This is used for two-phase borrows.
+        post_args: Option<Expr>,
+        /// Executed *inside* the function call, between the pre- and postcondition.
+        ///
+        /// We use this for the `atomically |update| { ... }` block of the atomic function call.
+        body: Option<Expr>,
```
tjhance, line 1081: Is this body always an `Atomically` expression? If so, we should make this obvious in the types, making Atomically its own type:

```
struct Atomically(AtomicCallInfo, VarIdent, Expr, bool)

...


 Call {
  target: CallTarget,
  args: Exprs,
  post_args: Option<Expr>,
  body: Option<Atomically>, // instead of Option<Expr>
}
```

source/vir/src/ast.rs
```diff
@@ -1135,14 +1155,26 @@ pub enum ExprX {
...
+        atomic_call: bool,
     },
     /// Open invariant
     OpenInvariant(Expr, VarBinder<Typ>, Expr, InvAtomicity),
+    /// Open Atomic Update
+    TryOpenAtomicUpdate(Expr, VarBinder<Typ>, bool, Expr),
+    /// Placeholder expression for the atomic update argument in the atomic function call
+    AtomicUpdateInitDummy,
+    /// Atomic function call
+    Atomically(AtomicCallInfo, VarIdent, Expr, bool),
```
tjhance: What's the meaning of all these arguments?

source/vir/src/ast.rs
```diff
@@ -1631,6 +1663,7 @@ pub struct FunctionX {
     pub fndef_axioms: Option<Exprs>,
     /// MaskSpec that specifies what invariants the function is allowed to open
     pub mask_spec: Option<MaskSpec>,
+    pub atomic_update: Option<Expr>,
```
tjhance, line 1674: I feel this should be a binder or a Param or something, rather than an arbitrary Expr.

source/vir/src/ast.rs
tjhance: And 'body' should probably be more specific, like `atomic_body`.

source/vir/src/ast.rs
```diff
@@ -1008,6 +1008,17 @@ pub enum AutospecUsage {
     Final,
 }
 
+pub type AtomicCallInfo = Arc<AtomicCallInfoX>;
+#[derive(Clone, Debug, Serialize, Deserialize, ToDebugSNode)]
+pub struct AtomicCallInfoX {
+    pub au_typ: Typ,
+    pub au_typ_args: Typs,
+    pub x_typ: Typ,
+    pub y_typ: Typ,
```
tjhance: This information looks a bit redundant. Is it always the case that `au_typ` is just `Datatype("vstd::atomic::AtomicUpdate", au_typ_args)`? And is `au_typ_args` always just `[x_typ, y_typ]`? In that case, we only need the `x_typ` and `y_typ` fields.

And `x_typ`/`y_typ` could have better names. `input_typ` and `output_typ`?

source/vir/src/headers.rs
```diff
@@ -291,10 +274,7 @@ pub fn read_header_block(block: &mut Vec<Stmt>, allows: &HeaderAllows) -> Result
     *block = block[n..].to_vec();
     let require = require.unwrap_or(Arc::new(vec![]));
     let recommend = recommend.unwrap_or(Arc::new(vec![]));
-    let (ensure_id_typ, ensure) = match ensure {
-        None => (None, (Arc::new(vec![]), Arc::new(vec![]))),
-        Some((id_typ, es)) => (id_typ, es),
-    };
+    let (ensure_id_typ, ensure) = ensure.unwrap_or_default();
```
tjhance: what's the reason for the changes with 'ensure' in this file?

source/vir/src/modes.rs
```diff
@@ -282,6 +281,13 @@ fn outer_reason_by_expr_kind(e: &Expr) -> Option<OuterProphReason> {
         ExprX::Return(..) => Some(OuterProphReason::Return),
         ExprX::BreakOrContinue { is_break: true, .. } => Some(OuterProphReason::Break),
         ExprX::BreakOrContinue { is_break: false, .. } => Some(OuterProphReason::Continue),
+
+        // todo
+        ExprX::TryOpenAtomicUpdate(..) |
+        ExprX::AtomicUpdateInitDummy |
+        ExprX::Atomically(..) |
+        ExprX::Update(..) |
+        ExprX::InvMask(..) => None,
```
tjhance: Anything stateful should be `Some(OuterProphReason::{something})`. Here, TryOpenAtomicUpdate and Update should definitely be Some.

source/vir/src/ast.rs
slerpyyy: Resolved during meeting: Once the `atomically` block has been placed into the body, we never make any assumptions about the body expression, we treat this as an arbitrary expression, and this is reflected in the variable name

source/vir/src/ast.rs
slerpyyy: We get this from the `verus_builtin::atomic_spec<A>(_a: A)` header function, so it is just an expression

source/vir/src/ast.rs
slerpyyy: Yes, you're right, I removed the `AtomicCallInfo` type entirely

source/vir/src/headers.rs
slerpyyy: I reverted the change

source/vir/src/modes.rs
slerpyyy: Done!

source/vir/src/ast.rs
slerpyyy: I've replaced the `bool` with an `AtomicallyKind` enum, so it should be easier to read now

parno:
FWIW, I tried out this branch on some of my code that would benefit from logical atomicity.  However, I ran into two syntax-related issues.  First, in the code below, if you uncomment the `where` clause, then Verus complains about `atomically (atomic_update)`, saying that it expected a `:` and underlining `atomic_update`.  This seems like it's probably an issue with the parser logic for the `atomically` clause.

Second, when the `where` clause is commented out, the code below leads to a proc macro panic with a help message that it expected a `,`.  

```
use vstd::prelude::*;
use vstd::logatom::*;

verus! {

pub fn f<Id, K, G, P: AtomicInvariantPredicate<K, u32, G>>(
    loc: &vstd::atomic_ghost::AtomicU32<K, G, P>,
    val: u32,
) -> (res: u32)
// where
//     P: AtomicInvariantPredicate<K, u32, G>,
    atomically (atomic_update) {
        (perm: PermissionU32) -> (res: Result<PermissionU32, (PermissionU32, OpenInvariantCredit)>),
        requires
            true,
        ensures
            true,
        outer_mask any,
        inner_mask none,
    },
    ensures true,
{
    42
}

} // verus!


fn main() {}
```

slerpyyy:
@parno Thank you for pointing that out, it should be fixed now

parno:
> @parno Thank you for pointing that out, it should be fixed now

Thanks for the fixes.  Will you be able to add some new pages to the concurrency guide?  I think that would make it more feasible for me to try it out further on my code.

slerpyyy:
> Will you be able to add some new pages to the concurrency guide? I think that would make it more feasible for me to try it out further on my code.

@parno I've added an initial draft of a logical atomicity section to the Verus Tutorial and Reference

https://github.com/verus-lang/verus/blob/adc265a0d49b201a24b4e0a500e82a4b3d918f77/source/docs/guide/src/SUMMARY.md?plain=1#L176-L179

parno:
Great, thanks for adding the documentation!  If you haven't yet, you should chat with @tjhance about whether the content should be in the main guide (where you put it) or in the [dedicated concurrency guide](https://verus-lang.github.io/verus/state_machines/), which lives in `/source/docs/state_machines/`.

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Main concepts
+
+The two functions above are logically atomic, because they have a **linearization point (LP)**, that is, a single atomic step which updates the state of the program.
+For the `increment` function, for example, the LP occurs when the `compare_exchange_weak` operations runs successfully.
+Identifying the LP in a function is the first important step in proving it is logically atomic.
+
+We can then specify the behavior of such a function using an *atomic* pre- and postcondition.
+These describe the state of the program immediately before and after the LP, and they live alongside the "regular" function pre- and postconditions.
+Together, they assert describe the state of the program at four distinct points in time:
```
parno: Cut "assert"

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+Following the Iris implementation of logical atomicity, we abstract the LP using a tracked ghost object called the **atomic update (AU)**.
+The atomic specification declares what the atomic update looks like for a particular function.
+This object is constructed by the client/caller, and it is destructed (or "opened") by the library/callee.
+
+## Previous & related work
+
+Formalizations of logical atomicity come in two flavors, which are commonly referred to as "HOCAP-style" and "TaDA-style" logical atomicity.
+
+HOCAP-style logical atomicity was first introduced by VeriFast ([Jacobs & Piessens POPL'11](https://dl.acm.org/doi/abs/10.1145/1926385.1926417)) and later refined by HOCAP ([Svendsen et al. ESOP'13](https://link.springer.com/chapter/10.1007/978-3-642-37036-6_11)).
+Here, logically atomic function take a "ghost callback" which is invoked at the linearization point to temporarily acquire the users permission and update the relevant ghost state.
```
parno: "a logically atomic function takes"

"the users permission" -> "the caller's permission"

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+This approach has been applied in Verus to verify CapybaraKV ([LeBlanc et al. OSDI'25](https://www.usenix.org/conference/osdi25/presentation/leblanc)) and the traits which have been used in this construction can be found in the [`vstd::logatom`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/index.html) module. While this approach has the advantage of being fully verified in Verus, it comes with significant downsides to ergonomics and flexibility.
```
parno: "and the traits which have been used in this construction can be found" -> ", and the traits they used can be found"

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+In particular, this encoding of logical atomicity is inherently higher-order and cumbersome to reason about, it requires a significant amount of boilerplate code -- both for the library/callee but also for the client/caller, it obfuscates the function specification in layers of abstraction, and it only allows ghost state to be updated at the LP, making this approach too weak for both of our running examples.
```
parno: This sentence is currently a run-on sentence.  I think most of the commas should be periods, but if some of them are meant to be continuations of the previous segment, you need something like ", because" or ", since".

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+
+TaDA-style logical atomicity is named after the TaDA logic ([da Rocha Pinto et al. ECOOP'14](https://link.springer.com/chapter/10.1007/978-3-662-44202-9_9)) which introduced atomic triples as a primitive part of its logic.
+Building on these ideas, Iris ([Jung et al. POPL'15](https://dl.acm.org/doi/abs/10.1145/2775051.2676980)) then presented similar notion of atomic specifications, which can be fully encoded in its separation logic.
```
parno: "presented a similar notion "

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,62 @@
...
+Our implementation of logical atomicity in Verus is based on the atomic specification as they have been verified in Iris.
```
parno, line 93: "is based on the atomic specification as they have been verified in Iris." -> "is based on the atomic specifications verified in Iris."

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+    ensures private_post(px, ax, ay, py),
+```
+
+Inside the `atomically` block, we find the `requires` and `ensures` clauses which denote the *atomic* pre- and postcondition, which describe the state of the program just before/after the linearization point (LP) respectively.
+
+To attach resources to the atomic pre- and postcondition, we can use the arrow `(ax: AX) -> (ay: AY)`.
+In what follows, we will refer to `ax: AX` as the "input type" of the atomic update, and `ay: AY` as the output type.
+Both `ax` and `ay` are always tracked-mode.
+
+This atomic specification binds an additional variable called `atomic_update` which has type `AtomicUpdate<AX, AY, PredType>`.
```
parno: I think it would help to be explicit that `atomic_update` can be any user-defined identifier (assuming that's the case).

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+The predicate type can be given a name using the optional `type PredType` clause as seen above.
```
parno: Am I right in thinking that `PredType`, regardless of name, is something that's manufactured by the `atomically` block?  I.e., I can't use an existing type here, right?  If that's true, I think it would help to make that explicit.

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+}
+
+pub trait UpdateTry {
+    spec fn branch(self) -> UpdateControlFlow;
+}
+```
+
+The Verus standard library provides implementations of this trait for two types:
+
+- `AY = Result<T, E>` specifies an atomic update which can be committed by outputting `Ok(t)` and aborted using `Err(e)`, this is equivalent to the atomic update as it can be found in Iris, except that we do not force `AX` and `E` to be the same.
```
parno: ", this is " -> "; this is "

"equivalent to the atomic update as it can be found in Iris," -> "equivalent an Iris atomic update,"

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+
+- `AY = Commit<T>` specifies an atomic update that cannot be aborted.
+
+In cases where one wants to differentiate between multiple different commit or abort cases, it can be helpful to implement this trait for custom types.
```
parno: This might be clearer as:

> To differentiate between multiple different commit or abort cases, it can be helpful to implement this trait for a custom type.

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+        outer_mask any,
+        inner_mask none,
+    },
+```
+
+```rs
+pub fn increment(var: &PAtomicU64) -> (out: u64)
+    atomically (atomic_update) {
+        (perm: PermissionU64)
+            -> (res: Result<PermissionU64, (PermissionU64, OpenInvariantCredit)>),
```
parno: Why does the spec return an `OpenInvariantCredit`?  And why doesn't the atomic postcondition say anything about it?

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
...
+When we open the atomic update, we are given the input `ax: AX` and learn the atomic precondition, and at the end of the macro call, we return the output `ay: AY` and prove the atomic postcondition.
+
+The macro outputs a value of type `Tracked<Result<(), AtomicUpdate<AX, AY, PredType>>>`.
+If the atomic update is committed, as indicated by the `UpdateTry` trait, then the atomic update is consumed and we get `Ok(())`.
+Otherwise, if the atomic update is aborted, we get back the same atomic update we put in `Err(au)`, allowing us to open it again later.
+
+When the atomic update is opened and committed, it is marked as resolved, meaning the value of `au.resolves()` becomes `true`.
+We must prove that the atomic update has resolved by the end of the logically atomic function.
+
+To open the atomic update, `au.outer_mask()` must be valid in the current scope.
```
parno: I think this mask discussion needs more context; at the very least a link to additional details elsewhere in the docs.

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
+# Opening atomic updates
```
parno: I don't think it will be clear to the reader what "Opening atomic updates" means, or why it's the logical follow on from the previous section on "Atomic specifications".  

I think it would be clearer to have a title like: "Implementing an atomic specification".  And then some discussion of how using/opening coincide.

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+When we open the atomic update, we are given the input `ax: AX` and learn the atomic precondition, and at the end of the macro call, we return the output `ay: AY` and prove the atomic postcondition.
+
+The macro outputs a value of type `Tracked<Result<(), AtomicUpdate<AX, AY, PredType>>>`.
+If the atomic update is committed, as indicated by the `UpdateTry` trait, then the atomic update is consumed and we get `Ok(())`.
+Otherwise, if the atomic update is aborted, we get back the same atomic update we put in `Err(au)`, allowing us to open it again later.
+
+When the atomic update is opened and committed, it is marked as resolved, meaning the value of `au.resolves()` becomes `true`.
+We must prove that the atomic update has resolved by the end of the logically atomic function.
```
parno: It's ambiguous where this proof obligation comes from.  If it's accurate, I think something like this would be clearer:

> Verus requires the body of an atomic function (like `reset` or `increment` in our running example) to prove that the atomic update has been resolved by the time the function returns.  Invoking `try_open_atomic_update!` discharges this proof obligation.

(Is that always enough to discharge the obligation?  Are there other ways to do so?)

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
...
+        requires
+            perm@.patomic == var.id(),
+        ensures
+            commit@.patomic == perm@.patomic
+            commit@.value == 0
+        outer_mask any,
+        inner_mask none,
+    },
+{
+    let tracked au = atomic_update;
```
parno: Does one always need to reassign like this?

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
...
+We must prove that the atomic update has resolved by the end of the logically atomic function.
+
+To open the atomic update, `au.outer_mask()` must be valid in the current scope.
+Inside the body of the `try_open_atomic_update` macro, `au.inner_mask()` becomes the new invariant mask.
+
+## Running examples
+
+Here are the full definitions of our two example functions:
+
+```rs
```
parno, line 45: We should eventually shift these code blocks into the `examples/guide` or `examples/state_machines`, so that they're covered by CI.

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
...
+    // assert atomic post
+    Tracked(ay)
+})
+```
+
+When we open the atomic update, we are given the input `ax: AX` and learn the atomic precondition, and at the end of the macro call, we return the output `ay: AY` and prove the atomic postcondition.
+
+The macro outputs a value of type `Tracked<Result<(), AtomicUpdate<AX, AY, PredType>>>`.
+If the atomic update is committed, as indicated by the `UpdateTry` trait, then the atomic update is consumed and we get `Ok(())`.
+Otherwise, if the atomic update is aborted, we get back the same atomic update we put in `Err(au)`, allowing us to open it again later.
```
parno, line 19: Can you add a bit more intuition for when the update should be committed/aborted?

source/docs/guide/src/logatom-open.md
```diff
@@ -0,0 +1,116 @@
+# Opening atomic updates
+
+To open the atomic update, we use the `try_open_atomic_update` macro as follows:
+
+```rs
+let res_au = try_open_atomic_update!(au, ax => {
```
parno: It looks like later you use `mut ax` here, so maybe it would be worth including `[mut | &mut] ax` in the syntax description?

source/docs/guide/src/logatom-call.md
```diff
@@ -0,0 +1,101 @@
+# Atomic function call
```
parno: I think the title would be clearer as "Calling an atomic function"

source/docs/guide/src/logatom-call.md
```diff
@@ -0,0 +1,101 @@
...
+
+To call a logically atomic function, we use the following syntax:
```
parno, line 3: I think it would be helpful for this to be:
> To call a logically atomic function (`my_atomic_fn`), we use the following syntax:

and then below use `my_atomic_fn` instead of `function`

source/docs/guide/src/logatom-call.md
```diff
@@ -0,0 +1,101 @@
...
+    let ay = update(ax);
+    // assume atomic post
+    // ...
+    break/continue
+}
+```
+
+The atomic function call binds an `update` function, which represents the effects of the `try_open_atomic_update` macro to the client.
+It is helpful to think of the macro as defining the `update` function, as their inputs and outputs, as well as their pre- and postcondition match precisely.
+The `update` function is `proof`-mode, allowing atomic invariants to be opened around it.
```
parno, line 21: Maybe clarify that the body of the `atomically` as a whole is `exec` mode (assuming that's true)

source/docs/guide/src/logatom-call.md
```diff
@@ -0,0 +1,101 @@
...
+
+Since the library function may open the atomic update multiple times, due to aborting, the client must prove that it can provide the required resources as many times as necessary, meaning the atomic function call must be a loop.
+
+If the atomic update is committed by the library, as indicated by the `UpdateTry` implementation on the output type, the client must `break` out of the loop.
```
parno, line 34: Maybe add a bit saying what happens if the client doesn't do so.

source/docs/guide/src/logatom-call.md
```diff
@@ -0,0 +1,101 @@
...
+Similarly, if the library aborts the atomic update, the loop must `continue`, either explicitly or implicitly.
+
+In cases where this aborting mechanism is not necessary (e.g. if the output type of the atomic update is `Commit<T>`), **the `loop` keyword can be removed** to disable this feature.
+This is equivalent to adding a `break` to the end of the `atomically` block, except it provides slightly better proof automation and eliminates the need to loop invariants in straight line code.
```
parno: "need to add loop"

parno: Also, this is the first mention of loop invariants.  It sounds like you're implying that by default, the client should plan to write invariants on the loop?  If so, you should say so and explain where they go.

slerpyyy:
Thank you for all the feedback!

I think this should go into the reference, since it is a language feature and does not directly depend on VerusSync, though it would make sense to add a small section there as well to show how the two features interact

source/docs/guide/src/logatom.md
```diff
@@ -0,0 +1,67 @@
...
+Following the Iris implementation of logical atomicity, we abstract the LP using a tracked ghost object called the **atomic update (AU)**.
+The atomic specification declares what the atomic update looks like for a particular function.
+This object is constructed by the client/caller, and it is destructed (or "opened") by the library/callee.
+
+## Previous & related work
+
+Formalizations of logical atomicity come in two flavors, which are commonly referred to as "HOCAP-style" and "TaDA-style" logical atomicity.
+
+HOCAP-style logical atomicity was first introduced by VeriFast ([Jacobs & Piessens POPL'11](https://dl.acm.org/doi/abs/10.1145/1926385.1926417)) and later refined by HOCAP ([Svendsen et al. ESOP'13](https://link.springer.com/chapter/10.1007/978-3-642-37036-6_11)).
+Here, a logically atomic function take a "ghost callback" which is invoked at the linearization point to temporarily acquire the caller's permission and update the relevant ghost state.
```
parno: "take" -> "takes"

source/docs/guide/src/logatom-spec.md
```diff
@@ -83,50 +84,31 @@ impl PredType {
...
+The output of the atomic update is wrapped in a `Commit<...>` which specifies that this AU cannot abort.
 
-    ensures
-        out == perm@.value,
+```rs
+{{#include ../../../../examples/guide/logatom.rs:increment_signature}}
 ```
+
+This specification is a bit more interesting.
+Here, the output if a `Result<...>` which means that the atomic update can be aborted.
```
parno: "if" -> "is"

source/docs/guide/src/logatom-spec.md
```diff
@@ -83,50 +84,31 @@ impl PredType {
...
+The atomic postcondition specifies that when the AU is aborted (i.e. when `res` is `Err`), the ID and value of the permission stays the same,
```
parno: Needs a comma after "i.e.", and a semicolon between "same" and "otherwise".

parno:
Thanks for all of the edits you've made; they're very helpful.  If we keep the docs in the main Guide, perhaps we could shift them to be after the "Unsafe code & complex ownership" segment, since that ends with a mention of concurrency?   The Guide's material is typically motivated by examples, while the Reference tends to be a bit more syntactic and examples are used to illustrate the operational detail of the feature (e.g., code A works; code B doesn’t work because of reason XYZ), not to motivate it.

slerpyyy:
That's a good point, I'll move the section

parno:
I ran into a surprising panic when using this branch.  Specifically, using a `matches` pattern in a loop invariant on an `atomically loop` causes an internal compiler panic during MIR building, but only if the bound variable is actually used in the conjunct.

Here's a (relatively) minimal repro:
```
use vstd::prelude::*;
use vstd::atomic::{PAtomicU32, PermissionU32, AtomicUpdate, Commit, try_open_atomic_update};

verus! {

fn store_zero(loc: &PAtomicU32)
    atomically (atomic_update) {
        type SPred,
        (perm: PermissionU32) -> (commit: Commit<PermissionU32>),
        requires perm@.patomic == loc.id(),
        ensures commit@@.patomic == perm@.patomic,
        outer_mask any,
        inner_mask none,
    },
{
    try_open_atomic_update!(atomic_update, mut perm => {
        loc.store(Tracked(&mut perm), 0);
        Tracked(Commit(perm))
    });
}

tracked struct Tok { ghost v: u32 }

fn caller() {
    let (loc, Tracked(mut perm)) = PAtomicU32::new(7);
    let tracked opt: Option<Tok> = Some(Tok { v: 5 });

    store_zero(&loc) atomically loop |update| -> (au: AtomicUpdate<
        PermissionU32, Commit<PermissionU32>, SPred,
    >)
        invariant
            // opt matches Some(t) && t.v == 5, // Panics
            opt matches Some(t), // Works
            opt is Some && opt->Some_0.v == 5,  // Works
            au.pred().args(&loc),
            perm@.patomic == loc.id(),
    {
        perm = update(perm).get();
        break;
    };
}

}
fn main() {}
```

And here's the output + stack trace I get:
```
[rustc_mir_build/src/builder/mod.rs:273:13] ids = (
    15,
    152,
)
[rustc_mir_build/src/builder/mod.rs:274:13] self.def_id = DefId(0:15 ~ repro_matches_workaround[5187]::caller::{closure#0})
[rustc_mir_build/src/builder/mod.rs:277:13] real_id.0.owner.def_id.local_def_index.as_usize() = 15
[rustc_mir_build/src/builder/mod.rs:278:13] self.fn_span = repro_matches_workaround.rs:28:22: 40:6 (#0)

thread 'rustc' (25043104) panicked at rustc_mir_build/src/builder/mod.rs:279:13:
Verus Internal Error: var_local_id failed: (15, 152)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <rustc_mir_build::builder::Builder>::var_local_id
   3: <rustc_mir_build::builder::Builder>::expr_as_place
   4: <rustc_mir_build::builder::Builder>::expr_as_place
   5: <rustc_mir_build::builder::Builder>::as_read_only_place
   6: <rustc_mir_build::builder::Builder>::expr_into_dest
   7: <rustc_mir_build::builder::Builder>::as_temp_inner
   8: <rustc_mir_build::builder::Builder>::as_operand
   9: <alloc::vec::Vec<rustc_middle::mir::syntax::Operand> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<rustc_middle::mir::syntax::Operand, core::iter::adapters::map::Map<core::iter::adapters::copied::Copied<core::slice::iter::Iter<rustc_middle::thir::ExprId>>, <rustc_mir_build::builder::Builder>::as_rvalue::{closure#1}>>>::from_iter
  10: <rustc_mir_build::builder::Builder>::as_rvalue
  11: <rustc_mir_build::builder::Builder>::expr_into_dest
  12: <rustc_mir_build::builder::Builder>::as_temp_inner
  13: <rustc_mir_build::builder::Builder>::as_operand
  14: <rustc_mir_build::builder::Builder>::as_call_operand
...
  47: rustc_mir_transform::ffi_unwind_calls::has_ffi_unwind_calls
      [... omitted 1 frame ...]
  48: rustc_mir_transform::mir_promoted
      [... omitted 1 frame ...]
  49: rustc_borrowck::borrowck_collect_region_constraints
  50: <rustc_borrowck::root_cx::BorrowCheckRootCtxt>::do_mir_borrowck
  51: rustc_borrowck::mir_borrowck
  52: rustc_query_impl::execution::try_execute_query::<rustc_data_structures::vec_cache::VecCache<rustc_span::def_id::LocalDefId, rustc_middle::query::erase::ErasedData<[u8; 8]>, rustc_middle::dep_graph::graph::DepNodeIndex>, false>
  53: <rustc_middle::ty::context::TyCtxt>::par_hir_body_owners::<<rust_verify::verifier::VerifierCallbacksEraseMacro>::run_lifetime_checks_on_verus_aware_items::{closure#0}>::{closure#0}
  54: rustc_data_structures::sync::parallel::par_for_each_in::<&rustc_span::def_id::LocalDefId, &[rustc_span::def_id::LocalDefId], <rustc_middle::ty::context::TyCtxt>::par_hir_body_owners<<rust_verify::verifier::VerifierCallbacksEraseMacro>::run_lifetime_checks_on_verus_aware_items::{closure#0}>::{closure#0}>
  55: <rust_verify::verifier::VerifierCallbacksEraseMacro>::run_lifetime_checks_on_verus_aware_items
  56: <rust_verify::verifier::VerifierCallbacksEraseMacro as rustc_driver_impl::Callbacks>::after_expansion
  57: rustc_interface::passes::create_and_enter_global_ctxt::<core::option::Option<rustc_interface::queries::Linker>, rustc_driver_impl::run_compiler::{closure#0}::{closure#2}>
  58: rustc_interface::interface::run_compiler::<(), rustc_driver_impl::run_compiler::{closure#0}>::{closure#1}
```

source/docs/guide/src/logatom-spec.md
```diff
@@ -0,0 +1,132 @@
...
+    },
+
+    requires private_pre(px),
+    ensures private_post(px, ax, ay, py),
+```
+
+Inside the `atomically` block, we find the `requires` and `ensures` clauses which denote the *atomic* pre- and postcondition, which describe the state of the program just before/after the linearization point (LP) respectively.
+
+To attach resources to the atomic pre- and postcondition, we can use the arrow `(ax: AX) -> (ay: AY)`.
+In what follows, we will refer to `ax: AX` as the "input type" of the atomic update, and `ay: AY` as the output type.
```
ahuoguo: nit: quote on "input type" but none on output type

source/vstd/atomic.rs
```diff
@@ -594,8 +595,452 @@ impl<T> PAtomicPtr<T> {
...
+unsafe impl<X: Send, Y: Send, Pred> Send for AtomicUpdate<X, Y, Pred> {}
+
+/// Unconditionally mark the `AtomicUpdate` as `Sync`.
+///
+/// # SAFETY
+/// A shared reference to an `AtomicUpdate` is pretty much useless.
+/// The only thing the user can do with an AU is open it, which requires full ownership.
+/// All methods provided by this type are spec-mode,
+/// meaning they can already be used with a much weaker ghost copy of the AU.
+unsafe impl<X, Y, Pred> Sync for AtomicUpdate<X, Y, Pred> {}
```
tjhance, line 624: can you add tests for this in rust_verify_test/tests/marker_traits.rs?

tjhance (approved):
I think we should add a disclaimer to the documentation that the feature is experimental.

But, I think it is ready to be merged.

PR #2331, opened 2026-04-14
Do not print verification results when --no_verify.

### LINKED ISSUE #2330: Wrong assumption in Verita about outputs.
When using verita to run verification for a crate depending on another verus-verified crate, verita will throw an error:
```
cannot parse verus output: expected value at line 1 column 1
got: verification results:: 0 verified, 0 errors
```

This also happens for ironkv in https://github.com/verus-lang/verus/actions/runs/24420764758/job/71341968760

ziqiaozhou:
Fixes #2330


parno (approved):
It seems reasonable to me that if the user specifies `--no-verify`, then we shouldn't produce a verification result report.

PR #2332, opened 2026-04-15
[cargo-verus] refactor: prepare to retire `fake-cargo`

mkovaxx:
- Refactor `cargo-verus` into two separate phases: planning and execution.
- Reorganize internals so that `main.rs` depends on `lib.rs`.
- Translate `test_build.rs` into `test_build_new.rs` to demonstrate how tests will be ported.

### DIFF
--- a/source/cargo-verus/src/main.rs
+++ b/source/cargo-verus/src/main.rs
@@ -20,20 +20,38 @@ pub mod test_utils;
 
 use crate::{
     cli::{CargoVerusCli, VerusSubcommand},
-    subcommands::CargoRunConfig,
+    subcommands::{CargoRunConfig, CargoRunPlan, NewCreationPlan},
 };
 
-pub fn main() -> Result<ExitCode> {
-    let parsed_cli = CargoVerusCli::from_args(env::args())?;
+fn main() -> Result<ExitCode> {
+    use ExecutionPlan::*;
+
+    let plan = plan_execution(env::args())?;
+
+    match &plan {
+        CreateNew(creation_plan) => subcommands::create_new_project(creation_plan),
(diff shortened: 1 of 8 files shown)


mkovaxx:
@parno @Chris-Hawblitzel @bsdinis The end goal here is to retire the `fake-cargo` binary and the `integration-tests` feature altogether.

This PR enables existing tests to be rewritten so that they no longer rely on that machinery. Only `test_build.rs` is translated here to keep the changes focused.

Note that this change leaves all existing tests intact to show that behavior is unaffected.

A follow-up change will translate all existing tests and also remove `fake-cargo` and the `integration-tests` feature.

source/cargo-verus/src/main.rs
```diff
@@ -12,58 +12,11 @@ use std::process::ExitCode;
...
-            compile_primary: false,
-            verify_deps: true,
-            warn_if_nothing_verified: true,
-        },
-    };
-
-    subcommands::run_cargo(cfg)
+fn main() -> Result<ExitCode> {
+    let args: Vec<String> = env::args().collect();
+    let plan = plan_execution(None, args.iter().map(String::as_str))?;
```
mkovaxx, line 19: The complexity is concentrated in `plan_execution` for testing to focus on.

source/cargo-verus/tests/test_build_new.rs
```diff
@@ -0,0 +1,114 @@
+use cargo_verus::{
```
mkovaxx, line 1: This file is added to show how all tests will be translated in the follow-up change.

parno (approved):
This looks like a nice move towards easier/faster tests.

PR #2343, opened 2026-04-16
Add Anvil to Verita

Catoverflow:
This PR adds Anvil to Verita as a part of the benchmark. Anvil's CI scripts are already updated to always build from the most recent Verus commit.


Catoverflow:
Currently to finish all proofs in Anvil it takes ~20mins on my laptop. I want to know if it's a concern @Chris-Hawblitzel @parno . If we want a smaller scale we can add `--verify-module vdeployment_controller` to check one out of all 4 controllers

parno:
Thanks for the PR and for the work you've done to get Anvil caught up with the latest version of Verus!  I think the current verita suite takes under 10 minutes on my laptop, so I think it'd be nice not to add an extra 20 minutes on top of that.  If the `vdeployment_controller` is representative of the other controllers, then maybe we should just focus on that one.  @Chris-Hawblitzel What do you think?

Catoverflow:
Thank you @parno! It should be very representative. I will also improve the proof speed of Anvil later.

Catoverflow:
I have updated the config to only verify VDeployment controller by default. Please let me know if there are other concerns. Thank you @parno !

Catoverflow:
<s>After Anvil main uses `1.95.0` I will remove draft tag</s>
It's ready

parno:
It looks like we ended up with a conflict between this PR and Ziqiao's PR to add svsm.  Would you mind resolving that?  Then we can merge this in.  Thanks for preparing it!

Catoverflow:
@parno I have resolved the conflict. Thanks!

PR #2353, opened 2026-04-19
Add stable id to VIR crate names to allow multiple crates with the same name

Chris-Hawblitzel:
This adds Rust's `u64` stable crate id to `CrateId` in `ast.rs`, allowing multiple crates with the same name:

```
/// Crate name, used at the beginning of a Path
#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub enum CrateId {
    // Verus-generated internal paths with no crate
    Internal,
    // Verus treats the vstd crate specially
    Vstd,
    // Verus treats the Rust core crate specially
    Core,
    // Verus treats the Rust alloc crate specially
    Alloc,
    // All other crates have Rust's stable crate id and a user-friendly name
    Id(u64, Ident),
}

/// A fully-qualified name, such as a module name, function name, or datatype name
pub type Path = Arc<PathX>;
#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub struct PathX {
    pub krate: CrateId,
    pub segments: Idents,
}
```

If there is only one crate with a name, the generated AIR code is the same as before this pull request.  If there are multiple crates with the same name, one crate will get the same AIR identifiers as before, while the others will have `!!2`, `!!3`, etc. inserted into their AIR identifiers.  For example, this is the generated AIR inside a cargo test included in this pull request that imports two crates, both named `spec_def`:

```
(= (spec_def!double.? (I x!)) (Sub (spec_def!!2!double.? (I x!)) 1))
```

In order to avoid unnecessary renamings, the `NameCtxt` assigns renamed crates lazily for any given AIR file, rather than preemptively generating crate names for all duplicate-named crates that rustc happens to import.

This pull request depends on three previous pull requests: https://github.com/verus-lang/verus/pull/2339 and https://github.com/verus-lang/verus/pull/2346 and https://github.com/verus-lang/verus/pull/2352 .


parno (approved):
Looks reasonable to me.  Thanks for working on this!

tjhance (approved):

PR #2359, opened 2026-04-22
Add wrapping shift to vstd

JakeGinesin:
This pull request adds support for `wrapping_shr` and `wrapping_shl` in vstd. 

### DIFF
--- a/source/rust_verify_test/tests/std.rs
+++ b/source/rust_verify_test/tests/std.rs
@@ -219,6 +219,26 @@ test_verify_one_file! {
     } => Err(err) => assert_one_fails(err)
 }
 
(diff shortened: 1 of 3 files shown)


tjhance (commented):
This looks fine, I think, but I think these should probably have tests.

the rust docs have lots of examples you can use https://doc.rust-lang.org/std/primitive.u16.html#examples-83

JakeGinesin:
Added some tests

tjhance (approved):

mmcloughlin:
Thanks for doing this @JakeGinesin! Could we merge it? Turns out I need it for a related project.

PR #2294, opened 2026-04-01
Allow const local variables

jaylorch:
Fixes #2023

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #2364, opened 2026-04-22
Support nested items (functions, consts, structs, enums) inside functions

Chris-Hawblitzel:
This replaces https://github.com/verus-lang/verus/pull/2294 .  Nested items in functions simply become top-level items in VIR, similar to how items inside modules work.  As discussed in https://github.com/verus-lang/verus/pull/2294 , this pull request extends the syntax macro to process nested items inside functions, ensuring that unerased proxies are generated.  This includes both the `verus!` syntax macro and the attribute-based syntax.


ziqiaozhou (commented):
What is the expected behavior for a nested function if its parent function is marked as external or external_body? Should that nested function also become external_body if no explicit #[verus_spec]?

tjhance:
The way external.rs determines this (which looks like it wasn't modified in this PR) is that any item nested inside an external or external_body item is also considered external.

ziqiaozhou (approved):
LGTM for #[verus_spec].

tjhance (approved):

PR #2365, opened 2026-04-23
Gate btree specs on alloc instead of alloc + std

tangruize:
`BTreeMap` and `BTreeSet` live in `alloc::collections`; `std` just re-exports them. Relax the feature gate from `cfg(all(alloc, std))` to `cfg(alloc)` so the specs are available in `no_std` environments.

- `std_specs/mod.rs`: change btree gate to `cfg(feature = "alloc")`
- `std_specs/btree.rs`: switch imports from `std::collections` to `alloc::collections`, add `use alloc::boxed::Box`

`hash` module stays `cfg(all(alloc, std))` since `HashMap` is std-only.

### DIFF
--- a/source/rust_verify_test/tests/btree.rs
+++ b/source/rust_verify_test/tests/btree.rs
@@ -5,6 +5,7 @@ use common::*;
 
 test_verify_one_file! {
     #[test] test_btree_map verus_code! {
+        extern crate alloc;
         use alloc::collections::BTreeMap;
         use vstd::prelude::*;
         fn test()
@@ -49,7 +50,7 @@ test_verify_one_file! {
 
 test_verify_one_file! {
     #[test] test_btree_set verus_code! {
-        use alloc::collections::BTreeSet;
+        use std::collections::BTreeSet;
         use vstd::prelude::*;
         fn test()
         {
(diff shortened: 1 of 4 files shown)


jaylorch:
I recommend also updating at least one of the BTreeMap tests in `source/rust_verify_test/tests/btree.rs` to use `alloc::collections::BTreeMap` instead of `std::collections::BTreeMap`. Same for `BTreeSet`.

jaylorch:
Also, the checks are failing. Remember to run `vargo fmt` before checking in.

tangruize:
@jaylorch Thanks! I've addressed these issues.

jaylorch:
Thanks! FYI, for future reference, in this repo we generally don't force-push updates; we just add additional commits to the PR.

jaylorch:
FYI, the test you added failed. (See the output of the workflow in the "Checks" tab above.)

tangruize:
@jaylorch Thanks! Added `extern crate alloc` to fix the test.

jaylorch (approved):
Thanks for your contribution!

bsdinis:
Thank you for this! I wonder if we should also put something like this for the HashMap sets!

jaylorch:
> Thank you for this! I wonder if we should also put something like this for the HashMap sets!

I don't think HashMap and HashSet are in alloc without std. See https://doc.rust-lang.org/stable/alloc/collections/

bsdinis:
Great point

PR #2367, opened 2026-04-23
Refactoring and readability improvements for specs on atomic ops

zero-to-nat:
Atomic ops now share common spec fns, and all orderings are consolidated into a single API.
- spec fns: `load_acquire`, `load_relaxed`, `store_release`, and `store_relaxed`
- helper specs for ^ to further improve reuse and readability

Other changes:
- Atomic ops output `Ghost` items into a single struct: `LoadData`, `StoreData`, `UpdateData`. This reduces the number of output arguments from these operations
- `History<T>` maps to `(T, View)` instead of `(T, Option<View>)` (when no view is required, we can simply use an empty view)
- the `View` trait is implemented on `ViewSeen`, `ReleaseViewSeen`, and `AcquireViewSeen`. I'm not sure if this will be confusing, since the term "view" is overloaded here....

### DIFF
--- a/source/vstd/atomic_relaxed.rs
+++ b/source/vstd/atomic_relaxed.rs
@@ -8,6 +8,7 @@ use core::sync::atomic::{
 #[cfg(target_has_atomic = "64")]
 use core::sync::atomic::{AtomicI64, AtomicU64};
 
+use super::pervasive::*;
 use crate::prelude::*;
 use crate::cell::CellId;
 use crate::pcm::*;
--- a/source/vstd/atomic_relaxed.rs
+++ b/source/vstd/atomic_relaxed.rs
@@ -377,6 +377,14 @@ pub axiom fn objective_from_acquire<T: Objective>(tracked a: Acquire<T>) -> (tra
 #[verifier::external_body]
 pub tracked struct ViewSeen;
 
+impl crate::view::View for ViewSeen {
+    type V = View;
+
+    open spec fn view(&self) -> View {
+        self.view()
+    }
+}
+
 impl ViewSeen {
     pub uninterp spec fn view(&self) -> View;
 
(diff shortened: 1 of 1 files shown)


source/vstd/atomic_relaxed.rs
```diff
@@ -377,6 +383,14 @@ pub axiom fn objective_from_acquire<T: Objective>(tracked a: Acquire<T>) -> (tra
 #[verifier::external_body]
 pub tracked struct ViewSeen;
 
+impl crate::view::View for ViewSeen {
```
LailaElbeheiry, line 387: yep that's very confusing. Can we maybe rename our `View` to `ThreadView`?

source/vstd/atomic_relaxed.rs
```diff
@@ -919,63 +961,60 @@ impl PWeakAtomicU8 {
...
-                    &&& pt.hist() == old(pt).hist()
-                    &&& vs_vr_opt.is_some()
-                    &&& vs_vr_opt.unwrap().view() == read_view
-                })
-            }),
+            match out.0 {
+                Ok(v) => {
+                    &&& current == v
+                    &&& out.2@.store_message_view.contains_strict(out.2@.load_message_view)
+                    &&& match success {
```
LailaElbeheiry, line 987: I think this post-condition is problematic because `load_acquire(..., old(v_sn)@, v_sn@, ..., out.2@.load_timestamp, ...)` will assert `load_timestamp_in_view` which in turn asserts that `out.2@.load_timestamp == v_sn@.get_timestamp(loc)`

Then `store_release(..., old(v_sn)@, v_sn@, ..., out.2@.load_timestamp + 1, ...)` will assert `store_timestamp_in_view` which in turn asserts that `out.2@.load_timestamp + 1 == v_sn@.get_timestamp(loc)`

I think you either need an intermediate view between the store and load (preferrable)
or, weaken `load_timestamp_in_view` maybe

(Same thing goes for the other ordering combinations)

LailaElbeheiry (changes requested):
Looks much cleaner. I think once the `compare_exchange` rule is fixed, it's ready to be merged.

source/vstd/atomic_relaxed.rs
zero-to-nat: thanks for catching this! i introduced an intermediate view as the "new view" in the load, and the "old view" in the store

LailaElbeheiry: Thanks! The specs now look good to me.

Issue #2366, opened 2026-04-23
Building Verus with custom target directory


slerpyyy:
It is quite common for Rust developers to configure one central target directory for all their Rust projects to keep project directories clean, allow build artifacts to be reused between projects, and to simplify managing disk space usage. My config file (`~/.cargo/config.toml`) looks like this, for example:
```toml
[build]
target-dir = "/home/aaron/.cargo/target/"
build-dir = "/home/aaron/.cargo/target/"
```
This setup currently breaks Verus, since the project expects all target directories to be in their default locations, relative to the project root. Verus relies on these assumptions without ever validating them.

To fix this, I can think of three approaches:

1. **Override the target directory**, ensuring all build artifacts are where we currently expect them to be.
This is the least amount of effort, but also the least satisfactory, since this means we have multiple target directories scattered throughout the project. Since there is currently no way to "unset" the target directory path (see [cargo#8687](https://github.com/rust-lang/cargo/issues/8687)), this would require setting the path explicitly in multiple places throughout the project.

2. **Respect the users choice**, make the build system flexible enough to work with the users preferred target directory. This would be ideal, but might be difficult since (afaik) the only reliable way to resolve the target directory path is using `cargo metadata --format-version 1 --no-deps`, and parsing the produced JSON object. There are also [multiple unit tests](https://github.com/verus-lang/verus/blob/29eadb3fba59c6d6236d5040405881b76fdd689d/source/rust_verify_test/tests/cargo.rs) that run `cargo clean` internally and needs to be reworked.

3. **Have one target directory for the entire project.** This might be a good middle ground between both solutions, it allows us to use hard-coded paths as before, and it's not too difficult to remove all build artifacts if you want to.

mkovaxx:
@slerpyyy Thanks for reporting this issue!

In the long term, I think Option 2 (Respect the users choice) would be ideal. As a first step, I'm making tests more robust to target directory overrides in #2490.

After that, I'll look into making changes to `vargo` itself.

PR #2368, opened 2026-04-23
Override target and build directory

slerpyyy:
Implements the "low effort" fix described in #2366


slerpyyy:
I'm reviving this PR

tjhance:
I don't understand the impact of this, so I'm adding @mkovaxx for review

slerpyyy:
**Status update:** This PR was originally like 50 LoC, but #2490 and #2686 have implemented almost everything I wanted, so now we're down to a single line of TOML.

This last line ensures that the build directory is configured correctly, even if it was overwritten by the environment. This is necessary because our build process for `verus` makes assumptions about where the build *artifacts* (not just the final executable) are located. Notably, `vargo` does not need this treatment, since we only assume the location of the executable.

mkovaxx (approved):
@slerpyyy Thanks for updating this!

My understanding is that `build-dir` defaults to `target-dir`, so this change has an effect only in the presence of an explicit override. Because of that, it should not affect usual Verus build flows in this repo.

CI passes too so I think this is safe to merge!

PR #2377, opened 2026-04-25
Add trait bound to spec == operator to improve type inference

Chris-Hawblitzel:
Note: this is a breaking change (although it appears to break only a few lines of code in the Verita benchmarks).

Ever since Verus's syntax macro was introduced, the `==` operator in spec mode has been translated into the `spec_eq` operator with the following signature:

```
pub fn spec_eq<Lhs, Rhs>(_lhs: Lhs, _rhs: Rhs) -> bool
```

This means that we asked the Rust type checker to accept any argument types `Lhs` and `Rhs` to `==`, and then after Rust type checking we applied our own additional checks to make sure that the argument types actually matched according to our rules, which allow direct equality comparisons between any integer types.  The primary reason for this arrangement was so that programmers didn't have to clutter their code with things like `x as int == y` -- they could just say `x == y`.  Unfortunately, this reduced the effectiveness of Rust's type inference.  For example, type inference couldn't handle expressions like `s == Seq::empty()` because the type of `Seq::empty()` was underconstrained, so you had to either write something like `s == Seq::<u8>::empty()` or use the relatively obscure `===` operator.  I would prefer if people could just write `s == Seq::empty()`.  This pull request enables this by making the type signature for spec_eq more precise:

```
pub fn spec_eq<Lhs: SpecEq<Rhs>, Rhs>(_lhs: Lhs, _rhs: Rhs) -> bool
```

where SpecEq is a trait that defines exactly what `==` operations are allowed.  This includes all instances where the left-hand and right-hand types are exactly equal (e.g. comparing `Box<(u8, u64)>` to `Box<(u8, u64)>`):

```
impl<A: ?Sized> SpecEq<A> for A {}
```

as well as allowances for one side having an extra `&`, `Ghost`, `Tracked`, etc., to match much of our previous policy (which simply ignored all decorations like `&`, `Ghost`, `Tracked`, etc.):

```
#[cfg(verus_keep_ghost)]
impl<A: ?Sized> SpecEq<&A> for A {}

#[cfg(verus_keep_ghost)]
impl<A: ?Sized> SpecEq<A> for &A {}

...
```

Finally, there are impls of SpecEq for all integer comparisons.

Using this, I was able to search and replace all occurrences of `===` with `==` in the examples directory and in vstd, and these all worked immediately.

The SpecEq trait isn't quite as permissive as the previous policy that ignored all decorations in `==`.  For example, it doesn't handle nested cases like comparing `Tracked<Ghost<T>>` to `T`.  For this, you have to write something like `*x == y` instead of just `x == y`.  But I think this matches user expectations, since you have to add `*` for normal Rust exec `==` comparisons anyway.  Also, I didn't add cases for `Box`, `Rc`, and `Arc`, because I wanted to avoid any complications of making `builtin` depend on `alloc` or `std` -- in this pull request, `builtin` remains strictly no_std.

Since SpecEq is more restrictive than the previous policy, this pull request is a breaking change.  However, only 6 lines of code in all of Verita broke ( https://github.com/verus-lang/verus/actions/runs/24913442522 ).  Furthermore, 4 of the 6 lines appear to be accidental code: 3 cases of using `&& a1 == b1 && a2 == b2 && a3 == b3 ...` instead of `&&& a1 == b1 &&& a2 == b2 &&& a3 == b3 ...` (causing the first `&& a1` to be parsed as two presumably unintentional address-of operations) and 1 case of a missing `.view()` that was previously allowed because the code appeared (superfluously) inside a `verifier::external` function.  The other 2 of the 6 lines simply needed an extra `*` to dereference a left-hand-side `Box<T>` when comparing to a right-hand-side `T`.

If the pull request is accepted, I will announce the pull request on Zulip at least a week before merging it.  I also have fixes ready for all 6 cases where the Verita benchmarks broke, for which I'll make pull requests to the respective repositories.


tjhance (commented):
can we add a check that people don't try to use SpecEq generically? People have been confused by SpecOrd in the past.

Chris-Hawblitzel:
> can we add a check that people don't try to use SpecEq generically? People have been confused by SpecOrd in the past.

Ok, I added this check.

parno (approved):
Seems like a nice upgrade to me, and the breaking changes look pretty minor.  After this PR, are there still use cases for `===`?

Chris-Hawblitzel:
> After this PR, are there still use cases for `===`?

As far as I know, after type checking `==` and `===` are translated to the same VIR, so I think the only use for `===` is to improve type inference.  In principle, there are still cases where `===` has more precise type inference, such as `proof fn test(i: int) { assert(i === arbitrary()); }`.  But I'm not sure there are places where it would still be used in practice.  I haven't done a thorough search.

parno:
> But I'm not sure there are places where it would still be used in practice. 

Should we consider deprecating/eliminating `===` then?

PR #2383, opened 2026-04-27
[cargo-verus] refactor: remove `fake-cargo` and update all tests

mkovaxx:
- Migrate old tests to new style that does not depend on `fake-cargo`.
- Update CI setup and documentation.
- Tiny refactor in `cli` to `return Err` instead of failing via `process::exit(2)`.
- Make unit test in `metadata.rs` stable using `--offline`.
- Address Clippy suggestions in `test_util.rs`.


.github/workflows/ci.yml
```diff
@@ -111,10 +111,10 @@ jobs:
         with:
           toolchain: 1.95.0
 
-      - name: cargo-verus integration tests
+      - name: cargo-verus tests
         if: needs.change_filter.outputs.cargo_verus == 'true'
         working-directory: ./source
-        run: cargo test -p cargo-verus --features integration-tests -- --skip z_flag_with_space_is_accepted
```
bsdinis, line 117: what is the problem with the test?

bsdinis (commented):
Thank you for putting this together!
It looks good to me but might be good to get @parno 's input

.github/workflows/ci.yml
mkovaxx: The `z_flag_with_space_is_accepted` test used to require a `nightly` toolchain to work, and that's fixed in this PR, so we no longer need to skip it.

parno (approved):
Looks good to me.

PR #2385, opened 2026-04-28
new-mut-ref: fix obscure loop invariant evaluation order issue

tjhance:
Now that borrow checking has to be aware of spec variables to check for prophecy-soundness, there's a small issue with loops, namely that the way they are encoded in Verus source / HIR
does not correspond to their conceptual "evaluation" points, e.g., a while-loop would be desugared
as thus:

```rust
loop {
    if cond() {
        ::verus_builtin::invariant(...);
        body;
    } else {
        break;
    }
}
```

when the conceptual evaluation point of the `invariant` is at the beginning/end of the loop.
Same issue for `invariant_except_break` / `ensures` / `decreases`
(and remember that the prophecy check is essential for soundness when it comes to decreases).

To resolve this, we have to handle all loop headers specially. First, we simply ignore them
when we encounter them during the main traversal (they should be marked EraseAbsolutely).
Then, when handling the Loop node itself, we insert the headers' shadow usages at the
appropriate program points.

The meaning of any loop header fn is given by this in erase.rs:

```rust
impl LoopSpecKind {
    fn loop_spec_evaluation_location(&self) -> LoopSpecEvaluationLocation {
        match self {
            LoopSpecKind::Invariant => LoopSpecEvaluationLocation::BodyStartAndPostLoop,
            LoopSpecKind::Decreases => LoopSpecEvaluationLocation::BodyStart,
            LoopSpecKind::Ensures => LoopSpecEvaluationLocation::PostLoop,
            LoopSpecKind::InvariantExceptBreak => LoopSpecEvaluationLocation::BodyStart,
        }   
    }   
}
```

Please check that this mapping makes sense.

I think the meaning of Invariant here (BodyStartAndPostLoop) is not quite right, as the meaning of 'invariant' for a while loop depends on whether or not there's a break statement.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -2325,6 +2325,12 @@ pub fn shadow_ghost_value<S, T>(_: S) -> T {
 }
 
 #[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::verus_erasure_get_first"]
+pub fn verus_erasure_get_first<S, T>(s: S, _: T) -> S {
+    s
+}
+
+#[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::verus_builtin::DummyCapture"]
 #[derive(Clone, Copy)]
 pub struct DummyCapture<'a> {
--- a/source/rust_verify/src/context.rs
+++ b/source/rust_verify/src/context.rs
@@ -57,7 +57,7 @@ pub enum HeaderSetting {
     /// Including closures
     Fn,
     /// Loops (invariants, ensures, etc.)
-    Loop,
+    Loop(HirId),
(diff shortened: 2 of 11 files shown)


Chris-Hawblitzel:
>            LoopSpecKind::Invariant => LoopSpecEvaluationLocation::BodyStartAndPostLoop,
>            LoopSpecKind::Decreases => LoopSpecEvaluationLocation::BodyStart,
>            LoopSpecKind::Ensures => LoopSpecEvaluationLocation::PostLoop,
>            LoopSpecKind::InvariantExceptBreak => LoopSpecEvaluationLocation::BodyStart,

> Please check that this mapping makes sense.

I think it corresponds exactly to how we translate loops with loop_isolation(true), and it's conservative for loop_isolation(false), which is fine.

Chris-Hawblitzel (approved):

PR #2393, opened 2026-04-30
new-mut-ref: release

tjhance:
* Move the new-mut-ref guide page to the actual guide, update cross-links with the migration page
 * Remove the "-V new-mut-ref" command line option
 * Update all tests and examples


source/rust_verify_test/tests/common/mod.rs
```diff
@@ -330,8 +330,9 @@ pub fn run_verus(
         } else if *option == "--disable-internal-test-mode" {
             use_internal_test_mode = false;
         } else if *option == "new-mut-ref" {
-            verus_args.push("-V".to_string());
-            verus_args.push("new-mut-ref".to_string());
+            // TODO remove new-mut-ref from the test harness
```
parno, line 333: Should this be deleted at this point?

parno (approved):
Exciting to see this moving into active production!  

I noticed that you removed the links to `requires-ensures-mut-ref.md` and `assert-mut-ref.md`.  Are they obsolete at this point?  If so, should we delete them?

tjhance:
actually the assert one might still be relevant. i'll take another look. you're right the files should be deleted if they aren't linked.

regarding the code: there is an _enormous_ amount of dead code, i will fix it in some follow-up, not in tgis PR.

Chris-Hawblitzel (approved):

PR #2395, opened 2026-04-30
Remove unnecessary assume in pcm_lib

jaylorch:


jaylorch:
I noticed that @bsdinis had left out `validate_3` in his upcoming PR porting `pcm_lib`, and wondered why. Then I noticed the likely reason: It had an `assume` in it. But this `assume` is unnecessary, so it should probably be ported as well.

bsdinis:
Oh, that's great! I think I'm going to merge my PRs first and then this can follow right after!

jaylorch:
OK, sounds good. Though I'll presumably have to make an update or two to it to make it compatible with the new pcm stuff.

jaylorch:
Oh, I see you already fixed the assume issue. I'll just close this redundant PR.

PR #2397, opened 2026-05-01
vstd: make persistent Map/Set duplication require shared refs

bsdinis:
`duplicate` on `GhostPersistent{Map,Set,..}` requires a mutable reference, which somewhat defeats the purpose. This removes that requirement.


tjhance:
can you summarize the change? there's a lot more new code than I expected given the PR title

bsdinis:
Sorry, the change should be self explanatory but I pushed more commits than I should've

bsdinis:
Updated the commits and the description

tjhance (approved):

PR #2403, opened 2026-05-03
[cargo-verus] fix: import .vir for transitive verified deps

Kab1r:
When a verified crate references items from a deeper transitive dependency by canonical path through a re-exporting facade, rustc loads the deeper crate's metadata even though it is not in `--extern`. Verus also needs the matching .vir, but `cargo-verus` only emitted `--import-dep-if-present` for direct deps and the verus driver only matched names against direct externs, so the deeper .vir was missing and Verus failed to recognize spec items declared there.

Walk the transitive dep closure when building the per-package args, and in the verus driver resolve the leftover names against the actual crate sources rustc loaded (`tcx.used_crate_source`), reading the .vir from alongside the artifact rustc actually picked.

Enable the `transitive3` cargo-test fixture, which previously had `test_ignore = true` to mark this scenario as known-broken.

### DIFF
--- a/source/cargo-verus/src/subcommands.rs
+++ b/source/cargo-verus/src/subcommands.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap as Map, BTreeSet as Set, VecDeque};
+use std::collections::{BTreeMap as Map, BTreeSet as Set};
 use std::env;
 use std::path::PathBuf;
 use std::process::{Command, ExitCode};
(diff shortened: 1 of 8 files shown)


mkovaxx:
@Kab1r Thanks for proposing a fix here!

As a first iteration, could you please add a test in `cargo-verus/tests` to confirm expected behavior?

source/cargo-verus/src/subcommands.rs
```diff
@@ -391,12 +391,40 @@ fn make_cargo_plan(
...
+                    };
+                    if let Some(import_name) = import_name {
+                        verus_driver_args_for_package.extend_from_slice(&[
+                            "--VIA-CARGO".to_owned(),
+                            format!("import-dep-if-present={import_name}"),
+                        ]);
+                    }
+                }
+                for d in dep_entry.deps() {
+                    queue.push_back((&d.pkg, None));
```
mkovaxx: The [`MetadataIndex::get_transitive_closure` method](https://github.com/verus-lang/verus/blob/fccc8bca49388563e84eb6f06003e31e28e93049/source/cargo-verus/src/metadata.rs#L84-L98) is responsible for computing transitive dependencies, so these additions should be made to `MetadataIndex` instead to separate the concerns of resolving and processing the dependency information.

https://github.com/verus-lang/verus/blob/fccc8bca49388563e84eb6f06003e31e28e93049/source/cargo-verus/src/metadata.rs#L84-L98

mkovaxx (changes requested):
I'll focus on the changes affecting `cargo-verus` for now and defer to more knowledgeable people about dependency logic in `rust_verify`.
CC @jsenn

source/rust_verify/src/cargo_verus.rs
jsenn, line 139: does the new code handling transitive dependencies have an equivalent of this line? I think this is important to make sure that cargo knows about the dependency's VIR.

Without this, I think you could have bugs where you change only the spec code, so the rmeta doesn't change. When you run `cargo verus verify`, cargo won't know that the VIR spec has changed, so it will report stale results until you do a clean or a full build.

source/rust_verify/src/cargo_verus.rs
```diff
@@ -149,6 +149,31 @@ pub(crate) fn handle_externs(
...
+            continue;
+        }
+        let source = tcx.used_crate_source(*cnum);
+        let artifact_path =
+            source.rlib.as_ref().or(source.rmeta.as_ref()).or(source.dylib.as_ref());
+        let Some(artifact_path) = artifact_path else { continue };
+        let vir_path = artifact_path.with_extension("vir");
+        if vir_path.exists() {
+            imports.push((name, vir_path.display().to_string()));
+        }
```
jsenn, line 175: if that path *doesn't* exist, should that be an error? I can't think of a case where that would actually happen, but `handle_externs` has an explicit error in this case

parno:
@jsenn and @mkovaxx, I believe @Kab1r has addressed your comments -- would you mind double checking?

@Kab1r After they check, can you run `vargo fmt`, so that we can get a clean run of the CI?

source/rust_verify/src/cargo_verus.rs
jsenn: latest LGTM, I can't close the comment

source/rust_verify/src/cargo_verus.rs
jsenn: latest LGTM

jsenn:
> @jsenn and @mkovaxx, I believe @Kab1r has addressed your comments -- would you mind double checking?

I don't have the power to resolve my comments or approve the PR but they were addressed!

Kab1r:
I ran `vargo fmt` so CI should be clean and marked the reviews that have been looked over a second time as resolved.

mkovaxx (approved):

parno:
Thanks everyone!

PR #2413, opened 2026-05-04
vstd: more borrow_mut like functions

tjhance:

### DIFF
--- a/source/vstd/map.rs
+++ b/source/vstd/map.rs
@@ -4,7 +4,8 @@ use super::pervasive::*;
 use super::prelude::*;
 use super::set::*;
 
-verus! {
+use verus as verus_; // skip verusfmt due to unhandled return-value-pattern
+verus_! {
 
 /// `Map<K, V>` is an abstract map type for specifications.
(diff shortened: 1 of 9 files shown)


source/vstd/map.rs
```diff
@@ -144,6 +144,26 @@ impl<K, V> Map<K, V> {
...
+    /// Index into a tracked map, getting a tracked mutable borrow of the value
+    pub axiom fn tracked_borrow_mut(tracked &mut self, key: K) -> (tracked v: &mut V)
+        requires
+            self.dom().contains(key),
+        ensures
+            *v === old(self).index(key),
+            *final(self) === old(self).insert(key, *final(v))
+    ;
+
+    /// Index into a tracked map, getting a tracked mutable borrow of the value
```
Chris-Hawblitzel: Is this the right comment?

Chris-Hawblitzel (approved):

source/vstd/map.rs
tjhance: no, good catch

PR #2417, opened 2026-05-05
implement an operator `shr_ref_struct_wrap`

tjhance:
Right now it's difficult to create ghost state based abstractions because if you have a struct Foo like:

```
tracked struct Foo {
    tracked bar: Bar
}
```

There is no way to go from `tracked &Bar` to `tracked &Foo`. This isn't a normal Rust operation, but it's perfectly fine in some cases. It's basically a transmute, but for tracked state, we don't have to worry about complicated things like layout.

Balthasar recently ran into this problem with his resource algebra changes and had to work around it with an ugly `ResourceRef` type in place of `&Resource`. There are also abstractions I want to build for which this is a blocker.

This PR implements an operator `shr_ref_struct_wrap` which lets you go from a shared reference to a field to a shared reference of a struct. Since it doesn't really correspond to a common Rust operation, there's no easy notation for it. I did the same thing we did for `get_variant_field` and created a builtin function that takes string literals for the variant & field names. Ugly, but I figure this feature is mostly for advanced library writers.

The PR also updates the resource algebra library to remove the `ResourceRef` and just use `&Resource`.

Example:

```
        tracked struct Y { 
            ghost y: int 
        }   

        tracked struct X { 
            ghost g: int,
            tracked y: Y,
        }   

        proof fn test(tracked y: &Y) -> (tracked x: &X) 
            ensures x.y == y && x.g == 0
        {   
            shr_ref_struct_wrap(y, &X { g: 0, y: arbitrary() }, "", "y")
        }   
```

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -2407,3 +2407,14 @@ pub fn after_borrow<T>(_: T) -> T {
 pub fn mut_ref_tracked<T>(_: &mut T) -> &mut Tracked<T> {
     unimplemented!()
 }
+
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::shr_ref_struct_wrap"]
+pub fn shr_ref_struct_wrap<'a, 'b, A, B>(
+    _: &'a A,
+    _: &'b B,
+    _variant: &'static str,
+    _field: &'static str,
+) -> &'a B {
+    unimplemented!()
+}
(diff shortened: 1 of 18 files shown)


bsdinis:
This is very cool! Is a good mental model that this is some sort of `tracked Ref/Deref`?

bsdinis:
I think it generally makes sense, have some just some questions:
- Could this have soundness implications?
- Is there an issue with lifetimes (effectively, we are "erasing" the super structure and holding onto the shared ref to the field)?

tjhance:
It's pretty trivially true in the VerusBelt model as Ghost doesn't own any resources, and there's no physical structure, so the two types are just isomorphic. The lifetime  is preserved by the signature of the builtin function (there's a test for this).

bsdinis:
Sounds good. I'll let Chris approve (I don't know anything about this part of the codebase).

Thank you so much for working on this!

Chris-Hawblitzel (approved):

PR #2420, opened 2026-05-06
Add specs for HashMap entry API

tjhance:
Examples (examples/entry_api.rs)

```rust
fn main() {
    let mut m = HashMap::<u64, u64>::new();

    // Use the entry to insert 0 => 20
    let entry0 = m.entry(0);
    entry0.insert_entry(20);

    // Use the entry to insert 1 => 30, then change it to 40
    let entry1 = m.entry(1);
    let mut occupied_entry = entry1.insert_entry(30);
    let value_ref = occupied_entry.get_mut();
    *value_ref = 40;

    assert(m@ =~= map![0 => 20, 1 => 40]);
}

/// Inserts the given key, value pair into the map if absent,
/// does nothing otherwise
fn insert_if_absent(m: &mut HashMap<u64, u64>, key: u64, value: u64)
    ensures
        !old(m)@.dom().contains(key) ==> final(m)@ =~= old(m)@.insert(key, value),
        old(m)@.dom().contains(key) ==> final(m)@ =~= old(m)@,
{
    m.entry(key).or_insert(value);
}

/// If the key is absent, insert 1; else, double it
fn try_double(m: &mut HashMap<u64, u64>, key: u64)
    requires
        m@.dom().contains(key) ==> m[key] * 2 <= u64::MAX,
    ensures
        !old(m)@.dom().contains(key) ==> final(m)@ =~= old(m)@.insert(key, 1),
        old(m)@.dom().contains(key) ==> final(m)@ =~= old(m)@.insert(key, (old(m)@[key] * 2) as u64)
{
    match m.entry(key) {
        Entry::Occupied(mut occ_entry) => {
            *occ_entry.get_mut() *= 2;
        }
        Entry::Vacant(vac_entry) => {
            vac_entry.insert(1);
        }
    }
}
```


source/vstd/std_specs/hash.rs
```diff
@@ -1428,6 +1428,236 @@ pub broadcast proof fn axiom_hashset_decreases<Key, S, A: Allocator>(m: HashSet<
...
+ensures
+    *value == entry.value();
+
+pub assume_specification<'a, 'b, K, V, A: Allocator> [ OccupiedEntry::get_mut ]
+    (entry: &'b mut OccupiedEntry::<'a, K, V, A>) -> (value: &'b mut V)
+ensures
+    *value == old(entry).value(),
+    final(entry).key() == old(entry).key(),
+    final(entry).value() == *final(value),
+    final(entry).final_value() == old(entry).final_value();
```
jaylorch: Should there also be the postcondition `*(entry).value() == *value`?

source/vstd/std_specs/hash.rs
```diff
@@ -1428,6 +1428,236 @@ pub broadcast proof fn axiom_hashset_decreases<Key, S, A: Allocator>(m: HashSet<
...
+ensures
+    *value === entry.value(),
+    entry.final_value() == Some(*final(value));
+
+pub assume_specification<'a, K, V, A: Allocator> [ OccupiedEntry::insert ]
+    (entry: &mut OccupiedEntry::<'a, K, V, A>, value: V) -> (old_value: V)
+ensures
+    old_value == old(entry).value(),
+    final(entry).key() == old(entry).key(),
+    final(entry).value() == value,
```
jaylorch: Should there also be `*(entry).value == value`?

source/rust_verify_test/tests/std.rs
```diff
@@ -745,3 +745,164 @@ test_verify_one_file! {
...
+                    assert(v == 40);
+                }
+                Entry::Vacant(_) => {
+                    assert(false);
+                }
+            }
+
+            assert(!m@.dom().contains(5));
+
+            assert(false); // FAILS
```
jaylorch, line 807: Why does the test have to end with `assert(false)`? Can't it just be a successful test?

source/rust_verify_test/tests/std.rs
```diff
@@ -745,3 +745,164 @@ test_verify_one_file! {
...
+            let entry = m.entry(5);
+            let mut occ_entry = entry.insert_entry(60);
+
+            let (removed_key, removed_value) = occ_entry.remove_entry();
+            assert(removed_key == 5);
+            assert(removed_value == 60);
+
+            assert(m@ =~= Map::empty());
+
+            assert(false); // FAILS
```
jaylorch, line 846: Why does the test have to end with `assert(false)`? Can't it just be a successful test?

source/vstd/std_specs/hash.rs
tjhance: it's on line 1605

source/rust_verify_test/tests/std.rs
tjhance: it checks that aren't any contradictory propositions in the specifications.

source/vstd/std_specs/hash.rs
tjhance: the value being returned is named `old_value`, and we have `old_value == old(entry).value()` on line 1619.

source/vstd/std_specs/hash.rs
jaylorch: Line 1605 is talking about `old(entry)`, which I thought meant "entry at the beginning of the function". Isn't that different from `*entry`, which I thought meant "entry at the end of the function, but not necessarily at the end of its lifetime"?

source/rust_verify_test/tests/std.rs
jaylorch: Ah, I see.

source/rust_verify_test/tests/std.rs
jaylorch: Ah, I see.

jaylorch:
Note also that there's a failing formatting check.

source/vstd/std_specs/hash.rs
tjhance: There's `*old(entry)` and `*final(entry)`, no plain `entry`.

So there are really four values of interest:

 - `old(entry).value()` - value of the entry at function start
 - `old(entry).final_value()` - value of the entry when the Entry expires
 - `final(entry).value()` - value of the entry when the mutable reference expires
 - `final(entry).final_value()` - value of the entry when the Entry expires (so identical to `old(entry).final_value()`)

tjhance: There's no way to reason about "entry at the end of the function, but not necessarily at the end of its lifetime" through the `entry` variable.

jaylorch: Oh, interesting, I hadn't realized that.

jaylorch: I was somewhat confused by this, so I asked Chris for clarification. He says that an `&mut ref` parameter like `entry` is reborrowed at the call site, so its expiration will always be coincident with the return from the function. So if you want to reason about `entry` upon return from the function, you can use `final(entry)`. After all, the lifetime of `entry` will expire at that point.

jaylorch:
Let me know when you've fixed the formatting issue and I'll approve.

source/vstd/std_specs/hash.rs
tjhance: That's true — provided the expiration _is_ coincident with the return from the function (which is true in the common case, i.e., all functions Verus has supported up until recently).

For this particular function, which has type `(entry: &'b mut OccupiedEntry::<'a, K, V, A>) -> (value: &'b mut V)`, the lifetime `'b` might expire at some point after the end of the function.

source/vstd/std_specs/hash.rs
```diff
@@ -1428,6 +1430,278 @@ pub broadcast proof fn axiom_hashset_decreases<Key, S, A: Allocator>(m: HashSet<
...
+    #[verifier::prophetic]
+    open spec fn final_value(self) -> Option<V> {
+        match self {
+            Entry::Occupied(occupied_entry) => occupied_entry.final_value(),
+            Entry::Vacant(vacant_entry) => vacant_entry.final_value(),
+        }
+    }
+}
+
+pub broadcast axiom fn axiom_has_resolved_occupied_entry<K, V>(entry: OccupiedEntry<K, V>)
```
parno: Should these broadcasts take an `A: Allocator` parameter to account for custom allocators?  Or is it deliberately omitted to limit the scope to the Global allocator?

parno (approved):
It's exciting to see what we can do with the new mutable reference support!

source/vstd/std_specs/hash.rs
Chris-Hawblitzel: One thing I was talking about with Jay is that if you have a function like `fn f(x: &mut t) { ... }`, then the `&mut t` is owned by `f` and is never passed back (since there's no return value), so the `&mut t` dies by the end of `f`'s execution and its final value can't change afterwards.  But I guess the more interesting case going on here is where `f` returns the reference or something derived from the reference, as in `fn f<'b>(x: &'b mut t) -> (r: &'b mut t) { ... modify *x...; x })`.  In this case, the `&mut t` outlives `f`'s execution and its final value may be different from what the value was at the end of `f`'s execution.  If I understand correctly, in this case we would not talk about the value referenced by the parameter `x` at the end of `f` -- there's no expression for that and `*x` wouldn't mean that -- but rather, we'd talk about the value referenced by the returned `r: &'b mut t` at the end of `f`, using `*r`.  Ultimately, we'll be able to relate this `*r` to whatever it was we originally borrowed from at the call to `f`.

jaylorch: Whoa, OK, this is a lot more complicated than I thought. But I understand now; thanks very much for explaining!

tjhance: It's worth making this explicit: the role of the `Entry` is to mimic a mutable reference to a particular entry, so having `&mut Entry` is a lot like having a `&mut &mut T`. That's part of what makes this example so complicated.

Ordinary, you wouldn't have a `&mut &mut T`, because usually if you had a `&mut T` and wanted to pass it mutably, you just reborrow and get another `&mut T`. But you can't reborrow an Entry, so the only way to pass it mutably is to take a mutable borrow, `&mut Entry`.

PR #2423, opened 2026-05-06
deprecate the set function for arrays/slices

tjhance:


parno (approved):
Exciting to see more payoff from the  new mutable reference support!

tjhance:
actually I think these have been redundant for a while

PR #2424, opened 2026-05-07
flesh out prophecy lib & docs

tjhance:
This adds a new Prophecy variable library based on the verifier::prophetic attribute to ensure soundness. Both libraries remain useful. I also wrote a lot of documentation that explains why there are 2 libraries, what the tradeoffs are, and what the soundness issues are.

### DIFF
--- a/source/docs/guide/src/reference-attributes.md
+++ b/source/docs/guide/src/reference-attributes.md
@@ -151,18 +151,76 @@ Cannot be used together with [`custom_err`](#verifiercustom_errtext-and-verifier
 
 ## `#[verifier::prophetic]`
 
-This can be applied to any `spec` function to indicate that the result is _prophecy-dependent_.
-This attribute is required on any `spec` function where the body is prophecy-dependent.
+This attribute is used to mark a value as _prophecy-dependent_, or _prophetic_, meaning that its value may depend
+on a value from the program's future, e.g., the `final` operator (used for [mutable references](./mutable-references.md)),
+or the [value of a prophecy variable](https://verus-lang.github.io/verus/verusdoc/vstd/proph/struct.ProphecyGhost.html#method.value).
+This attribute aids Verus in tracking which spec values are prophecy-dependent, which are restricted in certain contexts:
 
-For example, a spec function needs to be marked with this attribute if any of the following
-are true:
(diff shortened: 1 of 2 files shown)


source/docs/guide/src/reference-attributes.md
```diff
@@ -148,6 +149,21 @@ When a proof obligation (`requires`/`ensures`/`assert`) fails, then the `"text"`
 
 Cannot be used together with [`custom_err`](#verifiercustom_errtext-and-verifiercustom_errtext).
 
+## `#[verifier::prophetic]`
```
parno, line 152: This is a very operational explanation of the attribute.  Should we also provide at least a small hint about the conceptual side?  I.e., what/why you might use/encounter this attribute?

Chris-Hawblitzel (approved):
Should we also have `resolve_push` in `ProphecySeq`, with `old(self).seq() == final(self).seq().push(v)`?  Actually, it seems like any `spec_fn` on the `Seq<T>` would be ok.  Should we have a more general case like:

```
pub tracked struct ProphecyMut<T> {
    tracked var: ProphecyGhost<T>,
}

impl<T> ProphecyMut<T> {
    #[verifier::prophetic]
    pub closed spec fn value(&self) -> T {
        self.var.value()
    }

    pub proof fn new() -> (tracked proph_var: Self) {
        ProphecyMut { var: ProphecyGhost::new() }
    }

    pub proof fn resolve(tracked &mut self, f: spec_fn(T) -> T)
        ensures
            old(self).value() == f(final(self).value()),
    {
        let tracked mut var = ProphecyGhost::new();
        vstd::modes::tracked_swap(&mut var, &mut self.var);
        var.resolve_dependent(&self.var, f);
    }
}
```

tjhance:
Yes, ProphecyMut would be more general than ProphecySeq

tjhance:
I mostly added ProphecySeq for illustrative purposes (since it's a feature of the Future is Ours paper, but conspicuously missing from the Prophecy(non-Ghost) lib). In the Future is Ours paper, the prophecy sequence takes this shape for a reason. But it really doesn't have to take this shape here. We could define `ProphecyMut<T>` and then `ProphecySeq<T>` is just `ProphecyMut<Seq<T>>`.

PR #2425, opened 2026-05-07
Minor formatting tweak

jaybosamiya-ms:
Verusfmt recently removed an ugliness from some statements (https://github.com/verus-lang/verusfmt/pull/185).  This is _technically_ backwards-incompatible, but is clearly an improvement in readability.  Nonetheless, it makes for a minor change to a couple of lines vstd, so we need to coordinate a release.

### DIFF
--- a/source/vstd/string.rs
+++ b/source/vstd/string.rs
@@ -234,13 +234,13 @@ impl StrSliceExecFns for str {
             }
             if char_pos == to {
                 byte_end = Some(byte_pos);
-                break ;
+                break;
             }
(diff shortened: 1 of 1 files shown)


jaybosamiya-ms:
We need to coordinate the release of verusfmt 0.7.0 with the queue here, because otherwise we get CI failures.  I've kicked off a release on the verusfmt side, then will re-trigger the CI here once the release there succeeds, so that it all aligns up as expected.

Edit: retriggered + queued

PR #2435, opened 2026-05-11
Don't produce orphaned temporary variables for failing recommends checks in trait impls

### LINKED ISSUE #2434: Panic with opaque type in a recommends clause
I ran into this panic while adapting the verified-storage projects to the new mutable reference framework.  The code below is a minimized reproduction, and it produces:
```
thread '<unnamed>' (12766315) panicked at vir/src/sst_to_air.rs:172:25:
abstract datatype should be boxed Datatype(Path(Path(repro, ["opaque_mod" :: "Opaque"])), [TypParam("T")], [])
```
According to claude, the issue appears to be during SST -> poly lowering for the impl's `recommends_check`, the requires-clause processing in trait-method mode emits orphan `TempViaAssign` locals whose types are the opaque generic (here `Opaque<T>`). `update_temp_locals` skips locals still in `state.remaining_temps`, so their type is never coerced via `coerce_typ_to_native` (i.e. never boxed). Later, `typ_to_air` panics because the type is neither a monomorphic instance nor `Boxed`.

This seems plausible to me, and a fix based on this hypothesis does seem to clear up the issue, so I'll open a PR for that fix.

```
use vstd::prelude::*;

verus! {

mod opaque_mod {
    use vstd::prelude::*;
    pub tracked struct Opaque<T> {
        t: T,
    }
    impl<T> Opaque<T> {
        pub closed spec fn view(self) -> T { self.t }
    }
}
use opaque_mod::Opaque;

pub trait Tr<T> : Sized {
    proof fn f(s: Seq<Opaque<T>>, v: T)
        requires s[0]@ == v;
}

pub struct S<T>(core::marker::PhantomData<T>);

impl<T> Tr<T> for S<T> {
    proof fn f(s: Seq<Opaque<T>>, v: T) {
        // Any verification failure here triggers auto recommends-retry,
        // which is what surfaces the panic. Without the retry the bug is latent.
        assert(false);
    }
}
} // verus!

fn main() {}
```

parno:
Fixes #2434.


source/vir/src/poly.rs
```diff
@@ -1082,6 +1083,12 @@ fn update_temp_locals(
                 let typ = state.temp_types[&l.ident].clone();
                 Arc::make_mut(l).typ = typ;
                 updated_temps.insert(l.ident.clone());
+            } else if state.remaining_temps.contains(&l.ident) {
+                // Orphan temp: declared but never assigned to in the visited body.
+                // We still need to give it a valid native (or boxed) AIR type,
+                // matching the default coercion used for other locals.
+                let typ = coerce_typ_to_native(ctx, &l.typ);
+                Arc::make_mut(l).typ = typ;
```
tjhance: would it make more sense to use a different LocalDeclKind for whatever this case is? If the var was declared but never assigned, then "TempViaAssign" isn't an accurate description.

parno: That's true.  But looking at it more, we should probably try to just avoid creating these orphan definitions in the first place.  

I think the issue might be in `ast_to_sst_func.rs`.  If we're checking recommends, this this code:
https://github.com/verus-lang/verus/blob/c6f271700678ac2a31521e3f8803d2af064a2e49/source/vir/src/ast_to_sst_func.rs#L896-L905
puts the temp variables it allocates for the `s[0]` recommends-check (from the original issue example) into `state.pre_local_decls`, and the matching statements go into `req_stms`.

But then the subsequent code drops the `req_stms` when `inherit` is true; they're only used here when `inherit` is false:
https://github.com/verus-lang/verus/blob/c6f271700678ac2a31521e3f8803d2af064a2e49/source/vir/src/ast_to_sst_func.rs#L959-L972
Does that seem like a plausible culprit?

tjhance: This does seem plausible to me. I'm not 100% sure why this chunk of code is conditional on `!inherit` to begin with. Before [8db71d4ab6](https://github.com/verus-lang/verus/pull/1584) the condition was `trait_typ_substs.len() == 0`. It might be that the code was written this way because the trait-substitutions weren't implemented, rather than some fundamental reason having to do with traits, and we just need to fix that TODO comment.

tjhance: Also worth noting this if-block contains 2 distinct pieces of logic, the part of the decrease_by, and then the last 2 lines that add req_stms to the body stm. Maybe we just need to break them apart:

```
if ctx.checking_spec_preconditions() && !inherit {
...
}
if ctx.checking_spec_preconditions() {
        req_stms.push(stm);
        stm = stms_to_one_stm(&body.span, req_stms);
}
```

parno: > Also worth noting this if-block contains 2 distinct pieces of logic, the part of the decrease_by, and then the last 2 lines that add req_stms to the body stm.

That's true.  I tried out this change, and it fixes #2434, and the CI tests still pass.  Do you think that's a good enough fix?

tjhance: yes

tjhance: we should also add a test for something that these req_stms are supposed to do, like a recommends failure resulting from a requires clause in a trait impl method.

parno: Okay, I reverted the original change, applied the new one, and added a test.

tjhance (approved):

PR #2439, opened 2026-05-12
verita: change svsm to a verita branch

ziqiaozhou:


ziqiaozhou:
https://github.com/verus-lang/verus/actions/runs/25785546062

parno:
Do we still need the redirection to your branch, now that there's a recent release of Verus that should match what's on crates.io?

ziqiaozhou:
> Do we still need the redirection to your branch, now that there's a recent release of Verus that should match what's on crates.io?
We do not need the redirection If it is okay to see the verita failure for a few days.

parno:
> We do not need the redirection If it is okay to see the verita failure for a few days.

It looks like all of the other projects are now passing, so let's go ahead and merge this, and then we can swap it back when you're ready.  Thanks for your help!

PR #2440, opened 2026-05-12
.github: restore rolling releases and pages after merge-queue migration

mmcloughlin:
Restore the rolling-release and GitHub Pages pipelines, which both stopped working when the project adopted the merge queue.

## Problem

Commit 98cd82d removed the trigger on pushes to `main` from `ci.yml`. As a result, the `ci` workflow's `build-release` and `release` jobs stopped running since their `github.event_name == 'push' && github.ref == 'refs/heads/main'` gate is never true in the merge queue.  Since that point, the rolling release was not updated. The most recent weekly release (Monday, May 11) was also out of date, since the `release` workflow simply promotes the latest rolling release.

In addition, `pages.yml` has also stopped running, since it's triggered on `ci` workflow runs on `main` (but merge-queue runs are on a different ref, not `main`). Therefore GitHub Pages is stale.

## Approach

This PR fixes the issue by creating a new `rolling-release` workflow and updating the `pages` workflow to trigger on pushes to main. With the merge queue change, we now trust that all pushes to main have passed full tests (or were from trusted admins).

Related workflows:

* `rolling-release`: new workflow created from the `build-release` and `release` jobs in the existing `ci` workflow.
* `pages`: updated to trigger on pushes to main. Now runs the `build-docs` job from `ci`, making it self-contained. This means a small amount of repeated work.
* `ci`: removed the `build-release` and `release` jobs. The `build-docs` job is now for validation only and does not publish an artifact.
* `release`: unchanged, still promotes the latest rolling release

## Testing

Tested against my personal fork (https://github.com/mmcloughlin/verus).

Test-only changes on branch: https://github.com/mmcloughlin/verus/tree/restore-rolling-releases-fork-test (not included in this PR).

GitHub Actions runs:

* `rolling-release`: https://github.com/mmcloughlin/verus/actions/runs/25756579510
* `release`: https://github.com/mmcloughlin/verus/actions/runs/25757365044
* `pages`: https://github.com/mmcloughlin/verus/actions/runs/25757733204 (deploy failure expected since pages not enabled on fork)

Resulting releases:

* Rolling pre-release: https://github.com/mmcloughlin/verus/releases/tag/release/rolling/0.2026.05.12.733bb88
* Permanent release: https://github.com/mmcloughlin/verus/releases/tag/release/0.2026.05.12.7b6d88a

---


mmcloughlin:
Question for repository admin: what branch protection do we have on `main`?

* Are all pushes required to go via the merge queue?
* Is admin bypass allowed? If so, are we okay with the risk of a bad release due to a bad admin push?

(I don't have access to check these rules myself.)

parno (approved):
This looks like the right refactoring to me.  Thanks for setting this up!

parno:
> Question for repository admin: what branch protection do we have on `main`?

We currently have these features enabled for `main`:
- Require a pull request before merging
- Require status checks to pass before merging
- Require branches to be up to date before merging
- Require merge queue
    - Only merge non-failing pull requests

And then for the list of status checks, we have:
- fmt
- clippy
- cargo-verus-test
- full-test
- basic-test (for the various platforms)
- smoke-test (for various configs)
- change_filter

> * Are all pushes required to go via the merge queue?

I think, given the above settings, the answer to this is Yes.

> * Is admin bypass allowed? If so, are we okay with the risk of a bad release due to a bad admin push?

We do allow admin bypass at present.  I think this is probably okay at present.  I only use it very occasionally, e.g., when I make a change to the GitHub workflows, or the docs, where I know the CI isn't going to do anything useful.

mmcloughlin:
> We do allow admin bypass at present. I think this is probably okay at present. I only use it very occasionally, e.g., when I make a change to the GitHub workflows, or the docs, where I know the CI isn't going to do anything useful.

Yes seems fine. I was just raising it as a possible failure-mode of this design. We are assuming every commit to main is releasable.

Since the full release is only once a week, the cost of running the whole suite of tests again is relatively low. It might be nice to have the full release re-run the `ci` workflow rather than just promoting the last rolling release. But that would need more thought, not for this PR.

parno:
> But that would need more thought, not for this PR.

Agreed on both aspects.

PR #2447, opened 2026-05-14
Fix `struct_with_invariants` generic param detection for associated type prefixes

rikosellic:
This PR fixes a bug in `struct_with_invariants!` where generic type parameters are used only as an associated type prefix.
## Input
```
trait HasTarget {
    type Target;
}

struct_with_invariants!{
    pub struct S<A: HasTarget>  {
        a: AtomicBool<_,(),_>,
        _phantom: PhantomData<*const A::Target>,
    }

    closed spec fn wf(self) -> bool {
        invariant on a with (_phantom) is (v:bool,g:()) {
            true
        }
    }
}
```
## Output
It fails to compile with this error ([playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=f8d0ad53df430ed0eb949a9f1e6ab4dd)): 
```
error[[E0433]](https://doc.rust-lang.org/stable/error_codes/E0433.html): cannot find type `A` in this scope
  --> /playground/src/main.rs:14:38
   |
14 |         _phantom: PhantomData<*const A::Target>,
   |                                      ^ use of undeclared type `A`

```

## Fix
The fix updates the generic-usage collector in `struct_decl_inv.rs` so it also recognizes the first segment of multi-segment type paths. A regression test is also added.


rikosellic:
It turns out the current implementation doesn't carry enough type-bound information through the macro expansion. The generated field-type aliases and `InvariantPredicate_auto_*` structs now carry the relevant generic params along with their bounds, using `PhantomData` to anchor params that would otherwise appear only inside associated-type projections.

rikosellic:
The doc build error seems irrelevant.

source/rust_verify_test/tests/user_defined_type_invariants.rs
```diff
@@ -2722,3 +2722,28 @@ test_verify_one_file_with_options! {
         }
     } => Err(err) => assert_fails_type_invariant_error(err, 2)
 }
+
+test_verify_one_file! {
```
tjhance: wrong file (this is for the `#[verifier::type_invariant]` attribute) — can you create a new test file `struct_with_invariants.rs` for this?

tjhance (changes requested):

source/rust_verify_test/tests/user_defined_type_invariants.rs
rikosellic: Sure! I've created the file and moved the test into it. Also moved `struct_with_invariants_const_usage`.

tjhance (approved):

PR #2449, opened 2026-05-14
Add basic support for Iter::collect

parno:
So far, this just covers `Vec`, but it would be straightforward to add more collections if desired.


tjhance (changes requested):
1. I think the ensures clause needs to ensure 'will_return_none' on the iter.
2. Do we need the 'obeys_from_iterator_spec()' pattern for this? This seems like a simple enough trait that we could just specify the 'ensures' clause for each impl (like for Clone) without any additional spec fns.

parno:
> 1. I think the ensures clause needs to ensure 'will_return_none' on the iter.

Do you mean it needs to require `will_return_none`?  That does seem plausible.

> 2. Do we need the 'obeys_from_iterator_spec()' pattern for this? This seems like a simple enough trait that we could just specify the 'ensures' clause for each impl (like for Clone) without any additional spec fns.

I used the pattern so that we would have access to `FromIteratorSpec::from_iter_ensures` in the ensures clause of `collect()`.  Following your suggestion, I've tried rewriting the postcondition in terms of `call_ensures` on `FromIterator::from_iter`, but I can't find a version that Verus will accept.  Or did you have something else in mind?

tjhance:
> Do you mean it needs to require will_return_none? That does seem plausible.

No, I mean that it should ensure will_return_none.

The 'collect' function will call the iterator repeatedly until it returns None. Once it returns None, we know that will_return_none is true.

I don't remember where to find it now, but the original proof-of-concept I made many months ago ought to demonstrate the collect function.

> Following your suggestion, I've tried rewriting the postcondition in terms of call_ensures on FromIterator::from_iter, but I can't find a version that Verus will accept

Off the top of my head, I don't know why it would fail.

parno:
> Off the top of my head, I don't know why it would fail.

For example, this attempt:
```
    fn collect<B>(self) -> (collection: B)
        where
            B: FromIterator<<Self as Iterator>::Item>,
            Self: Sized + IntoIterator<Item = <Self as Iterator>::Item>,
        default_ensures
            call_ensures(B::from_iter::<Self>, (self,), collection),
    ;
 ```
results in many errors along this line:
```
error[E0277]: `Self` is not an iterator
  --> vstd/std_specs/iter.rs:90:16
   |
90 |             B: FromIterator<<Self as Iterator>::Item>,
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Self` is not an iterator
```
Changing the bound on Self to `Self: Sized + IntoIterator<Item = <Self as Iterator>::Item> + Iterator<Item = <Self as Iterator>::Item>,` leads to a stack overflow in Rust:
```
error[E0275]: overflow evaluating the requirement `<Self as std::iter::Iterator>::Item == _`
  --> vstd/std_specs/iter.rs:88:5
   |
88 | /     fn collect<B>(self) -> (collection: B)
89 | |         where
90 | |             B: FromIterator<<Self as Iterator>::Item>,
91 | |             Self: Sized + IntoIterator<Item = <Self as Iterator>::Item> + Iterator<Item = <Self as Iterator>::Item>,
92 | |         default_ensures
93 | |             call_ensures(B::from_iter::<Self>, (self,), collection),
94 | |     ;
   | |_____^
```
I tried a few other variations, but none successfully.

**Comment by parno** [comment]
I also tried ([95654c8](https://github.com/verus-lang/verus/pull/2449/commits/95654c8b59c7877d2a71c17aa659ac5b8741806e)) going through a layer of indirection via a spec function for the postcondition + a broadcast proof, but that also fails due to what appears to be a different Verus limitation.

At this point, I think the simplest approach would be to go back to using a trait extension for FromIterator.  The extension can define `from_iter_ensures` and give `from_iter` this postcondition:
```
  fn from_iter<T>(iter: T) -> (s: Self)
       where T: IntoIterator<Item = A>
        ensures
             Self::from_iter_ensures(into_iter_remaining(iter), s),
```
@tjhance Any objections to this plan?

tjhance (approved):

PR #2452, opened 2026-05-14
Add specs for SliceIndex get, get_mut, index, index_mut

zero-to-nat:
Add specs for `SliceIndex<T>::{get, get_mut, index, index_mut}` for `T` = {slices, `str`} and implementors `core::ops::{Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive}`. This requires defining a trait extension `SliceIndexSpec<T>` which adds spec fns for implementors of the trait.

The spec fns `valid_slice, valid_indices, index_result, index_result_mut` are _uninterp and defined with axioms_ instead of members of `SliceIndexSpec<T>`. I did this because these definitions change only based on the type `T` (slice or `str`) rather than the implementor of the trait (e.g. `Range`), and therefore they can be shared across all implementors for that type `T`.

There is an unrelated change to `raw_ptr.rs` to fix the build.


source/vstd/raw_ptr.rs
```diff
@@ -2822,7 +2822,7 @@ pub fn ptr_ref2<'a, T>(ptr: *const T, Tracked(perm): Tracked<&PointsTo<T>>) -> (
 }
 
 } // verus!
-/// Trusted wrapper around `ptr_ref`, due to
+/*/// Trusted wrapper around `ptr_ref`, due to
```
elanortang: was this causing problems, and that's why it's commented out?

source/vstd/std_specs/slice.rs
```diff
@@ -37,6 +42,249 @@ pub assume_specification[ core::hint::unreachable_unchecked ]() -> !
...
+    ensures
+        #[trigger] valid_slice(slice);
+
+#[cfg(not(verus_verify_core))]
+pub broadcast axiom fn valid_slice_str(slice: &str)
+    ensures
+        #[trigger] valid_slice(slice);
+
+/// True when the indices `start` and `end` (exclusive) are considered valid indices for `slice`.
+pub uninterp spec fn in_bounds<T: ?Sized>(start: int, end: int, slice: &T) -> bool;
```
elanortang: would a more accurate name be `valid_indices`, or something like that? Since for strings you need more than just being in bounds?

source/vstd/raw_ptr.rs
zero-to-nat: thanks for the help on fixing this! it is no longer commented out

source/vstd/std_specs/slice.rs
zero-to-nat: this is true -- I have updated the name. thanks!

zero-to-nat:
Hi @elanortang, thanks for the review! The specs for the `_mut` functions have been updated to make correct use `old` and `final`.

I did not end up changing the signatures for `spec_start` and `spec_end` to accept just the length of the slice instead of the full slice. In order to make that change, I would have needed to define another generic spec fn to return the length of the slice from any slice `&T`, and then provided implementations for `&[T]` and `&str`. Since there are quite a few such generic spec fns already, I opted to not add another one. But please let me know if you would still prefer that approach, I'm happy to make that change.

elanortang (approved):
lgtm!

PR #2459, opened 2026-05-16
remove named return value requirement

### LINKED ISSUE #2321: Name required for return value of `async fn` returning unit future
Verus unnecessarily requires naming the return value of an `async` function returning a unit-valued Future:

```rust

use vstd::prelude::*;

verus!{

    async fn set_zero(x: &mut usize) // works when adding: `-> (result: ())`
        ensures *x == 0,
    {
        *x = 0;
    }

}
```
[playground](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=5fe327fe529188e6d21efb1595103d83)

This produces:
```
error: the return value must be named in a function with an ensures clause
 --> /playground/src/main.rs:5:5
  |
5 |     async fn set_zero(x: &mut usize)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

This is likely because `set_zero`s return type is actually `impl Future<Output = ()>`. We could make this more ergonomic by exempting `impl Future<Output = ()>` from the named return value requirement.

tjhance:
Simple change, but adding both of you as reviewers since this is a long-standing restriction so I want to make sure I'm not missing anything.

This removes the requirement for a named return value when the return value is non-unit, thus allowing:

```rust
fn test() -> u8
    ensures true
{   
    1   
}  
```

This is motivated by https://github.com/verus-lang/verus/issues/2321 which is specific to async functions, but I think that we just don't need this restriction at all anymore (if we ever did).

fixes https://github.com/verus-lang/verus/issues/2321

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -1807,12 +1807,7 @@ pub(crate) fn check_item_fn<'tcx>(
     if header.ensure.0.len() + header.ensure.1.len() > 0 {
         match (is_async, &header.ensure_id_typ, ret_typ_mode.as_ref()) {
             (_, None, None) => {}
-            (_, None, Some(_)) => {
-                return err_span(
-                    sig.span,
-                    "the return value must be named in a function with an ensures clause",
-                );
-            }
+            (_, None, Some(_)) => {}
             (_, Some(_), None) => {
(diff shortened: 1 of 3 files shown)


parno (approved):
Looks sound to me.

Chris-Hawblitzel (approved):

PR #2461, opened 2026-05-17
Add external_trait_private_bound attribute to handle sealed traits

Chris-Hawblitzel:


ziqiaozhou (approved):
LGTM. I have tested this feature in my side, and it works well.
NIT: it would be nice to update the error message when "external_trait_specification trait bound mismatch" to help AI to understand that it would need to add #[verifier::external_trait_private_bound...]

PR #2463, opened 2026-05-18
bugfix: unhandled shadow case in get_manual_triggers

### LINKED ISSUE #2123: `variable shadowing not yet supported` when using `Result::ok` inside a closure
```rust
use vstd::prelude::*;

verus! {

pub enum A {
    Foo { xyz: Result<(), ()> },
}

pub open spec fn foo(a: A) -> Set<nat> {
    Set::new(|x: nat| {
        match a {
            A::Foo { xyz, .. } => {
                let _ = xyz.ok();
                true
            },
        }
    })
}

}
```

```
ξ verus --crate-type=lib test.rs
error: variable shadowing not yet supported
   --> /Users/m/Files/verus/source/target-verus/release/vstd/std_specs/result.rs:252:11
    |
252 |     match result {
    |           ^^^^^^

```
This can be worked around quite easily by just manually inlining the function. As far as I can tell there's not actually any shadowing, so either Verus introduces shadowing during the inlining process, or the check is incorrect. The `let _ = xyz.ok();` binding seems to be where Verus thinks shadowing is happening, just replacing that line with `xyz.ok();` makes the error disappear.

The bad error messages for inlined functions at first made this bug look more mysterious than it probably is.

```
ξ verus --version
Verus
  Version: 0.2026.01.27.aa436cd
  Profile: release
  Platform: macos_aarch64
  Toolchain: 1.93.0-aarch64-apple-darwin
```

### LINKED ISSUE #2342: Regression: variable shadowing not yet supported on spec_unwrap_or
We encountered this error when migrating Anvil to the most recent version of Verus from `0.2025.11.30.840fa61` in https://github.com/anvil-verifier/anvil/pull/789. <s>This is the last error blocking the migration.</s> I encapsulated most of the `unwrap_or` calls and the migration is unblocked.

```shell
error: variable shadowing not yet supported
 --> vstd/std_specs/option.rs:167:10
```

It points to

https://github.com/verus-lang/verus/blob/95f3cd1f51797cac80a34641775eeac2856bb4e4/source/vstd/std_specs/option.rs#L165-L171

I think it's a regression because verus@`0.2025.11.30.840fa61` was fine with them.

Full CI log: https://github.com/anvil-verifier/anvil/actions/runs/24532268781/job/71718351643

### Reproduction steps and temporary fix by rewriting the code

Because the error is reported on `spec_unwrap_or` and is likely caused by trigger + `unwrap_or`, I rewrote some functions to only call the wrapper of it in https://github.com/anvil-verifier/anvil/pull/789/changes/faeab1ae946e86b98b608b1d5be89804230a40ec. This fixed the error in module `vdeployment_controller::proof::liveness::resource_match`. If you run

```shell
verus -L dependency=src/deps_hack/target/debug/deps --crate-type=lib --multiple-errors=100 --extern=deps_hack="src/deps_hack/target/debug/libdeps_hack.rlib" src/esr_composition.rs --allow=unused --verify-only-module vdeployment_controller::proof::liveness::resource_match
```
It succeed after this commit but failed before. However, CI still fails in other places and we have a lof of `unwrap_or` calls in other places. It's hard to differentiate which are causing the problem because this commit didn't remove `unwrap_or` calls but moved it inside a `spec fn get_replicas` wrapper.

tjhance:
Fixes https://github.com/verus-lang/verus/issues/2342

Fixes https://github.com/verus-lang/verus/issues/2123 (it seems this was already fixed; this PR adds a test)

To make the fix, we changed `get_manual_triggers` to use the ScopeMap more systematically, rather than using a HashMap to the side.

The process of making this fix also made it obvious that the following case was unhandled:

```rust
forall |y: int| (|z: int| #[trigger] foo(y, z))(x)
```

where the trigger expression `foo(y, z)` would wrongly get picked up by the outer forall. So I fixed this too.

### DIFF
--- a/source/vir/src/triggers.rs
+++ b/source/vir/src/triggers.rs
@@ -5,6 +5,7 @@ use crate::ast::{
 use crate::context::Ctx;
 use crate::messages::{Span, error};
 use crate::sst::{BndX, Exp, ExpX, Exps, Trig, Trigs};
+use crate::sst_visitor::ScopeEntry;
 use crate::triggers_auto::AutoType;
 use crate::util::vec_map;
 use air::scope_map::ScopeMap;
@@ -191,7 +192,7 @@ fn get_free_vars(exp: &Exp) -> Result<HashSet<VarIdent>, VirErr> {
 fn check_trigger_expr(
     state: &mut State,
     exp: &Exp,
-    lets: &HashSet<VarIdent>,
+    scope_map: &ScopeMap<VarIdent, ScopeEntry>,
 ) -> Result<(), VirErr> {
     match &exp.x {
         ExpX::Call(..)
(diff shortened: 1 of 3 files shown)


source/vir/src/triggers.rs
```diff
@@ -214,169 +215,183 @@ fn check_trigger_expr(
...
+        ExpX::ArrayLiteral(_) => Err(error(&exp.span, "triggers cannot contain array literals")),
+        ExpX::Loc(..) | ExpX::VarLoc(..) => Ok(()),
+        ExpX::ExecFnByName(..) => Ok(()),
+        ExpX::Call(_, _typs, args) => {
+            check_trigger_expr_args(state, args);
+            Ok(())
+        }
+        ExpX::Var(x) => {
+            if let Some(entry) = scope_map.get(x)
+                && entry.bnd_kind.is_let()
```
tjhance, line 241: rustfmt went wild on this function. This is the important change: replacing `lets.contains(x)` with this scope_map condition.

Chris-Hawblitzel: To avoid this reformatting happening again, maybe we should hoist the `&mut |exp, _scope_map| match &exp.x { ... }` out of the call to `exp_visitor_check` and into a local variable that we pass to `exp_visitor_check`?

Chris-Hawblitzel (approved):

source/vir/src/triggers.rs
tjhance: I tried it, but it seems to have trouble inferring the right lifetime signature for the closure.

PR #2471, opened 2026-05-20
Add specifications for `NonZero`

rikosellic:
Add specifications for [`NonZero`](https://doc.rust-lang.org/std/num/struct.NonZero.html). This is made possible after #2461. However, I'm not very sure how to properly handle the unstable [`ZeroablePrimitive`](https://doc.rust-lang.org/std/num/trait.ZeroablePrimitive.html) trait.


source/vstd/std_specs/nonzero.rs
```diff
@@ -0,0 +1,225 @@
...
+        nonzero_from_primitive(self.get().bitor_spec(rhs))
+    }
+}
+
+pub assume_specification<T: ZeroablePrimitive + BitOr<Output = T>>[ <NonZero<T> as BitOr<
+    T,
+>>::bitor ](x: NonZero<T>, y: T) -> <NonZero<T> as BitOr<T>>::Output
+;
+
+impl<T: ZeroablePrimitive + BitOrSpec<Output = T>> BitOrSpecImpl<NonZero<T>> for NonZero<T> {
```
elanortang, line 140: What does it mean to impl `BitOrSpecImpl<NonZero<T>>` for `NonZero<T>`, as opposed to `BitOrSpecImpl<T>` for `NonZero<T>`?

elanortang (approved):
This looks good to me!

I see that here, `ExNonZero` contains a `NonZero<T>` instead of the `T: NonZeroInner` in the Rust standard library. This change makes sense to me, since the various `NonZeroInner` structs are a lot to deal with, and the `NonZeroInner` structs for a given primitive type hold all possible values in that type except for 0.

source/vstd/std_specs/nonzero.rs
dewert99: I'm not the author, but I'm guessing `BitOrSpecImpl<NonZero<T>>` for `NonZero<T>` would specify or-ing two `NonZero<T>`s and `BitOrSpecImpl<T> for NonZero<T>`  would specify or-ing a `NonZero<T>` with a regular `T`. In Rust `NonZero<T>` implements by `BitOr<NonZero<T>>` and `BitOr<T>`

elanortang: Ah makes sense, thanks!

dewert99:
@rikosellic would you be able to resolve the conflicts so this can be merged

rikosellic:
> @rikosellic would you be able to resolve the conflicts so this can be merged

Sure, I've merged the latest main.

dewert99:
Thank you

source/vstd/std_specs/nonzero.rs
```diff
@@ -0,0 +1,225 @@
...
+    open spec fn obeys_eq_spec() -> bool {
+        true
+    }
+
+    open spec fn eq_spec(&self, other: &Self) -> bool {
+        self.get().eq_spec(&other.get())
+    }
+}
+
+pub assume_specification<T: ZeroablePrimitive + PartialEq>[ <NonZero<T> as PartialEq>::eq ](
```
parno: Given that you already implemented `PartialEqSpecImpl` for `NonZero<T>`, I think you might not need this `assume_spec` (and similarly for the `PartialOrd` functions below).  You should inherit the trait's spec on `eq`.  This might not have worked before [PR 2540](https://github.com/verus-lang/verus/pull/2540) landed.

parno (commented):
Thanks for putting this together.  As long as we're adding this, should we include `BitAnd`, not just `BitOr`?

source/vstd/std_specs/nonzero.rs
rikosellic: That's a very neat feature! I've fixed it.

rikosellic:
> Thanks for putting this together. As long as we're adding this, should we include `BitAnd`, not just `BitOr`?

`BitAnd` is not implemented for [`NonZero`](https://doc.rust-lang.org/beta/std/num/struct.NonZero.html#impl-BitOr-for-NonZero%3CT%3E) in std because it is possible to get zero when performing bit-and.

parno:
> BitAnd is not implemented for [NonZero](https://doc.rust-lang.org/beta/std/num/struct.NonZero.html#impl-BitOr-for-NonZero%3CT%3E) in std because it is possible to get zero when performing bit-and.

Ah, of course, that makes sense.  Thanks for clarifying.

parno (approved):

PR #2472, opened 2026-05-20
guide: move Quantifiers to Fundamentals section

Chris-Hawblitzel:


parno (approved):
Seems reasonable to me.

tjhance (approved):
I still kinda think it makes more sense in the Proofs section, but my feelings aren't strong.

PR #2474, opened 2026-05-21
Emit output type constraints for top-level functions too

### LINKED ISSUE #2427: Missing type constraint for function closures used in trait implementations
In the example below, the two assertions in `use_it()` mirror the postconditions of `touch()`, but only one of the two functions succeeds.  The issue seems to be a mismatch between the generic `I::Item` in the `impl<I: HasItem> W<I>` verus the concrete `u32` type in `use_it()`.  In other words, at the call site, we learn that `f` has a function type, but not that the `Item` type in `HasItem` is `u32`.

This example came up when working on the `map()` iterator adapter.

```
use vstd::prelude::*;

verus! {

// A trait with an unconstrained associated type.
pub trait HasItem {
    type Item;
}

pub struct Ad<F>(pub F);

impl<F: FnOnce(u32) -> u32> HasItem for Ad<F> {
    type Item = F::Output;
}

// Generic wrapper using an unconstrained `I::Item`.
pub struct W<I: HasItem> {
    pub i: I,
}

impl<I: HasItem> W<I> {
    uninterp spec fn index(self) -> int;
    uninterp spec fn index2(self) -> int;
    uninterp spec fn seq(self) -> Seq<I::Item>;

    fn touch(&mut self)
        ensures
            final(self).index() == final(self).seq().len(),
            final(self).index() == final(self).index2(),
    {
        assume(false);
    }
}

fn use_it() {
    let f = |x: u32| -> u32 { x };
    let mut y = W { i: Ad(f) };
    y.touch();
    assert(y.index() == y.index2());      // succeeds
    assert(y.index() == y.seq().len());   // fails
}

}

fn main() {}
```

parno:
Fixes #2427 
(again)


source/vir/src/ast_simplify.rs
```diff
@@ -1031,10 +1032,64 @@ fn add_fndef_axioms_to_function(
...
-        &Arc::new(TypX::FnDef(fun.clone(), typ_args, None)),
+        &Arc::new(TypX::FnDef(fun.clone(), typ_args.clone(), None)),
         ExprX::ExecFnByName(fun.clone()),
     );
 
+    // Emit `FnDef : {Fn, FnMut, FnOnce}<Args>` and `<FnDef as FnOnce<Args>>::Output = Ret`.
+    //
+    // We emit a TraitImpl for each of the three Fn-family traits (not just Fn), because
+    // code that mentions only one through an associated-type projection (e.g. `Map::Item = F::Output`) 
+    // never creates a Fn term for the Fn-related axioms to trigger on.
```
tjhance, line 1043: Can you give an example of the issue being avoided by emitting all 3 trait impls?

Is it a matter of triggering, or are we missing the `T implement Fn ==> T implements FnMut` / `T implements FnMut ==> T implements FnOnce` axioms entirely?

tjhance: I'm trying to figure out why we need all 3 axioms for FnDef types but not for closures types (or _do_ we need all 3 for the closure types?)

parno: > Can you give an example of the issue being avoided by emitting all 3 trait impls?

This came up in this version of the test:
```
fn id<T>(x: T) -> T { x }

fn use_generic_fn() {
    let mut y = W { i: Ad(id::<u32>) };
    y.touch();
    assert(y.index() == y.seq().len());
}
```
I think the issue is related to triggering, since I believe Verus already emits quantified trait-bound axioms.  E.g., for the generic `id`, we get an axiom that quantifies over `tr_bound%Fn`, but in the VC for `use_generic_fn`, it uses `tr_bound%FnOnce`.  It's possible that another fix could be to adjust the triggering on the original axioms.

parno: I don't think we should have the same problem for closures, though, right?

tjhance: because we don't need to quantify over all instantiations of the closure's generics?

source/vstd/resource/map.rs
```diff
@@ -1306,8 +1306,10 @@ impl<K, V> GhostPointsTo<K, V> {
         self.lemma_map_view();
         self.submap.agree(auth);
         assert(self.submap@ <= auth@);
-        assert(self.submap@.contains_key(self.key()));
-        assert(self.submap@.contains_pair(self.key(), self.value()));
+        assert(self.submap@.dom().contains(self.key()));
+        assert(auth@.dom().contains(self.key()));
+        assert(self.submap@[self.key()] == self.value());
+        assert(auth@[self.key()] == self.value());
```
tjhance, line 1312: what's this change about?

parno: It’s a flaky proof that started to fail after this change, although only on the is-core test. These assertions seem to stabilize it.

source/vir/src/ast_simplify.rs
parno: Does Rust allow generic closures? I thought they were monomorphized on the spot

tjhance: Strictly speaking, they inherit the type parameters from the enclosing function. So if you have `f<A>() { let clos = || { }; }` the closure type would be like `AnonymousClosure<A>`.

But since the use of the type is private to the function, we only have to work with `Cl<A>` for the skolemized `A`. We don't have to quantify over all A.

tjhance (approved):

ziqiaozhou:
This PR causes panics in multiple Verita tests: https://github.com/verus-lang/verus/actions/runs/26284443033

I had to revert this commit in my local branch to get Verus working again.

parno:
Indeed.  There’s been some discussion on the Verus Zulip about it.

source/vir/src/ast_simplify.rs
```diff
@@ -1031,10 +1032,64 @@ fn add_fndef_axioms_to_function(
...
+        let args_tuple_typ = Arc::new(TypX::Datatype(
+            Dt::Tuple(arg_typs.len()),
+            Arc::new(arg_typs),
+            Arc::new(vec![]),
+        ));
+        let trait_typ_args = Arc::new(vec![self_typ, args_tuple_typ]);
+
+        let mk_impl_path = |kind: ClosureKind| {
+            Arc::new(crate::ast::PathX {
+                krate: CrateId::Internal,
```
mmcloughlin, line 1056: I think this introduced a bug: https://github.com/verus-lang/verus/pull/2559

PR #2477, opened 2026-05-22
Fn output fix

parno:
Commit 633baa8c started emitting an extra TraitImpl (for Fn/FnMut/FnOnce) and an AssocTypeImpl (for FnOnce::Output) for every function that needs fndef axioms.  However, this broke code in capybaraKV and svsm.

This PR makes two changes:
1. source/vir/src/ast_simplify.rs — register the tuple arity so the tuple datatype is actually emitted:
2. source/vir/src/prune.rs — give FnDef its own ReachedType variant keyed by (Fun, Vec<ReachedType>), so each trait-impl FnDef is a distinct prune key (and so the AssocTypeImpl is no longer "always reached")

### DIFF
--- a/source/rust_verify_test/tests/fndef_types.rs
+++ b/source/rust_verify_test/tests/fndef_types.rs
@@ -2006,5 +2006,34 @@ test_verify_one_file_with_options! {
             y.touch();
             assert(y.index() == y.seq().len());
         }
+
+
+        // Module `a` defines a struct with restricted visibility and a
+        // function whose signature mentions it. The function is used as an
+        // FnDef value, so Verus emits an auto-generated
+        // `<FnDef(takes_hidden) as FnOnce<(Hidden,)>>::Output = Hidden`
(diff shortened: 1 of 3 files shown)


tjhance (commented):
I forget how the ReachedType stuff works, but it seems a little suspicious to me that FnDef would be the only ReachedType to be recursive. @Chris-Hawblitzel should probably weigh in for prune.rs

Chris-Hawblitzel:
> I forget how the ReachedType stuff works, but it seems a little suspicious to me that FnDef would be the only ReachedType to be recursive. @Chris-Hawblitzel should probably weigh in for prune.rs

Right, I think you can just use `ReachedType::FnDef(Fun)`.  `prune.rs` doesn't track type arguments because this tracking wouldn't generally work if the arguments were type parameters.  You can think of `ReachedType::FnDef(Fun)` as an over-approximation that includes all possible type arguments to that particular `Fun`, which should be fine for what you need.

PR #2479, opened 2026-05-23
vstd/range: implement IteratorSpecImpl for RangeInclusive

ziqiaozhou:

### DIFF
--- a/source/rust_verify_test/tests/iterators.rs
+++ b/source/rust_verify_test/tests/iterators.rs
@@ -34,8 +34,6 @@ test_verify_one_file! {
             invariant
                 v.len() == iter.index(),
                 i <= 5,
-                iter.snapshot@@.start == 0,
-                iter.snapshot@@.end == 4,
             {
                 assert(i <= 4);
                 v.push(i);
--- a/source/rust_verify_test/tests/iterators.rs
+++ b/source/rust_verify_test/tests/iterators.rs
@@ -4,6 +4,48 @@ mod common;
 use common::*;
 
 test_verify_one_file! {
+    #[test] range_works verus_code! {
+        use vstd::prelude::*;
+
+        fn test()
+        {
+            let mut v = vec![];
+            for i in iter: 0..4
+            invariant
+                v.len() == iter.index(),
+                iter.index() <= 4,
+            {
+                assert(i < 4);
+                v.push(i);
+            }
+            assert(v.len() == 4);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] range_inclusive_works verus_code! {
+        use vstd::prelude::*;
+
+        fn test()
+        {
+            let mut v = vec![];
+            for i in iter: 0..=4
+            invariant
+                v.len() == iter.index(),
+                i <= 5,
+                iter.snapshot@@.start == 0,
+                iter.snapshot@@.end == 4,
+            {
+                assert(i <= 4);
+                v.push(i);
+            }
+            assert(v.len() == 5);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
     #[test] collect_works verus_code! {
         use vstd::prelude::*;
 
--- a/source/vstd/std_specs/range.rs
+++ b/source/vstd/std_specs/range.rs
@@ -93,11 +93,17 @@ pub assume_specification<Idx: PartialOrd<Idx>, U>[ RangeInclusive::<Idx>::contai
 // `#[verifier::when_used_as_spec(spec_range_inclusive_new)]` to the
 // specification for the executable `RangeInclusive::new` method and define
 // that spec function here.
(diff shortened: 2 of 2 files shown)


source/vstd/std_specs/range.rs
```diff
@@ -150,6 +150,61 @@ impl<A: core::iter::Step> super::iter::IteratorSpecImpl for Range<A> {
...
+                self,
+            )[i] == init@.start.spec_forward_checked_int(i).unwrap()
+    }
+
+    open spec fn decrease(&self) -> Option<nat> {
+        Some((self@.start.spec_steps_between_int(self@.end) + 1) as nat)
+    }
+
+    open spec fn peek(&self, index: int) -> Option<Self::Item> {
+        if 0 <= index <= self@.start.spec_steps_between_int(self@.end) {
```
parno: I think we need `self@.end + 1` here (note that `Range`'s `peek` is already inclusive of `end`, so `RangeInclusive` needs to go one more.  This facilitates proofs about the individual element we're iterating with (i.e., the `x` in `for x in 0..10`.

parno (commented):
Thanks for adding this.  Can you add an example of using this to `rust_verify_tests/tests/iterators.rs` to confirm that it works as expect (e.g., similar to the `collect` test and/or some of the range tests in `loops.rs` and/or `source/docs/migration-iterators.md`) ?

source/vstd/std_specs/range.rs
ziqiaozhou: I was unsure whether Range's peek really needs an extra element outside the valid range since Vector's peek does not and all existing tests still passed without it. But I realize that users may want to maintain invariants such as for x in 0..10 invariant x <= 10, which require a sentinel element at the end. I added tests to capture this behavior so it is not accidentally treated as a bug in the future.

source/rust_verify_test/tests/iterators.rs
```diff
@@ -3,6 +3,48 @@
...
+
+        fn test()
+        {
+            let mut v = vec![];
+            for i in iter: 0..=4
+            invariant
+                v.len() == iter.index(),
+                i <= 5,
+                iter.snapshot@@.start == 0,
+                iter.snapshot@@.end == 4,
```
parno: Are these last two invariants needed for the proof to work?  I'm guessing yes, which is unfortunate, since they're not needed in the normal Range case.  I think this is an indication that when writing the impl of `IteratorSpecImpl`, we should use `self.start` and `self.end`, not the view of `self`. 

I think we should also update the assume_specification for `RangeInclusive::<Idx>::new` to ensure that the resulting start and end match the input variables.  

If we do both of these items, then I think we can eliminate these two invariants and make RangeInclusive match the automation of Range.

source/vstd/std_specs/range.rs
parno: Indeed, with `Range`, `peek` can "cheat" a little bit for the sake of better automation.  Vec doesn't really have that option.  Thanks for adding this to the test.

source/rust_verify_test/tests/iterators.rs
ziqiaozhou: Unfortunately, the view is needed since start and end are private fields in RangeInclusive. 
`start..=end` is equivalent to `RangeInclusive::new(start, end)`, while `start..end` is equivalent to  `Range{start, end}`

parno: Ah, I see. I was looking at the [`std::range::RangeInclusive`](https://doc.rust-lang.org/std/range/struct.RangeInclusive.html) version, where they are public fields, but it looks like that's only in Rust 1.96 onwards.

In that case, to fix the automation, I think we need to follow the pattern used for `Vec`'s `into_iter`, which is paired with an uninterpreted `spec_into_iter` function via `assume_specification`.  If you do the same thing for `RangeInclusive::<Idx>::new` (including the broadcast axiom about `spec_into_iter`), then inside the `for` loop, I think should automatically know the info about start and end.

ziqiaozhou: start..=end uses [std::ops::RangeInclude](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html) instead of [std::range::RangeInclusive](https://doc.rust-lang.org/std/range/struct.RangeInclusive.html), and so it would need that axiom even in Rust 1.96.

I added a similar broadcast axiom for RangeInclusive::<Idx>::new to avoid that 2 invariants

parno: > so it would need that axiom even in Rust 1.96.

That's true.  It looks like [they plan](https://blog.rust-lang.org/2026/05/28/Rust-1.96.0/) to eventually make `std::range` the default, but that might take a while.

I'm glad the broadcast axiom was able to eliminate the extra invariants.

parno (approved):

PR #2480, opened 2026-05-23
vstd/iter: Add spec for Iterator::find

ziqiaozhou:

### DIFF
--- a/source/vstd/std_specs/iter.rs
+++ b/source/vstd/std_specs/iter.rs
@@ -95,7 +95,6 @@ pub trait ExIterator {
                 FromIteratorSpec::from_iter_ensures(self.remaining(), collection),
     ;
 
-    /// Users may need extra proofs to trigger the `old(self).peek(i)`
     fn find<P>(&mut self, predicate: P) -> (r: Option<Self::Item>)
         where Self: Sized,
             P: FnMut(&Self::Item) -> bool
@@ -112,7 +111,8 @@ pub trait ExIterator {
             // iterator's elements.
             final(self).obeys_prophetic_iter_laws() && r.is_none() ==> {
                 &&& final(self).remaining().len() == 0
-                &&& forall |i| (0 <= i < old(self).remaining().len()) ==> predicate.ensures((&(#[trigger]old(self).peek(i)).unwrap(),), false)
+                &&& forall |i| 0 <= i < old(self).remaining().len() ==>
+                    predicate.ensures((#[trigger]&old(self).remaining()[i],), false)
             },
             // If find returns Some, then the returned value satisfies the
             // predicate, and all previous elements did not satisfy the
@@ -122,8 +122,9 @@ pub trait ExIterator {
                 {
                     &&& old(self).remaining().len() > 0
                     &&& predicate.ensures((&r.unwrap(),), true)
-                    &&& old(self).peek(idx) == r
-                    &&& forall |i| (0 <= i < idx) ==> predicate.ensures((&(#[trigger]old(self).peek(i)).unwrap(),), false)
+                    &&& old(self).remaining()[idx] == r.unwrap()
+                    &&& forall |i| 0 <= i < idx ==>
+                        predicate.ensures((#[trigger] &old(self).remaining()[i],), false)
                 }
             };
 }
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/iter.rs
```diff
@@ -94,6 +94,38 @@ pub trait ExIterator {
...
+        where Self: Sized,
+            P: FnMut(&Self::Item) -> bool
+        default_ensures
+            // The iterator consistently obeys, completes, and decreases throughout its lifetime
+            final(self).obeys_prophetic_iter_laws() == old(self).obeys_prophetic_iter_laws(),
+            final(self).obeys_prophetic_iter_laws() ==> final(self).will_return_none() == old(self).will_return_none(),
+            final(self).obeys_prophetic_iter_laws() ==> (old(self).decrease() is Some <==> final(self).decrease() is Some),
+            // The final remaining elements are a subsequence of the old remaining.
+            final(self).obeys_prophetic_iter_laws() ==> {
+                &&& old(self).remaining().len() >= final(self).remaining().len()
```
parno: I think we could omit the comment and simplify this to:
```
final(self).remaining().is_suffix_of(old(self).remaining())
```

source/vstd/std_specs/iter.rs
```diff
@@ -94,6 +94,38 @@ pub trait ExIterator {
...
+                &&& final(self).remaining() == old(self).remaining().skip(old(self).remaining().len() - final(self).remaining().len())
+            },
+            // If the iterator returns None, then the final remaining elements are empty and
```
parno: I think this should be: "If `find` returns None, then the iterator has no remaining elements, and the predicate was false for all of the original iterator's elements."

source/vstd/std_specs/iter.rs
```diff
@@ -94,6 +94,38 @@ pub trait ExIterator {
...
+            // the predicate is false for all old remaining elements.
+            final(self).obeys_prophetic_iter_laws() && r.is_none() ==> {
+                &&& final(self).remaining().len() == 0
+                &&& forall |i| (0 <= i < old(self).remaining().len()) ==> predicate.ensures((&(#[trigger]old(self).peek(i)).unwrap(),), false)
+            },
+            // If the iterator returns Some, the last element is the one that satisfies the predicate,
```
parno: Similarly, I think this should be:
"If `find` returns Some, then the returned value satisfies the predicate, and all previous elements did not satisfy the predicate."

parno (commented):
Thanks this will be quite helpful.  Similar to my suggestion on #2479, could you please add a few examples of using this to the test suite?  It would be great to have examples that show find returning Some and examples showing it returning None.

source/vstd/std_specs/iter.rs
```diff
@@ -94,6 +94,38 @@ pub trait ExIterator {
...
+            final(self).obeys_prophetic_iter_laws() ==> (old(self).decrease() is Some <==> final(self).decrease() is Some),
+            final(self).obeys_prophetic_iter_laws() ==> {
+                final(self).remaining().is_suffix_of(old(self).remaining())
+            },
+            // If find returns None, then the iterator has no remaining
+            // elements, and the predicate was false for all of the original
+            // iterator's elements.
+            final(self).obeys_prophetic_iter_laws() && r.is_none() ==> {
+                &&& final(self).remaining().len() == 0
+                &&& forall |i| (0 <= i < old(self).remaining().len()) ==> predicate.ensures((&(#[trigger]old(self).peek(i)).unwrap(),), false)
```
parno: This: `(0 <= i < old(self).remaining().len())` can be written without parentheses (the `==>` is lower precedence).  In general, we should also try to write specs in terms of `remaining`, not `peek`, since all iterators implement `remaining`, but only some provide a useful `peek`.  So in this case, we'd have `old(self).remaining()[i]` instead of `old(self).peek(i)).unwrap()`.

Same set of comments for the `forall` in the ensures clause further below.

parno (approved):
Thanks for the update to remove `peek`.  In the test, would you mind using `assert(forall |i| 0 <= i < v.len() ==> v[i] >= 10);` instead of the `proof` block?  I think that might make the intent of the postcondition clearer.  I already checked, and the test still passes with that change.

PR #2481, opened 2026-05-23
[builtin_macros] desugar for-loops inside `#[verus_verify]` bodies

tangruize:
### Problem

A `for`-loop inside a `#[verus_verify]` function (without `#[verus_spec]` on the loop) triggers an ICE:

```
index out of bounds: the len is 0 but the index is 0 (generic_args.rs:54)
```

Repro:
```rust
use vstd::prelude::*;

#[verus_verify]
fn f() {
    for i in 0..10 {}
}
```

### Root Cause

Verus needs to desugar `for`-loops into an equivalent `loop` + `Iterator::next` form during macro expansion. #2272 (by @ziqiaozhou) fixed this for the `#[verus_spec]` path by adding `visit_expr_for_loop_mut` to `ExecReplacer`, but `#[verus_verify]` uses a different visitor (`VerusVerifyVisitor`) which was not updated.

Without desugaring, the raw for-loop passes through to rustc HIR, and `rust_verify` panics when resolving `IntoIterator::into_iter` with missing generic args.

This PR adds the same `visit_expr_for_loop_mut` to `VerusVerifyVisitor`, attaching an empty `#[verus_spec]` attribute to for-loops so the existing `verus_spec` proc macro handles the desugaring.

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -188,7 +188,11 @@ pub(crate) fn rewrite_verus_attribute(
     }
 
     // Special handling for impl blocks, add marker attribute to each method for `#[verus_spec]`.
-    let mut visitor = VerusVerifyVisitor { verify_const: true };
+    let mut visitor = VerusVerifyVisitor {
+        verify_const: true,
+        erase: erase.clone(),
+        inside_external_code: contains_external,
+    };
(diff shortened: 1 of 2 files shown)


ziqiaozhou (commented):
@tangruize I opened #2498 to solve the issue since I do not want verus_verify macro go to deep to visit the whole AST.

PR #2486, opened 2026-05-25
Make `Set` and `Map` finite, add `ISet` and `IMap`.

jaylorch:
`Set<A>` and `Map<A>` now represent finite sets and maps, respectively. `ISet<A>` and `IMap<A>` are the new names for the old-style possibly-infinite sets and maps.

This new design is motivated because (1) it allows recursive types, such as having an `enum T` with a `Set<T>` field; and (2) finite sets are quite common in user code, and it's easy to go down a wild goose chase trying to get an ambient broadcast property to instantiate that turns out to be a missing `.finite()`.

The simplest porting path for existing code is to use the infinite versions, which work like the previous single Set and Map. Replace Set with ISet, Map with IMap, `set!` with `iset!`, and `map!` with `imap!`.

To exploit the finite sets, use the Set and Map types. Where you might have specified a set domain with a predicate (boolean closure) before, now you might start with a finite constructor and transform it with map or filter.
```
let evens_lt_100 = Set::<int>::new(|x| 0 <= x < 100 && x%2==0); // old predicate constructor; now produces an `Option<Set<int>>` that you demonstrate is `Some` by proving it's finite.
let evens_lt_100 = Set::<int>::range(0, 100).filter(|x| x % 2 == 0); // new constructor starts with finite int range

let trees = Set::<int>::range(0, 5).map(|x| tree_constructor(x)); // Construct a set of trees, one per int [0,5)
```
Or, even better, use the recently added `set_build!` macro to produce a finite set. To produce a finite map, construct the finite domain as a `Set` and use it as the first parameter to the new `Map::new`.

This PR addresses issue #1512.


jaylorch:
Oh, I just realized I haven't made it possible to have a recursive type `A` using `Map<A, B>`. Converting to draft....

jaylorch:
I've created PRs for all the projects from veritas and I've tested that they work. (The vest projects don't seem to need a PR, since they don't use `Set` or `Map` and verify just fine with no changes.) Chris wrote the first drafts of the ones for verified-storage and verified-memory-allocator, and then I adapted them to the latest version of the finite-sets-and-maps PR.

https://github.com/microsoft/verified-storage/pull/53
https://github.com/verus-lang/verified-memory-allocator/pull/6
https://github.com/verus-lang/verified-ironkv/pull/5
https://github.com/verus-lang/verified-node-replication/pull/6
https://github.com/secure-foundations/human-eval-verus/pull/49
https://github.com/matthias-brun/verified-nrkernel/pull/10
https://github.com/ziqiaozhou/svsm/pull/12
https://github.com/anvil-verifier/anvil/pull/801

jonhnet (approved):
Approved on the general principle that I agree with the plan and the first fifty commits are from me. I definitely did not read the 7,000 line diff. :v(

source/vstd/iset.rs
```diff
@@ -0,0 +1,1090 @@
...
+            assert(s1.contains(a));
+            assert(false);
+        }
+        assert(s1 =~= s2);
+    }
+}
+
+pub broadcast proof fn lemma_iset_ext_equal_eq<A>(s1: ISet<A>, s2: ISet<A>)
+    ensures
+        #[trigger] (s1 =~= s2) ==> s1 == s2,
```
Chris-Hawblitzel: This should already be subsumed by the general axiom for `=~=`, which says that `=~=`  equals `==` for all types (in vir/src/prelude.rs):

```
        // Extensional equality
        (declare-fun [ext_eq] (Bool [typ] [Poly] [Poly]) Bool)
        (axiom (forall ((deep Bool) (t [typ]) (x [Poly]) (y [Poly])) (!
            (= (= x y) ([ext_eq] deep t x y))
            :pattern (([ext_eq] deep t x y))
            :qid prelude_ext_eq
            :skolemid skolem_prelude_ext_eq
        )))
```

jaylorch: Are you sure? When I remove `lemma_iset_ext_equal_eq` it from the broadcast group `group_iset_lemmas`, I get this error:
```
error: postcondition not satisfied
    --> vstd\resource\map.rs:1557:13
     |
1550 | /     pub proof fn agree(
1551 | |         tracked self: &GhostPersistentPointsTo<K, V>,
1552 | |         tracked auth: &GhostMapAuth<K, V>,
1553 | |     )
     | |_____- at the end of the function body
...
1557 |               auth@.contains_pair(self.key(), self.value()),
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ failed this postcondition
```

Chris-Hawblitzel: Yes.  The code in `resource::map::GhostPersistentPointsTo::agree` is unstable and `lemma_iset_ext_equal_eq` is just perturbing the SMT solver rather than fixing the instability.  The instability is due to `contains_pair`, which is defined as follows:

```
    pub open spec fn contains_pair(self, k: K, v: V) -> bool {
        self.dom().contains(k) && self[k] == v
    }
```

The problem is that when the code asserts `auth@.contains_pair(self.key(), self.value())`, the SMT solver may try to prove the two conjuncts `self.dom().contains(k)` and `self[k] == v` in either order.  If it proves `self.dom().contains(k)` first, it will trigger quantifiers based on `self.dom().contains(k)`, giving the necessary information for both conjuncts to succeed.  If it tries to prove `self[k] == v` first, though, it fails to trigger quantifiers based on `self.dom().contains(k)`, and the proof of `self[k] == v` immediately fails before even considering the first conjunct.  This is a known pitfall related to the SMT solver's relevancy handling (`smt.relevancy`).  In this case, the instability can be fixed by asserting the `contains` first:

```
    pub proof fn agree(
        tracked self: &GhostPersistentPointsTo<K, V>,
        tracked auth: &GhostMapAuth<K, V>,
    )
        requires
            self.id() == auth.id(),
        ensures
            auth@.contains_pair(self.key(), self.value()),
    {
        use_type_invariant(self);
        use_type_invariant(auth);

        self.lemma_map_view();
        self.submap.agree(auth);
        assert(self.submap@ <= auth@);
        assert(self.submap@.contains_key(self.key()));
+        assert(auth@.contains_key(self.key()));
    }
```

(This works because `assert(P)` in Verus is actually an assert(P) followed by an assume(P), creating an SMT obligation of the form `P && (P ==> ...)`, which can safely be processed in either order, `P && (P ==> ...)` or `(P ==> ...) && P`.)

I've seen this happen with `contains_pair` elsewhere.  I wonder if `contains_pair`, as currently written, is just an invitation for people to fall into this non-obvious trap.  We could rewrite it as:

```
    pub open spec fn contains_pair(self, k: K, v: V) -> bool {
        self.dom().contains(k) && (self.dom().contains(k) ==> self[k] == v)
    }
```

parno: That seems like a useful rewrite to me.

jaylorch: To me as well. And it seems to be working!

Chris-Hawblitzel:
I benchmarked the SMT solving on the two verita tests that I usually use for Verus SMT performance testing, mimalloc and capybarakv.  For both the old Verus Set/Map definitions (`orig`) and for this pull request (`iset`), I ran each benchmark twice to see the `smt-run` time in seconds when running on my local computer with `--num-threads 1` and `--time`:

```
smt-run

mimalloc
orig: 82, 83
iset: 100, 98

capybarakv
orig: 39, 38
iset: 38, 38
```

mimalloc is 20% slower, which is noticeable but not terrible, and anyway some impact on the performance is probably expected with a vstd change that is this sweeping.  It might be worth testing a version of mimalloc in which we prefer `ISet` over `Set`.

examples/state_machines/tutorial/fifo.rs
```diff
@@ -45,8 +45,8 @@ tokenized_state_machine!{FifoQueue<T> {
 
         // All the stored permissions
 
-        #[sharding(storage_map)]
-        pub storage: Map<nat, cell::PointsTo<T>>,
+        #[sharding(storage_imap)]
+        pub storage: IMap<nat, cell::PointsTo<T>>,
```
tjhance: can we make these use map instead of imap? no worries if it's too much trouble

source/rust_verify/src/externs.rs
```diff
@@ -33,16 +33,23 @@ pub enum VerusExtern {
...
+            VerusExtern::Macros,
+            "state_machines_macros",
+            format!("{LIB_PRE}verus_state_machines_macros.{LIB_DL}"),
+        ),
         (
             VerusExtern::Macros,
             "verus_state_machines_macros",
             format!("{LIB_PRE}verus_state_machines_macros.{LIB_DL}"),
         ),
+        (VerusExtern::Builtin, "builtin", format!("libverus_builtin.rlib")),
```
tjhance: what's the purpose of this change?

source/rust_verify_test/tests/state_machines.rs
```diff
@@ -6590,13 +6630,13 @@ test_verify_one_file! {
     #[test] persistent_set_codegen IMPORTS.to_string() + verus_code_str! {
```
tjhance, line 8304: I'd like to see both a persistent_set_codegen and persistent_iset_codegen, rather than just changing the existing test to use iset. Same for the other codegen tests.

source/state_machines_macros/src/simplification.rs
```diff
@@ -641,17 +646,19 @@ fn expr_add(stype: &ShardableType, cur: &Expr, elt: &MonoidElt) -> Expr {
...
+                }
                 ShardableType::PersistentCount => Expr::Verbatim(quote_vstd! { vstd =>
                     #vstd::state_machine_internal::nat_max(#cur, #e)
                 }),
                 ShardableType::PersistentBool => Expr::Verbatim(quote! {
                     ((#cur) || (#e))
                 }),
                 _ => {
-                    panic!("expr_can_add invalid case");
+                    panic!("expr_can_add invalid case jonh1 {:?}", stype);
```
tjhance: jon doesn't need to be in the error messages

source/vstd/multiset.rs
```diff
@@ -30,16 +32,23 @@ verus! {
...
 //   (1) representing multiplicity by an ordinal or cardinal or something
 //   (2) limiting each multiplicity to be finite
 // (1) would be complicated and it's not clear what the use would be; (2) has some
 // weird properties (e.g., you can't in general define a multiset `map` function
 // since it might map an infinite number of elements to the same one).
+// Also, note that if multiset were infinite, it couldn't accept_recursive_types.
+//
+// Perhaps someday Verus will be able to mark finite-Map as accept_recursive_types, then this file
+// can be rewritten as just a wrapper around Map rather than the present pile of trusted axioms.
+// See https://github.com/verus-lang/verus/discussions/1663
```
tjhance, line 50: Is this comment still accurate? We are using accept_recursive_types now, yeah?

tools/veritas/src/main.rs
```diff
@@ -2,7 +2,8 @@ const REPOS_CACHE_PATH_VAR: &str = "REPOS_CACHE_PATH";
 const WORKDIR_PATH_VAR: &str = "WORKDIR_PATH";
 const Z3_CACHE_PATH_VAR: &str = "Z3_CACHE_PATH";
 const OUTPUT_PATH_VAR: &str = "OUTPUT_PATH";
-const RUNNER_PATH: &str = "/root/veritas";
+const RUNNER_PATH_VAR: &str = "VERITAS_RUNNER_PATH";
+const RUNNER_PATH_DEFAULT: &str = "/root/veritas";
```
tjhance: What are all these veritas changes for?

tjhance (changes requested):

source/docs/guide/src/spec_lib.md
Chris-Hawblitzel: Can we list all 5 types here (Seq, Set, ISet, Map, IMap)?  Similar for below where it says "Seq, Set and Map", and "Seq and Set".

source/docs/guide/src/spec_lib.md
```diff
@@ -25,10 +35,28 @@ with particular contents:
...
+A `Set` can only be constructed if it's provably finite. There are several
+ways to do this, including the following:
+* For numeric types `T` (e.g., `int`, `usize`, `i32`, `u64`, etc.),
+  `Set::<T>::range(lo, hi)` produces a set of `i: T` such that `lo <= i <
+  hi`. (See the `FiniteRange` trait.)
+* For numeric types `T` other than `int` and `nat`,
+  `Set::<T>::full().unwrap()` produces a set of all values of type `T`. (See
+  the `FiniteFull` trait.)
+* You can use one of the above techniques, then modify the `Set` as desired
+  using `Set::map` and/or `Set::filter`.
```
Chris-Hawblitzel: Can we warn people here that `Set::map` relies on an `exists` (specifically, its meaning is defined via the formula `self.map(f).contains(b) <==> exists|a: A| self.contains(a) && b == f(a)`), which generally makes proof about the resulting set more difficult and less automated?  I think we should encourage people to consider alternatives to `Set::map`, such as `Set::map_by` and `set_build!` when possible.

(`Set::filter`, on the other hand, is fine.)

source/rust_verify_test/tests/regression.rs
```diff
@@ -612,32 +612,32 @@ test_verify_one_file! {
 test_verify_one_file! {
     #[test] test_multiset_finite_false_1 verus_code! {
         use vstd::{map::*, multiset::*};
-        proof fn test(mymap: Map<nat, nat>)
-            requires !mymap.dom().finite() {
+        // This test was originally intended to ensure we couldn't use an
+        // infinite Map to construct a multiset, which is axiomatically finite.
+        // Now, Multiset::from_map requires a Finite Map as its argument,
+        // so finiteness is preserved by type construction.
+        // TODO(discuss): so shall we delete the test?
```
Chris-Hawblitzel: Sounds fine to me to delete it.

Chris-Hawblitzel:
<del>When porting old code to use the new finite `Set` definition, I found it useful to have something like a `Set::TEMPORARY_ASSUME_NEW(f)` function that creates a "finite" set by assuming the function `f` used to create the set returns `true` on only finitely many inputs.  I used this as a stepping stone to first replace all the old `Set::new(f)` with `Set::TEMPORARY_ASSUME_NEW(f)`, fix all the type checking and incidental proof breakage, and then one by one replace each `Set::TEMPORARY_ASSUME_NEW(f)` with a real finite set construction, fixing any resulting broken proofs one by one.  Should we include such a function for the next month or two, marking it deprecated and clearly commenting it as temporary and unsafe?</del>

Oh, never mind, it's already there in `new_assuming_finite`.

Chris-Hawblitzel (approved):
I don't have any further comments or suggestions.

examples/state_machines/tutorial/fifo.rs
jaylorch: No problem, done!

source/rust_verify/src/externs.rs
jaylorch: I have no idea why @jonhnet did this. Maybe it used to be necessary when he was using `GSet`? In any event, I checked and it's not necessary, so I'm going to undo the three additions to this function `verus_builtin_std`. @jonhnet, let me know if that's unwise!

source/rust_verify_test/tests/state_machines.rs
jaylorch: OK, I've updated all the codegen tests that involve sets and/or maps to have two versions, one using the finite variants and one using the infinite variants.

source/state_machines_macros/src/simplification.rs
jaylorch: Good catch, done.

source/vstd/multiset.rs
jaylorch: Good point, fixed!

jaylorch: That is, I updated the comment to be more up to date. I haven't actually changed `multiset` yet; we can do that in a future PR.

tools/veritas/src/main.rs
jaylorch: I'm not sure what @jonhnet was up to here, but I agree that these changes to veritas should be in a separate PR. I left the fix that he made to the configuration settings for `verified-nrkernel`, though.

source/docs/guide/src/spec_lib.md
jaylorch: Sounds good, done.

source/docs/guide/src/spec_lib.md
jaylorch: Sure, done.

source/rust_verify_test/tests/regression.rs
jaylorch: OK, done.

Chris-Hawblitzel:
> mimalloc is 20% slower, which is noticeable but not terrible, and anyway some impact on the performance is probably expected with a vstd change that is this sweeping.

There is a lemma that is triggering unnecessarily:

```
pub broadcast proof fn lemma_to_iset<A>(s: Set<A>)
     ensures
        #![trigger s.to_iset()]
        s.to_iset().finite(),
        s.to_iset().len() == s.len(),
        forall|a: A| #[trigger] s.to_iset().contains(a) <==> s.contains(a),
```

Because `Set::contains` is defined in terms of `to_iset`, this broadcast triggers on every `Set::contains`.  Also, the `forall|a: A| #[trigger] s.to_iset().contains(a) <==> s.contains(a)` is not needed, since this is true by definition.  So I would suggest removing this lemma and replacing it with more focused lemmas:

```
pub broadcast proof fn lemma_to_iset_finite<A>(s: Set<A>)
    ensures
        #[trigger] s.to_iset().finite(),
{
...
}

pub broadcast proof fn lemma_to_iset_len<A>(s: Set<A>)
    ensures
        #[trigger] s.to_iset().len() == s.len(),
{
...
}
```

With this change, the SMT times are better:

*Correction: I originally posted times of <del>iset: 93, 92</del> but then realized this was from a wrong, edited version. The corrected numbers are below, also updated with the use-isets branch.*

```
smt-run

mimalloc
orig: 82, 83
iset: 97, 98 (finite-sets branch)
iset: 86, 86 (use-isets branch)
```
<del>iset: 93, 92</del>

```
capybarakv
orig: 39, 38
iset: 35, 35
```

<del>This is only 1% slower on average between the two benchmarks.</del>

This is 4% slower on average between the two benchmarks.  With use-isets for mimalloc, it's not slower on average at all.

tjhance:
> Because Set::contains is defined in terms of to_iset

Do we want to export this definition at all? It sounds like it unnecessarily introduces `s.to_iset()` into the context.

jaylorch:
A question for @tjhance: Should I switch `PointsToRaw` to use `ISet` rather than `Set` (e.g., in `split`)?

jaylorch:
> > Because Set::contains is defined in terms of to_iset
> 
> Do we want to export this definition at all? It sounds like it unnecessarily introduces `s.to_iset()` into the context.

@Chris-Hawblitzel suggested this change to further automate proofs that involve `to_iset()`.

tjhance:
> A question for @tjhance: Should I switch PointsToRaw to use ISet rather than Set (e.g., in split)?

I personally think Set makes more sense, but you might want to get @elanortang 's opinion since this will impact her more.

elanortang:
> > A question for @tjhance: Should I switch PointsToRaw to use ISet rather than Set (e.g., in split)?
> 
> I personally think Set makes more sense, but you might want to get @elanortang 's opinion since this will impact her more.

I agree that Set makes more sense, since PointsToRaw can only represent a finite number of addresses. I haven't run into many difficulties that I can remember offhand due to Set being finite.

Chris-Hawblitzel:
> > > Because Set::contains is defined in terms of to_iset
> > 
> > 
> > Do we want to export this definition at all? It sounds like it unnecessarily introduces `s.to_iset()` into the context.
> 
> @Chris-Hawblitzel suggested this change to further automate proofs that involve `to_iset()`.

My motivation for this is to try to share across various set/map terms that are commonly used as triggers, like `Set::contains` and `ISet::contains`, so that we reduce the potential for trigger mismatches when people convert between `Set` and `ISet`.  I just measured the performance difference between this and a version that keeps `Set::contains` closed, and there's no difference (<1% on both mimalloc and capybarakv).

jaylorch:
I've created PRs for all the projects from veritas and I've tested that they work. (Except the vest projects, which don't seem to need a PR, since they don't use `Set` or `Map` and verify just fine with no changes.) Chris wrote the first drafts of the ones for verified-storage and verified-memory-allocator, and then I adapted them to the latest version of the finite-sets-and-maps PR.

Here are those PRs:

https://github.com/microsoft/verified-storage/pull/53
https://github.com/verus-lang/verified-memory-allocator/pull/6
https://github.com/verus-lang/verified-ironkv/pull/5
https://github.com/verus-lang/verified-node-replication/pull/6
https://github.com/secure-foundations/human-eval-verus/pull/49
https://github.com/matthias-brun/verified-nrkernel/pull/10
https://github.com/ziqiaozhou/svsm/pull/12
https://github.com/anvil-verifier/anvil/pull/801

jaylorch:
I think I've addressed all the comments and change requests in this conversation. And I've created PRs for all the verita projects. So I think it's ready to go.

tjhance (approved):

tjhance:
thanks for finishing this up!

jaylorch:
@Chris-Hawblitzel and I talked about when to merge this, and we think it's best to give folks a few days to look over the PRs I submitted to their repos. So, the plan is that, unless some blocking issues come up, we'll merge around Tuesday, June 9.

source/docs/guide/src/spec_lib.md
```diff
@@ -1,34 +1,69 @@
...
 This allows specifications to talk about collections that
 are larger than could be contained in the physical memory of a computer.
 
-## Constructing and using Seq, Set, Map
+The possibly-infinite versions have one key benefit: You can construct them
+without proving they're finite. The finite versions have two key benefits.
+First, they can be used in recursive types; e.g., an `enum T` can contain
+fields of type `Set<T>` and `Map<T, U>` but not `ISet<T>` or `IMap<T, U>`.
+Second, the verifier knows the finiteness property always holds, which can
+prevent some SMT-time proof failure surprises. For instance, adding an element
```
parno: Perhaps "adding a new element"?  After all, adding 1 to `{ 1 }` doesn't increase the length even if it's well-defined.

source/docs/guide/src/spec_lib.md
```diff
@@ -1,34 +1,69 @@
...
+to a `Set` increases its `len()` by 1, but this doesn't always hold for an
+`ISet` since it might have an undefined length.
```
parno: Undefined or unbounded?

source/docs/guide/src/spec_lib.md
```diff
@@ -1,34 +1,69 @@
...
 
 ```rust
 {{#include ../../../../examples/guide/lib_examples.rs:macro}}
 ```
 
-The macros above can only construct finite sequences, sets, and maps.
-There are also functions `Seq::new`, `Set::new`, and `Map::new`,
-which can allocate both finite values and (for sets and maps) infinite values:
+The macros above can only construct finite (literal) sequences, sets, and maps.
+Sequences can be constructed with `Seq::new`.
```
parno: "can also be"?

parno: same on the next line

source/docs/guide/src/spec_lib.md
```diff
@@ -37,6 +72,21 @@ which can allocate both finite values and (for sets and maps) infinite values:
 Each `Map<Key, Value>` value has a domain of type `Set<Key>` given by `.dom()`.
 In the `test_map2` example above, `m`'s domain is the finite set `{0, 10, 20, 30, 40}`,
 while `m_infinite`'s domain is the infinite set `{..., -20, 10, 0, 10, 20, ...}`.
+Likewise, each `IMap<Key, Value>` has a domain of type `ISet<Key>`.
+
+Some constructors of `Set<T>` don't necessarily produce a finite set; they
+produce an `Option<Set<T>>` that's only `Some` when the set would be finite.
+So they're only useful if you can prove that they produce a finite set. If you
```
parno: Perhaps "Thus," to avoid starting with "So"?

parno: Also, can we give a bit of advice about how to write such a proof?  I don't think it'll be obvious to the reader how they would do this, even for something "obvious" like `let option_set_a = Set::new(|x : nat| x < 42);`

parno (approved):
Added a few minor comments/questions on the documentation.

source/docs/guide/src/spec_lib.md
jaylorch: Good point, done.

source/docs/guide/src/spec_lib.md
jaylorch: Good point, done. And I combined them into a single sentence.

source/docs/guide/src/spec_lib.md
jaylorch: Good idea. Done.

source/docs/guide/src/spec_lib.md
jaylorch: I decided to just say "infinite size". I didn't want to get into the fact that when an `ISet` is of infinite size, the result of `len` isn't meaningful. But maybe we should?

parno: I like "infinite size" here -- I think that's the core reason.

Catoverflow:
I noticed that some helper predicates do not exists for finite DS (e.g. `submap_of` in `Map`). Do you have plan to add them in the future? I can also help on that.

Catoverflow:
And as Anvil is added to verita, there will be more breaking changes in the future and I don't want to burden Verus developers (thank you @jaylorch!) with proofs on our side. You directly @ me or @marshtompsxd in the breaking PR, we will always be there to help.

jaylorch:
> And as Anvil is added to verita, there will be more breaking changes in the future and I don't want to burden Verus developers (thank you @jaylorch!) with proofs on our side. You directly @ me or @marshtompsxd in the breaking PR, we will always be there to help.

Thanks!!!

jaylorch:
> I noticed that some helper predicates do not exists for finite DS (e.g. `submap_of` in `Map`). Do you have plan to add them in the future? I can also help on that.

There is a `submap_of` in the new (and old) `map_lib.rs`.

Catoverflow:
> There is a submap_of in the new (and old) map_lib.rs.

I somehow missed it. Thanks!

Catoverflow:
I tried to fix broken proofs in Anvil and found this proof fails
```rust
use vstd::prelude::*;
verus! {
proof fn repro(m: Map<int, int>, k1: int, k2: int)
    requires
        k2 != k1,
{
    assert(m.insert(k1, 0)[k2] == m[k2]);
}
fn main() {}
}
```
```shell
$ verus repro.rs
note: recommendation not met
  --> repro.rs:18:12
   |
18 |     assert(m.insert(k1, 0)[k2] == m[k2]);
   |            ^^^^^^^^^^^^^^^^^^^
   |
  --> vstd/map.rs:87:12
   |
   = note: recommendation not met

error: assertion failed
  --> repro.rs:18:12
   |
18 |     assert(m.insert(k1, 0)[k2] == m[k2]);
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ assertion failed

note: recommendation not met
  --> repro.rs:18:35
   |
18 |     assert(m.insert(k1, 0)[k2] == m[k2]);
   |                                   ^^^^^
   |
  --> vstd/map.rs:87:12
   |
   = note: recommendation not met

verification results:: 1 verified, 1 errors
```

Previously this would pass verus, now without explicit `m.dom().contains()` the proof fails to use `lemma_map_insert_different`. I guess for finite map this is expected but it still causes some regressions in Anvil. Would it be better to document this? It took me some time to pinpoint the cause.

jaylorch:
Interesting observation, thanks! I looked into it more, and found this:

It's always been the case that indexing into a map using a key that isn't in its domain produces a meaningless value. The way `insert` used to be defined for `Map` (and is still defined for `IMap`) ensures that those meaningless results don't change after an `insert`. But that doesn't work for finite maps, which are defined differently.

So it's not just a triggering issue. For finite maps, indexing into keys outside the domain gives arbitrary results, and those results might be changed by an `insert`.

In general, it's good practice to not index into maps outside their domain for any type of map. Even for `IMap` or the old `Map`, it's still a violation of the `recommends` clause to do so.

jaylorch:
As for documenting it, that's a good idea. Where would it have helped to have documentation? On `Map::insert`, or in the `recommends` clause for `Map::index`? Or both?

jaylorch:
The reason for this behavior is the different signatures of the broadcast lemmas `lemma_map_insert_domain` and `lemma_imap_insert_domain`. But I imagine no one is going to look at the documentation for those lemmas to figure out what's going wrong with their proofs. So I just added documentation to `Map::insert` instead.

jaylorch:
The ultimate reason why the behavior differs is that the new `Map` relies on this axiom:
```
    broadcast axiom fn axiom_new(s: Set<K>, fv: spec_fn(K) -> V)
        ensures
            #![trigger Self::new(s, fv)]
            Self::new(s, fv).dom() == s,
            forall|k| s.contains(k) ==> #[trigger] Self::new(s, fv)[k] == fv(k),
    ;
```

I could weaken this to remove the `s.contains(k) ==>` part, for backwards compatibility. But my first inclination is to leave it as is, to make as few assumptions as possible in this trusted axiom.

jaylorch:
Ah, never mind, the new axiom `Map::axiom_new` is based on the old `axiom_mk_map_index`, which did have that requirement. So I should leave that axiom alone.

The issue is just that `Map::insert` works differently than `IMap::insert`. So I think it's best to just leave it as is and document the difference.

Catoverflow:
> For finite maps, indexing into keys outside the domain gives arbitrary results

I agree enforcing domain check is better.

> On Map::insert, or in the recommends clause for Map::index? Or both?

`recommends` would be a good place, however, I guess many developers are used to failed recommends in domain check like in this case and may not realize the difference. It may be better to hint for this when recommends fail and point to a explaining page on website.

jaylorch:
@Chris-Hawblitzel and I just talked, and decided to preserve the property of `Map::insert` and `Map::remove` that values outside the domain retain their values when indexed. This requires changing a couple of lemmas to axioms.

Catoverflow:
Great! A lot of fixing efforts are saved. Thank you @jaylorch. And I guess it may be helpful to this reproduction proof to tests in Verus to prevent such in the future?

Issue #2363, opened 2026-04-22
`cargo verus verify` fails on renamed crate imports in workspace


dschoepe:
Given a multi-crate workspace with the following layout (an archive with this example is attached):

```
.
├── Cargo.lock
├── Cargo.toml
├── consumer
│   ├── Cargo.lock
│   ├── Cargo.toml
│   └── src
└── dependency
    ├── Cargo.lock
    ├── Cargo.toml
    └── src
```

If `consumer` depends on `dependency` through the workspace level `Cargo.toml` where `dependency` is made available under a different name using Cargo's [dependency renaming](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml) feature, `cargo verus verify` fails, while `cargo verus build` succeeds. `cargo verus verify` prints the following error:

```
error: extern location for renamed_dependency does not exist: /Volumes/workplace/cargo-verus-bug/target/debug/deps/libdependency-e20ca9acdcdbeb01.rmeta
 --> consumer/src/lib.rs:2:5
  |
2 | use renamed_dependency;
``` 

The dependency renaming looks as follows; in `./Cargo.toml`:

```
[workspace]
resolver = "2"
members = [
    "consumer", "dependency"
]

[workspace.dependencies]
renamed_dependency = { version = "0.1.0", path = "dependency", package = "dependency" }
```

And in `consumer/Cargo.toml`:

```
[package]
name = "consumer"
version = "0.1.0"
edition = "2021"

[dependencies]
vstd = "=0.0.0-2026-04-05-0114"
renamed_dependency.workspace = true

[package.metadata.verus]
verify = true

[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = [
  'cfg(verus_only)',
] }
```

`cargo verus build` succeeds.

dschoepe:
[cargo-verus-bug.zip](https://github.com/user-attachments/files/26986126/cargo-verus-bug.zip)

mkovaxx:
@dschoepe Thanks for the detailed report!

I tried to repro this issue at 99b6383baa59fd6e45a7ff73e0f01a47a777834a, and it seems to have been resolved. (The fix was likely #2403.)

Could you please try and see if it works on your end as well?

dschoepe:
I can confirm that this has been fixed now.

PR #2489, opened 2026-05-26
[cargo-verus] test: confirm expected behavior around package aliases

mkovaxx:
- Add a test to confirm `cargo-verus` behavior in the presence of workspace-level package aliases.
- Extend `Mock*` test utilities to allow creating such workspaces.


mkovaxx:
Example of why this is important: #2363

PR #2491, opened 2026-05-26
Simplify output fix from #2477

parno:
Implements @Chris-Hawblitzel's suggested simplification from the discussion of #2477.

### DIFF
--- a/source/vir/src/prune.rs
+++ b/source/vir/src/prune.rs
@@ -35,7 +35,7 @@ enum ReachedType {
     Float(u32),
     SpecFn(usize),
     Datatype(Dt),
-    FnDef(Fun, Vec<ReachedType>),
+    FnDef(Fun),
     StrSlice,
     Array,
     Primitive,
@@ -132,9 +132,7 @@ fn typ_to_reached_type(typ: &Typ) -> ReachedType {
         TypX::AnonymousClosure(..) => ReachedType::None,
         TypX::Datatype(dt, _, _) => ReachedType::Datatype(dt.clone()),
         TypX::Dyn(..) => ReachedType::None,
-        TypX::FnDef(fun, typs, _) => {
-            ReachedType::FnDef(fun.clone(), typs.iter().map(typ_to_reached_type).collect())
-        }
+        TypX::FnDef(fun, _, _) => ReachedType::FnDef(fun.clone()),
         TypX::Decorate(_, _, t) => typ_to_reached_type(t),
         TypX::Boxed(t) => typ_to_reached_type(t),
         TypX::TypParam(_) => ReachedType::None,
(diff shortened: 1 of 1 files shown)


parno:
@Chris-Hawblitzel This branch passes all of the tests, including some I added in #2477, but when I try running verita, both capybara and svsm panic.  I asked Claude to create a small, standalone example that fails on this PR but succeeds on `main`, and it produced the following:
```
pub trait HasItem { type Item; }

pub struct Ad<F>(pub F);
impl<F: FnOnce(u32) -> u32> HasItem for Ad<F> {
  type Item = F::Output;
}

pub struct W<I: HasItem> { pub i: I }

impl<I: HasItem> W<I> {
  pub uninterp spec fn seq(self) -> Seq<I::Item>;
}

// A trait with a single static method shared by every impl.
pub trait MyTrait {
  fn op(x: u32) -> u32;
}

impl MyTrait for u32 {
  fn op(x: u32) -> u32 { x }
}

mod a {
  mod inner {
      pub(super) struct Hidden { pub x: u32 }
  }

  // Hidden's impl of `MyTrait::op` shares the trait-method
  // `Fun` identifier with `u32`'s impl. The auto-generated
  // FnDef AssocTypeImpls are:
  //   <FnDef(MyTrait::op, [u32])   as FnOnce<(u32,)>>::Output = u32
  //   <FnDef(MyTrait::op, [Hidden]) as FnOnce<(u32,)>>::Output = u32
  // - same `Fun`, distinguished only by the typ_args vector.
  impl crate::MyTrait for inner::Hidden {
      fn op(x: u32) -> u32 { x }
  }
}

mod b {
  fn use_op() {
      // Binds `MyTrait::op` (for u32) as an FnDef value, so
      // pruning reaches `FnDef(MyTrait::op, [u32])`. The `W`/`Ad`
      // wrapper reaches the `FnOnce::Output` associated-type
      // decl via the `F::Output` projection.
      //
      // - With `ReachedType::FnDef(Fun)` (the first attempt),
      //   `FnDef(MyTrait::op)` matches both impls' AssocTypeImpls
      //   (the typ_args are dropped from the key), so Hidden's
      //   impl is pulled into this bucket. `Hidden` isn't
      //   visible to `b`, so `decoration_for_datatype` panics.
      // - With `ReachedType::FnDef(Fun, Vec<ReachedType>)` the
      //   keys are `[Int(U(32))]` vs `[Datatype(Hidden)]`, so
      //   only the u32 impl's AssocTypeImpl is reached and
      //   verification succeeds.
      let f = <u32 as crate::MyTrait>::op;
      let _y = crate::W { i: crate::Ad(f) };
  }
}
```
Is there another way you would fix this, rather than what's used in #2477?

PR #2494, opened 2026-05-27
[cargo-verus] feat: support forwarding verbosity settings to Cargo and Verus

mkovaxx:
- Support "stacked" `-v` flags i.e. `-vv`, `-vvv`, etc.
- When verbosity level >= 2, forward one less to Cargo, and exactly one to Verus (which does not support stacking).

### DIFF
--- a/source/cargo-verus/src/subcommands.rs
+++ b/source/cargo-verus/src/subcommands.rs
@@ -134,7 +134,7 @@ pub fn plan_cargo_run(cfg: VerusConfig) -> Result<CargoRunPlan> {
     //////////////////////////////////////////////////
     let metadata_args = {
         let for_cargo_metadata = true;
-        make_cargo_args(&cfg.options.cargo_opts, for_cargo_metadata)
+        make_cargo_args(&cfg.options.cargo_opts, for_cargo_metadata, cfg.options.verbosity)
     };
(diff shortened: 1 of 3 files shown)


source/cargo-verus/src/subcommands.rs
```diff
@@ -220,9 +226,19 @@ WARNING: You asked for verification, but cargo did not find any crates that opte
...
 }
 
-fn make_cargo_args(opts: &CargoOptions, for_cargo_metadata: bool) -> Vec<String> {
+fn make_cargo_args(opts: &CargoOptions, for_cargo_metadata: bool, verbosity: u8) -> Vec<String> {
     let mut args = vec![];
 
+    for _ in 1..verbosity {
+        args.push("-v".to_owned());
+    }
+    if verbosity > 0 {
```
bsdinis, line 235: Is this a leftover debug print?

bsdinis (approved):
thanks for this! generally lgtm except i think you might have left over a stray debug print.

source/cargo-verus/src/subcommands.rs
mkovaxx: No, that's totally intentional, so that in verbose mode, `cargo-verus` also explains how it propagates verbosity settings. :)

mkovaxx:
@bsdinis Thanks for the review!

PR #2496, opened 2026-05-27
implement if_trait_bound_then_redirect_to attribute

tjhance:
This:

```rust
#[verifier::if_trait_bound_then_redirect_to(helper)]
spec fn foo<T,...>(args...) { fallback_body }

spec fn helper<T,...>(args...) where BOUNDS { body }
```

is equivalent to this:

```rust
spec fn foo<T,...>(args...) {
    if BOUNDS { helper(args) } else { fallback_body }
}

spec fn helper<T,...>(args...) where BOUNDS { body }
```

We can add good syntax that desugars to a use of `verifier::if_trait_bound_then_redirect_to` later.

### DIFF
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -287,6 +287,8 @@ pub(crate) enum Attr {
     NoAutoTrigger,
     // when used in a ghost context, redirect to a specified spec method
     Autospec(String),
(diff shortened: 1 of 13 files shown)


tjhance:
Marking this as a draft because I realized this has cycle-checking problems (see the last commit).

PR #2500, opened 2026-05-28
Include auto-deref adjustments when handling patterns.

### LINKED ISSUE #2495: Verus crashes with tuple-destructuring of reference type in spec code
## Summary

A `for` loop with both a tuple-destructuring pattern and a named `iter: v` (as shown below) causes Verus to crash with an internal compiler error. Either of these by itself works fine.

## Minimal reproduction
[Playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=7081f5bd5b56a006ea70eb5b4ff84289)
```rust
use vstd::prelude::*;
verus! {
fn foo(v: &Vec<(u32, u32)>) {
    for (a, b) in iter: v
        invariant iter.index() >= 0,
    {
        let _ = (a, b);
    }
}
} // verus!
fn main() {}
```
Error:
```
error: internal compiler error: /rustc-dev/59807616e1fa2540724bfbac14d7976d7e4a3860/compiler/rustc_middle/src/mir/statement.rs:217:21: deref projection of non-dereferenceable ty PlaceTy { ty: (u32, u32), variant_index: None }


thread 'rustc' (26) panicked at /rustc-dev/59807616e1fa2540724bfbac14d7976d7e4a3860/compiler/rustc_middle/src/mir/statement.rs:217:21:
Box<dyn Any>
note: [run with `RUST_BACKTRACE=1` environment variable to display a backtrace](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=7081f5bd5b56a006ea70eb5b4ff84289#)
```

parno:
Fixes #2495

In #2495, the Verus for-loop desugaring wraps each user invariant `inv` in an invariant that introduces the bound terms:  `let (a, b) = IteratorSpec::peek(...).unwrap_or(arbitrary()); <inv>`.  In the example, since `v: &Vec<(u32, u32)>`, the RHS has type `&(u32, u32)`, and Rust's pattern lowering inserts an auto-deref into the THIR pattern:
  `PatKind::Deref { sub: Leaf{(a, b)} }`, where the outer pattern's ty is `&(u32, u32)`.

When Verus erases this spec let statement, it builds an `erased_ghost_value::<T>(())` placeholder using `T =
  cx.typeck_results.node_type(pat.hir_id)`, but `node_type` returns the pattern's type as written (`(u32, u32)`), not the adjusted type (`&(u32, u32)`). MIR building then tries to apply the `PatKind::Deref` against a place of type `(u32, u32)`, which crashes in `multi_projection_ty`.

The updated code uses `pattern.ty` (the adjusted THIR pattern type) instead of the HIR pattern's `node_type`.

### DIFF
--- a/source/rust_verify_test/tests/loops.rs
+++ b/source/rust_verify_test/tests/loops.rs
@@ -1844,3 +1844,17 @@ test_verify_one_file! {
       }
   } => Ok(())
 }
+
+test_verify_one_file! {
+  // https://github.com/verus-lang/verus/issues/2495
+  #[test] test_for_loop_tuple_destructure verus_code! {
+      use vstd::prelude::*;
+      fn foo(v: &Vec<(u32, u32)>) {
+          for (a, b) in iter: v
+              invariant iter.index() >= 0,
+          {
+              let _ = (a, b);
+          }
+      }
+  } => Ok(())
+}
--- a/source/rustc_mir_build_additional_files/verus.rs
+++ b/source/rustc_mir_build_additional_files/verus.rs
@@ -1085,7 +1085,8 @@ fn erase_let_for_pattern_checking<'tcx>(
 
     let pattern = erase_pat_all_binders(cx.pattern_from_hir(pat));
 
-    let init_ty = cx.typeck_results.node_type(pat.hir_id);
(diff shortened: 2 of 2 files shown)


parno:
I just saw that Travis implemented his own fix in #2497.

PR #2509, opened 2026-05-30
[rust_verify] fix: skip codegen when `--compile` is not set

### LINKED ISSUE #2504: Verus compiles even without `--compile`
Ever since #2280, when you run `verus` directly on a Rust file, it produces an executable, even when you don't pass in the `--compile` flag.  We also end up seeing Rust compilation warnings that we didn't see when we terminated the compilation pipeline early.  It would be nice to restore the old behavior for bare `verus` calls, while preserving the improvements for `cargo verus` that #2280 provided.

mkovaxx:
- Add an `after_expansion` hook to stop before codegen when `--compile` is not set.
- Add tests to confirm `--compile` flag behavior (both positive and negative), via a new `run_verus_raw` helper.

### DIFF
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -138,13 +138,12 @@ impl rustc_driver::Callbacks for CompilerCallbacksEraseMacro {
         }
     }
 
-    fn after_expansion<'tcx>(
+    fn after_analysis<'tcx>(
         &mut self,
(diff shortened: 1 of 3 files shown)


parno (approved):
This looks good to me, and a [run of verita](https://github.com/verus-lang/verus/actions/runs/26755883306/job/78855531693) succeeds, so I think we're in good shape.  Thanks!

PR #2511, opened 2026-05-31
Updates for refactored pointer arithmetic specs

zero-to-nat:
Updates for [https://github.com/secure-foundations/rustlib/pull/42](https://github.com/secure-foundations/rustlib/pull/42).

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -98,11 +98,13 @@ impl Provenance {
 }
 
 /// Allocations do not "wrap around" the address space.
-// Q: Where in the documentation do we get this fact?
-// A: Should be self-evident, comes from allocations being contiguous regions of memory
+/// See: https://doc.rust-lang.org/std/ptr/index.html#allocation
 pub broadcast axiom fn alloc_bound(p: Provenance)
     ensures
-        #[trigger] p.start_addr() + #[trigger] p.alloc_len() <= usize::MAX + 1,
+        #![trigger p.start_addr()]
+        #![trigger p.alloc_len()]
+        p.start_addr() + p.alloc_len() <= usize::MAX,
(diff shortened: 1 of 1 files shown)


elanortang (approved):
Looks good to me, except for my question to Chris about the typical Verus way to approach modulo.

source/vstd/raw_ptr.rs
```diff
@@ -1254,7 +1259,7 @@ impl PointsTo<str> {
     // Note that even for ZSTs, pointers need to be aligned.
     pub axiom fn is_aligned(tracked &self)
         ensures
-            self.ptr()@.addr as nat % spec_align_of_val::<str>(self.value()) == 0,
+            self.ptr()@.addr as int % spec_align_of_val::<str>(self.value()) as int == 0,
```
elanortang, line 1262: It seems like these two are equivalent in terms of the SMT encoding, as per Chris's response - I would lean towards the old version to reduce the number of casts cluttering things up? But can go either way if you have a strong preference/argument for the change.

zero-to-nat: It appears that there are some inconsistencies within this file about whether to use `nat` or `int` when doing mod. I made this change to be more compatible with the way that a particular safety condition is written in rustlib. It is probably best to make sure that we make a consistent choice of `nat` or `int` everywhere, but I'm not sure if this is in scope for this PR

elanortang: Gotcha, yeah I agree - I think I wasn't as careful about checking if double int casts are necessary in past PR reviews, and yes this does seem like its own clean-up Pr at some point. Maintaining compatibility with rustlib (especially for ensuring proper triggering) sounds good to me!

zero-to-nat: sounds good, thank you for the review!

PR #2515, opened 2026-06-01
Add specs for Iterator::filter

parno:
In the presence of possibly non-deterministic closure arguments to filter, it's hard to give a clean spec to `filter` as a whole.  Here's what I came up with.  It entailed adding a new Seq function, `filter_index`; the existing `Seq::filter` filters based on values, while the new `filter_index` filters based on the index within the sequence.

Suggestions for improvement are most welcome.


Chris-Hawblitzel (approved):

source/vstd/std_specs/iter.rs
```diff
@@ -373,6 +388,92 @@ impl <I> IteratorSpecImpl for &mut I
...
+
+// Ghost accessor for the inner iterator
+pub uninterp spec fn filter_iter<I, F>(r: Filter<I, F>) -> I;
+
+// Ghost accessor for the inner predicate
+pub uninterp spec fn filter_fun<I, F>(r: Filter<I, F>) -> F;
+
+// Ghost accessor for the sequence of predicate decisions made for the (prefix of) the
+// inner iterator's elements
+pub uninterp spec fn filter_keep<I, F>(r: Filter<I, F>) -> Seq<bool>;
```
tjhance, line 409: this should be `prophetic`

tjhance (changes requested):

source/vstd/std_specs/iter.rs
parno: Good point.  Done.

tjhance (approved):

PR #2516, opened 2026-06-01
Add support for transmuting SharedReference

zero-to-nat:
Support for transmuting shared references via the `SharedReference` type.

In order to verify a `transmute::<&Src, &Dst>(src, dst)`, you need to:
1. Create an exec `SharedReference<Src>`
2. Transmute the `&PointsTo<Src>` (obtained from the `SharedReference<Src>`) to `&PointsTo<Dst>`
3. Create a (tracked) `SharedReference<Dst>`
4. Invoke `transmute::<SharedReference<Src>, SharedReference<Dst>>` with the values from 1. and 3. Note that `SharedReference<T>` now has `#[repr(transparent)]`, meaning that its type layout is now the same as that of `&T`.
5. Create the `&Dst` from the `SharedReference<Dst>`

Summary of important changes:

[These axioms](https://github.com/verus-lang/verus/pull/2516/changes#diff-1ba496d71c04224de9c43c70abcceb63989b4c497d54f5960f2ac962facf9225R2849) are the proof-only versions of [this function](https://github.com/verus-lang/verus/pull/2516/changes#diff-1ba496d71c04224de9c43c70abcceb63989b4c497d54f5960f2ac962facf9225R2830). When transmuting `&Src` to `&Dst`, the axioms are needed to create `Tracked<SharedReference<Dst>>` without creating an exec `SharedReference<Dst>` (because that is what the transmute is supposed to do).

As discussed offline, the precondition for transmuting `PointsTo`'s has been [weakened here](https://github.com/verus-lang/verus/pull/2516/changes#diff-b3c452698e240396f4a0a3d0587c4f66cda45693236a13557b3fca27b3c27a4dR33).

### DIFF
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -2846,33 +2846,41 @@ pub fn ptr_ref2<'a, T>(ptr: *const T, Tracked(perm): Tracked<&PointsTo<T>>) -> (
 
 /// Same as [`ptr_ref2`], but operates on ghost values.
 /// Because this doesn't constitute a retag, the returned value's pointer has the same provenance as the original pointer.
-pub axiom fn ptr_ref2_ghost<'a, T>(ptr: *const T, tracked perm: &PointsTo<T>) -> (tracked v: SharedReference<'a, T>)
+pub axiom fn ptr_ref2_ghost<'a, T>(ptr: *const T, tracked perm: &PointsTo<T>) -> (tracked v:
+    SharedReference<'a, T>)
     requires
         perm.ptr() == ptr,
         perm.is_init(),
     ensures
         v.value() == perm.value(),
-        v.ptr() == ptr;
+        v.ptr() == ptr,
+;
 
 /// Same as [`ptr_ref2`], but operates on ghost values.
 /// Because this doesn't constitute a retag, the returned value's pointer has the same provenance as the original pointer.
-pub axiom fn ptr_ref2_str_ghost<'a>(ptr: *const str, tracked perm: &PointsTo<str>) -> (tracked v: SharedReference<'a, str>)
+pub axiom fn ptr_ref2_str_ghost<'a>(ptr: *const str, tracked perm: &PointsTo<str>) -> (tracked v:
+    SharedReference<'a, str>)
     requires
         perm.ptr() == ptr,
         perm.is_init(),
     ensures
         v.value() == perm.value(),
-        v.ptr() == ptr;
+        v.ptr() == ptr,
+;
 
 /// Same as [`ptr_ref2`], but operates on ghost values.
 /// Because this doesn't constitute a retag, the returned value's pointer has the same provenance as the original pointer.
-pub axiom fn ptr_ref2_slice_ghost<'a, T>(ptr: *const [T], tracked perm: &PointsTo<[T]>) -> (tracked v: SharedReference<'a, [T]>)
+pub axiom fn ptr_ref2_slice_ghost<'a, T>(
+    ptr: *const [T],
+    tracked perm: &PointsTo<[T]>,
+) -> (tracked v: SharedReference<'a, [T]>)
     requires
         perm.ptr() == ptr,
         perm.is_init(),
     ensures
         v.value()@ == perm.value(),
-        v.ptr() == ptr;
+        v.ptr() == ptr,
+;
 
 } // verus!
 /// Trusted wrapper around `ptr_ref`, due to
--- a/source/vstd/raw_ptr.rs
+++ b/source/vstd/raw_ptr.rs
@@ -2633,6 +2633,7 @@ pub fn deallocate(
 #[verifier::external_body]
 #[verifier::accept_recursive_types(T)]
 #[cfg_attr(verus_keep_ghost, rustc_diagnostic_item = "verus::vstd::raw_ptr::SharedReference")]
+#[repr(transparent)]
 pub struct SharedReference<'a, T: ?Sized>(&'a T);
 
 impl<'a, T: ?Sized> Clone for SharedReference<'a, T> {
@@ -2683,7 +2684,7 @@ impl<'a, T: ?Sized> SharedReference<'a, T> {
     }
 
     #[verifier::external_body]
-    const fn as_ref(self) -> (t: &'a T)
+    pub const fn as_ref(self) -> (t: &'a T)
         ensures
             t == self.value(),
     {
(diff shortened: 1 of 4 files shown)


source/vstd/raw_ptr.rs
```diff
@@ -2843,6 +2844,44 @@ pub fn ptr_ref2<'a, T>(ptr: *const T, Tracked(perm): Tracked<&PointsTo<T>>) -> (
     SharedReference(unsafe { &*ptr })
 }
 
+/// Same as [`ptr_ref2`], but operates on ghost values.
+/// Because this doesn't constitute a retag, the returned value's pointer has the same provenance as the original pointer.
```
elanortang, line 2853: Why doesn't this constitute a retag? Is it because we view ptr+PointsTo and SharedReference as equivalent?

source/vstd/transmute.rs
```diff
@@ -285,10 +285,10 @@ impl<T: AbstractByteRepresentation> PointsTo<T> {
 }
 
 impl PointsTo<str> {
-    pub axiom fn transmute_shared<'a>(tracked &'a self, target: &[u8]) -> (tracked ret:
+    pub axiom fn transmute_shared<'a>(tracked &'a self, tracked target: &[u8]) -> (tracked ret:
```
elanortang, line 292: add documentation

source/vstd/transmute.rs
```diff
@@ -299,4 +299,25 @@ impl PointsTo<str> {
     ;
 }
 
+impl PointsTo<[u8]> {
+    pub axiom fn transmute_shared<'a>(
```
elanortang, line 311: add documentation

source/vstd/raw_ptr.rs
zero-to-nat: My thinking is that this is not a retag because this is a ghost operation, but I could be wrong.

Right now, having the provenance stay the same is a requirement for verifying transmutes between `SharedReference` types. Our spec for transmute on `SharedReference`s requires that the provenance of the src and dst pointers be the same (or that the dst pointer have an empty provenance). This is because the provenance is part of the `AbstractByte` encoding. 

However, if the `ptr_ref2_ghost` is unsound as written, I can think of how to rework this.

elanortang: @tjhance does this constitute a retag or no, since it's an axiom fn instead of an exec fn?

@zero-to-nat could you give more details on how the `SharedReference` transmute spec provenance requirement means that this operation must preserve provenance?

source/vstd/transmute.rs
```diff
@@ -285,6 +285,10 @@ impl<T: AbstractByteRepresentation> PointsTo<T> {
 }
 
 impl PointsTo<str> {
+    /// Creates a `&PointsTo<[u8]>` from a `&PointsTo<str>`.
+    /// Requires that it is possible to transmute between the pointed-to value of `self`
+    /// and the provided value `target`. This value (`target`) will become the pointed-to value
+    /// of the resuling `&PointsTo<[u8]>`.
```
elanortang: typo resuling -> resulting

source/vstd/transmute.rs
```diff
@@ -300,6 +304,10 @@ impl PointsTo<str> {
 }
 
 impl PointsTo<[u8]> {
+    /// Creates a `&PointsTo<str>` from a `&PointsTo<[u8]>`.
+    /// Requires that it is possible to transmute between the pointed-to value of `self`
+    /// and the provided value `target`. This value (`target`) will become the pointed-to value
+    /// of the resuling `&PointsTo<str>`.
```
elanortang: and here

source/vstd/raw_ptr.rs
tjhance: it's a proof fn, not an exec fn, so by definition it can't have any operational meaning like a retag

zero-to-nat: For example, to transmute `x: &str` to `y: &[u8]`:

```
let src: SharedReference<str> = SharedReference::new(x);
let tracked src_pts_to: &PointsTo<str> = src.points_to();
let tracked dst_pts_to: &PointsTo<[u8]> = src_pts_to.transmute_shared(x.as_bytes());
  // ensures dst_pts_to.value()@ == src_pts_to.value()@
let ghost dst_ptr = src.ptr() as *const [u8];
  // ensures dst_ptr == ptr_mut_from_data(
  //      PtrData::<[T]> { addr: src_ptr@.addr, provenance: src_ptr@.provenance, metadata: src_@.metadata },
  //   )
  // ^this is from cast_slice_ptr_to_str_ptr
let tracked dst_ghost: SharedReference<[u8]> = ptr_ref2_slice_ghost(dst_ptr, dst_pts_to);
  // ensures: ??? (postcondition in question)
let dst: SharedReference<[u8]> = transmute(src, Tracked(dst_ghost));
  // requires: forall |bytes: Seq<AbstractByte>| encode(src, bytes) ==> decode(bytes, dst_ghost)
let y: &[u8] = dst.as_ref();
```

Thus, the postcond of `ptr_ref2_slice_ghost` must allow us to satisfy the transmute precond.

By knowing `encode(src, bytes)`, we have that each `AbstractByte` in `bytes` has the same provenance as `src.ptr()@.provenance`.

And to show `decode(bytes, dst_ghost)`, we need to show that each `AbstractByte` in `bytes` has the same provenance as `dst_ghost.ptr()@.provenance`. We therefore need `src.ptr()@.provenance == dst_ghost.ptr()@.provenance`.

If `ptr_ref2_slice_ghost` did not guarantee that `dst_ghost.ptr()@.provenance == dst_ptr@.provenance`, then we would have no way to prove this. (note we already have that `src.ptr()@.provenance == dst_ptr@.provenance`)

zero-to-nat: @tjhance, more broadly, is it sound if `let dst = transmute::<SharedReference<Src>, SharedReference<Dst>>(src, dst_ghost);` ensures that `dst.ptr()@.provenance == src.ptr()@.provenance`? I'm not sure if/how transmutes are modeled in tree/stacked borrows

tjhance: as i understand it, that condition should be enforced by the _precondition_

source/vstd/type_representation.rs
```diff
@@ -1007,6 +1076,66 @@ raw_ptr_encoding_from_type_representation! {
...
+/// It is not sound to **directly** transmute values of type `&T` in Verus.
+/// In Verus, `&T` is interpreted simply as the pointed-to value of type `T`.
+/// However, the layout for `&T` is the same as that of raw pointers.
+/// Thus, according to the Rust specification for transmute,
+/// it is safe to transmute any value of type `&T` to any value of type `&U`, as long as they have the same pointer layout.
+/// However, in Verus, this would mean that you could transmute any value of type `T` to any value of type `U`, and this is unsound.
+pub broadcast axiom fn shared_ref_cannot_be_encoded<T: ?Sized>()
+    ensures
+        !(#[trigger] abs_can_be_encoded::<&T>()),
+;
```
tjhance, line 1088: do you need this axiom for anything?

tjhance (approved):

source/vstd/raw_ptr.rs
zero-to-nat: okay got it, thanks! This was also my understanding

source/vstd/type_representation.rs
zero-to-nat: It is not strictly necessary. I added it to remind ourselves that we should not try to transmute directly on &T. But I can remove it if you think it will be confusing

PR #2522, opened 2026-06-04
Add a blanket impl of IteratorSpecImpl for &mut references

parno:
so that clients don't see mysterious proof failures when working with &mut Iterator values.  I was quite puzzled when code similar to (but larger than) the test this PR adds failed.  Hopefully this blanket impl will prevent such surprises in the future.

### DIFF
--- a/source/vstd/std_specs/iter.rs
+++ b/source/vstd/std_specs/iter.rs
@@ -258,6 +258,40 @@ impl <I> DoubleEndedIteratorSpecImpl for Rev<I>
     }
 }
 
+// Forwarding spec impl for the Rust-supplied blanket `impl<I> Iterator for &mut I`.
+// Without this, bare method-call syntax on a `i: &mut I` receiver (e.g. `i.remaining()`)
+// resolves to these (otherwise uninterpreted) functions on `&mut I` rather than on `I`,
+// silently disconnecting clients from `I`'s actual specs.
+impl <I> IteratorSpecImpl for &mut I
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/iter.rs
```diff
@@ -258,6 +258,40 @@ impl <I> DoubleEndedIteratorSpecImpl for Rev<I>
...
 }
 
+// Forwarding spec impl for the Rust-supplied blanket `impl<I> Iterator for &mut I`.
+// Without this, bare method-call syntax on a `i: &mut I` receiver (e.g. `i.remaining()`)
+// resolves to these (otherwise uninterpreted) functions on `&mut I` rather than on `I`,
+// silently disconnecting clients from `I`'s actual specs.
+impl <I> IteratorSpecImpl for &mut I
+    where I: Iterator {
+    open spec fn obeys_prophetic_iter_laws(&self) -> bool {
+        (**self).obeys_prophetic_iter_laws()
```
tjhance: can you use fully qualified syntax for all these, to make it easier to confirm the type arguments?

`<I as IteratorSpec>::obeys_prophetic_iter_laws(**self)`

tjhance (changes requested):

source/vstd/std_specs/iter.rs
parno: Done

tjhance (approved):

PR #2523, opened 2026-06-04
rust_verify: build bucket AIR context once, shared across spun-off contexts

mmcloughlin:
Refactor bucket AIR context construction to reduce duplication and ensure equivalence in spin-off queries.

The verifier builds a per-bucket AIR context and populates it with background declarations: fuel, datatypes, traits, etc.  However, this sequence was constructed twice: once for the main context and repeated in `new_air_context_with_bucket_context` used for spinoff queries.  The code duplication risks them getting out of sync, and indeed the spinoff version does not have the "Opaque-Type-Constructors" context.

This PR refactors to avoid duplication and therefore fix the bug.  The bucket context is built as a list of titled command batches (`CommandBatch`) which is run at the start of the main bucket, and then replayed into each spun-off context. This also has the benefit that the per-query context ops are just appended to the same list.  As a side-benefit, `new_air_context_with_bucket_context` now has seven fewer parameters (17 to 10).

This change should be behavior-preserving. The only difference is that prelude and fuel commands are now counted under `time_air` timing data (which I think is intended behavior).

Adds a regression test for the opaque type constructors bug. On the main branch this test fails with `ill-typed AIR code: ...`.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1,5 +1,5 @@
 use crate::boundary_suggestions::build_boundary_suggestion;
-use crate::commands::{Op, OpGenerator, OpKind, QueryOp, Style};
+use crate::commands::{OpGenerator, OpKind, QueryOp, Style};
 use crate::config::{Args, CargoVerusArgs, ShowTriggers};
 use crate::context::{ContextX, ErasureInfo};
 use crate::debugger::Debugger;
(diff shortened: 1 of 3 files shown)


tjhance:
What exactly does `-V spinoff-all` do and what's it for?

tjhance (approved):
seems like a straightforward improvement regardless

mmcloughlin:
> What exactly does `-V spinoff-all` do and what's it for?

It causes all queries to be spunoff.

That option predates this PR so I'm unsure of the history. Based on its introduction in #863, it looks like it was motivated by profiling.

mmcloughlin:
> seems like a straightforward improvement regardless

For background, I'm working on an experiment that emits "slim" `bit_vector` queries without the prelude and bucket context.

This refactor was motivated by making that plumbing a bit cleaner to implement, and I thought it was probably worthwhile change whether we adopt these slim `bit_vector` queries or not.

PR #2530, opened 2026-06-05
verita: apply project filters before pre-flight checks

mmcloughlin:
The `--project` / ignore filtering happened inline in the run loop, after the Singular and cargo-verus precondition checks. So for example if you do not have singular, you would still be blocked from running `--project vest` because a different project in the configuration requires singular.

This PR changes the order: compute the active project set up front and base the precondition checks on it.


parno (approved):
Seems like a sensible refactoring.  Thanks!

PR #2531, opened 2026-06-05
rust_verify: apply per-query solver tuning in the verifier

mmcloughlin:
Move the `nonlinear` and `bit_vector` prover options out of the SST command stream and into the verifier.

This removes duplication (options are currently created twice in `sst_to_air.rs`), and also provides a central point at which options can be tuned based on other factors available to the verifier.

Refactor should be behavior-neutral, since it only makes a minor ordering change to the SMT output.


parno (approved):
It does seem useful to consolidate these prover configuration settings into one location.

PR #2534, opened 2026-06-07
spinoff_prover_all for parallel verification at function granularity

### LINKED ISSUE #2458: Parallel verification at function granularity
Verus's parallel verifier dispatches one module (bucket) per worker thread, so a module with many functions is verified serially by a single worker on one core. As a result, oversized modules dominate the overall wall-clock verification time. I found that simply moving the work-unit boundary from module to function gives a 2.28× wall-time speedup on our controller code, and I'm opening this issue to discuss how it could be supported natively.

## Measurement

I (with the help of Claude) added a flag `-V spinoff-prover-all` that treats every function as if it carried `#[verifier::spinoff_prover]`, reusing the existing spinoff dispatch path (`new_air_context_with_bucket_context`) so each function becomes its own top-level bucket scheduled by the existing FIFO worker pool. I then verified [Anvil](https://github.com/anvil-verifier/anvil)'s VDeployment controller with and without the flag:

| Mode | Buckets | Total SMT work | Wall (makespan) | Speedup |
|------|--------:|---------------:|----------------:|--------:|
| Default | 23 modules | 679 s | 351 s | — |
| `-V spinoff-prover-all` | 356 functions | 916 s (+35%) | **154 s** | **2.28×** |

### Per-thread breakdown

Default (module granularity): one module on T2 carries the entire makespan; most other threads finish in under 10 s.

Image

`-V spinoff-prover-all` (function granularity): the same work spreads evenly across all eight threads.

Image

Each colored segment is one function's SMT time, extracted from `--time-expanded --output-json` logs. The per-thread layout is reconstructed by simulating the existing scheduler (lexicographic FIFO + first-idle dispatch) rather than instrumenting the workers directly; the simulated makespan (black dashed line) tracks the real verify-crate wall-clock (red dashed line) within 4%, so the simulation faithfully reflects what Verus actually does.

### Possible regression

**1 out of 130 proof broke** after `spinoff_prover` is applied to all functions and is fixed by simply giving more rlimit to Z3 (the overall rlimit increased from `29,737,854, default` to `32,326,105, spin_off for all`). Claude suspected this to be caused by additional rlimit spent on initializing the context for each function.

Catoverflow:
This PR closes #2458

### DIFF
--- a/source/rust_verify/src/buckets.rs
+++ b/source/rust_verify/src/buckets.rs
@@ -85,16 +85,18 @@ impl Bucket {
 /// Arrange the given modules into buckets.
 /// Typically, this means 1 bucket per module;
 /// However, any functions marked 'spinoff_prover' get their own bucket.
+/// If `force_spinoff_all` is true, every function is treated in this way
 pub fn get_buckets(
     krate: &Krate,
     modules_to_verify: &Vec<vir::ast::Module>,
+    force_spinoff_all: bool,
 ) -> Vec<(BucketId, Bucket)> {
     let mut map: HashMap<BucketId, Vec<Fun>> = HashMap::new();
     let module_set: HashSet<&Path> = modules_to_verify.iter().map(|m| &m.x.path).collect();
     for func in &krate.functions {
         if let Some(owning_module) = &func.x.owning_module {
             if module_set.contains(owning_module) {
-                let bucket_id = if func.x.attrs.spinoff_prover {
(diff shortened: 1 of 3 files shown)


Chris-Hawblitzel (changes requested):
Rather than using a command-line flag, can you make this a crate-level / module-level attribute `#![verifier::spinoff_prover_all]`, which would apply spinoff_prover_all to the whole crate or the whole module and child modules?

PR #2535, opened 2026-06-08
Add documentation about `cargo verus`

parno:
I think this probably replaces the section heading for "Working with crates", but it's possible whoever wrote that had something else in mind. 

On a related note, what is "Invoking Verus code from Rust" supposed to cover?

### DIFF
--- a/source/docs/guide/src/cargo_verus.md
+++ b/source/docs/guide/src/cargo_verus.md
@@ -31,6 +31,23 @@ This creates a new directory with a correctly-configured `Cargo.toml`, an initia
 `src/main.rs` (or `src/lib.rs`), and a `.gitignore`. The generated project already has
 `vstd` as a dependency and the `[package.metadata.verus]` section described below.
 
+## Updating an existing Rust project to support Verus
+
+To enable verification for one or more of the crates in your project,
+add the following to each of their `Cargo.toml` files:
+
+```toml
+[package.metadata.verus]
+verify = true
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(verus_only)'] }
+```
+See below for more details.
+
+Note that you can still use normal `cargo` commands (e.g., `cargo build`) on
+crates and projects that include Verus annotations.
+
 ## Cargo.toml configuration
 
 Crates that should be verified must opt in by adding a `[package.metadata.verus]`
--- a/source/docs/guide/src/cargo_verus.md
+++ b/source/docs/guide/src/cargo_verus.md
@@ -92,6 +92,10 @@ changed crates are re-verified.
 Like `verify`, but skips re-verification of dependencies. Useful when you are only
 iterating on root crates and their dependencies have not changed.
 
+Unlike `verify`, `focus` stores its artifacts in a separate target directory
+under `target/verus-partial`, so that a full `cargo verus verify` isn't misled
+by the cached partial results.
+
 ```bash
 # Verify all of the root crates, but not their dependencies
 cargo verus focus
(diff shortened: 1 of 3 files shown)


tjhance:
> On a related note, what is "Invoking Verus code from Rust" supposed to cover?

Pretty much what it says, how do you set up a project if one crate is a verified verus crate and the parent crate is a normal Rust crate. Though, it sounds like this pretty much will "just work"? So it probably doesn't need its own page.

tjhance:
Probably a better page would be "considerations for a project that has an arbitrary mix of verified and unverified crates"

source/docs/guide/src/cargo_verus.md
```diff
@@ -0,0 +1,161 @@
...
+# Binary project
+cargo verus new --bin my_project
+
+# Library project
+cargo verus new --lib my_library
+```
+
+This creates a new directory with a correctly-configured `Cargo.toml`, an initial
+`src/main.rs` (or `src/lib.rs`), and a `.gitignore`. The generated project already has
+`vstd` as a dependency and the `[package.metadata.verus]` section described below.
```
tjhance, line 32: I think we should have a section "## Updating an existing cargo project to support verus" that just says "Paste the following lines into Cargo.toml", separate from the sections that explain what these lines do.

parno:
> Though, it sounds like this pretty much will "just work"? 

Yes, at least in theory, running `cargo verus verify` and running `cargo build` should just work on a mixed verified/unverified project.  The former should ignore crates that haven't opted-into verification, and the latter shouldn't see any of the Verus annotations/types.

source/docs/guide/src/cargo_verus.md
parno: Okay, I'll add it as a sibling to "Starting a new project".

parno: Done.

source/docs/guide/src/cargo_verus.md
```diff
@@ -0,0 +1,178 @@
...
+
+- **Before `--`**: forwarded to `cargo` 
+- **After `--`**: forwarded to every `verus` invocation
+
+```bash
+cargo verus verify -p my_crate --release -- --rlimit 60 --expand-errors
+#                  ^^^^^^^^^^^^^^^^^^^^^    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#                  Cargo arguments            Verus arguments
+```
+
```
mkovaxx, line 135: Might be helpful to document the sharp edges of Cargo arg forwarding.

```suggestion
Verus-relevant Cargo options must come before generic Cargo flags such as `--release`.

The Verus-relevant options are:
- `-p`/`--package`, `--workspace`, `--all`, `--exclude`
- `--features`, `--all-features`, `--no-default-features`
- `--manifest-path`, `--target-dir`
- `--frozen`, `--locked`, `--offline`
- `--config`, `-Z`

The `-Z` option must be written with a space, e.g. `-Z unstable-options`, the compact form is not supported.
```

source/docs/guide/src/cargo_verus.md
```diff
@@ -0,0 +1,178 @@
...
+Use `verify` for day-to-day proof development. Incremental builds mean that only
+changed crates are re-verified.
+
+### `cargo verus focus`
+
+Like `verify`, but skips re-verification of dependencies. Useful when you are only
+iterating on root crates and their dependencies have not changed.
+
```
mkovaxx, line 94: Should we add a short paragraph here about the `TARGET_DIR` behavior of `focus`?

```suggestion
Unlike `verify`, `focus` stores its artifacts in a separate target directory under `target/verus-partial`, so that partially verified results cannot interfere with a full `cargo verus verify`.
```

source/docs/guide/src/cargo_verus.md
```diff
@@ -0,0 +1,178 @@
...
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(verus_only)'] }
+```
+
+`cargo verus new` adds this automatically. See [Ghost Erasure](./erasure.md) for a full
+explanation of `verus_only` and when to use it.
+
+## Subcommands
+
+### `cargo verus verify`
```
mkovaxx, line 77: Should this section mention `cargo verus check` as well?

mkovaxx (approved):
Looks good to me overall! I've added a few non-blocking comments/suggestions.

Chris-Hawblitzel (approved):

source/docs/guide/src/cargo_verus.md
parno: Good idea -- done.

source/docs/guide/src/cargo_verus.md
parno: Good idea -- done.

source/docs/guide/src/cargo_verus.md
parno: At the moment, `cargo verus check` is redundant, I believe.  Unless we come up with a use case for it, I think we should eliminate it soon.

PR #2537, opened 2026-06-08
Add documentation for traits

parno:
While working on Iterator specs, I realized that we were missing basic documentation on how Verus thinks about traits, so I filled some in.


source/docs/guide/src/traits.md
```diff
@@ -0,0 +1,174 @@
...
+To implement `View` for a custom type, choose an appropriate abstraction and
+then define `type V` and `spec fn view`:
+
+```rust
+{{#include ../../../../examples/guide/traits.rs:view_impl}}
+```
+
+Because `Stack`'s `data` is a private field, `view` is `closed` — callers cannot see its
+definition, but they can still reason about the effect each function has on that view,
+as illsutrated by the postconditions on `push` and `is_empty`.  If you want
```
Chris-Hawblitzel: illustrated

source/docs/guide/src/traits.md
```diff
@@ -0,0 +1,174 @@
+# Traits
+
+Verus supports writing specifications for trait methods, including `requires` and
```
Chris-Hawblitzel: It's probably better to use the term "trait function" rather than "trait method".  Internally, Verus is sloppy about this distinction, but my understanding is that not all trait functions are referred to as methods.

Chris-Hawblitzel (approved):

source/docs/guide/src/traits.md
parno: Good catch -- fixed.

source/docs/guide/src/traits.md
parno: Okay, I've swapped them all to "function" instead of "method".

PR #2539, opened 2026-06-09
Add specs for Iterator::all and Iterator::any

parno:
These mostly follow the same pattern as `find`.  In the tests, it's a bit annoying that to prove `assert(exists |i| 0 <= i < v.len() && v[i] >= 10);`, you need to explicitly construct the witness via `let ghost idx = g.remaining().len() - it.remaining().len() - 1;`.  I don't an easy way to avoid that.  Someday when it's feasible to add extra ghost/tracked arguments to external functions, I could imagine returning an extra ghost `idx` value, which would be the witness, if appropriate, so a client could simply pull it out without worrying about how it was computed.

### DIFF
--- a/source/rust_verify_test/tests/iterators.rs
+++ b/source/rust_verify_test/tests/iterators.rs
@@ -104,6 +104,62 @@ test_verify_one_file! {
 }
 
 test_verify_one_file! {
+    #[test] all_works verus_code! {
+        use vstd::prelude::*;
+        use vstd::std_specs::iter::IteratorSpec;
+
+        fn test(v: Vec<u32>)
+        {
+            let mut it = v.into_iter();
+            let ghost g = it;
+            let v_result = it.all(
+                |i: u32| -> (ret: bool)
+                    ensures ret == (i < 10)
+                {i < 10}
+            );
+            if v_result {
+                // If `all` returned true, every element was below 10.
+                assert(forall |i| 0 <= i < v.len() ==> v[i] < 10);
+            } else {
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel:
`find`, `any`, and `all` are implemented on top of `try_fold`.  Do we need an extra requirement that `try_fold` obeys the expected behavior?  Otherwise, someone could implement a bad `try_fold` while inheriting the default `find`, `any`, and `all`.

tjhance:
Should we suppose that `obeys_prophetic_iter_laws` implies that `try_fold` is implemented correctly?

parno:
@Chris-Hawblitzel That's a good point that I hadn't considered.  We could add a spec for `Iterator::try_fold` that uses `ensures`, rather than `default_ensures` to try to pin down minimal requirements for what `try_fold` should do.  I'm also open to Travis's suggestion of considering it implicit in `obeys_prophetic_iter_laws`, at least for now.

Chris-Hawblitzel:
> We could add a spec for `Iterator::try_fold` that uses `ensures`, rather than `default_ensures` to try to pin down minimal requirements for what `try_fold` should do. I'm also open to Travis's suggestion of considering it implicit in `obeys_prophetic_iter_laws`, at least for now.

These all sound fine to me, as long as we have something now so that we remember to handle `try_fold` carefully in the future.

parno:
> These all sound fine to me, as long as we have something now so that we remember to handle try_fold carefully in the future.

Okay, I've added a note in iter.rs to keep track of this.

PR #2540, opened 2026-06-09
Liberalize our policy on specifications for trait and impl fn specs

parno:
Liberalize our policy on trait specifications: A specification must be provided for a trait implementation function either via
a spec on the trait function's definition, or on the impl (whereas before we required both, which introduced unnecessary redundancy).  Providing a spec on both is now optional and only necessary if the impl wants to provide a stronger postcondition than the trait definition spec.


Chris-Hawblitzel:
Specifically, this covers the case where an external trait `T` has a verus `#[verifier::external_trait_specification]` declaration, but a specific external impl of `T::f` has no `#[verifier::external_fn_specification]`, and Verus code is calling that specific impl of `T::f`.  Previously, the call was unconditionally considered an error.  Now, it is allowed and is interpreted as a Dynamic call to `T::f` rather than a DynamicResolved call to the impl.

Chris-Hawblitzel (approved):

PR #2527, opened 2026-06-05
verita: add vstd as a benchmark project

mmcloughlin:
Add `vstd` as a verita project. This is configured in the normal way without any special-casing, besides the `--is-vstd` extra verus options. This does mean that it's possible for the verus version under test and the vstd version to be different.

Tested locally:

```
> cat vstd.toml
verus_git_url = "https://github.com/verus-lang/verus.git"
verus_refspec = "main"
verus_extra_args = ["--no-report-long-running"]

# Successes

[[project]]
name = "vstd"
git_url = "https://github.com/verus-lang/verus.git"
refspec = "main"
crate_roots = ["source/vstd/vstd.rs"]
extra_verus_args = [
    "--crate-type=lib",
    "--is-vstd",
    "--cfg", "feature=\"std\"",
    "--cfg", "feature=\"alloc\"",
]

> cargo run --release -- --verus-repo ../.. --label vstd vstd.toml
    Finished `release` profile [optimized] target(s) in 0.04s
     Running `target/release/verita --verus-repo ../.. --label vstd vstd.toml`
   0.002375196s  INFO running project vstd
   5.774846623s  INFO running target source/vstd/vstd.rs (1 of 1)

--- Run Summary ---
Total projects: 1
Succeeded (1): vstd
Output: output/2026-06-05-19-51-03-323-vstd
```

The output `output/2026-06-05-19-51-03-323-vstd/vstd.json` looked reasonable.

<small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>

PR #2542, opened 2026-06-09
Slim bit-vector queries

mmcloughlin:
By design, `bit_vector` queries are self-contained and assert one obligation.
Therefore, they do not require Verus's prelude, however the current
implementation includes it anyway.  This PR updates bit-vector queries to verify
in a "slim" context without the standard prelude, recommended-options preset,
or bucket background.

The motivation is not just the size of the SMT files, but also the logic mix
involved in the prelude potentially throws off heuristics.  Emitting minimal SMT
additionally opens up the possibility of logic-specific tuning, or even using
dedicated solvers in future with limited theory support (e.g. bitwuzla).

One consequence is that `smt_get_model`, which localizes to a failing assertion,
is inapplicable to a single non-incremental check over one obligation. Reporting
at the query level is the existing behavior for bit-vector queries anyway. This
PR renames the `disable_incremental_solving` flag to `single_check_query` and
skips `smt_get_model` when it is set, which is the case for every bit-vector
query. This preserves existing bit-vector error reporting behavior.

The verita evaluation suggest this change is both correct and beneficial.
Details below.

## Evaluation

Run `verita` suite (plus `vstd` #2527) on main and this branch.

<details>
<summary>Script: <code>verita-compare.sh</code></summary>

```sh
#!/usr/bin/env bash

set -exuo pipefail

REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
export PATH="${REPO_ROOT}/tools/vargo/target/release:${PATH}"

for branch in main slim-bitvector; do
    git switch "${branch}"

    ## Rebuild.
    pushd "${REPO_ROOT}/source"
    vargo build --release --features singular
    popd

    ## Run verita.
    pushd "${REPO_ROOT}/tools/verita"
    cargo run --release -- \
        --verus-repo "${REPO_ROOT}" \
        --label "${branch}" \
        --output-dir "output/${branch}" \
        run_configuration_all.toml
    popd
done
```

</details>

Adhoc script for data analysis based on the deterministic `rlimit` metrics:

<details>
<summary>Script: <code>slim_bv_compare.py</code></summary>

```py
#!/usr/bin/env python3

import glob, json, os, sys


def stats(path):
    data = json.load(open(path))
    smt = data["times-ms"]["smt"]
    results = data["verification-results"]
    funcs = {}
    for module in smt.get("smt-run-module-times", []):
        for entry in module.get("function-breakdown", []):
            funcs[entry["function"]] = funcs.get(entry["function"], 0) + entry.get(
                "rlimit", 0
            )
    return (
        results["verified"],
        results["errors"],
        smt["rlimit-init"] + smt["rlimit-run"],
        funcs,
    )


def pct(off, on):
    return (on - off) / off * 100 if off else 0


off_dir, on_dir = sys.argv[1], sys.argv[2]
for path in sorted(glob.glob(os.path.join(off_dir, "*.json"))):
    name = os.path.basename(path)
    off_verified, off_errors, off_rlimit, off_funcs = stats(path)
    on_verified, on_errors, on_rlimit, on_funcs = stats(os.path.join(on_dir, name))

    # rlimit is deterministic, so a function whose rlimit changed is exactly a bit_vector one.
    bv = [f for f in off_funcs if off_funcs[f] != on_funcs.get(f)]
    bv_off = sum(off_funcs[f] for f in bv)
    bv_on = sum(on_funcs.get(f, 0) for f in bv)
    bv_report = (
        f"bit_vector {pct(bv_off, bv_on):+.1f}% over {len(bv)} funcs ({bv_off:,}->{bv_on:,})"
        if bv
        else "no bit_vector queries"
    )

    print(
        f"{name[:-5]}|"
        f"verified {off_verified}->{on_verified}|errors {off_errors}->{on_errors}|"
        f"rlimit {pct(off_rlimit, on_rlimit):+.1f}%|{bv_report}"
    )
```

</details>

Run it on the output directories:

```
python3 slim_bv_compare.py tools/verita/output/{main,slim-bitvector} | column -t -s"|"
```

Full results:

```
anvil                        verified 140->140    errors 0->0  rlimit +0.0%   no bit_vector queries
human-eval                   verified 366->366    errors 0->0  rlimit +0.0%   no bit_vector queries
ironkv                       verified 312->312    errors 0->0  rlimit +0.0%   no bit_vector queries
memory-allocator             verified 736->736    errors 0->0  rlimit -3.4%   bit_vector -61.0% over 27 funcs (29,945,087->11,672,831)
node-replication             verified 253->253    errors 0->0  rlimit +0.0%   no bit_vector queries
page-table                   verified 597->597    errors 0->0  rlimit -1.2%   bit_vector -4.4% over 47 funcs (474,963,660->454,036,575)
svsm                         verified 240->240    errors 0->0  rlimit -2.8%   bit_vector -94.0% over 8 funcs (2,904,769->174,831)
verified-storage-capybaraKV  verified 722->722    errors 0->0  rlimit -0.2%   bit_vector -59.9% over 9 funcs (1,341,602->538,251)
verified-storage-multilog    verified 164->164    errors 0->0  rlimit +0.0%   no bit_vector queries
verified-storage-pmemlog     verified 80->80      errors 0->0  rlimit +0.0%   no bit_vector queries
vest-vest-examples           verified 2555->2555  errors 0->0  rlimit +0.0%   no bit_vector queries
vest-vest                    verified 506->506    errors 0->0  rlimit -24.7%  bit_vector -39.4% over 20 funcs (19,111,962->11,590,153)
vstd                         verified 1858->1858  errors 0->0  rlimit -20.3%  bit_vector -54.9% over 51 funcs (23,693,137->10,675,833)
```

Summary:

* No regressions on any project
* Unchanged for the non-bitvector projects
* Significant speedups for bitvector queries
* Bitvector heavy projects benefit overall: 24% for `vest`, 20% for `vstd`

---

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1155,6 +1155,7 @@ impl Verifier {
         is_rerun: bool,
         prelude_config: vir::prelude::PreludeConfig,
         profile_file_name: Option<&std::path::PathBuf>,
+        prover_choice: vir::def::ProverChoice,
     ) -> Result<air::context::Context, VirErr> {
         let mut air_context =
             air::context::Context::new(message_interface.clone(), self.args.solver);
@@ -1223,27 +1224,37 @@ impl Verifier {
             air_context.set_smt_transcript_log(Box::new(file));
         }
 
-        // air_recommended_options causes AIR to apply a preset collection of Z3 options
-        air_context.set_z3_param("air_recommended_options", "true");
+        let bv_slim = self.is_bv_slim(prover_choice);
+        // air_recommended_options applies AIR's preset Z3 options; a prelude-free (slim) query omits
+        // it, along with the prelude and bucket background.
+        if !bv_slim {
+            air_context.set_z3_param("air_recommended_options", "true");
+        }
         self.set_default_rlimit(&mut air_context);
         for (option, value) in self.args.smt_options.iter() {
             air_context.set_z3_param(&option, &value);
         }
-        if self.args.axiom_usage_info {
-            air_context.enable_usage_info();
+        if !bv_slim {
+            if self.args.axiom_usage_info {
+                air_context.enable_usage_info();
+            }
+            self.run_command_batch(
+                bucket_id,
+                diagnostics,
+                &mut air_context,
+                &CommandBatch::new("Prelude", ctx.prelude(prelude_config)),
+            );
         }
-
-        self.run_command_batch(
-            bucket_id,
-            diagnostics,
-            &mut air_context,
-            &CommandBatch::new("Prelude", ctx.prelude(prelude_config)),
-        );
 
         air_context.blank_line();
         air_context.comment(&("MODULE '".to_string() + &bucket_id.friendly_name() + "'"));
 
         Ok(air_context)
+    }
+
+    /// A `by(bit_vector)` query runs in a prelude-free context under `-V slim-bitvector`.
+    fn is_bv_slim(&self, prover_choice: vir::def::ProverChoice) -> bool {
+        self.args.slim_bitvector && prover_choice == vir::def::ProverChoice::BitVector
     }
 
     /// Per-query SMT tuning options.
@@ -1253,6 +1264,7 @@ impl Verifier {
         prover_choice: vir::def::ProverChoice,
     ) {
         match prover_choice {
+            vir::def::ProverChoice::BitVector if self.is_bv_slim(prover_choice) => {}
             vir::def::ProverChoice::BitVector => match self.args.solver {
                 air::context::SmtSolver::Z3 => {
                     air_context.set_z3_param("sat.euf", "true");
@@ -1284,6 +1296,7 @@ impl Verifier {
         span: &vir::messages::Span,
         profile_file_name: Option<&std::path::PathBuf>,
         spinoff_reason: &str,
+        prover_choice: vir::def::ProverChoice,
     ) -> Result<air::context::Context, VirErr> {
         let mut air_context = self.new_air_context_with_prelude(
             ctx,
@@ -1294,6 +1307,7 @@ impl Verifier {
             is_rerun,
             PreludeConfig { arch_word_bits: ctx.arch_word_bits, solver: self.args.solver },
             profile_file_name,
+            prover_choice,
         )?;
 
         // Write the span of spun-off query
@@ -1301,8 +1315,10 @@ impl Verifier {
         air_context.blank_line();
         air_context.comment(&format!("query spun off because: {}", spinoff_reason));
 
-        // set up bucket context
-        self.run_command_batches(bucket_id, diagnostics, &mut air_context, bucket_context);
+        // set up bucket context (skipped for a prelude-free slim query)
+        if !self.is_bv_slim(prover_choice) {
+            self.run_command_batches(bucket_id, diagnostics, &mut air_context, bucket_context);
+        }
 
         Ok(air_context)
     }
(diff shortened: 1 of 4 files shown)


tjhance:
Does this even need to be optional experiment? If you already ran it on verita and found no regressions, it seems like a clear improvement.

mmcloughlin:
> Does this even need to be optional experiment? If you already ran it on verita and found no regressions, it seems like a clear improvement.

Doesn't have to be. When I began this experiment I thought it might be a risky change, so I added the flag in an abundance of caution. But the evaluation seems clear-cut.

A couple of things to confirm:

* Bit-vector queries are constructed to be isolated, by design? There isn't some lurking edge case where it would reference the prelude?
* Are there any other projects we should check that aren't included in `verita`?

source/rust_verify/src/verifier.rs
```diff
@@ -1223,36 +1224,47 @@ impl Verifier {
...
-        );
-
         air_context.blank_line();
         air_context.comment(&("MODULE '".to_string() + &bucket_id.friendly_name() + "'"));
 
         Ok(air_context)
     }
 
+    /// A `by(bit_vector)` query runs in a prelude-free context under `-V slim-bitvector`.
+    fn is_bv_slim(&self, prover_choice: vir::def::ProverChoice) -> bool {
```
parno: It looks like this function is defined here, but used above on line 1227.  Can we move the definition up above the first use?

parno:
I agree with Travis that I don't think this needs to be gated behind a feature flag, given the existing testing.

> Bit-vector queries are constructed to be isolated, by design? 

Correct.

> Are there any other projects we should check that aren't included in verita?

I think verita covers all the most bv-heavy instances that we have ready access to.

parno:
It looks like this PR eliminates the Z3 set-options that we traditionally used for BV queries.  Did you try the slim version with/without them?  More generally, I guess we could try a broader collection of options now that the BV queries are in a purer state.

mmcloughlin:
Updated to remove the feature flag gate. Full CI tests pass.

> It looks like this PR eliminates the Z3 set-options that we traditionally used for BV queries. Did you try the slim version with/without them? More generally, I guess we could try a broader collection of options now that the BV queries are in a purer state.

Yes. The previous set of options was the `air_recommended_options` pre-prelude followed by four bit-vector specific options before the query.

https://github.com/verus-lang/verus/blob/3e443e7291230f2cf3a56b600eb476d7df70c706/source/air/src/context.rs#L311-L319

https://github.com/verus-lang/verus/blob/3e443e7291230f2cf3a56b600eb476d7df70c706/source/rust_verify/src/verifier.rs#L1258-L1261

The experiments I ran suggest that compared to no options at all (as in this PR), these options are approximately 70-80% worse on QF formulae. They do offer a ~15% improvement on quantified bit-vector (e.g. in memalloc) but that's relatively rare.

I think just removing the options makes sense for this PR, but I agree a follow-up to try bit-vector specific options would be worth it. I updated the code to include a comment and a TODO about this.

The other thing I'd like to follow up with is the bit-vector queries providing the minimal logic they need. Most are `QF_BV`, but some need quantifiers, `FP`, or even Reals. So the minimal logic can be `(QF_)?BV(FP)?` or `ALL`. I have some experimental local changes where the `bitvector_to_air.rs` translator computes this as it builds the query. Given this we can:

* Try emitting `set-logic` and seeing if Z3's per-logic tuning is better
* Have bespoke `set-option` lists for each of the logic kinds
* Dispatch to alternative solvers based on logic (e.g. bitwuzla can handle `(QF_)?BV(FP)?` but not `ALL`)

All of this is optional follow-ups. I think the PR is a clear win as-is.

source/rust_verify/src/verifier.rs
mmcloughlin: This function is gone now we're not gating on the `-V slim-bitvector` option as well.

parno (approved):

PR #2548, opened 2026-06-11
Update example path for Linux and macOS

MelchizedekTP:


parno (approved):
Thanks for the fix!

PR #2550, opened 2026-06-12
[cargo-verus] feat: toolchains

mkovaxx:
- Adds `cargo-verus-toolchains` to define toolchain manifests and operations.
- Extend `cargo-verus/build.rs` to parse `toolchain-manifests/*.toml` and embed data into the binary.
- Add *test* toolchain manifest files to confirm that `build.rs` works in CI.
- Implement a simple `cargo verus toolchains list` command to demo usage.


mkovaxx:
This is step 1 of the plan for implementing the version checking part of the toolchain functionality:
1. Extend `cargo-verus` to embed toolchain data.
2. Extend the `release.yml` CI flow to create a permanent `cargo-verus/toolchain-manifests/$VERUS_VERSION.toml` file before building the archive.
3. Extend `cargo-verus` to check that `verus --version` and the resolved `vstd` version are compatible.
4. Extend the `rolling-release.yml` CI flow to create a temporary `cargo-verus/toolchain-manifests/rolling-release.toml` file before building the archive.
5. Merge the `crate-updates.yml` flow into the `release.yml` flow.

source/cargo-verus/src/toolchains.rs
```diff
@@ -0,0 +1,31 @@
+/// A set of Verus components meant to be used together.
```
parno: Should we be sharing these definitions with those in `source/cargo-verus-toolchains/src/lib.rs`?

parno (approved):
Looks good me.  For the overall plan, which phase will add the `external-deps.toml` file?

mkovaxx:
> Looks good me. For the overall plan, which phase will add the `external-deps.toml` file?

Step 2, extending the Release workflow.

A helper executable will be added in `cargo-verus-toolchains` to gather information and create the `toolchain-manifests/$VERUS_VERSION.toml` file. As part of that, it'll need to read the Z3 version from the `external-deps.toml` file, which will be added in the same change set.

source/cargo-verus/src/toolchains.rs
mkovaxx: I've merged the two `Toolchain` types. This is slightly more complex and less self-documenting than before (e.g. `version` is no longer a named field), but not having to keep the two sides in sync is likely a win overall.

PR #2551, opened 2026-06-12
resource: replicate Map/Set changes in GhostMap/GhostSet

bsdinis:

### DIFF
--- a/source/vstd/resource/combinators/agree.rs
+++ b/source/vstd/resource/combinators/agree.rs
@@ -1,6 +1,6 @@
-use super::super::prelude::*;
-use super::algebra::Resource;
-use super::algebra::ResourceAlgebra;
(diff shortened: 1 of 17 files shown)


jaylorch:
Looks good! But is it possible to not introduce the new directory level, for greater backward compatibility (and simpler paths)?

bsdinis:
> Looks good! But is it possible to not introduce the new directory level, for greater backward compatibility (and simpler paths)?

The directory level does not impact backward compatibility or the paths. On the `resource/mod.rs` there are things like `pub use impls::map`, for instance, retaining the path `vstd::resource::map` from before. It does make the difference for contributors about where combinators land and where the user-facing resources live.

jaylorch (approved):

PR #2552, opened 2026-06-12
source/tools: collect `rust_verify_test` code coverage

mmcloughlin:
Adds a `source/tools/coverage.sh` script to generate a test-coverage report for the `rust_verify_test` test suite.

Example HTML report:

Screenshot 2026-06-12 at 2 45 36 PM

---

### DIFF
--- a/source/.gitignore
+++ b/source/.gitignore
@@ -3,4 +3,5 @@ target/
 /cvc5
 .verus-log/
 doc/
+coverage/
 /target-verus/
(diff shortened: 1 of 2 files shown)


parno (approved):
Seems like a useful addition, even though it will be some time before we have the resources necessary to focus on increasing coverage.

mmcloughlin:
> Seems like a useful addition, even though it will be some time before we have the resources necessary to focus on increasing coverage.

Oh yes I imagine so. The use case for me was checking whether code-paths I was editing in the bit-vector query generator had test coverage.

PR #2559, opened 2026-06-14
Fix panic from cross-crate `FnDef` impl-path collision

mmcloughlin:
The synthetic `FnDef` trait-impl that Verus generates for each `clone` method built its impl path from the function's path segments only, dropping the crate and placing it under `CrateId::Internal`.

Two crate-root impls in different crates can share a `rustc` disambiguator. For example `verus_builtin` has `Clone for Ghost` assigned to `impl&%7`, which can collide with a definition in a user crate.

This PR uses the function's krate in the path instead of `CrateId::Internal`. Adds a regression test.

Related: #2474

---

### DIFF
--- a/source/rust_verify_test/tests/fndef_types.rs
+++ b/source/rust_verify_test/tests/fndef_types.rs
@@ -2037,3 +2037,28 @@ test_verify_one_file_with_options! {
         }
     } => Ok(())
 }
+
+test_verify_one_file_with_options! {
+    // Regression test for a panic due to fndef impl path collision across
+    // crates.
+    //
+    // The test defines 12 `Clone` impls in the crate root, which are enough to
+    // collide with the `Clone` impls for `Ghost` and `Tracked` in
+    // `verus_builtin` at the same disambiguator (brought in by `["vstd"]`).
+    #[test] fndef_impl_path_includes_crate ["vstd"] => verus_code! {
+        use vstd::prelude::*;
+
+        struct S0;  impl Clone for S0  { fn clone(&self) -> Self { S0  } }
+        struct S1;  impl Clone for S1  { fn clone(&self) -> Self { S1  } }
+        struct S2;  impl Clone for S2  { fn clone(&self) -> Self { S2  } }
+        struct S3;  impl Clone for S3  { fn clone(&self) -> Self { S3  } }
+        struct S4;  impl Clone for S4  { fn clone(&self) -> Self { S4  } }
+        struct S5;  impl Clone for S5  { fn clone(&self) -> Self { S5  } }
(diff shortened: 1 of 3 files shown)


source/rust_verify_test/tests/fndef_types.rs
```diff
@@ -2037,3 +2037,28 @@ test_verify_one_file_with_options! {
...
     } => Ok(())
 }
+
+test_verify_one_file_with_options! {
+    // Regression test for a panic due to fndef impl path collision across
+    // crates.
+    //
+    // The test defines 12 `Clone` impls in the crate root, which are enough to
+    // collide with the `Clone` impls for `Ghost` and `Tracked` in
+    // `verus_builtin` at the same disambiguator (brought in by `["vstd"]`).
```
mmcloughlin, line 2047: This is quite a fragile test in that it relies on a disambiguator collision with definitions in `verus_builtin`.

The ideal test would setup two synthetic crates to force a collision. However, I'm not sure there's an easy way to setup multi-crate tests? Most `rust_verify_test` tests are single file.

tjhance (commented):
The definition of `impl_fndef` still seems a little dubious to me:

```
pub(crate) fn impl_fndef(fun: &Fun, kind: ClosureKind) -> Ident {
    let joined =
        fun.path.segments.iter().map(|s| s.as_str()).collect::<Vec<_>>().join(PATH_SEPARATOR);
    Arc::new(format!("{}{}{}", PREFIX_IMPL_FNDEF, kind, joined))
}
```

e.g. it seems like there could be a collision if the first segment of joined starts with "Once" or "Mut"?

Can we come up with a different strategy entirely here? There doesnt' seem to be any reason to collapse the segments. We could do something like:

```
fun.path.push_segment(format!("{}{}", PREFIX_IMPL_FNDEF, kind))
```

mmcloughlin:
> Can we come up with a different strategy entirely here? There doesnt' seem to be any reason to collapse the segments. We could do something like:
> 
> ```
> fun.path.push_segment(format!("{}{}", PREFIX_IMPL_FNDEF, kind))
> ```

Much nicer! Implemented. Please take another look @tjhance.

tjhance (approved):

source/rust_verify_test/tests/fndef_types.rs
parno: [This directory](https://github.com/verus-lang/verus/tree/main/source/rust_verify_test/tests/cargo-tests) is setup to support multi-crate `cargo verus` tests.  Want to add something more stable there?

parno:
Thanks for noticing and fixing this.

PR #2561, opened 2026-06-15
Add verifier::shadow_data

Chris-Hawblitzel:
This adds tracking of shadow data associated with exec function parameters, return values, and local variables.  This is useful when the SMT encoding of these values omits some data that is irrelevant to ordinary Verus verification but that may be useful to lower-level code.  In particular, while the SMT encoding of a shared reference `&t` is just the SMT encoding of `t`, with no underlying pointer information, the shadow data for a `&t` parameter, return value, or local variable can contain the extra pointer information.

To opt into this feature, a function uses the attribute `#[verifier::shadow_data]`.  This creates extra shadow variables in the SST/AIR/SMT representation of that function.  Non-`shadow_data` functions that call `shadow_data` functions pass in uninterpreted shadow data for parameters and ignore shadow data for return values.  Likewise, when a `shadow_data` function calls a non-`shadow_data` function, the parameter shadow data is ignored and the return shadow data is uninterpreted.

In general, each parameter, return value, and local variable `x` of type `t` gets shadow data `shadow_data(x)` of the entirely abstract type `ShadowData<t>`.  For example:

```
#[verifier::shadow_data]
fn ident(x: &u64) -> (r: &u64)
    ensures shadow_data(r) == shadow_data(x)
{
    x
}

#[verifier::shadow_data]
fn test_ident1(x: &u64, y: &u64) {
    let xx = ident(x);
    let yy = ident(y);
    assert(shadow_data(xx) == shadow_data(x));
    assert(shadow_data(yy) == shadow_data(y));
}

fn test_ident2(x: &u64, y: &u64) {
    let xx = ident(x);
    let yy = ident(y);
}
```

The shadow data feature is independent of any particular types -- it is not specialized in any way to `&t` types.  To use shadow data to get pointer information in `&t` types, you can declare an uninterp spec function that extracts pointer information from the type `ShadowData<&T>` for all `T`.

You can just as easily have shadow data for values of types other than `&t`, like `Option<&t>` or `(&t, &t)`.  To use shadow data inside an Option, for example, you could do the following:

```
uninterp spec fn option_some_shadow_data<T>(s: ShadowData<T>) -> ShadowData<Option<T>>;
uninterp spec fn option_unwrap_shadow_data<T>(s: ShadowData<Option<T>>) -> ShadowData<T>;

#[verifier::shadow_data]
#[verifier::external_body]
fn option_some_with_shadow<T>(x: T) -> (r: Option<T>)
    ensures
        r == Some(x),
        shadow_data(r) == option_some_shadow_data(shadow_data(x)),
        // TODO: this should be a separate axiom:
        shadow_data(x) == option_unwrap_shadow_data(option_some_shadow_data(shadow_data(x))),
{
    Some(x)
}

#[verifier::shadow_data]
#[verifier::external_body]
fn option_unwrap_with_shadow<T>(x: Option<T>) -> (r: T)
    requires
        x.is_some(),
    ensures
        r == x.unwrap(),
        shadow_data(r) == option_unwrap_shadow_data(shadow_data(x)),
{
    x.unwrap()
}

#[verifier::shadow_data]
fn test_option(x: &u64) {
    let o = option_some_with_shadow(x);
    let y = option_unwrap_with_shadow(o);
    assert(shadow_data(y) == shadow_data(x));
}
```

The implementation of shadow data builds on https://github.com/verus-lang/verus/pull/2560 , using extra SST parameters and return values to represent the shadow data values.  For now, the `call_requires` and `call_ensures` axioms are disabled for any functions marked `#[verifier::shadow_data]`.  Within a function, tracking of shadow data is currently very conservative: direct local variable assignment, function call arguments, and function call return values are tracked precisely, but all other operations generate uninterpreted shadow data.

### DIFF
--- a/source/builtin/src/lib.rs
+++ b/source/builtin/src/lib.rs
@@ -2361,6 +2361,27 @@ pub use decreases_to;
 #[cfg(verus_verify_core)]
 pub use decreases_to_internal;
 
+/// Type of shadow data that accompanies parameters, return values, and local variables
+/// in verifier::shadow_data functions.
+/// This does not contain the value A,
+/// but instead may contain extra data associated with a value of type A.
+/// In particular, it contains pointer data for shared reference types &t,
+/// for the purpose of verifying low-level code in the Rust standard library.
+#[cfg_attr(verus_keep_ghost, rustc_diagnostic_item = "verus::verus_builtin::ShadowData")]
+#[cfg_attr(verus_keep_ghost, verifier::external_body)]
+#[cfg_attr(verus_keep_ghost, verifier::reject_recursive_types(A))]
+pub struct ShadowData<A> {
+    phantom: PhantomData<A>,
+}
+
+/// Construct a spec ShadowData for a parameter, return value, or local variable
+#[cfg(verus_keep_ghost)]
+#[rustc_diagnostic_item = "verus::verus_builtin::shadow_data"]
+#[verifier::spec]
+pub fn shadow_data<A>(_a: A) -> ShadowData<A> {
+    unimplemented!()
+}
+
 #[cfg(verus_keep_ghost)]
 #[rustc_diagnostic_item = "verus::verus_builtin::infer_spec_for_loop_iter"]
 #[verifier::spec]
--- a/source/rust_verify/src/attributes.rs
+++ b/source/rust_verify/src/attributes.rs
@@ -396,6 +396,7 @@ pub(crate) enum Attr {
     MigratePostconditionsWithMutRefs(bool),
     TrackedSwap,
     TrackedTakeOption,
+    ShadowData,
 }
 
 fn get_trigger_arg(span: Span, attr_tree: &AttrTree) -> Result<u64, VirErr> {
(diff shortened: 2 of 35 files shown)


elanortang:
I only ran into a couple issues when I was trying this out with `run_utf8_validation`:
- The branch `verify-rustlib-shadow-data` panics when I try to build it, with this error message:
```
Some checks are taking longer than 2s (diagnostics for these may be reported out of order)
• vstd/arithmetic/div_mod.rs:503:1: 503:69 has finished in 18s
  ⌝ function body check for vstd::arithmetic::div_mod::lemma_div_denominator                                                                  
thread '<unnamed>' (23727199) panicked at rust_verify/src/verifier.rs:972:21:
unexpected output from solver: vstd/arithmetic/div_mod.rs:503:1: 503:69 (#0) (:reason-unknown "")

thread '<unnamed>' (23727199) panicked at rust_verify/src/verifier.rs:2238:29:
worker thread panicked

thread 'main' (23727073) panicked at vstd_build/src/main.rs:159:9:
vstd build failed with exit code Some(1)
```
Adding a temporary `verifier::external_body` to `lemma_div_denominator` suppresses the error.

PR #2563, opened 2026-06-15
Add `VERUS_USE_RUSTUP` flag to allow running without `rustup`

lschuermann:
This removes a hard dependency on `rustup` being available for the `verus` binary crate wrapper. It does not affect any build-time dependencies on `rustup` or any default behavior.

While relying on, and calling out to `rustup` is sane default behavior, having it as a hard dependency unnecessarily restricts the environments that Verus can work in. Provided that a user sets up an environment with the correct Rust toolchain in path, Verus will happily work without `rustup`.

This is particularly useful for running Verus in hermetic environments like a Bazel or Nix build sandbox. I verified that the `VERUS_USE_RUSTUP=0` mode works in a pure Nix devshell, with the proper Rust nightly toolchain provided (and the `librustc_driver` in the `LD_LIBRARY_PATH`).

I've tried to keep the diff as small and mechanical as possible (reviewing [without whitespace changes](https://github.com/verus-lang/verus/pull/2563/changes?w=1) shows only the `Command` constructor moved and conditionals added). At a later point, some of this logic could probably be cleaned up or deduplicated.

### DIFF
--- a/source/verus/src/main.rs
+++ b/source/verus/src/main.rs
@@ -64,6 +64,10 @@ fn warning(msg: &str) {
 }
 
 fn run() -> Result<std::process::ExitStatus, String> {
+    // If instructed, skip the rustup toolchain sanity check & trust the user to
+    // have set up their environment to point to the correct toolchain.
+    let use_rustup: bool = std::env::var("VERUS_USE_RUSTUP").map_or(true, |v| v != "0");
+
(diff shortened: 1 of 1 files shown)


parno (approved):
Looks reasonable to me.  As a small simplification, can we assign `std::env::var("VERUS_USE_RUSTUP").as_ref().map(String::as_str) != Ok("0")` to a variable, so we don't repeat that clause twice?

PR #2568, opened 2026-06-16
Add support for nuanced no-cheating policies

parno:
This implements [the policy we converged on](https://github.com/verus-lang/verus/discussions/1292) ~2 years ago.  To summarize, we use familiar-looking Rust syntax (e.g., the various [lint levels](https://doc.rust-lang.org/rustc/lints/levels.html)) to make the notation accessible, but Verus imposes additional policy on top of this when run with `--no-cheating`. The policy enforce:

1. The crate root must start with `#![deny(verus::assumptions)]`.
2. The `#[allow(verus::assumptions)]` annotation can only appear in the crate root, and only on the `mod name` lines.
3. An `#[allow(verus::assumptions)]` module must correspond to a file-level module (to make it easy to find and audit trusted code)
4. Modules on the allow list must be transitively closed; they can't use any non-allow modules, although they can reference external crates.

### DIFF
--- a/source/docs/guide/src/assert_assume.md
+++ b/source/docs/guide/src/assert_assume.md
@@ -418,7 +418,11 @@ verification results:: verified: 1 errors: 0
 ```
 
 It works!
-Now we've eliminated all the `assume`s, so we've completed the verification:
+Now we've eliminated all the `assume`s, so we've completed the verification.
+To guarantee no stray `assume`s remain, Verus offers a [no-cheating
+mode](./tcb.md#no-cheating-mode) that enforces this.
+
+Here is the finished proof:
 
 ```rust
 pub proof fn lemma_len_intersect<A>(s1: Set<A>, s2: Set<A>)
--- a/source/docs/guide/src/calling-unverified-from-verified.md
+++ b/source/docs/guide/src/calling-unverified-from-verified.md
@@ -5,6 +5,10 @@ verified code to call unverified code. To do this, we need to make Verus
 aware of the unverified code, and we need to tell Verus what it should
 **assume without proof**.
 
+The mechanisms below (`external_body`, `assume_specification`) introduce unverified assumptions,
+so they are rejected under [no-cheating mode](./tcb.md#no-cheating-mode) unless they appear in an
+explicitly-allowed module.
+
 ## Specifications without proof
 
 One way to apply an assumption to an unverified function is to use the `#[verifier::external_body]` attribute.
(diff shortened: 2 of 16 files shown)


bsdinis:
From my POV, the file restriction is too restrictive:
1. If people want it, it is easily auditable (check that #[allow(...)] is followed by `mod` or better yet checking it's the #![allow])
2. Having it on statements seems useful to me:
3. Allows for running no-cheating to check that added `assumes` for proof development were fixed
4. Permits other forms of auditing/trusted specs where someone does not want to fully separate the top level spec from the code, and allows for other disciplines on how to make the TCB clear

mmcloughlin:
Proposal: `assume`, `external_body`, ... are treated and implemented as standard `rustc` lints. They use existing `rustc` libraries to define lints and check effective lint level.  Therefore developers can enable/suppress them with `allow`, `deny`, ... attributes, command-line options, and Cargo configuration.

Under this design, `--no-cheating` effectively becomes `--forbid verus::assumptions`. This also doubles as a report because it would list all of the places the lint fires.

I think this benefits from simplicity, and it will match developers existing experience of how the `#[allow(...)]` etc. attributes behave. To @bsdinis's point, this wouldn't have the file restriction, and it could be added at the function or statement level.

My understanding is this doesn't satisfy the auditability goal. Could that be served by an extension or a separate tool?

parno:
After some additional discussion at recent Verus meetings, I think we're converging on the following design:

You still indicate which modules/files are trusted (and hence should be audited) in the root of your crate.  In the longer term, we might try to distinguish trusted imports (e.g., `assume_specification`) from trusted exports (e.g., the spec for an external API that may be called by unverified code), but we plan to defer that for now.  

In a trusted file, the code is expected to be transitively closed (not the module imports), but you can add a function-level annotation that says "the auditor needs to review the signature/specification but can ignore the body".  The bodies of such functions are ignored for transitivity checking purposes, but the signatures are not.  This makes it feasible to connect untrusted code to trusted specifications.

We would also instrument Verus (or a tool that ships with Verus) to be able to compute a snapshot of the trusted code for a code base (crate?).  The snapshot would capture all of the existing trusted code (as defined above) and emit either the code itself or a hash of it.  You could then rerun the snapshot to see if someone (e.g., an AI agent) tampered with the trusted definitions.  This is an important use case for which people are already using ad hoc solutions, so it would be useful to provide a common, supported way to do it.

PR #2576, opened 2026-06-17
Add impl, is/isnt, and deref support to exec_spec

JakeGinesin:
Extending the `exec_spec_verified!` and `exec_spec_unverified!` macros to support impl blocks for structs and enums, `is` and `isnt` expressions, and deref. 

### impl blocks for structs and enums

`compile_impl()` translates `impl T { spec fn myfun(&self) -> { ... } ... }` into a parallel `impl ExecT { exec_myfun(&self) -> { ... } ... } `. The original spec impl is kept. Each generated exec method gets a postcondition like `response.deep_view() <eq> self.deep_view().spec_name(args)`. 

To exec-ify user-written `recommends/decreases` clauses on specs with `self`, the generated exec functions will be preluded with `let __exec_spec_self_view: T = self.deep_view()`, and all references to `self` in the clauses are replaced with`__exec_spec_self_view`. This is done because the user's spec clauses are written assuming `self: &T`, but the generated exec method will be `self: &ExecT`, so we must create a local binding within the generated exec function to reference `T`.  

Generic and trait implementations are still unsupported. 

### `is/isn't` exprs
For `Option` and `Result`, a direct `matches!(e, Variant(_))` is emitted. For user-defined enums, one `exec_is_<Variant>(&self) -> bool` predicate per-variant is emitted. 

### deref (`*x`)
Compiles `Expr::Unary(UnOp::Deref)` by requesting the inner expression with `VarMode::Owned`, then re-borrowing with `get_ref()` if the caller wants a reference. 

### Other improvements
- The `Expr::MethodCall` arm has a new fallthrough that handles unknown method names by rewriting `recv.myfun(args)` to `recv.exec_myfun(args)`, with `recv` and each argument exec-compiled in `VarMode::Ref`. This fallthrough is needed to support user-defined functions inside of impl blocks. If `exec_myfun` doesn't exist on the receiver type, `rustc` will error, "`exec_myfun` instead of `myfun`". 
- `compile_sig` is now generalized from taking in an `&ItemFn` to taking in a `(&Signature, &Visibility, Option<&Ident>)`. The `Option<&Ident>` is the spec self type when compiling an impl method. 
- Support for self-prefixed paths (e.g. `Self::Variant` and `Self::mymethod`) was also added. 

Many tests exercising the above behavior were added, and no existing tests were modified. 


source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -692,19 +726,87 @@ fn compile_enum(item_enum: &ItemEnum) -> Result<TokenStream2, Error> {
...
+///
+/// `replace_self_tokens` exists to allow `compile_sig` to support specs with
+/// the self type. `compile_sig` emits deep-view snapshot at the top of the emitted exec
+/// function's body:
+///
+/// ```ignore
+/// let __exec_spec_self_view: <SelfTy> = self.deep_view();
+/// ```
+///
+/// then runs ever clause expression through this function with
```
zero-to-nat: suspected typo: "ever clause expression"

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -692,19 +726,87 @@ fn compile_enum(item_enum: &ItemEnum) -> Result<TokenStream2, Error> {
...
+                    receiver,
+                    "`&mut self` is not supported in exec_spec impl methods",
+                ));
+            }
+
+            has_receiver = true;
+            let exec_spec = Ident::new(&format!("Exec{}", self_ty_ident), self_ty_ident.span());
+            ctx.add(Ident::new("self", receiver.self_token.span), VarMode::Ref);
+            let span = receiver.span();
+            Some(quote_spanned! { span => self: &#exec_spec })
```
zero-to-nat, line 797: Just curious: is it necessary to rewrite the receiver argument as `self: &Exec...`, or would keeping it as `&self` also work?

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -2226,6 +2381,15 @@ fn compile_expr(
                 let expr = compile_expr(ctx, &expr_unary.expr, VarMode::Ref, unverified)?;
                 quote! { #op #expr }
             }
+            UnOp::Deref(..) => {
```
zero-to-nat, line 2384: Is this currently covered by any of the test cases? If not, would you be able to add some?

source/builtin_macros/src/contrib/exec_spec.rs
```diff
@@ -3037,12 +3289,108 @@ fn compile_spec_fn(item_fn: &ItemFn, unverified: bool) -> Result<TokenStream2, E
     })
 }
 
-/// Compiles a fn/struct/enum item.
+/// Compiles an impl block. Each spec method becomes an exec method on the corresponding
+/// `Exec<T>` type. The original `impl` block is preserved verbatim so spec-mode
```
zero-to-nat: nitpick: what is "spec-mode verification"?

zero-to-nat (approved):
This looks great! Thanks so much for the additions. I just had a couple comments. In addition, would you be able to add some notes about the newly supported features to the documentation [here](https://github.com/verus-lang/verus/blob/main/source/docs/guide/src/exec_spec.md)?

source/builtin_macros/src/contrib/exec_spec.rs
JakeGinesin: I believe `&self` also works. Do you have any preference between the options?

source/builtin_macros/src/contrib/exec_spec.rs
JakeGinesin: added a test case!

source/builtin_macros/src/contrib/exec_spec.rs
JakeGinesin: I was thinking "spec-mode verification" would imply the reasoning about specs in proof functions, but on second look this isn't too clear. 

I think this sentence can be changed to: "The original `impl` block is preserved verbatim so verification still sees the original spec methods."

JakeGinesin:
Updated the docs!

PR #2584, opened 2026-06-23
[cargo-verus] feat: check verus and vstd compatibility

mkovaxx:
- Add a new (and temporary) `--check-toolchain` option to check `verus` and `vstd` compatibility.
- Parse the output of `verus --version` to get the version.
- Extend `MetadataIndex` with the ability to collect `vstd` instances in the verified dep tree.
- Add unit tests for `MetadataIndex::collect_vstd_metadata`.
- Extend `test_utils`: more dep types, ability to spoof a local package to appear as `is-vstd`.


source/cargo-verus/src/metadata.rs
```diff
@@ -132,15 +136,93 @@ impl<'a> MetadataIndex<'a> {
...
+            .collect();
+
+        assert!(
+            vstd_metadata.len() <= 1,
+            "The `vstd` versioning scheme prevents multiple instances in a resolve set.",
+            // This is a consequence of the current `vstd` versioning scheme.
+            // In the current scheme, each version matches `0.0.0-*`.
+            // By definition, *all* such versions are semver-compatible.
+            // Cargo disallows different *compatible* versions in a resolve set.
+            // https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility
```
mkovaxx, line 175: @parno @Chris-Hawblitzel The main thing I'd like you to confirm if this property of the versioning scheme is intentional.

source/cargo-verus/src/subcommands.rs
```diff
@@ -156,6 +156,16 @@ pub fn plan_cargo_run(cfg: VerusConfig) -> Result<CargoRunPlan> {
...
 
+    if cfg.options.toolchain_checks {
+        let vstd_metadata = metadata_index.collect_vstd_metadata(packages_to_verify);
+        println!("`vstd` metadata instances:");
+        for metadata in vstd_metadata {
+            println!("version = {:?}", metadata.version.to_string());
+            println!("source = {:?}", metadata.source);
+            println!();
+        }
+    }
```
mkovaxx: @parno @Chris-Hawblitzel This is just a demo of collecting `vstd` metadata. It will be used to implement toolchain version checks once all the CI machinery is in place.

source/cargo-verus/src/metadata.rs
```diff
@@ -132,15 +136,93 @@ impl<'a> MetadataIndex<'a> {
...
+                (git_source, None)
+            };
+            // Trim the query part of the URL.
+            let url = url.split_once('?').map_or(url, |(trimmed_url, _query)| trimmed_url);
+            PackageSource::Git { url: url.to_string(), rev }
+        } else {
+            PackageSource::Unsupported
+        }
     }
 }
```
mkovaxx: Main caveat here is that this code relies on Cargo internals because there's no stable API. The tests in `test_vstd_sources.rs` should be able to detect if changes were to happen.

source/cargo-verus/src/metadata.rs
parno: We're using the hyphenated version to indicate pre-release status (i.e., from the [SemVer docs](ttps://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility): "A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version.")  We're definitely not at the stage of promising any compatibility from one vstd release to another.  In this case, it sounds like it's having the unintended but useful side effect of preventing people from mixing different versions of vstd.

parno (approved):

source/cargo-verus/src/subcommands.rs
mkovaxx: A simple check is now implemented using the `toolchains` module.

bsdinis (approved):
lgtm

Issue #2323, opened 2026-04-13
Crash when passing Future to function with additional bound


dschoepe:
Passing a future into a function that adds a `Send` bound results in an internal compiler error.

```
pub fn consumes_opaque<F>(f: F)
    where
    F: Future<Output = usize> + Send,
{}

async fn ret42() -> (res: usize)
{
    42
}
async fn verified_join_set_client() {
    let f1 = ret42();
    consumes_opaque(f1);
}
```
[playground](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=7eb138d8ac9279478e5399fc33b585c9)

```
thread 'rustc' (26) panicked at rustc_mir_build/src/../../rustc_mir_build_additional_files/verus.rs:156:21:
Internal Verus Error: The thir_body query is running for item DefId(0:7 ~ main[e366]::ret42) which may require erasure, but the VerusErasureCtxt has not been initialized. Please file a github issue for this error and consider using `--no-lifetime` as a temperary measure to work around the issue.
note: [run with `RUST_BACKTRACE=1` environment variable to display a backtrace](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=7eb138d8ac9279478e5399fc33b585c9#)
note: no errors encountered even though delayed bugs were created

note: those delayed bugs will now be shown as internal compiler errors

error: internal compiler error: {OpaqueTypeKey { def_id: DefId(0:11 ~ main[e366]::verified_join_set_client::{opaque#0}), args: [] }: ProvisionalHiddenType { span: /playground/src/main.rs:15:36: 15:36 (#15), ty: ?0t }}
  |
  = note: delayed at /rustc-dev/4a4ef493e3a1488c6e321570238084b38948f6db/compiler/rustc_infer/src/infer/opaque_types/table.rs:127:43 - disabled backtrace


thread 'rustc' (26) panicked at /rustc-dev/4a4ef493e3a1488c6e321570238084b38948f6db/compiler/rustc_errors/src/lib.rs:1845:9:
Box<dyn Any>
```

Removing the `Send` bound or passing `--no-lifetime` prevents the crash.

FeizaiYiHao:
Seems like the problem is similar to opaque types in traits and only happens with auto traits (`Send`, `Sync`, `Unpin`).

[Additional trait bounds works](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=787ef45f428fadefc216f6fb88ce574d)

[...]

PR #2589, opened 2026-06-23
Use proxy async function to allow auto traits on futures returned by async function

FeizaiYiHao:
Currently passing `impl Future` to a function that requires auto traits (e.g., `Send`) will cause the lifetime checks to run early #2323. This PR adds support to generate a proxy function for each async function with all the ghost code erased. 

The auto traits depend on the async function's body. Specifically, if a variable that is `!Send` is held across an `await`, the returned future is `!Send`. The proxy function will mirror this except for the tracked variables. Therefore, in `resolution_inference.rs`, we check to make sure no tracked variables are held across `await`s as a temporary fix. The users can wrap the tracked variables in `Tracked<>` to work around this. 


source/vir/src/resolution_inference.rs
```diff
@@ -1160,6 +1205,23 @@ impl<'a> Builder<'a> {
...
                 } else {
+                    // In async functions, we also record the read places for tracked variables.
+                    if let Some(p) = p.get_place_for_read() {
+                        let var_mode = match &p.local {
+                            LocalName::Named(var) => self.locals.var_modes.get(&var).copied(),
+                            LocalName::Temporary(ast_id, _) => {
+                                self.locals.temporary_modes.get(&ast_id).copied()
+                            }
+                        };
+                        if self.is_async_fn && matches!(var_mode, Some(Mode::Proof)) {
```
tjhance, line 1216: can you put the `if self.is_async_fn` around the outer if function?

source/vir/src/resolution_inference.rs
```diff
@@ -3312,6 +3387,153 @@ impl BasicBlock {
...
+            record(found, &pl.span);
+            Ok(())
+        },
+    )
+    .unwrap();
+    found
+}
+
+/// Find the first access instruction for the given variable, scanning from the entry BB forward.
+fn find_first_reference(
```
tjhance, line 3501: can you use the dataflow functionality already in this file instead of reimplementing reachability? It should be straightforward to implement the DataflowState trait to compute variable liveness.

source/rustc_mir_build/src/thir/cx/expr.rs
```diff
@@ -59,6 +59,8 @@ impl<'tcx> ThirBuildCx<'tcx> {
 
         let mut expr = self.make_mirror_unadjusted(hir_expr);
 
+        // println!("mirrored expr: {:#?}", expr);
```
tjhance, line 62: I presume this was left in accidentally

source/vir/src/resolution_inference.rs
```diff
@@ -269,11 +270,26 @@ pub(crate) fn infer_resolution(
...
+        &var_modes,
+        temporary_modes,
+        is_async_fn,
+    )?;
     //println!("{:}", pretty_cfg(&cfg));
-    let resolutions = if dual_mode { vec![] } else { get_resolutions(&cfg, rtypes, asserts) };
+    let (resolutions, initialization_analyses) =
+        if dual_mode { (vec![], vec![]) } else { get_resolutions(&cfg, rtypes, asserts) };
+
+    if is_async_fn && !no_lifetime {
```
tjhance, line 290: Is there a reason to guard this by no_lifetime?

source/vir/src/resolution_inference.rs
```diff
@@ -415,9 +445,13 @@ enum InstructionKind {
     /// or partially initialized. Whatever was there before gets dropped.
     /// The place is subsequently uninitialized.
     DropFrom(FlattenedPlace),
+    /// Read to a tracked variable.
+    /// This is only used when the function is an async function
+    Read(FlattenedPlace),
```
tjhance, line 450: call it ReadTracked if it's only for tracked stuff

source/vir/src/resolution_inference.rs
```diff
@@ -1160,6 +1205,23 @@ impl<'a> Builder<'a> {
...
                     Ok(bb)
                 } else {
+                    // In async functions, we also record the read places for tracked variables.
+                    if let Some(p) = p.get_place_for_read() {
+                        let var_mode = match &p.local {
+                            LocalName::Named(var) => self.locals.var_modes.get(&var).copied(),
+                            LocalName::Temporary(ast_id, _) => {
+                                self.locals.temporary_modes.get(&ast_id).copied()
+                            }
+                        };
```
tjhance, line 1215: var_mode should never be None, I think, so it would be best to throw an internal error in that case

source/vir/src/resolution_inference.rs
```diff
@@ -3312,6 +3387,153 @@ impl BasicBlock {
     }
 }
 
+/// Check that no tracked-mode (Mode::Proof) variable is alive across an await point.
+fn check_tracked_vars_alive_across_await(
+    cfg: &CFG,
+    body: &Expr,
+    initialization_analyses: &[DataflowOutput<InitializationPossibilities>],
+) -> Result<(), VirErr> {
+    println!("cfg.await_bbs {:#?}", cfg.await_bbs);
```
tjhance, line 3396: left in by accident?

tjhance (changes requested):
This mostly seems good!

My main comment on the implementation is that resolution_inference already contains a framework for this kind of reachability analysis, which should be re-used. In particular, you should be able to use a reverse dataflow analysis by implementing the `DataflowState` trait and calling `do_dataflow`. For the error message, you can compute the spans however is convenient, but for the main check I would prefer to reuse the existing dataflow analysis.

Also, note that not all the InstructionKinds imply liveness. `Drop` or `Overwrite` makes the variable non-live, while `Read`, `Move`, or `Mutate` make it live.

source/vir/src/ast_util.rs
```diff
@@ -27,6 +31,12 @@ impl PathX {
         Arc::new(PathX { krate: self.krate.clone(), segments: Arc::new(segments) })
     }
 
+    pub fn replace_first(&self, ident: Ident) -> Path {
```
tjhance, line 34: Where is this used?

matthai:
I'm working on a verified service where a database transaction is represented as a linear ghost
capability (minted by `BEGIN IMMEDIATE`, borrowed by each query, consumed by `COMMIT`) so
that a query outside a transaction fails to typecheck. Since every database call is an
`.await`, this depends on tracked values and async interacting well. Some measurements that
may be useful here, all on 0.2026.07.20.83719d5.

**Tracked values cross `.await` fine.** Both as locals and as parameters, used after
resumption. No barrier there.

**What is rejected is a raw tracked value in the return position of a default-exec async
function:**

```rust
async fn step(tracked t: Tx) -> (tracked out: Tx)   // error: cannot access
    ensures out.r.loc() == t.r.loc()                //   proof-mode place in
{ t }                                               //   executable context
```

**But the same diagnostic appears with no async involved**: a non-async function returning a
raw tracked value is rejected identically. That makes me unsure whether what I hit is an async
gap at all, or just the general executable/proof return policy showing up in an async
signature.

**The `Tracked<T>` form works**, which is what I have used:

```rust
async fn begin(db: u64) -> (ret: Tracked<Tx>)
    ensures ret@.value() == ExclusiveRA::Exclusive(db)
```

This resulted in 8 verified, 0 errors, and the negative control behaves as expected: handing transaction A's token to a query requiring B fails the precondition, so the identity discipline survives the wrapper.

So I guess I have the following narrow (I think!) question:

**Is a raw tracked return rejected by the general executable/proof return policy, does the
async signature path need an output-mode or erasure rule that is currently missing, or is
there a separate soundness invariant here related to cancellation?**

On that last one, for what it is worth: I could not find a cancellation barrier by
experiment. Verus accepts discarding a tracked token, and accepts dropping an unpolled future
that owns one; only a second move is rejected. Which reads as affine-and-therefore-leaky
rather than unsound, and matches what `fdec29dd` seems to assume.

Happy to contribute the probes as test cases if any of them are useful, including the
non-async control, since it is the one that distinguishes the three possibilities.

parno:
> Is a raw tracked return rejected by the general executable/proof return policy?

Yes, that's unrelated to async.  You can see more in the [mode documentation](https://verus-lang.github.io/verus/guide/reference-var-modes.html), e.g., the cheatsheet section on "Exec function, return a tracked and ghost value".

PR #2591, opened 2026-06-24
add loop_isolation_boundary attribute

tjhance:
the intent is that this will be added to the desugaring of for-loops, as discussed today

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -2146,6 +2146,17 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
                 let block = block_to_vir(bctx, body, &expr.span, &expr_typ()?)?;
                 if crate::attributes::is_proof_in_spec(bctx.ctxt.tcx.hir_attrs(expr.hir_id)) {
                     mk_expr(ExprX::ProofInSpec(block))
+                } else if crate::attributes::is_loop_isolation_boundary(
+                    bctx.ctxt.tcx.hir_attrs(expr.hir_id),
+                ) {
+                    let wrap = loop_isolation_boundary_check(expr.span, &block)?;
+                    let block = if wrap {
+                        let x = ExprX::Unary(UnaryOp::LoopIsolationBoundary, block);
+                        bctx.spanned_typed_new(expr.span, &expr_typ()?, x)
+                    } else {
+                        block
+                    };
+                    Ok(ExprOrPlace::Expr(block))
                 } else {
                     Ok(ExprOrPlace::Expr(block))
                 }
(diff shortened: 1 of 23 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -4336,3 +4347,13 @@ pub(crate) fn simplify_place_by_cancelling(place: &Place) -> Place {
         }
     }
 }
+
+fn loop_isolation_boundary_check(span: Span, block: &vir::ast::Expr) -> Result<bool, VirErr> {
+    let ExprX::Block(_stmts, Some(e)) = &block.x else {
+        crate::internal_err!(span, "loop_isolation_boundary expected Block")
```
Chris-Hawblitzel: From this, it looks like loop_isolation_boundary is an internal feature, not something users are expected to write directly.  If so, could we use `#[verus::internal(loop_isolation_boundary)]` instead of `#[verifier::loop_isolation_boundary]`?

source/vir/src/sst.rs
```diff
@@ -257,6 +257,15 @@ pub enum StmX {
     Air(Arc<String>),
     /// Sequential composition of statements
     Block(Stms),
+    /// Wraps a Loop with statements whose effects should be visible from the inner query
+    LoopIsolationBoundary {
+        pre_stms: Stms,
+        /// This must be a Loop with loop_isolation=true
```
Chris-Hawblitzel: Another way to do this would be to not have `StmX::LoopIsolationBoundary` as a case separate from `StmX::Loop`, but instead change `loop_isolation` from a `bool` to an `Option<LoopIsolation>`, with `LoopIsolation` holding `pre_stms`, or change `loop_isolation` to a `LoopIsolation` enum holding `pre_stms` in the isolated case.

source/vir/src/sst.rs
```diff
@@ -257,6 +257,15 @@ pub enum StmX {
...
+        loop_stm: Stm,
+        /// Same as Loop's `pre_modified_params` but only vars modified
+        /// before this LoopIsolationBoundary statement evaluates.
+        pre_modified_params: Option<Arc<crate::sst_vars::HavocSet>>,
```
Chris-Hawblitzel: Is it true that if we use this `pre_modified_params`, then the `pre_modified_params` in `StmX::Loop` does not get used?  If so, this would be an argument in favor of either declaring `pre_modified_params` in only one place, or of merging `StmX::LoopIsolationBoundary` into `StmX::Loop` as mentioned in the comment above (and perhaps putting `pre_modified_params` in each of the two cases of the `LoopIsolation` enum if we really wanted to keep them distinct for some reason).

Chris-Hawblitzel (approved):

tjhance:
Converting to draft as:

 * I want to confirm that this really helps with IterMut before moving forward
 * Lots of "jank" needs fixing

source/vir/src/sst.rs
tjhance: I combined the LoopIsolationBoundary into the Loop node, renamed the two `pre_modified_params` fields to be clearer about the difference & documented which one is used when. However, I left `loop_isolation` as a boolean.

PR #2600, opened 2026-06-26
Add mut ref + `spec_from_from_spec` axioms

elanortang:
This PR adds/updates several trusted axioms which return `&mut`. It also adds an axiom relating the control flow `spec_from` to `From::from_spec`. 

---

`raw_vec.rs`:
- `into_seq_pt_shared` converts an `&PointsToUnaligned<[T]>` to an `&SeqPointsTo<T>`.
- `into_seq_pt_mut` converts an `&mut PointsToUnaligned<[T]>` to an `&mut SeqPointsTo<T>`.
- `seq_into_slice_mut` converts an `&mut SeqPointsTo<T>` to an `&mut PointsToUnaligned<[T]>`.
- Also adds verified proof fns to borrow the underlying `Seq<PointsTo<T>>` from a `SeqPointsTo<T>`.
- (Ignore lines 1811-1939 - they're temporary until the finalized `SeqPointsTo` transmute axioms land).

`seq.rs`:
- `tracked_mut_borrow` returns a tracked `&mut` to a single sequence element.
- Also adds verified proof fns to split and rejoin tracked `Seq<T>`.

`option.rs`:
- Add `tracked_mut_borrow` to the `OptionAdditionalFns` trait to returns a tracked `&mut T` to the underlying value, provided that `Option<T>` is Some. 
- Currently implemented as an axiom, since Verus only supports mut ref binding in a pattern in exec code.

---

`control_flow.rs`: 
- Add broadcast axiom `spec_from_from_spec` connecting `control_flow::spec_from` and `FromSpec::from_spec`, so that Verus can automatically treat expressions like
```
let required_cap = len.checked_add(additional).ok_or(CapacityOverflow)?;
```
like this
```
let required_cap = match len.checked_add(additional).ok_or(CapacityOverflow) {
    Ok(res) => res,
    Err(err) => return Err(From::from(err)),
};
```


source/vstd/std_specs/option.rs
```diff
@@ -49,6 +49,16 @@ pub trait OptionAdditionalFns<T>: Sized {
             t == self->0,
     ;
 
+    #[allow(deprecated)]
```
parno, line 52: Why is this attribute needed?

source/vstd/std_specs/option.rs
```diff
@@ -107,6 +117,15 @@ impl<T> OptionAdditionalFns<T> for Option<T> {
         }
     }
 
+    axiom fn tracked_mut_borrow(tracked &mut self) -> (tracked t: &mut T);
```
parno, line 122: This seems redundant with tracked_mut_borrow?

source/vstd/std_specs/option.rs
```diff
@@ -107,6 +117,15 @@ impl<T> OptionAdditionalFns<T> for Option<T> {
...
+
+    // Proof version fails because a mut ref binding in a pattern is only allowed in exec mode
```
parno: Cut the commented code?

source/vstd/raw_ptr.rs
```diff
@@ -1658,6 +1808,135 @@ impl<T> SeqPointsTo<T> {
...
+    /// (1) The pointer's address is aligned to `T`.
+    ///
+    /// (2) The length is exactly `capacity * layout::size_of::<T>()`.
+    ///
+    /// (3) It is possible to decode the memory contents as a `Seq<MemContents<T>`.
+    pub axiom fn cast_to_type<T>(tracked self, capacity: usize) -> (tracked out: SeqPointsTo<T>)
+        requires
+            self.ptr()@.addr as nat % align_of::<T>() == 0,
+            self.len() == capacity * layout::size_of::<T>(),
+            // self.is_fully_uninit() || exists|s: Seq<MemContents<T>>| self.mem_contents() == encode(s),
```
parno, line 1867: Should this be removed?

parno (approved):

source/vstd/std_specs/option.rs
elanortang: Do you mean the `proof fn` on line 53? That is the trait declaration, and this is the trait implementation.

source/vstd/std_specs/option.rs
elanortang: Copying `tracked_borrow`, this spec uses the deprecated function `is_Some`. Is there a different function I should use?

source/vstd/raw_ptr.rs
elanortang: That part is only there temporarily - Natalie put together the final version which is a separate PR.

source/vstd/std_specs/option.rs
elanortang: Removed the commented-out code and added a note saying that we can make this a proof fn if Verus supports mut ref bindings in a pattern in proof mode.

source/vstd/std_specs/option.rs
parno: You should be able to use `x is Some` instead, unless that doesn't work in rustlib for some reason.

source/vstd/std_specs/option.rs
parno: Oh, I see.  I didn't realize that when looking at the diff.

source/vstd/std_specs/option.rs
elanortang: It looks like it doesn't work (and not because of rustlib) - as a function defined on the trait `OptionAdditionalFns<T>: Sized`, Verus has no way of knowing at this point that `self` is an `Option` datatype. I'm guessing this is why the other trait functions still use the deprecated attribute.

parno: Ah, I see.  That makes sense.

PR #2607, opened 2026-06-28
Add AbstractBytes to PointsTo model and PointsTo transmute axioms

zero-to-nat:
Add `uninterp spec fn abstract_bytes(): Seq<AbstractByte>` to `PointsTo<T: ?Sized>` and assume that any initialized `MemContents<T>` can be decoded from the abstract bytes. Modify transmute axioms to utilize the abstract bytes, and add axioms which allow for a roundtrip `PointsTo<T> -> PointsTo<[u8]> -> PointsTo<T>` which preserves the initialized `MemContents<T>` from the original permission.

This also builds up transmute infrastructure for `SeqPointsTo`, but none of that is trusted.

AbstractBytes:
- `abs_decode_mem_contents, abs_decode_mem_contents_unsized`: defines `abs_decode::<MemContents<T>>`. Initialized mem contents must have the in-memory `T` value satisfy `abs_decode::<T>`; uninitialized mem contents correspond to any abtract bytes.
- `abstract_bytes_decode` for `PointsTo<T>`, `PointsTo<[T]>`, `PointsTo<str>`. Assumes tha mem contents can always be decoded from the abstract bytes for the permission.
- `PointsTo<T>:: cast_to_untyped`. There is a subtle soundness argument regarding the `TypedMemory<T>` output by this function.
- `PointsTo<[u8]>::{cast_to_typed, cast_to_typed_uninit}`. There is a subtle soundness argument regarding the `TypedMemory<T>` required by `cast_to_typed`.
- `cast_to_X_shared_inner` for `PointsTo<[u8]>`, `PointsTo<str>`. These are updated versions of the previous `cast_to_X_shared` axioms (which are no longer axioms) that use the abstract bytes.

Other changes:
- Conversion axioms between different types of PointsTo's now preserve/update the abstract bytes, analogously to the mem contents.
- Add `provenance_non_null` in `PointsToUnaligned<T>` and `PointsToUnaligned<[T]>` to say that if the type is not a ZST, `Provenance` is not null.
- Add `PointsToRaw` axioms to ensure sound cast from `PointsToRaw` to `PointsTo<T>`.
  - `ptr_bounds`
  - `provenance_non_null`
  - `is_disjoint`

This PR also contains a fix to the `AbstractByte` encoding for `[u8]`. Previously, this encoding did not take endianness into account.

NOTE: More changes to come - this is just a partial update.


source/vstd/raw_ptr.rs
```diff
@@ -1786,6 +1834,7 @@ impl<T> SeqPointsTo<T> {
             );
             let tracked mut mut_spt = self;
             mut_spt.perm.tracked_push(zs_pt);
+            Self::abstract_bytes_len_helper(mut_spt.seq_perm());
```
elanortang, line 2184: I think the proof goes through without this?

elanortang (commented):
generally looks good so far (pending updates)

PR #2610, opened 2026-06-29
vstd: add no_unwind to Result/Option methods that don't unwind

bsdinis:


source/vstd/std_specs/option.rs
```diff
@@ -252,6 +257,7 @@ pub assume_specification<T: core::default::Default>[ Option::<T>::unwrap_or_defa
     ensures
         option.is_some() ==> res == option.unwrap(),
         option.is_none() ==> T::default.ensures((), res),
+    no_unwind
```
tjhance: this one is wrong because the `T::default` function might unwind

tjhance (changes requested):

source/vstd/std_specs/option.rs
bsdinis: right, i removed it

tjhance (approved):

PR #2578, opened 2026-06-18
Upgrade to Z3 4.16.0

parno:
Z3 4.12.5 has some known memory-corruption bugs that [have affected Verus users](https://github.com/verus-lang/verus/issues/1986).  The Z3 4.12.5 release for Linux on Arm (useful for running containers on Macs) also bundles the wrong executable, which breaks our `get-z3.sh` installation path, requiring users on that platform to build Z3 from source.  Various soundness bugs have been repaired in newer versions of Z3, although we don't know of any that affected Verus queries.

Given the above, in this PR, I made the mechanical changes necessary to migrate to the latest release of Z3, 4.16.0.  They were relatively minor, but they (as expected) broke various proofs in vstd and in our tests, which this PR also repairs.  In the verita projects, several projects had zero proofs break.  The rest had 1 (anvil, svsm), 3 (capybaraKV), 4 (page-table), 5 (memory-allocator), and 7 (human-eval).  I put together branches/forks of each with repaired proofs (see the updates to the verita `run_configuration_all.toml`).

To evaluate the performance impact of the Z3 change, I ran three iterations of `verita` with 4.12.5 and three with 4.16.0 on my desktop that I was otherwise not using.  This is not a fully rigorous evaluation, but as you can see below, the rlimit and time limit variations within a Z3 version are small, so I think we can take them as representative.  One minor quirk is that one page-table proof failed in the 4.12.5 runs; I'm not sure why, given that it's passing the nightly verita tests.

As you can see, moving to 4.16.0 modestly improved performance on a few projects (anvil, vest), mildly hurt performance on a few (node-replication, vstd), modestly hurt performance on a few more (ironkv, svsm, verified-storage), and hurt performance on human-eval (+407%), memory-allocator (+116%), page-table (+228%).

## Comparison between two 4.12.5 runs:

| Project | Crate Root | Status | Verified | Errors | Resources | Total Time |
|---------|------------|--------|----------|--------|-----------|------------|
| anvil |  | OK | 140 → 140 (+0) | 0 → 0 (+0) | 828.8M → 828.8M (+0, +0.0%) | 166.14 → 164.04 s (-2.11 s, -1.3%) |
| human-eval |  | OK | 404 → 404 (+0) | 0 → 0 (+0) | 74.9M → 74.9M (+0, +0.0%) | 4.56 → 4.56 s (+0.00 s, +0.1%) |
[...]

PR #2618, opened 2026-07-01
ci: upgrade runners to Ubuntu 24.04

mmcloughlin:
Upgrade CI and support policy to Ubuntu 24.04.

The immediate motivation is the Z3 4.16.0 upgrade (#2578), since its release binaries require a glibc version unsupported on Ubuntu 22.04. Since Ubuntu 22.04 reaches end-of-life in April next year, this upgrade is required soon anyway.

---

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,12 +1,12 @@
 # Support
 
 Verus provides first-tier support to relatively recent OS distributions (MacOS, Windows and Ubuntu).
-The general rule of thumb is that we lag Github's `<os>-latest` by one version.
+The general rule of thumb is that we lag one version behind the latest offered on Github.
 These are the architectures that have release artifacts prebuilt:
 - MacOS 14 (arm)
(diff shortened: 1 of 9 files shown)


INSTALL.md
```diff
@@ -1,12 +1,12 @@
 # Support
 
 Verus provides first-tier support to relatively recent OS distributions (MacOS, Windows and Ubuntu).
-The general rule of thumb is that we lag Github's `<os>-latest` by one version.
+The general rule of thumb is that we lag one version behind the latest offered on Github.
```
mmcloughlin, line 4: Slight policy update here. Currently, `ubuntu-latest` is 24.04 but Github does offer 26.04.

https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

parno (approved):
Looks good to me.  Thanks for setting this up!

PR #2619, opened 2026-07-01
docs: add examples and documentation in the guide about HOCAP style logatom

bsdinis:

### DIFF
--- a/source/docs/guide/src/SUMMARY.md
+++ b/source/docs/guide/src/SUMMARY.md
@@ -102,6 +102,7 @@
   - [Atomic specifications](./logatom-spec.md)
   - [Invoking the linearization point](./logatom-open.md)
   - [Calling a logically atomic function](./logatom-call.md)
+  - [HOCAP style linearization](./logatom-hocap.md)
 
 - [Verifying a container library: Binary Search Tree](./container_bst.md)
   - [First draft](./container_bst_first_draft.md)
--- a/source/docs/guide/src/logatom-hocap.md
+++ b/source/docs/guide/src/logatom-hocap.md
@@ -1,0 +1,153 @@
+# Logical Atomicity in the HOCAP style
+
+In Verus, there is another way to prove [linearizability](https://doi.org/10.1145/78969.78972), based on the [HOCAP](https://kasv.dk/articles/hocap.pdf)
+(Higher-Order Concurrent Abstract Predicates) program logic.
+HOCAP promotes a proof design pattern, rather than a syntactic language feature. This makes it harder
+(but still possible) to [open invariants around logically atomic code](https://research.ralfj.de/iris/talk-iris2019.pdf).
+
+To use this pattern, one can use the [`logatom` traits](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/index.html).
+The main concept is that there are `Linearizer`s and `Operations` (both mutable and read-only versions).
+A `Linearizer` is a ghost callback that has pre/post-conditions.
+It must be called to linearize an `Operation` at a valid [linearization point](https://dl.acm.org/doi/pdf/10.1145/78969.78972),
+where it is consumed and a `Completion` is obtained.
+
+In our examples, we will focus on [`MutLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.MutLinearizer.html),
+but the [`ReadLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.ReadLinearizer.html) is very similar.
+
+## How the logical atomicity proof works
+
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
+This resource captures the concurrent object being talked about.
+The logical atomiciy proof is achieved by calling the `apply` function in succession,
+which requires showing the atomic pre-/post-conditions in the `apply` function itself.
+
+We are going to give intuition for how to use HOCAP, showing different
+elements that are required for its successful application. As a running example,
+we will show how to verify the logical atomicity of a monotonic counter example.
+For the full version, refer to the examples directory on the repository.
+
+## Constructing the Operations
+
+Logical atomicity and linearizability deal in operations being linearized.
+The `ReadOperation` and `MutOperation` model these operations that are linearized.
+
+In the counter example, we can specify an increment operation, `IncOp`:
+
+```rust
+/// Operation for incrementing the counter
+struct IncOp {
+    /// Location of the GhostVarAuth
+    id: Loc,
+}
+
+impl logatom::MutOperation for IncOp {
+    type Resource = GhostVarAuth<u64>;
+
+    type ExecResult = u64;
+
+    type NewState = ();
+
+    closed spec fn requires(
+        self,
+        pre: Self::Resource,
+        new_state: Self::NewState,
+        e: Self::ExecResult,
+    ) -> bool {
+        &&& pre.id() == self.id
+        &&& pre@ == e
+    }
+
+    closed spec fn ensures(
+        self,
+        pre: Self::Resource,
+        post: Self::Resource,
+        new_state: Self::NewState,
+    ) -> bool {
+        &&& post.id() == self.id
+        &&& post@ == pre@.wrapping_add(1)
+    }
+}
+```
+
+Here, we force the resource being passed in to refer to the same ghost location
+as the counter's resource (by having the location in the operation itself). Moreover,
+we model the expected atomic behaviour of the increment (the counter maches the resource
+state and is incremented).
+
+## Constructing a Linearizer
+
+The linearizer typically holds the resource itself. The apply function can either
+express agreement (for read only operations) or update the resource (for mutable operations).
+
+We can express additional pre and post conditions on the linearizer. Centrally, in the `apply`
+function we must be able to show that from both the operation and the linearizer preconditions
+we can derive both post conditions:
+
+```rust
+struct IncPerm {
+    pub tracked var: GhostVar<u64>,
+}
+
+impl logatom::MutLinearizer<IncOp> for IncPerm {
+    type Completion = GhostVar<u64>;
+
+    closed spec fn namespaces(self) -> ISet<int> { ISet::empty() }
+
+    closed spec fn pre(self, op: IncOp) -> bool {
+        op.id == self.var.id()
+    }
+
+    closed spec fn post(self, op: IncOp, exec_res: u64, completion: Self::Completion) -> bool {
+        &&& op.id == self.var.id()
+        &&& op.id == completion.id()
+        &&& exec_res.wrapping_add(1) == completion@
+    }
+
+    proof fn apply(
+        tracked self,
+        op: IncOp,
+        tracked resource: &mut GhostVarAuth<u64>,
+        new_state: (),
+        exec_res: &u64,
+    ) -> (tracked result: Self::Completion) {
+        let tracked mut var = self.var;
+        resource.update(&mut var, exec_res.wrapping_add(1));
+        var
+    }
+
+    proof fn peek(tracked &self, op: IncOp, tracked resource: &GhostVarAuth<u64>) {}
+}
+
+```
+
+## Calling `apply`
+
+Consider the example of a monotonic counter (as seen in the examples dir).
+The ghost state is modelled after a [`GhostVarAuth`](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html) held in an [`AtomicInvariant`](https://verus-lang.github.io/verus/verusdoc/vstd/invariant/struct.AtomicInvariant.html),
+and a [`GhostVar`](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVar.html) held by the `MutLinearizer`.
+
+```rust
+open_atomic_invariant!(inv => v => {
+    let tracked CounterInvariant { perm, auth } = v; // destructure
+
+    // curr is previously loaded value,
+    // next is the value we are incrementing to,
+    // and res has a result indicating whether the CAS succeeded
+    res = self.atomic.compare_exchange_weak(Tracked(&mut perm), curr, next);
+
+    proof {
+        if res is Ok {
+            let op = IncOp { id: self.auth_id() };
+            compl = Some(lin.apply(op, &mut auth, (), &curr));
+
+            v = CounterInvariant { perm, auth };
+        } else {
+            v = CounterInvariant { perm, auth };
+        }
+    }
+});
+```
+In the CAS loop to increment the counter, we open the invariant, call `compare_exchange_weak` and,
+on success, use the `GhostVarAuth` to update the ghost state within `apply`.
+
--- a/examples/logatom.rs
+++ b/examples/logatom.rs
@@ -1,0 +1,387 @@
+//! Example usage of vstd::logatom
+//!
+//! We develop a linearizable counter, modelled with a GhostVarAuth resource
+use vstd::atomic::*;
+use vstd::invariant::*;
+use vstd::logatom;
+use vstd::prelude::*;
+use vstd::resource::ghost_var::*;
+use vstd::resource::*;
+
+verus! {
+
+/// The owned version of the counter
+struct OwnedCounter {
+    atomic: PAtomicU64,
+    inv: Tracked<AtomicInvariant<CounterPredicate, CounterInvariant, CounterPredicate>>,
+    counter_id: Ghost<Loc>,
+}
+
+impl OwnedCounter {
+    #[verifier::type_invariant]
+    spec fn inv(self) -> bool {
+        &&& self.inv@.constant().perm_id == self.atomic.id()
+        &&& self.inv@.constant().auth_id == self.counter_id@
+    }
+
+    /// id of the authoritative resource
+    spec fn auth_id(self) -> Loc {
+        self.counter_id@
+    }
+
+    /// id of the atomic invariant
+    spec fn inv_id(self) -> int {
+        self.inv@.namespace()
+    }
+
+    /// Creating a counter with a particular initial value
+    /// We get out of it the counter and a half-permission to change the counter
+    fn new(initial_value: u64) -> (r: (Self, Tracked<GhostVar<u64>>))
+        ensures
+            r.0.auth_id() == r.1@.id(),
+            r.1@@ == initial_value,
+    {
+        let (atomic, Tracked(perm)) = PAtomicU64::new(initial_value);
+
+        let tracked var;
+        let tracked inv;
+        proof {
+            let tracked (auth, v) = GhostVarAuth::new(initial_value);
+            var = v;
+            let ghost counter_pred = CounterPredicate { perm_id: perm.id(), auth_id: auth.id() };
+            let tracked counter_inv = CounterInvariant { perm, auth };
+            inv = AtomicInvariant::new(counter_pred, counter_inv, 0);
+        }
+
+        let counter = OwnedCounter { atomic, inv: Tracked(inv), counter_id: Ghost(var.id()) };
+
+        (counter, Tracked(var))
+    }
+
+    /// We can get a reference view over the counter
+    fn as_ref<'a>(&'a self) -> (r: Counter<'a>)
+        ensures
+            r.auth_id() == self.auth_id(),
+            r.inv_id() == self.inv_id(),
+    {
+        proof {
+            use_type_invariant(self);
+        }
+        Counter { atomic: &self.atomic, inv: &self.inv, counter_id: self.counter_id }
+    }
+}
+
+/// The counter that threads will interact with
+struct Counter<'a> {
+    atomic: &'a PAtomicU64,
+    inv: &'a Tracked<AtomicInvariant<CounterPredicate, CounterInvariant, CounterPredicate>>,
+    counter_id: Ghost<Loc>,
+}
+
+/// The invariant that is maintained
+tracked struct CounterInvariant {
+    perm: PermissionU64,
+    auth: GhostVarAuth<u64>,
+}
+
+/// The predicate over the invariant (i.e., the constant part of the invariant)
+struct CounterPredicate {
+    perm_id: int,
+    auth_id: Loc,
+}
+
+impl InvariantPredicate<CounterPredicate, CounterInvariant> for CounterPredicate {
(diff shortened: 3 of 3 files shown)


source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
+# Logical Atomicity in the HOCAP style
+
+In Verus, there is another way to prove linearization, based on the [HOCAP](https://kasv.dk/articles/hocap.pdf).
+HOCAP is a proof discipline, rather than a language feature and is [strictly weaker than approaches like atomic update](https://research.ralfj.de/iris/talk-iris2019.pdf),
```
ahuoguo: I wouldn't say HOCAP is strictly weaker. It's rather a proof pattern that can't have a way to talk about the invariant rule in a syntactic way (you can still open invariants as if your code is atomic).

bsdinis: Good point, lmk if this now looks good

ahuoguo:
I think Travis might have mentioned this at the retreat, but maybe making this a module-level rustdoc for `vstd::logatom` makes more sense, and leaving a link to the rustdoc in the logical atomicity chapter. I think my process of going through this guide requires me to jump to the `vstd::logatom` code frequently.

tjhance:
We should definitely have verusdoc style docs for logatom in vstd, but we also desperately need more "guide" material on the subject of concurrency verification, which should include an overview of logical atomicity approaches, among other topics. So I think this material, as written, is appropriate for the guide. It's probably fine as-is; we can organize it more later once we have more material.

source/docs/guide/src/SUMMARY.md
```diff
@@ -96,6 +96,12 @@
   - [Pointers](./pointers.md)
   - [Concurrency](concurrency.md)
 
+- [Logical atomicity](./logatom.md)
+  - [Atomic specifications](./logatom-spec.md)
```
slerpyyy, line 102: You can remove the file paths from the links here, e.g. `- [Atomic specifications]()`, to make the sections appear grayed out in the index

slerpyyy: At least if you want this to be merged before #2326, I think that's the right move. If you want this to be merged after, we should probably join forces on this.

bsdinis: Aye, I was trying to prevent you having many merge conflicts but I'll take your suggestion. Indeed I think this should go first the atomic PR

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
+# Logical Atomicity in the HOCAP style
+
+In Verus, there is another way to prove linearization, based on the [HOCAP](https://kasv.dk/articles/hocap.pdf).
```
parno: Can we spell out HOCAP at first use, so readers have more context?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+HOCAP is a proof pattern, rather than a syntactic language feature. This makes it harder
+(but still possible) to [open invariants around logical atomic code](https://research.ralfj.de/iris/talk-iris2019.pdf).
```
parno: "logic" -> "logically"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+
+To use this, one needs to use the [`logatom` traits](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/index.html).
```
parno: "this" -> "this pattern"?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+The main concept is that there are `Linearizer`s and `Operations` (both mutable and read-only versions).
+A `Linearizer` is a ghost callback that has pre- and post- conditions.
+It must be called to linearize an `Operation` at a valid [linearization point](https://dl.acm.org/doi/pdf/10.1145/78969.78972),
+where it is consumed and a `Completion` is obtained.
+
+In our examples, we will focus on [`MutLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.MutLinearizer.html),
+but the [`ReadLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.ReadLinearizer.html) is very similar.
+
+## How the logically atomicity proof works
```
parno: I think this should be "logically atomic" or "logical atomicity"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
+This resource captures the concurrent object being talked about.
+The linearization proof comes from the fact that we can encode the pre and post conditions of the operations
+(which can be thought about as the logically atomic triple)
```
parno: "about" -> "of"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+to express that the ordering that the object is going through (expressed via the resource) makes sense.
```
parno: This clause is hard to follow.  Can we say something more precise/clear w.r.t. "the ordering that the object is going through"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Calling `apply`
+
+Consider the example of a monotonic counter (as seen in the examples dir).
+The ghost state is modelled after a [`GhostVarAuth`](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html) held in an [`AtomicInvariant`](https://verus-lang.github.io/verus/verusdoc/vstd/invariant/struct.AtomicInvariant.html),
+and a [`GhostVar`](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVar.html) which the `MutLinearizer` holds.
```
parno: "which the `MutLinearizer` holds" -> "held by the `MutLinearizer`"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+
+```rust
+open_atomic_invariant!(inv => v => {
+    let tracked CounterInvariant { perm, auth } = v; // destructure
+
+    // curr is the loaded value
```
parno: Can we make the comment a bit clearer and cover `next` and `res`?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
+This resource captures the concurrent object being talked about.
+The linearization proof comes from the fact that we can encode the pre and post conditions of the operations
+(which can be thought about as the logically atomic triple)
+to express that the ordering that the object is going through (expressed via the resource) makes sense.
+
+## Calling `apply`
+
+Consider the example of a monotonic counter (as seen in the examples dir).
```
parno: Ideally, the Guide examples should be readable in the guide, without having to refer to the Verus files elsewhere in the repo.  Can you use some anchor tags to include relevant snippets from the file here instead?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            v = CounterInvariant { perm, auth };
+        }
+    }
+});
+```
+In the CAS loop to increment the counter, we open the invariant, call `compare_exchange_weak` and,
+on success, use the `GhostVarAuth` to update the ghost state within `apply`.
+
+## Constructing the Operations
```
parno, line 30: Should this come before the section above, since the code above is constructing an IncOp that wasn't  previously introduced?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+    ) -> bool {
+        &&& post.id() == self.id
+        &&& post@ == pre@.wrapping_add(1)
+    }
+}
+```
+
+Here, we force the resource being passed in to refer to the same ghost location
+as the counter's resource (by having the location in the operation itself) and
+models the logically atomic pre and post conditions.
```
parno: I think we're missing a word or phrase before "models"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Constructing a Linearizer
+
+The linearizer tipically holds the resource itself. The apply function can either
```
parno: "tipically" -> "typically"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
+# Logical Atomicity in the HOCAP style
+
+In Verus, there is another way to prove linearization, based on the [HOCAP](https://kasv.dk/articles/hocap.pdf).
+HOCAP is a proof pattern, rather than a syntactic language feature. This makes it harder
+(but still possible) to [open invariants around logical atomic code](https://research.ralfj.de/iris/talk-iris2019.pdf).
+
+To use this, one needs to use the [`logatom` traits](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/index.html).
+The main concept is that there are `Linearizer`s and `Operations` (both mutable and read-only versions).
+A `Linearizer` is a ghost callback that has pre- and post- conditions.
```
parno: Looking at the rest of the guide, I think we generally use either "pre-/post-conditions" or "pre-  and post-conditions"

parno (commented):
Thanks for working on this!  I realize it's still work in progress, but I was curious to take a look.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,144 @@
+# Logical Atomicity in the HOCAP style
+
+In Verus, there is another way to prove linearization, based on the [HOCAP](https://kasv.dk/articles/hocap.pdf).
+HOCAP is a proof pattern, rather than a syntactic language feature. This makes it harder
```
slerpyyy: I think "HOCAP is a proof pattern" is a bit misleading; HOCAP is the program logic that popularized this design pattern for representing logically atomic functions. As far as I'm aware, the first paper to describe this pattern was [Jacobs & Piessens POPL'11](https://dl.acm.org/doi/abs/10.1145/1926385.1926417) for the VeriFast verification tool. HOCAP (released at [ESOP’13](https://link.springer.com/chapter/10.1007/978-3-642-37036-6_11)) cites that paper in their related work section.

source/docs/guide/src/logatom_hocap.md
```diff
@@ -0,0 +1 @@
+# HOCAP style linearization
```
slerpyyy: This file can go

source/docs/guide/src/logatom-hocap.md
slerpyyy: nit: you don't "need to use the `logatom` traits", they mostly just help guide you through the construction; there is nothing magical about them, you could also define everything manually

source/docs/guide/src/logatom-hocap.md
slerpyyy: If the example is too large, we could also hide it behind an HTML [`<detail>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) element so it doesn't interrupt the flow of the explanation too much. The mdbook does seem to support this, but it would be the first time we do this in the guide.

source/docs/guide/src/logatom-hocap.md
slerpyyy: nit: "linearization" -> "linearizability"

optional: good place to cite [Herlihy & Wing TOPLAS'90](https://doi.org/10.1145/78969.78972)

source/docs/guide/src/logatom-hocap.md
bsdinis: I think the point here is to guide through the HOCAP parts of the example, and the reference to the guide is just for reference (pardon my tautology).

If we do want to insist on having the full thing, the way to go is the `<detail>`, but at the end of the example

source/docs/guide/src/SUMMARY.md
```diff
@@ -102,6 +102,7 @@
   - [Atomic specifications](./logatom-spec.md)
   - [Invoking the linearization point](./logatom-open.md)
   - [Calling a logically atomic function](./logatom-call.md)
+  - [HOCAP style linearization](./logatom-hocap.md)
```
parno, line 105: I think we need a hyphen here: "HOCAP-style".

I think we also need to work on the organization of this chapter.  The top-level chapter page (`logatom.md`) focuses almost entirely on the new TaDa-style atomicity, with a brief mention of HOCAP style under "Previous & related work".  If we want to present HOCAP and TaDa as two equal alternatives, we should revamp that opening page a bit.  Even if we think the new style is going to be the dominant approach moving forward, we should at least make the availability of HOCAP-style more prominent on that page and add a link to the new `logatom-hocap.md` page.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+A `Linearizer` is a ghost callback that has pre/post-conditions.
+It must be called to linearize an `Operation` at a valid [linearization point](https://dl.acm.org/doi/pdf/10.1145/78969.78972),
+where it is consumed and a `Completion` is obtained.
+
+In our examples, we will focus on [`MutLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.MutLinearizer.html),
+but the [`ReadLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.ReadLinearizer.html) is very similar.
+
+## How the logical atomicity proof works
+
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
```
parno, line 19: There's an extra space after "linearizer"

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
+This resource captures the concurrent object being talked about.
+The logical atomiciy proof is achieved by calling the `apply` function in succession,
```
parno, line 22: atomiciy -> atomicity

parno: What do you mean by "in succession"?  That sounds like you're sitting there calling it a bunch of times in a row, which doesn't seem right.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+It must be called to linearize an `Operation` at a valid [linearization point](https://dl.acm.org/doi/pdf/10.1145/78969.78972),
+where it is consumed and a `Completion` is obtained.
+
+In our examples, we will focus on [`MutLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.MutLinearizer.html),
+but the [`ReadLinearizer`](https://verus-lang.github.io/verus/verusdoc/vstd/logatom/trait.ReadLinearizer.html) is very similar.
+
+## How the logical atomicity proof works
+
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
```
parno, line 20: "happens" sounds very passive/vague here.  I think it would be clearer and connect better with the previous sentence with something like: "When calling `MutLinearizer::apply`, we present ..."

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+This resource captures the concurrent object being talked about.
```
parno, line 21: "being talked about" is unclear.  Perhaps "being manipulated at the linearization point"?  Although that doesn't seem accurate either, since `apply` might also be manipulating its own resources too.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+The logical atomiciy proof is achieved by calling the `apply` function in succession,
+which requires showing the atomic pre-/post-conditions in the `apply` function itself.
+
+We are going to give intuition for how to use HOCAP, showing different
+elements that are required for its successful application. As a running example,
+we will show how to verify the logical atomicity of a monotonic counter example.
+For the full version, refer to the examples directory on the repository.
```
parno, line 28: "on" -> "in"

And we should add a link to the right file.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+
+When we have a linearizer , we need to call `MutLinearizer::apply` at a valid linearization point.
+When `apply` happens, we present a [`Resource`](https://docs.rs/vstd/latest/vstd/resource/algebra/struct.Resource.html).
+This resource captures the concurrent object being talked about.
+The logical atomiciy proof is achieved by calling the `apply` function in succession,
+which requires showing the atomic pre-/post-conditions in the `apply` function itself.
+
+We are going to give intuition for how to use HOCAP, showing different
+elements that are required for its successful application. As a running example,
+we will show how to verify the logical atomicity of a monotonic counter example.
```
parno, line 27: Cut "example" at the end.  It's redundant with the "running example" earlier.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+For the full version, refer to the examples directory on the repository.
+
+## Constructing the Operations
+
+Logical atomicity and linearizability deal in operations being linearized.
+The `ReadOperation` and `MutOperation` model these operations that are linearized.
+
+In the counter example, we can specify an increment operation, `IncOp`:
+
+```rust
```
parno, line 37: Can we update these code snippets to be pulled from the code in the examples directory, so that it remains under CI?

parno: We also need to introduce the `GhostVarAuth`

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+        post: Self::Resource,
+        new_state: Self::NewState,
+    ) -> bool {
+        &&& post.id() == self.id
+        &&& post@ == pre@.wrapping_add(1)
+    }
+}
+```
+
+Here, we force the resource being passed in to refer to the same ghost location
```
parno, line 73: "we force the" -> "we force (via our definition of `requires`) the" ?

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+as the counter's resource (by having the location in the operation itself). Moreover,
+we model the expected atomic behaviour of the increment (the counter maches the resource
```
parno, line 75: I would rephrase this as: "in the `ensures` clause, we guarantee that the resource with the monotonic counter has been incremented appropriately."

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+state and is incremented).
+
+## Constructing a Linearizer
+
+The linearizer typically holds the resource itself. The apply function can either
```
parno, line 80: This is confusing, since the impl of `IncOp` above shows it defining/holding a `GhostVarAuth<u64>` resource.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+express agreement (for read only operations) or update the resource (for mutable operations).
+
+We can express additional pre and post conditions on the linearizer. Centrally, in the `apply`
```
parno, line 83: Elsewhere, I think we use "preconditions and postconditions"

source/docs/guide/src/logatom-hocap.md
parno: Maybe it would help to either in the code of the impl (as a comment) or in the text describing it, to say that the `type Resource` defines the type of resource the operation plans to operate on, rather than something it holds.

source/docs/guide/src/logatom-hocap.md
parno: Same with the next line.

source/docs/guide/src/logatom-hocap.md
parno, line 80: We should also (briefly) introduce the `Completion` type.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+In the counter example, we can specify an increment operation, `IncOp`:
+
+```rust
+/// Operation for incrementing the counter
+struct IncOp {
+    /// Location of the GhostVarAuth
+    id: Loc,
+}
+
+impl logatom::MutOperation for IncOp {
```
parno, line 44: Also need to introduce the `ExecResult` and `NewState` types.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    proof fn peek(tracked &self, op: IncOp, tracked resource: &GhostVarAuth<u64>) {}
+}
+
+```
+
+## Calling `apply`
+
+Consider the example of a monotonic counter (as seen in the examples dir).
```
parno, line 126: This sentence seems unnecessary; we've already introduced it as a running example.

source/docs/guide/src/logatom-hocap.md
```diff
@@ -0,0 +1,153 @@
...
+The ghost state is modelled after a [`GhostVarAuth`](https://verus-lang.github.io/verus/verusdoc/vstd/resource/ghost_var/struct.GhostVarAuth.html) held in an [`AtomicInvariant`](https://verus-lang.github.io/verus/verusdoc/vstd/invariant/struct.AtomicInvariant.html),
```
parno, line 127: This seems like it belongs at the start of this page, along with a more concrete example of the actual executable code we're trying to verify here.

parno: Put another way, the description below is accurate but low-level.  It doesn't convey what the programmer was attempting to accomplish at a high-level.

parno (commented):
Thanks for working on this.  I think there's still some work to do in order to make it accessible to a reader unfamiliar with these concepts.  I've tried to add suggestions along the way.

PR #2623, opened 2026-07-04
Add call body to VIR and SST

slerpyyy:
This PR adds a `body` field to the VIR-AST `ExprX::Call` and the VIR-SST `StmX::Call`. The body of a call is evaluated *inside* the call, after the precondition is asserted but before the postcondition is assumed.

We need this change for the atomic function call, as part of the logical atomicity extension (#2326), to allow the `atomically` block to be evaluated during the call, ensuring the atomic pre- and postcondition are scheduled between the regular function pre- and postcondition.

**Note:** We never actually construct a call with a body in this PR, so the CI will not tell you much about whether it works or not. Instead, refer to the PR #2326, which does make use of the call body and has been rebased on top of this branch.

### DIFF
--- a/source/vir/src/resolution_inference.rs
+++ b/source/vir/src/resolution_inference.rs
@@ -744,7 +744,7 @@ impl<'a> Builder<'a> {
                 Ok(bb)
             }
 
-            ExprX::Call(call_target, es, post_args) => {
+            ExprX::Call { target: call_target, args: es, post_args, body } => {
                 assert!(post_args.is_none());
 
                 // Can skip the expression in CallTarget because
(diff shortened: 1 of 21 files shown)


parno:
I think Travis and/or Chris would be better reviewers for this one, than me.

source/vir/src/resolution_inference.rs
```diff
@@ -744,7 +744,7 @@ impl<'a> Builder<'a> {
                 Ok(bb)
             }
 
-            ExprX::Call(call_target, es, post_args) => {
+            ExprX::Call { target: call_target, args: es, post_args, body: _ } => {
```
tjhance: should recurse on `body` here

tjhance (dismissed):

source/vir/src/resolution_inference.rs
slerpyyy: Should be fixed now

tjhance (approved):

PR #2624, opened 2026-07-04
Generalize invariant masks for logical atomicity

slerpyyy:
This PR refactors the way invariant masks are parsed and lowered, making them flexible enough to be used in the logical atomicity extension (#2326).

In short, we change the way the `opens_invariants [...]` family of function specification items is represented using built-in functions as follows:

```
opens_invariants_none()         ~>  opens_invariant_mask(inv_mask_none())
opens_invariants_any()          ~>  opens_invariant_mask(inv_mask_any())
opens_invariants([...])         ~>  opens_invariant_mask(inv_mask_list([...]))
opens_invariants_except([...])  ~>  opens_invariant_mask(inv_mask_list_compl([...]))
opens_invariants_set(...)       ~>  opens_invariant_mask(inv_mask_set(...))
```

This allows the `inv_mask_*` functions to be used independently of the `opens_invariant_mask` header, meaning they can now be used as regular `ISet<int>` spec-mode expressions and places into functions or constants by the `verus!` macro.

This PR also changes the way error messages for mask-related verification conditions are produced, in anticipation for the atomic function call and the `try_open_atomic_update!` macro introduced in the logical atomicity extension.

**Note:** Just like #2623, the logical atomicity PR #2326 has been rebased on this PR.


source/builtin/src/lib.rs
```diff
@@ -315,6 +283,44 @@ pub fn get_union_field<Adt, Field>(_a: Adt, _field: &str) -> Field {
     unimplemented!();
 }
 
+// TODO: These `inv_mask` functions are essentially a copy
```
tjhance: do you need this comment anymore?

source/rust_verify/src/fn_call_to_vir.rs
```diff
@@ -594,44 +580,102 @@ fn verus_item_to_vir<'tcx, 'a>(
...
+                        SpecItem::InvMaskList => MaskSpec::InvariantOpens(span, args),
+                        SpecItem::InvMaskListCompl => MaskSpec::InvariantOpensExcept(span, args),
+                        _ => unreachable!(),
+                    }))
+                }
+                SpecItem::InvMaskSet => {
+                    fn typ_is_int_iset(typ: &Typ) -> bool {
+                        // NOTE: This validation check is a little bit fuzzy on purpose, since there
+                        // are some tests (e.g. `tests/core_special_setup.rs`) that import vstd as a
+                        // module rather than a crate, so we need to accept either.
```
tjhance: Usually to deal with this, we add the rustc_diagnostic_item and declare it as a verus item, which makes the path stable

source/vir/src/ast_to_sst.rs
```diff
@@ -2707,6 +2716,32 @@ pub(crate) fn expr_to_stm_opt(
...
+            let mut exps = Vec::new();
+            let mut stms = Vec::new();
+
+            for expr in exprs.as_ref() {
+                let (exp_stms, ret_val) = expr_to_stm_opt(ctx, state, expr)?;
+                stms.extend(exp_stms);
+
+                let value = unwrap_or_return_never!(ret_val, stms);
+                exps.push(value.to_exp());
+            }
```
tjhance, line 2769: for this sort of pattern you need to use the Sequencer to avoid reordering issues (there are a bunch of examples in this file)

tjhance (dismissed):

source/vir/src/modes.rs
```diff
@@ -3070,6 +3071,22 @@ fn check_expr(
...
+                    record,
+                    &mut typing,
+                    Mode::Spec,
+                    expr,
+                    Mode::Spec,
+                    outer_proph,
+                )?;
+            }
+
+            Ok((Mode::Spec, Proph::No))
```
tjhance: you need to join the proph results returned by `check_expr_has_mode`

source/vir/src/modes.rs
```diff
@@ -3070,6 +3071,22 @@ fn check_expr(
 
             Ok((Mode::Exec, Proph::No))
         }
+        ExprX::InvMask(mask_spec) => {
```
tjhance, line 3102: also for this case you need to set `in_pure` to be true (this has to be done for any node that uses `record_spec_fn_pure_args_only` in fn_call_to_vir)

source/builtin/src/lib.rs
slerpyyy: No, that's outdated, I'll remove it

source/vir/src/modes.rs
slerpyyy: Fixed using `try_fold`

source/vir/src/ast_to_sst.rs
slerpyyy: github doesn't mark this as outdated because the `}` didn't change, but this is also fixed

tjhance (approved):

Issue #2212, opened 2026-02-25
Mode Checking of `Tracked` in spec context


rikosellic:
Verus currently rejects `Tracked` even if it will only be used in the spec context, which is considered a bug according to this [Zulip conversation](https://verus-lang.zulipchat.com/#narrow/channel/399078-help/topic/An.20attempt.20to.20define.20.60Seq.60.20with.20.60enum.60/with/575582940).

**Example**
```
spec fn test(f: spec_fn(nat)->nat)
{
    let n = Tracked(f(0));
}
```

**Output**
```
error: cannot perform operation with mode proof
 --> /playground/src/main.rs:7:13
  |
7 |     let n = Tracked(f(0)); 
  |             ^^^^^^^^^^^^^

error: aborting due to 1 previous error
```

PR #2625, opened 2026-07-04
vstd: change Seq to Enum and verify axioms

bsdinis:
This PR updates the code in #988 by changing Seq into a cons list construction and verifying its axioms.

It mostly went fine with the exception of `axiom_seq_len_decreases`, which I couldn't prove. Happy to take ideas on how to achieve it.

One main thing we should decide on is what to do with the naming. The previous axioms are now lemmas, but I've left them named as `axiom_*`. The obvious argument for changing is that the naming is better. The argument against is that it might break users. The counter to that is that most people use them via the broadcast group and the change (while breaking) is quite easy to fix.

I also implemented some of the precondition weakening @tchajed had implemented. Couldn't make it work neatly for `axiom_seq_add_index2`, though.

Interesting note:
- In a lot of cases, verus failed to pierce through the isolation of `struct Seq { inner: SeqInner }`, forcing the lemma to be proven on `SeqInner` instead (with the exception of the ext_equality lemmas that had to be proven on the full Seq). This is why most methods are defined on `SeqInner` and then shallow inserted into `Seq`

---

### DIFF
--- a/source/vstd/seq.rs
+++ b/source/vstd/seq.rs
@@ -68,14 +68,18 @@ impl<A> SeqInner<A> {
             0 <= i < self.len(),
         decreases self.len(),
     {
-        match self {
-            SeqInner::Nil => arbitrary(),
-            SeqInner::Cons { head, tail } => if i == 0 {
-                SeqInner::Cons { head: a, tail }
-            } else {
-                let new_tail = tail.update(i - 1, a);
-                SeqInner::Cons { head, tail: Box::new(new_tail) }
-            },
+        if !(0 <= i < self.len()) {  // this supports weakening some preconditions
+            self
+        } else {
+            match self {
+                SeqInner::Nil => arbitrary(),
+                SeqInner::Cons { head, tail } => if i == 0 {
+                    SeqInner::Cons { head: a, tail }
+                } else {
+                    let new_tail = tail.update(i - 1, a);
+                    SeqInner::Cons { head, tail: Box::new(new_tail) }
+                },
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


bsdinis:
Is there any advice on circumventing the `no_alloc` issue? Do we have another indirection mechanism for recursive datatypes at the ghost level?

rikosellic:
> Is there any advice on circumventing the `no_alloc` issue? Do we have another indirection mechanism for recursive datatypes at the ghost level?

I've tried the following definition earlier that does not have the alloc issue:
```
enum SeqInner<A> {
    Nil,
    Cons(A, Tracked<SeqInner<A>>),
}
```

However, it gets stuck because of a mode-checking bug. You can find the details in [this Zulip discussion](https://verus-lang.zulipchat.com/#narrow/channel/399078-help/topic/An.20attempt.20to.20define.20.60Seq.60.20with.20.60enum.60/with/575582940).

bsdinis:
Thank you! I'll give it a look!

bsdinis:
Apparently this is blocked on #2212. 
The `Box`es should be switched with `Tracked`s and this would also allow us to verify the `tracked_` interface.

Travis mentioned on Zulip that it should be allowed

cc @tjhance

bsdinis:
On second thought, it might not be blocked. We can do the proof with `Ghost`, but this still doesn't allow for verifying the `tracked` interface. Opened an issue (#2627) to track proving the `tracked_` functions, which *is* blocked on #2212

source/vstd/seq.rs
```diff
@@ -25,35 +139,48 @@ verus! {
 ///
 /// To prove that two sequences are equal, it is usually easiest to use the
 /// extensional equality operator `=~=`.
-#[verifier::external_body]
 #[verifier::ext_equal]
 #[verifier::accept_recursive_types(A)]
-pub tracked struct Seq<A> {
-    dummy: marker::PhantomData<A>,
+pub struct Seq<A> {
```
tjhance: we need to keep the `tracked` here or it could break some clients using `tracked &mut Seq<T>`

tjhance (changes requested):

source/vstd/seq.rs
bsdinis: good point, that shoud;nt have made it in

tjhance (approved):

PR #2630, opened 2026-07-06
rust_verify: rust_to_vir_expr: don't emit no-op ToDyn

### LINKED ISSUE #2629: Verus rejects established pre-conditions on owned `Box<dyn Trait>` parameters
Verus rejects the following code:

```rust
use vstd::prelude::*;

verus! {

pub trait T {
    spec fn t(&self) -> int;
}

struct Ty {}

impl T for Ty {
    open spec fn t(&self) -> int { 0 }
}

fn borrowed<'a>(x: &'a Box<dyn T>)
    requires
        x.t() == 0,
{}

fn owned(x: Box<dyn T>)
    requires
        x.t() == 0,
{}

fn repro() {
    let x: Box<dyn T> = Box::new(Ty {});

    assert(x.t() == 0);

    // succeeds
    borrowed(&x);
    
    // We can even add an assume and still get the below failure:
    assume(x.t() == 0);
    
    // expected: should use the assumed precondition; actual: fails
    owned(x);
}

fn main() {}

}
```

Expected: the `assert(x.t() == 0)` or `assume(x.t() == 0)` should discharge the pre-condition on `owned`, just like it does for `borrowed`.

Actual: Verus rejects the call to `owned` due to an unsatisfied pre-condition:
```
error: precondition not satisfied
  --> /playground/src/main.rs:37:5
   |
22 |         x.t() == 0,
   |         ---------- failed precondition
...
37 |     owned(x);
   |     ^^^^^^^^

```

Verus playground: https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=4372750ee5ca89c46ee78e18716e1b9a

lschuermann:
rustc sometimes emits no-op pointer unsize adjustments. When Verus generates a `ToDyn` coercion for two identical types, this breaks verification because facts about the original value are not propagated through `ToDyn`. Skip emitting these coercions when the source and target types are identical.

Fixes: https://github.com/verus-lang/verus/issues/2629

I want to caution and disclose: I'm way out of my depth on this one. I traced this issue from the SMT queries up through the AIR and VIR, and looked at the HIR. I used an LLM to help me trace down the exact code where this is translated, and fudged the code til it worked. I have little confidence whether this change is correct and doesn't break anything else, so I'd appreciate the appropriate caution during review. 😄 

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -1372,6 +1372,14 @@ pub(crate) fn expr_to_vir_with_adjustments<'tcx>(
 
             let arg = expr_to_vir_with_adjustments(bctx, expr, adjustments, adjustment_idx - 1)?;
 
+            // rustc can emit no-op Pointer(Unsize) adjustments. If this is
+            // lowered into a visible `ToDyn`, facts about the original value
+            // are not propagated through the coercion, as in #2629.
+            // Don't emit anything for no-op coercions.
+            if ty1 == ty2 {
+                return Ok(ExprOrPlace::Expr(arg.consume(bctx, ty1)));
+            }
+
             let (tyr1, tyr2) = remove_decoration_typs_for_unsizing(bctx.ctxt.tcx, ty1, ty2);
             let op = match (tyr1.kind(), tyr2.kind()) {
                 (_, TyKind::Dynamic(_, _)) => {
--- a/source/rust_verify_test/tests/traits_dyn.rs
+++ b/source/rust_verify_test/tests/traits_dyn.rs
@@ -431,3 +431,35 @@ test_verify_one_file! {
         }
     } => Err(err) => assert_one_fails(err)
 }
+
+test_verify_one_file! {
+    #[test] test_dyn_owned_precond verus_code! {
+        use vstd::prelude::*;
+        verus! {
+            pub trait T {
+                spec fn t(&self) -> int;
+            }
+            struct Ty {}
+            impl T for Ty {
+                open spec fn t(&self) -> int { 0 }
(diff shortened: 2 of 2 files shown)


tjhance (changes requested):
This fix looks fine to me, but the repro from #2629 should be added as a test case (rust_verify_test/tests/traits_dyn.rs)

lschuermann:
Thanks! Added the test case. Please let me know if there's anything I should change!

tjhance:
this looks fine, I think, but I also want to review a couple of things related to rust's handling of the Unsize coercion before final approval, which I will try to do in the next week

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -1372,6 +1372,14 @@ pub(crate) fn expr_to_vir_with_adjustments<'tcx>(
 
             let arg = expr_to_vir_with_adjustments(bctx, expr, adjustments, adjustment_idx - 1)?;
 
+            // rustc can emit no-op Pointer(Unsize) adjustments. If this is
+            // lowered into a visible `ToDyn`, facts about the original value
+            // are not propagated through the coercion, as in #2629.
+            // Don't emit anything for no-op coercions.
+            if ty1 == ty2 {
+                return Ok(arg);
```
tjhance: OK, I can't find a way for this to actually matter in practice, but to be correct, we should make sure that the `arg` is coerced from a Place to an Expr, if necessary:

```
return Ok(ExprOrPlace::Expr(arg.consume(bctx, ty1)));
```

tjhance (changes requested):
We should be good to go after this small fix.

tjhance (approved):

PR #2645, opened 2026-07-12
air: handle unparseable (get-model) output without aborting

zoratu:
If Z3 returns get-model output that isn't a parseable s-expression, verus aborts the whole process instead of reporting an error. smt_get_model calls lines_to_model, which unwraps/expects on the parse; the panic then unwinds through a PanicOnDropVec whose destructor panics again, so it turns into a non-unwinding abort with no useful message.

This is reachable whenever Z3 hands back something unexpected for get-model, e.g. an (error ...) reply, or garbage from a broken or mismatched solver binary. The nearby "model is not available" case is already handled gracefully; this is the same situation with different output.

The fix makes parse_sexpression and lines_to_model return Result (matching the Result<_, String> style already used throughout the parser), and degrades gracefully in the two callers: smt_get_model reports the obligation as invalid like the "model is not available" branch does, and smt_get_rlimit_count returns UnexpectedOutput. No change on the normal path.

I ran into this with a misconfigured Z3. To reproduce it deterministically I used a small proxy that corrupts only the get-model reply, triggered by a failing assert so verus asks for a model: before the change it aborts, after it reports the failure and exits normally. A set of real proofs verify identically with and without the patch.

### DIFF
--- a/source/air/src/parser.rs
+++ b/source/air/src/parser.rs
@@ -875,18 +875,18 @@ impl Parser {
         }
     }
 
-    pub fn lines_to_model(&self, lines: &Vec<String>) -> ModelDefs {
-        let node = parse_sexpression(lines);
-        self.node_to_model(&node).expect("failed to parse SMT model")
+    pub fn lines_to_model(&self, lines: &Vec<String>) -> Result<ModelDefs, String> {
+        let node = parse_sexpression(lines)?;
+        self.node_to_model(&node)
     }
 }
 
-pub(crate) fn parse_sexpression(lines: &Vec<String>) -> Node {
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel (changes requested):
Thanks for the pull request.  This definitely improves handling of unexpected SMT output.  The only change I'd like to see is for `smt_get_model` to return `ValidityResult::UnexpectedOutput` rather than degrading gracefully with `ValidityResult::Invalid`.  For expected missing models, like the "model is not available" case, it's fine to return `ValidityResult::Invalid`, but if it's an unexpected error I'd rather have `ValidityResult::UnexpectedOutput`.

PR #2646, opened 2026-07-13
fix: pin version in cargo-verus vstd test

mkovaxx:
Fixes failure due to new compatible `vstd` versions becoming available.
See: https://github.com/verus-lang/verus/actions/runs/29175180600


mkovaxx:
The problem was that Cargo treats all `0.0.0-*` versions as compatible [regardless of the pre-release version](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#pre-releases) (in our case, the date). So even though the test fixture asks for `vstd = "0.0.0-2026-06-14-0213"`, Cargo figures that (the recently-published) `0.0.0-2026-07-12-0122` is also fine, but the test needs to confirm a specific version number.

The fix is to use an [equality requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#comparison-requirements), i.e. `vstd = "=0.0.0-2026-06-14-0213"`.

PR #2648, opened 2026-07-13
make_spec_type: support tuple/unit structs and add a closed option

amaurremi:
## Summary

The `#[make_spec_type]` contrib macro (`builtin_macros/src/contrib/spec_derive.rs`) previously assumed named-field structs. This PR makes it handle all three struct kinds and adds a `closed` option for types whose fields are not fully `pub`.

## What changed

1. **Tuple and unit structs.** `generate_struct_impls` called `f.ident.as_ref().unwrap()`, which panics on a tuple struct (unnamed fields have no ident) and mishandles unit structs. It now matches on all three `Fields` kinds:
   - `Named`   → `XSpec { field: <T as DeepView>::V, ... }`
   - `Unnamed` → `XSpec(<T0 as DeepView>::V, ...)` (project the view as `x@.0`)
   - `Unit`    → `XSpec` (unit spec type)

   Field exclusion stays named-only, consistent with tuple enum variants.

2. **`closed` option.** `#[make_spec_type(closed)]` generates a `closed spec fn deep_view` instead of `open`. An `open spec` body that reads a field less visible than the function is rejected by Verus (opaque-datatype error), so without `closed` the macro cannot be applied to a type with `pub(crate)`/private fields. The generated spec *type* keeps its fields `pub`, so callers can still project the view (`x@.0`); only the view *computation* is hidden in the defining module. `View::view` forwards to `deep_view` and stays `open` regardless.

3. **Attribute parsing.** Now accepts a comma-separated list of `exclude(...)` and/or `closed`, composes them, and rejects unknown options and non-existent excluded fields with a clear compile error (instead of `panic!`).

## Tests

Added to `rust_verify_test/tests/spec_derive.rs`: tuple newtype (with lifetime), multi-field tuple, unit struct, `closed` + `pub(crate)` field, `closed` + `exclude`, and unknown-option rejection.

## Notes

- `rustfmt`-clean (default settings).
- No change to named-struct or enum behavior; existing tests unchanged.

### DIFF
--- a/source/builtin_macros/src/contrib/spec_derive.rs
+++ b/source/builtin_macros/src/contrib/spec_derive.rs
@@ -32,16 +32,9 @@ use syn::{
 /// }
 /// ```
 pub fn make_spec_type(attr: TokenStream, item: TokenStream) -> TokenStream {
-    let excluded_fields: HashSet<String> = if attr.is_empty() {
-        HashSet::new()
-    } else {
-        match syn::parse(attr) {
-            Ok(meta) => match parse_excluded_fields(meta) {
-                Ok(fields) => fields,
-                Err(err) => return err.to_compile_error().into(),
-            },
-            Err(err) => return err.to_compile_error().into(),
-        }
+    let (excluded_fields, closed) = match parse_make_spec_type_attr(attr) {
+        Ok(parsed) => parsed,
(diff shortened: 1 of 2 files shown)


parno (approved):
Looks reasonable to me.

PR #2650, opened 2026-07-14
workspace: update dependencies

bsdinis:
With `cargo`, it is possible to set the dependencies at the workspace level
and just set the dependency in the workspace member's Cargo.toml with
`dep = { workspace = true }`.

This has several advantages:
- It is easier to audit the dependencies and update them
- We ensure that dependencies are unique (i.e., we are not downloading 5
  different compatible versions of the same library without needing to)

I looked into all our dependencies, and we had 66 different dependencies
(crate + version) corresponding to only 44 unique crates.

This PR moves us to a centralized workspace dependency management
style.

Dependencies are declared in the top level Cargo.toml.
Crates always set their dependencies to be `{ workspace = true }`.

This PR also updates dependencies.

As a passerby, added bump_crate_versions to the workspace and fixed
clippy warnings.

This PR is structured as follows:
- centralize dependencies
- update dependencies that have security advisories as set out with `cargo audit`
- update other dependencies

Two extra thoughts:
- we should consider fixing `cargo audit` and running it with some regularly
- we depend on unmaintained/problematic crates, that also prevent us from updating some other dependencies. The offenders are `smt-scope` (which has too many old dependencies) and `kuchiki` (fully unmaintained, and follow up `kuchikiki` also seems unmaintained)


mkovaxx:
@bsdinis Good idea, thanks for making this happen! :)

mkovaxx (changes requested):
It seems that the current state has some unrelated changes mixed in...

Would it be possible to make the changes more conservative, updating only `Cargo.toml` files first, and making other changes in a follow-up PR?

mkovaxx:
@bsdinis I've split out the changes restricted to `Cargo.toml` files here: https://github.com/verus-lang/verus/pull/2653

source/air/src/emitter.rs
```diff
@@ -2,7 +2,6 @@ use crate::ast::{Decl, Expr, Ident, Query};
 use crate::context::SmtSolver;
 use crate::printer::{NodeWriter, Printer, macro_push_node};
 use crate::{node, nodes};
-use sise::Node;
```
mkovaxx, line 5: I think these changes should use import aliases to reduce churn.
```rust
use sise::TreeNode as Node;
```

mkovaxx (approved):
LGTM, added a non-blocking comment about reducing churn.

source/rustc_mir_build/src/builder/matches/mod.rs
```diff
@@ -9,7 +9,7 @@ use std::borrow::Borrow;
 use std::sync::Arc;
 use std::{debug_assert_matches, mem};
 
-use itertools::{Itertools, Position};
+use itertools::Itertools;
```
tjhance: It would be better to not modify forked crates unnecessarily.

tjhance (changes requested):

source/rustc_mir_build/src/builder/matches/mod.rs
tjhance: In particular, it would also be best not to redirect anything in rustc_mir_build/Cargo.toml to the workspace so that we can easily stay in sync with whatever version rustc uses.

bsdinis: Gotcha, is it just the `rustc_` crates?

tjhance: yeah

source/air/src/emitter.rs
bsdinis: I don't feel too strongly about it, but part of the problem here is that it was very confusing in the code what is our Node and other nodes. Moreover, we have ~4 Node types with our dependencies that then emit very noisy warnings. My inclination is to leave as is unless someone who works a lot with these parts of air/vir has very strong opinions

source/rustc_mir_build/src/builder/matches/mod.rs
bsdinis: okay, will do that

source/air/src/emitter.rs
mkovaxx: Thanks for explaining the intent in more detail. I agree that changing `Node` to `TreeNode` likely helps disambiguate existing code, even though I'm not personally familiar with those parts.

That being said, I think that goal is sufficiently different from the stated goal of this PR for it to be done in a separate follow-up PR, making focused review easier on both PRs.

bsdinis: done

source/rustc_mir_build/src/builder/matches/mod.rs
bsdinis: done

bsdinis:
@tjhance I think I've fixed the issue with the rustc_ crates

bsdinis:
@tjhance ping on this, I'll fix the conflicts in the meantime

bsdinis:
> @tjhance ping on this, I'll fix the conflicts in the meantime

fixed problems

tjhance (approved):

tjhance:
make sure to re-request review so it gets back on my queue!

PR #2654, opened 2026-07-14
Upgrade Verus to use Rust 1.97

dschoepe:
(AI-assisted)


bsdinis:
Rust 1.97.0 [has a bug](https://blog.rust-lang.org/2026/07/16/Rust-1.97.1/), we should instead update to 1.97.1

source/rust_verify/src/boundary_suggestions.rs
```diff
@@ -150,7 +151,9 @@ pub(crate) fn build_fn_assume_specification_suggestion<'tcx>(
     let mut region_renamer: RegionRenamer<'_> =
         build_region_renamer(ctxt, external_def_id, generics)?;
     let fn_sig = fn_sig.fold_with(&mut region_renamer);
-    let inst_predicates = inst_predicates.fold_with(&mut region_renamer);
+    // TODO(1.97.0): figure out how to replace this. InstantiatedPredicates no longer implements TypeFoldable
+    // let predicates = predicates.fold_with(&mut region_renamer);
+    // let inst_predicates = inst_predicates.fold_with(&mut region_renamer);
```
tjhance: https://doc.rust-lang.org/1.97.0/nightly-rustc/rustc_middle/ty/struct.InstantiatedPredicates.html

fold on each Clause individually?

dschoepe:
> Rust 1.97.0 [has a bug](https://blog.rust-lang.org/2026/07/16/Rust-1.97.1/), we should instead update to 1.97.1

Done

tjhance (approved):

source/rust_verify/src/boundary_suggestions.rs
dschoepe: Thanks, I think I discovered that in the meantime too. (I think the comment is on a previous commit)

Chris-Hawblitzel (approved):

PR #2660, opened 2026-07-15
clean up: well_formed error returns

tjhance:
The interface to well_formed.rs was greatly confusing me on account of: (i) the confusing names NonBlockingError and NonFatalError, (ii) the fact that both were passed into the diagnostics, which is usually just for warnings and notes, resulting in scenarios where we would return Ok for an erroring scenario.

This PR removes NonBlockingError and NonFatalError; transmits the "boundary errors" through Err like normal, and uses another channel for the deferred errors.


parno (approved):
Seems reasonable to me.

PR #2665, opened 2026-07-15
Implement IndexMut spec with &mut semantics

sarah-quinones:

this pr rewrites `IndexMut` with a `&mut`-based spec api, similar to the immutable `Index` version, and removes the previous `index_set` hack.

it also removes the (soft) restriction that disallows mutable references in function parameters in combination with `allow_in_spec` and `when_used_as_spec`. note that this restriction was an artificial limitation, and is implemented in a shallow way that disallows `&mut T`, but allows `Wrapper<&mut T>` for example. if this is an objectionable change, im open to reverting it and using a different workaround.


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -3096,9 +3088,22 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
...
-                    let typ_args =
-                        Arc::new(vec![undecorate_typ(&tgt_vir.typ), idx_vir.typ.clone()]);
-                    let fun = vir::fun!(CrateId::Core => "ops", "index", "Index", "index");
+
+                    let mut ty = undecorate_typ(&tgt_vir.typ);
+                    if mutbl {
+                        if let TypX::MutRef(inner) = &*ty {
+                            ty = inner.clone();
+                        } else {
+                            panic!()
```
tjhance: use the `internal_error!` macro for this (EDIT: you should be able to get rid of this entirely - see my above comment)

source/vstd/std_specs/core.rs
```diff
@@ -50,16 +60,33 @@ pub trait ExIndex<Idx> where Idx: ?Sized {
 
     type Output: ?Sized;
 
-    spec fn index_req(&self, index: &Idx) -> bool;
+    spec fn index_requires(&self, index: &Idx) -> bool;
```
tjhance: add a note here that this is also uses as the precondition in `IndexMut`

source/vstd/std_specs/core.rs
```diff
@@ -28,19 +28,29 @@ pub trait ExFn<Args: core::marker::Tuple>: FnMut<Args> {
...
+            self.deref_ensures(output);
 }
 
 #[verifier::external_trait_specification]
+#[verifier::external_trait_extension(DerefMutSpec via DerefMutSpecImpl)]
 pub trait ExDerefMut: core::ops::Deref + PointeeSized {
     type ExternalTraitSpecificationFor: core::ops::DerefMut;
 
-    fn deref_mut(&mut self) -> &mut Self::Target;
+    spec fn deref_mut_ensures(&mut self, output: &mut Self::Target) -> bool;
```
tjhance: Why do we need this spec function?

Typically you just use `call_ensures` to get the ensures clause of each trait impl without needing to write an explicit spec fn for it.

I could see adding it for convenience, but I don't see it used anywhere else in this PR.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -3375,9 +3367,22 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
                     // general Index trait case
                     let (impl_paths, target_kind) =
                         resolve_index_call(bctx, *tgt_ty, idx_ty, false, expr.span)?;
-                    let typ_args =
-                        Arc::new(vec![undecorate_typ(&tgt_vir.typ), idx_vir.typ.clone()]);
```
tjhance: FYI you'll need to rebase over this: https://github.com/verus-lang/verus/pull/2670/changes

I didn't like the way the typ_args were computed (before _or_ after your change (the use of undecorate_typ here is incredibly dubious)) but with #2670 it should compute typ_args that work with or without mutable references.

tjhance (changes requested):

source/vstd/std_specs/core.rs
sarah-quinones: ah, was not aware of `call_ensures`, and this was an unrelated change anyway. so im happy to revert it

source/rust_verify/src/rust_to_vir_expr.rs
sarah-quinones: i'll take a look and rebase

source/vstd/std_specs/core.rs
```diff
@@ -28,19 +28,21 @@ pub trait ExFn<Args: core::marker::Tuple>: FnMut<Args> {
...
 
 #[verifier::external_trait_specification]
+#[verifier::external_trait_extension(DerefSpec via DerefSpecImpl)]
 pub trait ExDeref: PointeeSized {
     type ExternalTraitSpecificationFor: core::ops::Deref;
 
     type Target: ?Sized;
 
-    fn deref(&self) -> &Self::Target;
+    fn deref(&self) -> (output: &Self::Target);
```
Chris-Hawblitzel: You don't need to name the output here.

source/vstd/std_specs/core.rs
```diff
@@ -28,19 +28,21 @@ pub trait ExFn<Args: core::marker::Tuple>: FnMut<Args> {
 }
 
 #[verifier::external_trait_specification]
+#[verifier::external_trait_extension(DerefSpec via DerefSpecImpl)]
```
Chris-Hawblitzel: I don't think you need to add external_trait_extension here.

source/vstd/std_specs/core.rs
```diff
@@ -28,19 +28,21 @@ pub trait ExFn<Args: core::marker::Tuple>: FnMut<Args> {
...
     type ExternalTraitSpecificationFor: core::ops::Deref;
 
     type Target: ?Sized;
 
-    fn deref(&self) -> &Self::Target;
+    fn deref(&self) -> (output: &Self::Target);
 }
 
 #[verifier::external_trait_specification]
+#[verifier::external_trait_extension(DerefMutSpec via DerefMutSpecImpl)]
```
Chris-Hawblitzel: I don't think you need to add external_trait_extension here.

source/vstd/std_specs/core.rs
```diff
@@ -28,19 +28,21 @@ pub trait ExFn<Args: core::marker::Tuple>: FnMut<Args> {
...
 pub trait ExDerefMut: core::ops::Deref + PointeeSized {
     type ExternalTraitSpecificationFor: core::ops::DerefMut;
 
-    fn deref_mut(&mut self) -> &mut Self::Target;
+    fn deref_mut(&mut self) -> (output: &mut Self::Target);
```
Chris-Hawblitzel: You don't need to name the output here.

source/vstd/std_specs/core.rs
```diff
@@ -50,16 +52,24 @@ pub trait ExIndex<Idx> where Idx: ?Sized {
 
     type Output: ?Sized;
 
-    spec fn index_req(&self, index: &Idx) -> bool;
+    // NOTE: this used as a precondition for both `Index` and `IndexMut`,
+    // since both share the same `s[i]` syntax.
+    spec fn index_requires(&self, index: &Idx) -> bool;
```
Chris-Hawblitzel: We've been using the `_req` suffix in general in std_specs (e.g. in ops.rs), so I would leave this as `index_req`.

source/vstd/std_specs/slice.rs
```diff
@@ -58,35 +40,59 @@ pub assume_specification<T>[ <Range<usize> as SliceIndex<[T]>>::index ](i: Range
...
-    where [T]: Index<I>,
+pub assume_specification<T, I, const N: usize>[ <[T; N]>::index_mut ](array: &mut [T; N], index: I) -> (output: &mut <[T; N] as Index<I>>::Output)
+    where
+        [T]: IndexMut<I>,
     ensures
-        call_ensures(<[T] as Index<I>>::index, (array, index), output),
+        forall|slice: &mut [T]| #![trigger slice@] {
+            &&& slice@ == old(array)@
+            &&& final(slice)@ == final(array)@
+        } ==> call_ensures(<[T]>::index_mut, (slice, index), output),
```
Chris-Hawblitzel: Instead of using a forall here, I would suggest the following ensures:

```
    ensures
        exists|slice: &mut [T]| {
            &&& #[trigger] slice@ == old(array)@
            &&& final(slice)@ == final(array)@
            &&& call_ensures(<[T]>::index_mut, (slice, index), output)
        }
;
```

This pattern mimics a call to `ref_mut_array_unsizing_coercion(r: &mut [T; N]) -> (out: &mut [T])`, resulting in the existence of a particular `slice` value, which is then passed on to `<[T]>::index_mut`.

Here are some test cases that show this in action:

```
fn test1(a: &mut [u8; 5]) {
    *a.index_mut(2) = 7;
    assert(a[2] == 7);
}

fn test2(a: &mut [u8; 5]) {
    let r: &mut [u8] = &mut a[1..3];
    r[1] = 7;
    assert(r[1] == 7);
    assert(a[2] == 7);
}
```

source/vstd/slice.rs
```diff
@@ -122,11 +122,16 @@ pub trait ExSliceIndex<T> where T: ?Sized {
 
     type Output: ?Sized;
 
-    spec fn index_req(&self, slice: &T) -> bool;
+    spec fn index_requires(&self, slice: &T) -> bool;
 
-    fn index(self, slice: &T) -> &Self::Output
+    fn index(self, slice: &T) -> (output: &Self::Output)
```
Chris-Hawblitzel: You don't need to name the output here.

source/vstd/slice.rs
```diff
@@ -122,11 +122,16 @@ pub trait ExSliceIndex<T> where T: ?Sized {
...
         requires
-            self.index_req(slice),
+            self.index_requires(slice),
+    ;
+
+    fn index_mut(self, slice: &mut T) -> (output: &mut Self::Output)
```
Chris-Hawblitzel: You don't need to name the output here.

Chris-Hawblitzel (changes requested):
Overall, this looks really good.  I have some detailed comments.

Chris-Hawblitzel (approved):

tjhance (approved):

sarah-quinones:
oh whoops, looks like i missed one last `index_requires` in the test file

tjhance:
sorry, I was messing with the Index code some more, so this needs conflicts to be resolved yet again.

PR #2669, opened 2026-07-16
vstd: finish proving lemmas for `Seq`

### LINKED ISSUE #2627: Proving the `Seq::tracked_*` functions
A recent PR (#2625) verifies Verus' Sequences using a cons-list approach.

However, due to a bug in the mode-checking (#2212 ) we cannot use `Tracked` to define the `SeqInner` enum, which would in turn enable, in theory, verifying the `tracked_` functions of `Seq`.

Once that issue is solved, we can revisit and verify those.

bsdinis:
This PR finishes the work on verified `Seq`.

- It proves the `tracked_` functions (adding two new ones that I find would be useful -- `tracked_split_at` and `tracked_add`)
- It adds a new lemma for composing the `subrange` operation.
- It moves the old `axioms` to be deprecated and instead call the actual lemmas.

To do this, we had to allow tracked bindings to proof code.

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -1026,12 +1026,11 @@ fn add_pattern_rec(
 
 fn check_binding(span: &Span, by_ref: &ByRef, mode: Mode) -> Result<(), VirErr> {
     match (by_ref, mode) {
-        (ByRef::MutRef, Mode::Spec | Mode::Proof) => {
-            // Supporting this for Mode::Proof would be nice but requires thought for how
-            // to implement.
-            Err(error(span, "a 'mut ref' binding in a pattern is only allowed for exec mode"))
+        (ByRef::MutRef, Mode::Spec) => {
+            Err(error(span, "a 'mut ref' binding in a pattern is not allowed in spec mode"))
         }
         (ByRef::No | ByRef::ImmutRef, _) => Ok(()),
+        (_, Mode::Proof) => Ok(()),
         (_, Mode::Exec) => Ok(()),
     }
 }
--- a/source/vstd/seq.rs
+++ b/source/vstd/seq.rs
@@ -1174,7 +1174,17 @@ impl<A> Seq<A> {
     }
 }
 
+#[deprecated(note = "Seq axioms have been verified, and are now lemmas")]
 pub broadcast proof fn axiom_seq_index_decreases<A>(s: Seq<A>, i: int)
+    requires
+        0 <= i < s.len(),
+    ensures
+        #[trigger] (decreases_to!(s => s[i])),
+{
+    lemma_seq_index_decreases(s, i)
+}
+
+pub broadcast proof fn lemma_seq_index_decreases<A>(s: Seq<A>, i: int)
     requires
         0 <= i < s.len(),
     ensures
@@ -1194,6 +1204,7 @@ pub axiom fn axiom_seq_len_decreases<A>(s1: Seq<A>, s2: Seq<A>)
         decreases_to!(s1 => s2),
 ;
 
+#[deprecated(note = "Seq axioms have been verified, and are now lemmas")]
 pub broadcast proof fn axiom_seq_subrange_decreases<A>(s: Seq<A>, i: int, j: int)
     requires
         0 <= i <= j <= s.len(),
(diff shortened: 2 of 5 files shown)


rikosellic:
Is there a `tracked_borrow_mut` method?

bsdinis:
> Is there a `tracked_borrow_mut` method?

Extremely pertinent point.

The reason there isn't is because I thought of adding it, but forgot.

bsdinis:
@rikosellic It is now implemented

source/vir/src/modes.rs
```diff
@@ -1026,12 +1026,11 @@ fn add_pattern_rec(
 
 fn check_binding(span: &Span, by_ref: &ByRef, mode: Mode) -> Result<(), VirErr> {
     match (by_ref, mode) {
-        (ByRef::MutRef, Mode::Spec | Mode::Proof) => {
-            // Supporting this for Mode::Proof would be nice but requires thought for how
-            // to implement.
+        (ByRef::MutRef, Mode::Spec) => {
             Err(error(span, "a 'mut ref' binding in a pattern is only allowed for exec mode"))
```
parno: Should this message now say "only allowed in proof or exec mode"?

source/vstd/seq.rs
```diff
@@ -1064,34 +1857,36 @@ pub broadcast proof fn lemma_seq_add_index2_alt<A>(s1: Seq<A>, s2: Seq<A>, i: in
...
-    axiom_seq_add_index2,
+    lemma_seq_add_len,
+    lemma_seq_add_index1,
+    lemma_seq_add_index2,
 }
 
+// TODO(bsdinis): rename to group_seq_lemmas?
 // Expensive lemmas not in the default group (may slow down verification)
-pub broadcast group group_seq_lemmas_expensive {
+pub broadcast group group_seq_axioms_expensive {
```
parno: This was previously `group_seq_lemmas_expensive`, so I don't think it needs to change.

source/vstd/seq.rs
```diff
@@ -1064,34 +1857,36 @@ pub broadcast proof fn lemma_seq_add_index2_alt<A>(s1: Seq<A>, s2: Seq<A>, i: in
     ensures
         (#[trigger] s1.add(s2))[i + s1.len()] == #[trigger] s2[i],
 {
-    broadcast use axiom_seq_add_index2;
+    broadcast use lemma_seq_add_index2;
 
 }
 
+// TODO(bsdinis): rename to group_seq_lemmas?
```
parno: I guess we ought to mark it deprecated and add a second group with the same members named `group_seq_lemmas`

parno (approved):
Thanks for working on this!  It's great to have the Seq library on a sounder footing.  I only had a few minor comments.

source/vir/src/modes.rs
bsdinis: changed it to `not allowed in spec mode`

Chris-Hawblitzel:
@bsdinis Some axioms got added that don't need to exist because they didn't exist before: `axiom_seq_add_index1_alt`, `axiom_seq_add_index2_alt`, `axiom_seq_two_subranges_index`, `axiom_seq_subrange_index_alt`, `axiom_seq_update_different_alt`, `axiom_seq_update_same_alt`, `axiom_seq_push_index_different_alt`

bsdinis:
> @bsdinis Some axioms got added that don't need to exist because they didn't exist before: `axiom_seq_add_index1_alt`, `axiom_seq_add_index2_alt`, `axiom_seq_two_subranges_index`, `axiom_seq_subrange_index_alt`, `axiom_seq_update_different_alt`, `axiom_seq_update_same_alt`, `axiom_seq_push_index_different_alt`

Good catch, I'll remove them in a later PR

Chris-Hawblitzel:
@bsdinis I'm also seeing this warning during `vargo build`:

```
verification results:: 2062 verified, 0 errors
warning: use of deprecated function `seq::group_seq_axioms`: Seq axioms have been verified, and are now lemmas
 --> vstd\std_specs\iter.rs:3:5
  |
3 |     group_seq_axioms, lemma_seq_empty, lemma_seq_subrange_index, lemma_seq_subrange_len,
  |     ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default

warning: 1 warning emitted
```

PR #2672, opened 2026-07-16
Rescue some help text with information about how to suppress warnings

parno:
In commit 457dca62be432, we added the ability to suppress warnings.  When a warning is displayed, it also displays help text telling you how to suppress it.  However, that didn't work for certain warnings that supplied their own help text, since that overwrote the suppression suggestion.  We fix that by using a help_append call instead.


Chris-Hawblitzel (approved):
Should we go a step further and make `help` a `Vec<String>` instead of an `Option<String>`?  Is there ever any need to overwrite `help` instead of appending to it?

parno:
> Should we go a step further and make help a Vec<String> instead of an Option<String>? Is there ever any need to overwrite help instead of appending to it?

That's a good point.  I'll double check that the proof_notes/custom_err attributes don't overwrite the help; assuming they don't I'll update this to use `Vec`.

parno:
> Should we go a step further and make help a Vec<String> instead of an Option<String>?

I implemented this, and it ended up cleaning up a few other minor warts, so definitely a net win.  Thanks for the suggestion!

PR #2677, opened 2026-07-17
Add a policy on AI-based contributions

parno:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,13 @@
 
 
+<!--
+Thanks for your contribution!
+
+Please note: 
+  - All code produced via agentic AI **must be disclosed in the PR**. 
+  - **The PR description, along with subsequent PR discussion comments, must be human-written.**
+
+See our [guide on Verus contributions](https://github.com/verus-lang/verus/blob/main/CONTRIBUTING.md) for more details.
+--> 
+
 <small>By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).</small>
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,7 @@
 Thanks for your contribution!
 
 Please note: 
-  - All code produced via agentic AI **must be disclosed in the PR**. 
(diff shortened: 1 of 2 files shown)


tjhance:
Another thing that occurred to me: should we have a policy regarding listing Claude as a "co-author" on commits?

Personally, I dislike the practice, as it seems to me to be against the spirit of taking responsibility for one's submission.

mmcloughlin:
> Another thing that occurred to me: should we have a policy regarding listing Claude as a "co-author" on commits?
> 
> Personally, I dislike the practice, as it seems to me to be against the spirit of taking responsibility for one's submission.

I also dislike `Co-authored-by` attribution for AI. I think the [Linux Kernel AI Coding Assistants Policy](https://docs.kernel.org/process/coding-assistants.html) makes more sense, recommending:

```
Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]
```

Example: https://github.com/torvalds/linux/commit/c88eb7e8d8397a8c1db59c425332c5a30b2a1682

For the kind of language we could adopt for Verus, see [Bytecode Alliance AI Tool Use Policy](https://github.com/bytecodealliance/governance/blob/2067ac933589ef1c4bb511f7306aa93a13a23e28/AI_TOOL_POLICY.md) which allows but does not require `Assisted-by`, and disallows `Co-authored-by`:

> As a consequence of our position that the contributor proposing changes is fully accountable for their contributions, contributors aren't required to explicitly mention the use of any tools, AI or otherwise. If a contributor thinks there is value in highlighting tool use, they *may* do so in their pull request description, commit message, or using a commit message trailer like `Assisted-by: <tool>:<model>`. They **MUST NOT**, however, list any tool, AI or otherwise, as a co-author: only human contributors can be authors and take on the accountability that comes with it.

tjhance:
I agree that 'Assisted-by' is much better than 'Co-authored-by'

parno:
I added some text to both files indicating that AI shouldn't be a co-author by can instead be cited via "Assisted-by".

Any further thoughts before we merge this?

tjhance (approved):
looks fine to me.

by the way, the rust team recently published their own code of conduct for LLMs, in case you want to compare: https://blog.rust-lang.org/inside-rust/2026/08/05/rust-langrust-is-adopting-an-llm-policy/

Chris-Hawblitzel (approved):

PR #2680, opened 2026-07-18
Delete stale builtin/builtin_macros crate in cargo-verus example

Marti2203:
The example/cargo-verus Cargo.toml files still referenced the pre-rename crate names (builtin, builtin_macros) as dependency path targets. The actual package names in `source/builtin/Cargo.toml` and `source/builtin_macros/Cargo.toml` are `verus_builtin/verus_builtin_macros`, so following this example verbatim fails immediately - before any compilation or verification is even attempted. `cargo verus check` (or `verify`, or plain `cargo build`) first shells out to `cargo metadata` to plan the build, and that step itself errors out:
```
  Error: `cargo metadata` exited with an error: error: no matching package named `builtin` found
  location searched: .../source/builtin
  required by package `test v0.1.0 (.../examples/cargo-verus/test)`
```
The rename occurred in 2496084da ("Prepare Verus-related crates for posting to crates.io", #1815, 2025-07-30) - packages were prefixed with `verus_` to use the crates.io namespace responsibly. The example itself was last touched two months earlier (0135ab541, a path fixup) and has not been updated since.

examples/cargo-verus isn't exercised by anything automated: it's not a member of source/Cargo.toml's workspace, vargo never builds it, and CI's cargo-verus-test job only runs `cargo test -p cargo-verus` (the tool's own unit tests). It's a downstream-consumer demo for someone using an already-built verus via the separate cargo-verus integration (docs/CARGO-VERUS.md) - a different, untested pathway from vargo/CI.

Fixed with Cargo's dependency-renaming syntax so the local names used throughout the example (builtin, builtin_macros) still work, now pointing at the real package names. Confirmed working end-to-end, with the identical command that failed above: `cargo verus check --manifest-path test/Cargo.toml -p test` now verifies both the library and test crates successfully (1 verified, 0 errors each), after vstd itself verifies clean.


parno:
@Chris-Hawblitzel @mkovaxx Chris, I believe you first introduced this example.  Now that we have a more thorough set of tests and some documentation for `cargo verus`, do we still need this?  Mate, what do you think?

Chris-Hawblitzel:
> Chris, I believe you first introduced this example. Now that we have a more thorough set of tests and some documentation for `cargo verus`, do we still need this?

It's fine with me if we delete it.

Marti2203:
Updated the PR as discussed

parno (approved):
Thanks for cleaning this up!

PR #2681, opened 2026-07-18
Reject ident subpatterns in function parameter position instead of panicking

### LINKED ISSUE #1933: Ident subpatterns in argument position crash Verus
I found a funny way to make Verus panic today:
```rs
fn double(a @ b: i32) -> i32 { a + b }
```
This function compiles perfectly fine in Rust, it takes one argument, binds it to two different variable names, and adds one to the other, so this function simply doubles its input.

Verus does not support pattern matching in argument position, but also doesn't reject this case properly, so the error message we get is as follows:
```
thread 'rustc' panicked at vir/src/modes.rs:349:55:
internal error: missing mode
```

Marti2203:
Fixes #1933. Verus doesn't support pattern matching in argument position; e.g., the function `fn double(a @ b: i32) -> i32 { a + b }` wasn't rejected properly: pat_to_mut_var only ever returns the outer name (a) for a PatKind::Binding, silently discarding any subpattern, so `b` never got registered as a parameter at all. The first time the function body referenced `b`, mode-checking hit a `let-else` on `self.vars.get(x)` and panicked with `"internal error: missing mode"`, since `b` was never in the map.

Fixed at the one call site that processes function parameters specifically (`rust_to_vir_func.rs`) rather than in `pat_to_mut_var` itself - that helper is also used for match/let-else patterns, where ident subpatterns are legitimately supported and already handled correctly elsewhere (confirmed by running `mut_refs_patterns.rs`'s full suite, which covers several `x @ pattern` match/let-else cases, both before and after this change).

Verified: the repro now gives a clear error instead of panicking; plain and `mut` parameters are unaffected; mut_refs_patterns.rs (72 tests), exec_closures.rs, closures.rs, proof_closures.rs, mut_refs_closures.rs, and mutable_params.rs all still pass. Added a regression test to mutable_params.rs covering the exact repro from the issue.

I would like to check with you whether it makes sense to make this a supported behavior, i.e., generating the corresponding VIR code and making it pass? I think it would be useful and can prepare a subsequent PR.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -1653,6 +1653,15 @@ pub(crate) fn check_item_fn<'tcx>(
             let Body { params, value: _ } = body;
             let mut ps = Vec::new();
             for Param { hir_id, pat, ty_span: _, span } in params.iter() {
+                // Check the parameter's pattern is a plain identifier (e.g., `x: T`).
+                // VIR expects each param to be associated with exactly one identifier so complex params
+                // (e.g., `(x, y): T` or `_: T` or `a @ b: T`) are unsupported
+                if !matches!(pat.kind, rustc_hir::PatKind::Binding(_, _, _, None)) {
+                    return err_span(
+                        *span,
+                        "function parameters must be a plain identifier pattern (e.g. `x: T` or `mut x: T`)",
+                    );
+                }
                 let (is_mut_var, name) = pat_to_mut_var(pat)?;
                 // is_mut_var: means a parameter is like `mut x: X`
                 // is_mut: means a parameter is like `x: &mut X` or `x: Tracked<&mut X>`
--- a/source/rust_verify_test/tests/mutable_params.rs
+++ b/source/rust_verify_test/tests/mutable_params.rs
@@ -587,3 +587,41 @@ test_verify_one_file_with_options! {
         }
     } => Err(err) => assert_fails(err, 1)
 }
+
+// https://github.com/verus-lang/verus/issues/1933
+// An ident subpattern (`a @ b: T`) in parameter position used to silently
+// register only the outer name (`a`), leaving `b` unregistered - referencing
+// `b` in the body then panicked in mode-checking instead of being rejected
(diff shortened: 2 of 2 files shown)


source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -1653,6 +1653,21 @@ pub(crate) fn check_item_fn<'tcx>(
...
+                // An `@` binding pattern (`name @ subpattern`) binds `name` to
+                // the whole value, but `pat_to_mut_var` below only ever
+                // registers `name` - any names bound inside `subpattern` (be
+                // it as trivial as another identifier, e.g. `a @ b: T`, or a
+                // real destructure, e.g. `whole @ (a, b): (T, U)`) would
+                // silently never get registered, then panic later in
+                // mode-checking the first time the body referenced them.
+                // Rejected generically here regardless of what `subpattern`
+                // itself looks like.
+                if let rustc_hir::PatKind::Binding(_, _, _, Some(_)) = pat.kind {
```
tjhance: Since nontrival patterns of any kind are unsupported, I would simply do:

```rust
if !matches!(path.kind, rustc_hir::PatKind::Binding(_, _, _, None)) { /* error */ }
```

Also, while I ordinarily appreciate code comments, this one is fairly long-winded for a fairly straightforward and self-explanatory check. I would just write:

```
// Check the parameter's pattern is a plain identifier (e.g., `x: T`).
// VIR expects each param to be associated with exactly one identifier so complex params
// (e.g., `(x, y): T` or `_: T` or `a @ b: T`) are unsupported
```

tjhance (changes requested):

source/rust_verify/src/rust_to_vir_func.rs
Marti2203: Hi! Will fix it up and reduce the comments

Marti2203: May I check why wildcard or tuple patterns would be rejected? Is this a design choice or missing functionality?

tjhance (approved):

PR #2682, opened 2026-07-18
Give loop-invariant/decreases checks a real AssertId instead of None

Marti2203:
Loop-invariant preservation and decrease-clause checks (INV_FAIL_LOOP_END/INV_FAIL_LOOP_FRONT/DEC_FAIL_LOOP_*) are synthesized directly in `sst_to_air.rs` during `SST->AIR` lowering - unlike an ordinary user assert/requires/ensures, there's no pre-existing SST-level `StmX::Assert` node (with its own `ast_to_sst.rs`-assigned `AssertId`) to forward for these. Every such site hardcoded `StmtX::Assert(None, ...)`, meaning `default_prover_failed_assert_ids` (in `rust_verify::verifier`) was never populated for a function whose *only* failures were loop-invariant or decrease checks - completely invisible to anything keyed off AssertId.

Today, the only in-tree consumer of that id is `--expand-errors`, and its behavior is unchanged by this PR either way (see below) - so this isn't fixing a user-visible bug on its own. The value is foundational: a real id here is a prerequisite for `--expand-errors` (or any other tooling built around per-assertion ids) to ever support this failure class - currently impossible since there's nothing to key off.

Fixed by adding State::next_synthetic_assert_id, minting a real, structurally distinct AssertId (a two-element vec![u64::MAX, n], vs. the always-one-element `vec![n]` `ast_to_sst.rs` assigns) at the 5 call sites that previously hardcoded None. The two-element shape guarantees no collision with any real id without needing any cross-module counter coordination.

This surfaced one real regression, fixed in the same commit: --expand-errors (expand_errors_driver.rs) hardcodes `assert_id.len() == 1` and crashed the whole process (`assertion failed: assert_id.len() == 1`) whenever a function's first failing assertion was one of these newly-id'd loop-invariant/decreases checks, since it was never designed to receive anything but a real, one-element id. Fixed by exposing a small `vir::sst_to_air::is_synthetic_assert_id` predicate and skipping the `--expand-errors` call entirely when the first failing assertion is synthetic - exactly matching that assertion's pre-fix behavior (it previously had no id at all, so this code path was never reached for it). Not attempting to teach `--expand-errors` to actually support these ids - real follow-up work if wanted, out of scope here.

Verified: a minimal loop-invariant-only failure now sees a real, non-empty default_prover_failed_assert_ids where it previously saw none; `--expand-errors` no longer crashes on this case and still produces its full expansion diagnostic for ordinary (non-synthetic) failures, unaffected. Full expand_errors.rs (16, +1 new regression test), triggers.rs (37), mut_refs_patterns.rs (72), mutable_params.rs (11), loops.rs (75, 3 ignored), loops_havoc.rs (11), loops_no_spinoff.rs (60), and mut_refs_loops.rs (29) test suites all pass; vstd still fully verified.

I understand that this PR may introduce a new identifier structure, and I may not be familiar with the conventions. I am happy to change the identifier structure if needed.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1784,6 +1784,9 @@ impl Verifier {
                             if any_invalid
                                 && self.args.expand_errors
                                 && default_prover_failed_assert_ids.len() > 0
+                                && !vir::sst_to_air::is_synthetic_assert_id(
+                                    &default_prover_failed_assert_ids[0],
+                                )
                             {
                                 function_opgen.start_expand_errors_if_possible(
                                     &op,
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1784,8 +1784,9 @@ impl Verifier {
                             if any_invalid
                                 && self.args.expand_errors
                                 && default_prover_failed_assert_ids.len() > 0
-                                && !vir::sst_to_air::is_synthetic_assert_id(
+                                && vir::sst_to_air::has_assert_stm_node(
                                     &default_prover_failed_assert_ids[0],
+                                    &func_check_sst.as_ref().and_then(|f| f.last_minted_id.clone()),
                                 )
                             {
                                 function_opgen.start_expand_errors_if_possible(
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -1773,6 +1773,9 @@ impl Verifier {
                             if any_invalid
                                 && self.args.expand_errors
                                 && default_prover_failed_assert_ids.len() > 0
+                                && !vir::sst_to_air::is_synthetic_assert_id(
+                                    &default_prover_failed_assert_ids[0],
+                                )
                             {
                                 function_opgen.start_expand_errors_if_possible(
                                     &op,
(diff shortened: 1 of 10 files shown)


tjhance:
This doesn't seem to solve a real problem to me. Do you have in mind a specific use-case for AssertIds beyond expand-errors? If no, I don't think there's any reason to make this change. If yes, we can probably come up with something more natural than the synthetic AssertId encoding.

Marti2203:
Hi @tjhance, it doesn't solve a user-facing issue. The change is internal to Verus. I am experimenting with automated fault localization for Verus verification failures — analyzing why a specific assertion failed, its trace and dependencies, and extracting a concrete counterexample.

To do that, given a target assertion, I isolate just that assertion's own query out of the function's full verification query using `air::focus::focus_commands_on_assert_id` — the same AIR-level mechanism `--expand-errors` itself already relies on for the same kind of isolation. That mechanism works purely by `AssertId` equality (walking the `Stmt` tree, keeping only the branch/prefix leading to a matching Assert node, dropping everything else) — so it requires every assertion that could be a target to actually carry one.

Would be happy to discuss further if needed.

tjhance:
In that case, I think we should just give all asserts a normal (length 1) AssertId.

Marti2203:
Sure, will update the PR in a couple of hours and make them have normal identifiers. Thanks for the feedback!

Marti2203:
Hi @tjhance , any update on this PR? It's not blocking anything, just wanted to check

tjhance:
i will get to it soon (i am currently in the middle of a move so very occupied)

Marti2203:
Ah, no problem! :) Good luck, I know the feeling!

PR #2683, opened 2026-07-18
vstd: relate char::len_utf8 to encode_utf8, for byte-offset/char-index reasoning

Marti2203:
vstd already models per-scalar UTF-8 byte width via `encode_scalar`/`encode_utf8` (used to define `str::spec_bytes()`/`as_bytes()`'s own trusted spec), but exposed no way to relate that model to the real, callable `char::len_utf8()`. Without that link, downstream code that needs to relate a UTF-8 byte offset into a string (e.g. from `str::match_indices/char_indices`) to the character index it corresponds to, or vice versa, had no way to *prove* that conversion correct - it could only trust an unverified implementation.

Adds:

- `char::len_utf8`'s own assume_specification (returns encode_scalar(c as u32).len() as usize) - justified since Rust's real implementation uses the exact same scalar-value boundaries has_width_N_encoding already defines, and a char can never be a surrogate, so encode_scalar's surrogate exclusion never actually applies here. No looser a trust than str::as_bytes's existing spec.
- `lemma_encode_utf8_len_additive`: encode_utf8 distributes over Seq concatenation.
- `lemma_encode_utf8_push_len`: the common-case specialization for appending one char, tying directly to encode_scalar's width.
- `lemma_encode_utf8_len_monotonic` / `lemma_encode_utf8_len_pos` / `lemma_encode_utf8_len_strictly_monotonic`: a growing prefix's encoded byte length never decreases, is always positive for a nonempty prefix, and strictly increases when the prefix grows by at least one character - lets a caller conclude an ordering (or rule out ambiguity) between character indices from an ordering between the byte offsets they produced, and vice versa.

Together, these let downstream code prove a per-char byte-offset invariant via a running sum of `char::len_utf8()` calls (tied back to `encode_utf8(prefix).len()` at each step), rather than trusting an unverified conversion or reasoning about the whole-string encode_utf8 at once. vstd: 2062 -> 2067 verified, 0 errors.

I am happy to close this PR if the lemmas are trivial or not generally useful.

### DIFF
--- a/source/vstd/utf8.rs
+++ b/source/vstd/utf8.rs
@@ -463,7 +463,6 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
 // See `vstd::std_specs::char` for the `char::len_utf8`/`char::is_whitespace`
 // `assume_specification`s that relate this module's model to those real
 // std-library methods.
-
 /// [`encode_utf8`] distributes over sequence concatenation.
 pub broadcast proof fn encode_utf8_concat(a: Seq<char>, b: Seq<char>)
     ensures
--- a/source/vstd/utf8.rs
+++ b/source/vstd/utf8.rs
@@ -482,7 +482,8 @@ pub open spec fn is_white_space(c: char) -> bool {
 }
 
 pub assume_specification[ char::is_whitespace ](c: char) -> (res: bool)
-    returns is_white_space(c),
+    returns
+        is_white_space(c),
 ;
 
 /// [`encode_utf8`] distributes over sequence concatenation.
--- a/source/vstd/utf8.rs
+++ b/source/vstd/utf8.rs
@@ -471,34 +471,45 @@ pub assume_specification[ char::len_utf8 ](c: char) -> usize
         encode_scalar(c as u32).len() as usize,
 ;
 
-/// For an ASCII `c`, `is_whitespace()` is exactly the 6-character C0 set
-/// (confirmed by exhaustively checking all 128 ASCII code points against
-/// real `rustc`) - Unicode's full `White_Space` property, which also covers
-/// non-ASCII code points, isn't modeled here, so `res` is unconstrained
-/// outside the ASCII case.
+/// Unicode's `White_Space` property:
+/// <https://www.unicode.org/reports/tr44/#White_Space>.
+pub open spec fn is_white_space(c: char) -> bool {
+    c == '\u{9}' || c == '\u{A}' || c == '\u{B}' || c == '\u{C}' || c == '\u{D}' || c == '\u{20}'
+        || c == '\u{85}' || c == '\u{A0}' || c == '\u{1680}' || c == '\u{2000}' || c == '\u{2001}'
+        || c == '\u{2002}' || c == '\u{2003}' || c == '\u{2004}' || c == '\u{2005}' || c
+        == '\u{2006}' || c == '\u{2007}' || c == '\u{2008}' || c == '\u{2009}' || c == '\u{200A}'
+        || c == '\u{2028}' || c == '\u{2029}' || c == '\u{202F}' || c == '\u{205F}' || c
+        == '\u{3000}'
+}
+
 pub assume_specification[ char::is_whitespace ](c: char) -> (res: bool)
     ensures
-        c as u32 <= 0x7F ==> res == (c == ' ' || c == '\t' || c == '\n' || c == '\x0B' || c
-            == '\x0C' || c == '\r'),
+        res == is_white_space(c),
 ;
 
 /// [`encode_utf8`] distributes over sequence concatenation - lets a per-`char`
-/// byte offset be computed as a running sum instead of the whole string at once.
-pub proof fn lemma_encode_utf8_len_additive(a: Seq<char>, b: Seq<char>)
+/// byte offset be computed as a running sum instead of the whole string at
+/// once. Stated as full sequence equality (not just matching lengths), since
+/// the length fact then follows directly from `Seq`'s own additive-length
+/// property (`(a + b).len() == a.len() + b.len()`) at any call site that
+/// only needs it, with no separate length-specific lemma required.
+pub proof fn lemma_encode_utf8_concat(a: Seq<char>, b: Seq<char>)
     ensures
-        encode_utf8(a + b).len() == encode_utf8(a).len() + encode_utf8(b).len(),
+        encode_utf8(a + b) == encode_utf8(a) + encode_utf8(b),
     decreases a.len(),
 {
     if a.len() == 0 {
         assert(a + b =~= b);
     } else {
         assert((a + b).drop_first() =~= a.drop_first() + b);
-        lemma_encode_utf8_len_additive(a.drop_first(), b);
+        lemma_encode_utf8_concat(a.drop_first(), b);
+        assert(encode_scalar(a[0] as u32) + (encode_utf8(a.drop_first()) + encode_utf8(b)) =~= (
+        encode_scalar(a[0] as u32) + encode_utf8(a.drop_first())) + encode_utf8(b));
     }
 }
 
-/// Specialization of [`lemma_encode_utf8_len_additive`] for appending one
-/// `char` - ties `encode_utf8(chars.push(c))`'s length to `c`'s own width.
+/// Specialization of [`lemma_encode_utf8_concat`] for appending one `char` -
+/// ties `encode_utf8(chars.push(c))`'s length to `c`'s own width.
 pub proof fn lemma_encode_utf8_push_len(chars: Seq<char>, c: char)
     ensures
         encode_utf8(chars.push(c)).len() == encode_utf8(chars).len() + encode_scalar(
@@ -506,7 +517,7 @@ pub proof fn lemma_encode_utf8_push_len(chars: Seq<char>, c: char)
         ).len(),
 {
     assert(chars.push(c) =~= chars + seq![c]);
-    lemma_encode_utf8_len_additive(chars, seq![c]);
+    lemma_encode_utf8_concat(chars, seq![c]);
     assert(seq![c].drop_first() =~= Seq::<char>::empty());
     assert(encode_utf8(Seq::<char>::empty()) =~= Seq::<u8>::empty());
     assert(encode_utf8(seq![c]) =~= encode_scalar(c as u32) + encode_utf8(Seq::<char>::empty()));
@@ -521,7 +532,7 @@ pub proof fn lemma_encode_utf8_len_monotonic(s: Seq<char>, i: int, j: int)
         encode_utf8(s.subrange(0, i)).len() <= encode_utf8(s.subrange(0, j)).len(),
 {
     assert(s.subrange(0, i) + s.subrange(i, j) =~= s.subrange(0, j));
-    lemma_encode_utf8_len_additive(s.subrange(0, i), s.subrange(i, j));
+    lemma_encode_utf8_concat(s.subrange(0, i), s.subrange(i, j));
 }
 
 /// A nonempty `char` sequence always encodes to at least one byte.
@@ -542,7 +553,7 @@ pub proof fn lemma_encode_utf8_len_strictly_monotonic(s: Seq<char>, i: int, j: i
         encode_utf8(s.subrange(0, i)).len() < encode_utf8(s.subrange(0, j)).len(),
 {
     assert(s.subrange(0, i) + s.subrange(i, j) =~= s.subrange(0, j));
-    lemma_encode_utf8_len_additive(s.subrange(0, i), s.subrange(i, j));
+    lemma_encode_utf8_concat(s.subrange(0, i), s.subrange(i, j));
     assert(s.subrange(i, j).len() == j - i);
     lemma_encode_utf8_len_pos(s.subrange(i, j));
 }
--- a/source/vstd/utf8.rs
+++ b/source/vstd/utf8.rs
@@ -460,6 +460,84 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
     }
 }
 
+/* Relating encode_utf8 to the real char::len_utf8, for byte-offset/char-index reasoning */
+
+/// Matches Rust's real `char::len_utf8()` width to this module's own model:
+/// same scalar boundaries as [`encode_scalar`], and a `char` can never be a
+/// surrogate, so `encode_scalar`'s surrogate exclusion never applies here.
+#[verifier::allow_in_spec]
+pub assume_specification[ char::len_utf8 ](c: char) -> usize
+    returns
+        encode_scalar(c as u32).len() as usize,
+;
+
+/// [`encode_utf8`] distributes over sequence concatenation - lets a per-`char`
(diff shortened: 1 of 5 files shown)


source/vstd/utf8.rs
```diff
@@ -460,6 +460,95 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+pub assume_specification[ char::len_utf8 ](c: char) -> usize
+    returns
+        encode_scalar(c as u32).len() as usize,
+;
+
+/// For an ASCII `c`, `is_whitespace()` is exactly the 6-character C0 set
+/// (confirmed by exhaustively checking all 128 ASCII code points against
+/// real `rustc`) - Unicode's full `White_Space` property, which also covers
+/// non-ASCII code points, isn't modeled here, so `res` is unconstrained
+/// outside the ASCII case.
```
tjhance: I wasn't familiar with the term "C0" but I looked it up and it seems to correspond to all codes in the range `00-1F`? That's not a "6 character" set.

It would be best to just say it implements the White_Space property, and that our spec only applies to ASCII characters, and ideally link to something in the unicode standard that defines White_Space.

Finally, information about testing should be left to internal comments (`//`). Speaking of testing, there should also be a test in rust_verify_tests.

source/vstd/utf8.rs
```diff
@@ -460,6 +460,95 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+    ensures
+        c as u32 <= 0x7F ==> res == (c == ' ' || c == '\t' || c == '\n' || c == '\x0B'
+            || c == '\x0C' || c == '\r'),
+;
+
+/// [`encode_utf8`] distributes over sequence concatenation - lets a per-`char`
+/// byte offset be computed as a running sum instead of the whole string at once.
+pub proof fn lemma_encode_utf8_len_additive(a: Seq<char>, b: Seq<char>)
+    ensures
+        encode_utf8(a + b).len() == encode_utf8(a).len() + encode_utf8(b).len(),
```
tjhance: It might make more sense to just have this be:

```
encode_utf8(a + b) == encode_utf8(a) + encode_utf8(b)
```

as the desired properties on `len` will just follow immediately given the existing sequence axioms.

tjhance (changes requested):

source/vstd/utf8.rs
Marti2203: I have enhanced it to cover the proper whitespace definitions according to the UTF-8 standard

source/vstd/utf8.rs
Marti2203: Updated, thanks for the feedback!

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
     }
 }
 
+/* Relating encode_utf8 to the real char::len_utf8, for byte-offset/char-index reasoning */
```
zero-to-nat: Nitpick: what does "real" mean here? I agree that having a specification for this function is important, but seems like whether `char::len_utf8` or `str::as_bytes` (or both) are needed depends on the context.

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+
+/// Matches Rust's real `char::len_utf8()` width to this module's own model:
```
zero-to-nat: Same as above

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+/// same scalar boundaries as [`encode_scalar`], and a `char` can never be a
+/// surrogate, so `encode_scalar`'s surrogate exclusion never applies here.
+#[verifier::allow_in_spec]
+pub assume_specification[ char::len_utf8 ](c: char) -> usize
```
zero-to-nat: Question for @parno or @tjhance: is there a desired placement within vstd for `assume_specification`? For example, should this go in `vstd/string.rs` with the existing `string` and `String` trusted specs? Or a new file, like `std_specs/char.rs`? (Currently, our utf8 module does not define any trusted specs)

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+    c == '\u{9}' || c == '\u{A}' || c == '\u{B}' || c == '\u{C}' || c == '\u{D}' || c == '\u{20}'
+        || c == '\u{85}' || c == '\u{A0}' || c == '\u{1680}' || c == '\u{2000}' || c == '\u{2001}'
+        || c == '\u{2002}' || c == '\u{2003}' || c == '\u{2004}' || c == '\u{2005}' || c
+        == '\u{2006}' || c == '\u{2007}' || c == '\u{2008}' || c == '\u{2009}' || c == '\u{200A}'
+        || c == '\u{2028}' || c == '\u{2029}' || c == '\u{202F}' || c == '\u{205F}' || c
+        == '\u{3000}'
+}
+
+pub assume_specification[ char::is_whitespace ](c: char) -> (res: bool)
+    ensures
```
zero-to-nat, line 468: Nitpick: could use `returns` clause instead of `ensures`

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+        res == is_white_space(c),
+;
+
+/// [`encode_utf8`] distributes over sequence concatenation - lets a per-`char`
```
zero-to-nat: I think that the first part of this documentation comment matches the style used in the rest of the file, but it seems like the second part starting with "lets a per-`char`" is mostly relevant to a single use casesof this lemma. Since this comment will show up on the public API documentation [here](https://verus-lang.github.io/verus/verusdoc/vstd/utf8/index.html) and could be used in many ways, I would be inclined to remove the second part.

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+        assert((a + b).drop_first() =~= a.drop_first() + b);
+        lemma_encode_utf8_concat(a.drop_first(), b);
+        assert(encode_scalar(a[0] as u32) + (encode_utf8(a.drop_first()) + encode_utf8(b)) =~= (
+        encode_scalar(a[0] as u32) + encode_utf8(a.drop_first())) + encode_utf8(b));
+    }
+}
+
+/// Specialization of [`lemma_encode_utf8_concat`] for appending one `char` -
+/// ties `encode_utf8(chars.push(c))`'s length to `c`'s own width.
+pub proof fn lemma_encode_utf8_push_len(chars: Seq<char>, c: char)
```
zero-to-nat: Should this lemma be generalized to `encode_utf8(chars.push(c)) == encode_utf8(chars) + encode_scalar(c as u32)`, analogously to the above lemma?

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+    ensures
+        encode_utf8(chars.push(c)).len() == encode_utf8(chars).len() + encode_scalar(
+            c as u32,
+        ).len(),
+{
+    assert(chars.push(c) =~= chars + seq![c]);
+    lemma_encode_utf8_concat(chars, seq![c]);
+    assert(seq![c].drop_first() =~= Seq::<char>::empty());
+    assert(encode_utf8(Seq::<char>::empty()) =~= Seq::<u8>::empty());
```
zero-to-nat: Are all of these `assert` statements necessary?

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+        0 <= i <= j <= s.len(),
+    ensures
+        encode_utf8(s.subrange(0, i)).len() <= encode_utf8(s.subrange(0, j)).len(),
+{
+    assert(s.subrange(0, i) + s.subrange(i, j) =~= s.subrange(0, j));
+    lemma_encode_utf8_concat(s.subrange(0, i), s.subrange(i, j));
+}
+
+/// A nonempty `char` sequence always encodes to at least one byte.
+pub proof fn lemma_encode_utf8_len_pos(s: Seq<char>)
```
zero-to-nat: If this lemma is trivial to prove, is it necessary to invoke anywhere?

Marti2203:
Hi @zero-to-nat , thanks for all the feedback, answering just to keep the flow cleaner. Will respond to all the questions and handle the feedback. Thanks!

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+    assert(chars.push(c) =~= chars + seq![c]);
+    lemma_encode_utf8_concat(chars, seq![c]);
+    assert(seq![c].drop_first() =~= Seq::<char>::empty());
+    assert(encode_utf8(Seq::<char>::empty()) =~= Seq::<u8>::empty());
+    assert(encode_utf8(seq![c]) =~= encode_scalar(c as u32) + encode_utf8(Seq::<char>::empty()));
+    assert(encode_utf8(seq![c]).len() == encode_scalar(c as u32).len());
+}
+
+/// The `encode_utf8` byte length of a growing prefix never decreases.
+pub proof fn lemma_encode_utf8_len_monotonic(s: Seq<char>, i: int, j: int)
```
zero-to-nat: I think this lemma may not be necessary if we also have `lemma_encode_utf8_len_strictly_monotonic`. In the case where `i = j`, the result `encode_utf8(s.subrange(0, i)).len() == encode_utf8(s.subrange(0, j)).len()` should follow trivially from `s.subrange(0, i) == s.subrange(0, j)`.

source/vstd/utf8.rs
```diff
@@ -460,6 +460,104 @@ pub open spec fn encode_utf8(chars: Seq<char>) -> Seq<u8>
...
+        res == is_white_space(c),
+;
+
+/// [`encode_utf8`] distributes over sequence concatenation - lets a per-`char`
+/// byte offset be computed as a running sum instead of the whole string at
+/// once. Stated as full sequence equality (not just matching lengths), since
+/// the length fact then follows directly from `Seq`'s own additive-length
+/// property (`(a + b).len() == a.len() + b.len()`) at any call site that
+/// only needs it, with no separate length-specific lemma required.
+pub proof fn lemma_encode_utf8_concat(a: Seq<char>, b: Seq<char>)
```
zero-to-nat: Would it also be possible to make these proofs `broadcast` with appropriate triggers, and add them to `group_utf8_lib`?

zero-to-nat (commented):
Looks good to me, pending my feedback. Thank you for adding these additional specs and lemmas!

I saw that @tjhance requested an additional review from @parno, but I believe that Bryan is on vacation for the next two weeks. Please let me know if I can help with any additional review

source/vstd/utf8.rs
tjhance: Good question. Typically assume_specification goes in std_specs. It usually doesn't matter so much though, since the user never actually uses the path.

tjhance:
Oh, I added bryan without realizing he had already added @zero-to-nat. Your review should be sufficient

source/vstd/utf8.rs
tjhance: I agree; stylistically this doc comment reads strangely. Typically there's no need to reiterate "this is the spec for function X" in the spec for function X; that's what `assume_specification` is _always_ for.

tjhance: I don't think it's necessary to say anything about surrogates, either, that's just a characteristic of the encoding and true of all chars, not more or less important to this function than any other aspect of the encoding.

source/vstd/utf8.rs
Marti2203: In this case "real" was about the standard library. Have updated the comment!

source/vstd/utf8.rs
Marti2203: Cleaned it up! Thanks!

source/vstd/utf8.rs
Marti2203: Updated

source/vstd/utf8.rs
Marti2203: Reworded

source/vstd/utf8.rs
Marti2203: Thanks for the catch! Fixed!

source/vstd/utf8.rs
Marti2203: Dropped most of them

source/vstd/utf8.rs
Marti2203: Removed

source/vstd/utf8.rs
Marti2203: Removed

source/vstd/utf8.rs
Marti2203: I have added `broadcast` to the proofs but have shied away from adding it to `lemma_encode_utf8_len_strictly_monotonic` as I do not think that makin such a property broadcasted immediately useful.

source/vstd/utf8.rs
Marti2203: I haven't done anything about this issue. Personally, I think having a `char.rs` would be better for separating the character logic, but I will await your decision.

zero-to-nat: Okay, makes sense! In that case, to align with the organization with most of vstd, would you be able to create a new `std_specs/char.rs` to add these specifications?

zero-to-nat (approved):
All of the comments have been resolved, except for a pending reorganization on trusted specs for char, so I will mark this as approved. Thanks so much!

source/vstd/utf8.rs
Marti2203: Yes, created! Also added some more tests to be complete. Will let you resolve this comment and merge. Thanks!

zero-to-nat: Awesome, thank you!!

Marti2203:
my bad @zero-to-nat i rebased and that stopped the automerge...

zero-to-nat:
Hi @tjhance, would you be able to approve your requested changes? It seems this is blocking the PR from merging.

tjhance (approved):

wood-ghost:
Hi @Marti2203 , I notice that spec for `char::len_utf8` uses `#[verifier::allow_in_spec]` but this property is not tested in the test file.
In `option::expect` they use 
```
assert(ok_option2.expect("option was built with Some") == 1);
```
Something like `assert('a'.len_utf8() ==1` would be better from my point of view.
Also I'm wondering whether it will be helpful to also allow `is_whitespace` in spec.

Marti2203:
Thanks for catching this @wood-ghost  — you're right: no test exercised allow_in_spec directly.  Will open a PR adding that coverage, plus `allow_in_spec` on `is_whitespace too`, per your suggestion.

PR #2686, opened 2026-07-20
build: use Cargo instead of Vargo

mkovaxx:
## Summary

The Verus Toolchain's executables, libraries, and (verified) `vstd` can now all be built without `vargo` as follows:
```sh
cd source
cargo clean
cargo build
cargo run -p cargo-verus -- build --manifest-path vstd/Cargo.toml
touch target-verus/debug/verus-root
```

## Details

- Add Cargo workspace configuration to reproduce the Vargo build env and place artifacts in `target-verus`.
- Extend `cargo-verus build` to detect a `vstd` build and hoist its required artifacts into the output directory.
- Generate `rust_verify`'s version, revision, profile, and platform metadata directly in a `build.rs` file.
- Migrate the primary CI, documentation, and feature-build workflows from Vargo to Cargo and `cargo-verus`.
- Isolate Cargo and `cargo-verus` invocations in tests using a temp dir as working directory.
- Isolate `cargo build` invocations in `bump_crate_versions` using a temp dir as working directory.


.github/workflows/ci-vargo.yml
```diff
@@ -0,0 +1,341 @@
+name: ci
```
mkovaxx: This file is a copy of the original `ci.yml` workflow, and it's only here to show that Vargo still works. It should be removed before merging the PR.

source/tools/line_count/.cargo/config.toml
```diff
@@ -1,5 +1,8 @@
 [build]
-rustflags = "--cfg proc_macro_span --cfg span_locations"
```
mkovaxx, line 2: Needed so that Cargo can merge the settings of this file with `source/.cargo/config.toml`.

source/tools/bump_crate_versions/src/main.rs
```diff
@@ -272,12 +272,17 @@ fn update_workspace_dependencies(manifest: &Path, dependencies: &Vec<&Crate>) {
...
     let mut cmd = Command::new("cargo");
     cmd.arg("publish");
+    cmd.arg("--manifest-path").arg(dir.join("Cargo.toml"));
     if dry_run {
         cmd.arg("--dry-run");
     }
-    let status = cmd.current_dir(dir).status().expect("Failed to execute cargo publish");
+    let temp_dir = tempfile::tempdir().expect("create fresh workdir for cargo publish");
+    let status =
+        cmd.current_dir(temp_dir.path()).status().expect("Failed to execute cargo publish");
```
mkovaxx, line 285: Ensures that the Cargo builds during `bump_crate_versions publish` aren't affected by the settings in `source/.cargo/config.toml`.

source/rust_verify/build.rs
```diff
@@ -0,0 +1,30 @@
+use std::process::Command;
+
+use cargo_verus_toolchains::versions::get_verus_version;
+
+fn main() {
+    let (default_version, default_sha) = get_verus_version(true).expect("version info");
```
mkovaxx, line 6: Inject env vars that are otherwise set by Vargo and are required by the build.

source/tools/bump_crate_versions/src/main.rs
```diff
@@ -272,12 +272,17 @@ fn update_workspace_dependencies(manifest: &Path, dependencies: &Vec<&Crate>) {
 fn publish(dir: &Path, dry_run: bool) {
```
mkovaxx, line 272: @parno Please review these changes from the context of the `crate-updates.yml` flow.

mkovaxx:
@parno @Chris-Hawblitzel @bsdinis @dschoepe The updated CI flows seem stable over a few days of merging changes from `main` on a daily basis. The design doc for more context is here: [LINK](https://docs.google.com/document/d/11qJW9bkZCRfaZ4Zp_mahTix3D-yZA4hGsBtAvHx3Ob0/view).

## Rollout Plan

- **[this PR] Phase 0: Prove feasibility by changing the CI flow to use Cargo-only tooling.**
- Phase 1: Use Cargo-only tooling in the Rolling Release and Release flows.
- Phase 2: Ask the wider Verus dev community to help stress test Cargo-only tooling.
- --- DECISION POINT ---
- Phase 3: Extend Vargo to emit deprecation warnings with a concrete date.
- Phase 4: Retire Vargo, deleting it from the repo.

## Additional Notes

- The `ci-vargo.yml` flow is meant to be temporary and removed before the PR is merged. Its purpose is to show that this PR doesn't break existing `vargo` functionality needed to build and release Verus.
- This new Cargo-based build flow currently needs a `touch target-verus/{debug,release}/verus-root` step. The ergonomics of that will be improved in a follow-up change. Current idea is to check for `vstd.vir` instead of `verus-root`, but this is still TBD. Please chime in if you have thoughts on this.

Also CC @mmcloughlin because this will be relevant to the Release build flow re-design.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: ci-cargo
```
mkovaxx: TODO: Change this back to just `ci` before merging.

.github/workflows/ci.yml
```diff
@@ -40,9 +40,7 @@ jobs:
 
       - name: check rustfmt/verusfmt
         working-directory: ./source
-        run: |
-          . ../tools/activate
-          vargo fmt -- --check
+        run: cargo fmt -- --check
```
parno: I think `vargo fmt` also ran `verusfmt`, not just `rustfmt`.  If that's the case, I think this needs adjustment.

bsdinis:
Thank you so much for starting this undertaking!

Did a prelim pass over it.
Q: does this enable `cargo install verus` ?

mkovaxx:
> Did a prelim pass over it. Q: does this enable `cargo install verus` ?

@bsdinis Thanks for taking a look!

Re: `cargo install verus`, while that command could technically work, it wouldn't produce the correct result. In particular, it doesn't ensure that `rust_verify` and a (verified) `vstd` are also installed alongside `verus`.

Current thinking is that `cargo install cargo-verus` would become the recommended way to install Verus, as put forward in this discussion: https://github.com/verus-lang/verus/discussions/1790

A few rough steps to get there after this PR is landed.

1. Do some more legwork on the Verus Toolchains effort. Design doc here: [LINK](https://docs.google.com/document/d/1E_1XwSpFfoNhK-ZNJNLFHV2VWvCyo7ucEq-GYuVYKwc/view).
2. Extend `cargo-verus` so that it automatically manages the local Verus Toolchains, i.e. it can internally `cargo install {verus,rust_analyzer}` and `vstd`, cache multiple versions locally, ensure compatibility, etc.

source/tools/bump_crate_versions/src/main.rs
parno: In `main()`, the paths to the various Cargo.toml, the paths for the various crates that might need updating are all provided via paths that are relative to the base of the repo.  It looks like `publish` is now calling canonicalize before changing directories into a tempdir, so I think that should be okay.  The `update_crates` hasn't been updated equivalently, presumably b/c it only reads/writes files without invoking `cargo`.  I think that should be fine.  Another option would be to add some code to `main` to pre-process the `crates` Vec to pre-emptively canonicalize the paths, so we don't have to worry about forgetting to do it elsewhere.

source/tools/docs-cargo.sh
parno, line 1: Should we have a note somewhere, so that we remember to eventually use this file to replace tools/docs.sh?

parno (approved):
Generally looks good to me.  It's exciting to see this progressing!

.github/workflows/ci.yml
mkovaxx: Thanks for catching this! Indeed `vargo fmt` runs `rustfmt` on `vstd`. I've updated this PR and opened a related issue: https://github.com/verus-lang/verus/issues/2758

source/tools/docs-cargo.sh
mkovaxx: Done!

source/tools/bump_crate_versions/src/main.rs
mkovaxx: I've opened an issue to track the defensive measure of canonicalizing paths in `main`: https://github.com/verus-lang/verus/issues/2759

ahuoguo:
Can you also add some instructions in https://github.com/verus-lang/verus/blob/main/BUILD.md?

mkovaxx:
> Can you also add some instructions in https://github.com/verus-lang/verus/blob/main/BUILD.md?

@ahuoguo PR underway! https://github.com/verus-lang/verus/pull/2807

PR #2687, opened 2026-07-20
vstd: fix unsound RangeInclusive::end_bound spec for exhausted ranges

### LINKED ISSUE #2674: Unsound `assume_specification` for vstd's `RangeInclusive::end_bound`
When `RangeInclusive` is exhausted via iteration, the Rust standard library will return `Excluded(&range.end)`. However, the `assume_specification` for `<RangeInclusive<T> as RangeBounds<T>>::end_bound` and its `RangeBoundsSpecImpl` currently assume the end bound is always `Included(&range@.end)`, allowing Verus to assert behavior about `end_bounds` that is not true. 

[Rust stdlib implementation of RangeInclusive::end_bound](https://doc.rust-lang.org/src/core/ops/range.rs.html#1124)

[Verus playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=a599ae69eea1ac16b93b840b83cf67aa)

```rust
use core::ops::{Bound, RangeBounds};
use vstd::pervasive::runtime_assert;
use vstd::prelude::*;

verus! {

fn main() {
    let mut r = 1u8..=1u8;

    // Exhausts the range: `1..=1` yields Some(1) on the first next().
    // `Iterator::next`'s trait spec has no requires, so this is freely callable
    let _ = r.next();

    // std returns Excluded(&1) for an exhausted range,
    // so this is false
    let end_is_included = match r.end_bound() {
        Bound::Included(_) => true,
        _ => false,
    };

    // vstd's assume_specification says end_bound() is always
    // Included(&end)
    assert(end_is_included);

    // dies at runtime
    runtime_assert(end_is_included);
}

} // verus!
```

Marti2203:
Fixes https://github.com/verus-lang/verus/issues/2674. `RangeInclusive`'s end bound was unconditionally specced as `Included`, but real Rust returns `Excluded` once the range's iterator is exhausted (confirmed directly against rustc, including that it's specifically the exhausted flag that gates this, not is_empty()/an inverted range - an inverted 5..=1 range that was never iterated still reports `Included`). This let Verus statically prove false facts about an exhausted range's end bound.

Fixed in both places this was specced: the `end_bound` `assume_specification` itself, and the separate `RangeBoundsSpecImpl` used by `slice_range_end` (and so <[T]>::copy_within's own pre/postcondition) - confirmed the latter needed its own fix by reverting each independently and watching the same-shaped test fail. A fresh range's end bound is unaffected by either fix.

Also added a spec for `RangeInclusive::is_empty()` - a range is empty once its iterator is exhausted, or if it was never valid (start > end).

Verified: vstd still verifies fully; slices.rs regression suite passes, including 8 new tests pinning the exhausted-bound behavior, its downstream `slice_range_end` consequence, and `is_empty()`'s spec (checked by reverting each fix and watching the corresponding test fail).

### DIFF
--- a/source/rust_verify_test/tests/slices.rs
+++ b/source/rust_verify_test/tests/slices.rs
@@ -893,6 +893,137 @@ test_verify_one_file! {
     } => Ok(())
 }
 
+// verus-lang/verus#2674: a fresh range's end bound is still Included.
+test_verify_one_file! {
+    #[test] test_range_inclusive_fresh_end_bound_is_included verus_code! {
+        use vstd::prelude::*;
+        use std::ops::{Bound, RangeBounds};
+
+        fn test() {
+            let r = 1u8..=5u8;
+            let end_is_included = match r.end_bound() {
+                Bound::Included(_) => true,
+                _ => false,
+            };
+            assert(end_is_included);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_range_inclusive_exhausted_end_bound_is_not_always_included verus_code! {
+        use vstd::prelude::*;
+        use std::ops::{Bound, RangeBounds};
+
+        fn test() {
+            let mut r = 1u8..=1u8;
+            let _ = r.next();
+            let end_is_included = match r.end_bound() {
+                Bound::Included(_) => true,
+                _ => false,
+            };
+            assert(end_is_included); // FAILS
+        }
+    } => Err(err) => assert_one_fails(err)
+}
+
+test_verify_one_file! {
+    #[test] test_slice_range_end_treats_exhausted_range_as_excluded verus_code! {
+        use std::ops::RangeInclusive;
+        use vstd::prelude::*;
+        use vstd::std_specs::range::{slice_range_end, RangeInclusiveView};
+
+        proof fn test(r: RangeInclusive<usize>)
+            requires
+                r@ == (RangeInclusiveView { start: 2, end: 2, exhausted: true }),
+        {
+            assert(slice_range_end(&r, 5) == 2);
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] test_slice_range_end_treats_exhausted_range_rejects_included verus_code! {
+        use std::ops::RangeInclusive;
+        use vstd::prelude::*;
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/range.rs
```diff
@@ -323,11 +339,14 @@ pub assume_specification<'s, T>[ <RangeInclusive<T> as RangeBounds<T>>::start_bo
...
 
+// `end_bound()` returns `Included` while the range is not exhausted and
+// returns `Excluded` after it is exhausted.
 pub assume_specification<'s, T>[ <RangeInclusive<T> as RangeBounds<T>>::end_bound ](
     range: &'s RangeInclusive<T>,
 ) -> (result: Bound<&'s T>)
     ensures
-        spec_bound(result) == SpecBound::Included(&range@.end),
+        range@.exhausted ==> spec_bound(result) == SpecBound::Excluded(&range@.end),
+        !range@.exhausted ==> spec_bound(result) == SpecBound::Included(&range@.end),
```
parno: Can this instead be phrased in terms of `spec_end_bound`, which seems to have similar logic?

source/rust_verify_test/tests/slices.rs
parno, line 1: Can we simplify/consolidate the tests?  LLMs often seem overly eager to add lots of tests, which then adds to the CI/testing burden.

parno (commented):
Thanks for working on this!

@JakeGinesin I see that you added yourself as a reviewer as well.  Did you have a chance to take a look and or test it out with verus-pbt?

source/rust_verify_test/tests/slices.rs
Marti2203: Sure, will do! I would be also happy to check out if tests can be parallelized to help reduce testing times.

source/vstd/std_specs/range.rs
Marti2203: Will investigate, thank you for the feedback!

Marti2203:
Hi @Chris-Hawblitzel  and @parno - Thank you for the feedback and interactions! I have responded to the comments. I will not do the work in #2801 so that the fix can be merged on top of it.

Chris-Hawblitzel (approved):

PR #2691, opened 2026-07-22
Fix false borrow-check error on reborrow closures under `#[verus_verify]`

tangruize:
## Problem

Under the attribute form `#[verus_verify]`, an expression-bodied closure whose body is a mutable reborrow — such as `|x| &mut *x` — is rejected with a false `used binding isn't initialized` error. `verus! {}` already normalizes every closure body to a block, so the reborrow is handled fine there; the attribute form leaves the body as a bare expression, and Verus's borrow check trips on it. 

## Reproduce

```rust
use vstd::prelude::*;

#[verus_verify]
struct Item { marker: bool }

#[verus_verify(external_body)]
fn take(_x: Option<&mut Item>) -> u32 { unimplemented!() }

#[verus_verify]
fn caller(mut x: Option<&mut Item>, retry: bool) -> u32 {
    let first = take(x.as_mut().map(|r| &mut **r));
    if retry { take(x) } else { first }
}

fn main() {}
```

This example reproduces the `used binding isn't initialized` error under `#[verus_verify]`. Replacing `#[verus_verify]` with an equivalent `verus! { ... }` block lets it pass.

## Approach

`ExecReplacer` (the visitor the attribute path already runs over the exec body) now rewrites an expression-bodied closure `|..| EXPR` to a block-bodied `|..| { EXPR }`, matching what the `verus! {}` path does in `syntax::Visitor::handle_closures`. The rewrite is semantics-preserving; closures that are already block-bodied and closures in spec/ghost tokens are left unchanged.

### DIFF
--- a/source/builtin_macros/src/attr_rewrite.rs
+++ b/source/builtin_macros/src/attr_rewrite.rs
@@ -813,13 +813,16 @@ pub(crate) fn rewrite_verus_spec_on_fun_or_loop(
                 fun = proxy; // Add proof and spec on proxy func.
             }
 
+            // Parse and replace proof_xxx!() inside function and replace panic.
+            // Do this before injecting the generated requires/ensures statements so
+            // the exec rewriter only touches the user's exec body, not generated
+            // spec code (matching how the closure branch above is handled).
+            let inside_external_code = has_external_code_syn(&fun.attrs);
+            replace_block(erase, fun.block_mut().unwrap(), inside_external_code);
+
             // Add the spec/proof (requires/ensures) to the function body.
             let new_stmts = spec_stmts.into_iter().map(|s| parse2(quote! { #s }).unwrap());
             let _ = fun.block_mut().unwrap().stmts.splice(0..0, new_stmts);
-
-            // Parse and replace proof_xxx!() inside function and replace panic.
-            let inside_external_code = has_external_code_syn(&fun.attrs);
-            replace_block(erase, fun.block_mut().unwrap(), inside_external_code);
             fun.to_tokens(&mut new_stream);
             proc_macro::TokenStream::from(new_stream)
         }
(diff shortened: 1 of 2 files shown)


tjhance:
Why does having a block make a difference?

tangruize:
> Why does having a block make a difference?

Thanks @tjhance for reviewing!

This PR follows how `verus! {}` handles closures: it wraps every exec closure body in a block before emitting it: https://github.com/verus-lang/verus/blob/89690b39873cd4c6f26142a8a5378fbf769d4a72/source/builtin_macros/src/syntax.rs#L3631-L3644

The attribute style left the body as a bare expression, and that's what triggers the false error. Verus runs `mir_borrowck` on its own forked MIR, whose time-travel-prevention pass rewrites every `&mut place` into `mutable_reference_tie(&mut place, &mut place_shadow)` and injects the `place_shadow` declarations by wrapping the body:
https://github.com/verus-lang/verus/blob/89690b39873cd4c6f26142a8a5378fbf769d4a72/source/rustc_mir_build_additional_files/verus_time_travel_prevention.rs#L322-L372

For a bare-expression closure body it effectively produces:

```rust
// |c| &mut *c  =>
|c: &mut u32| {
    let c_shadow = arbitrary_ghost_value();          // injected on the SAME HIR node as the reborrow
    mutable_reference_tie(&mut *c, &mut *c_shadow)
}
```

The injected `let c_shadow` and the reborrow that uses it land on the same HIR node / region scope, so borrowck can't see `c_shadow` initialized before its use and reports `used binding isn't initialized` (E0381) on `&mut *c`.

Wrapping the body in a block (what this PR does, matching `verus! {}`) gives the reborrow its own nested scope, so the shadow is initialized first:

```rust
// |c| { &mut *c }  =>
|c: &mut u32| {
    let c_shadow = arbitrary_ghost_value();
    {
        mutable_reference_tie(&mut *c, &mut *c_shadow)   // nested scope: shadow initialized first
    }
}
```

tjhance:
Thanks for identifying the problem & providing the explanation, this was very helpful. However, it's better to fix the rustc_mir_build code that produces the bad scopes, rather than patching the syntax macro to avoid the bad case. So instead I'm going to fix this with: https://github.com/verus-lang/verus/pull/2698

tangruize:
Thanks! Since it's already been addressed by another PR, I'll close this one.

PR #2694, opened 2026-07-22
vstd: add axiom `axiom_slice_get_range` for slice get with a `Range` index

arxgy:
## Problem

`axiom_slice_get_usize` interprets `spec_slice_get` only for a `usize` index. A `Range<usize>` index is left uninterpreted, so `slice.get(a..b)` has no usable specification — one cannot conclude `Some`/`None`, nor relate a `Some` to a subrange:

```rust
fn f(v: &[u8]) requires v@.len() >= 5 {
    let s = v.get(1..3);
    assert(s matches Some(sub) && sub@ == v@.subrange(1, 3)); // fails: spec_slice_get(v, 1..3) is uninterpreted
}
```

## Fix

Add the companion `axiom_slice_get_range` (mirrors `axiom_slice_get_usize`):

- in bounds (`a <= b <= v.len()`): `Some(s)` with `s@ == v@.subrange(a, b)`;
- out of bounds (`a > b` or `b > v.len()`): `None`.

Added to `group_slice_axioms`, plus a `test_get_range` regression test.

### DIFF
--- a/source/rust_verify_test/tests/slices.rs
+++ b/source/rust_verify_test/tests/slices.rs
@@ -569,3 +569,34 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] test_get_range verus_code! {
+        use vstd::{slice::*, prelude::*};
+
+        broadcast use group_slice_axioms;
+
+        fn in_bounds(v: &[u8])
+            requires v@.len() == 5,
+        {
+            let x = v.get(1..3);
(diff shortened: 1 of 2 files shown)


Chris-Hawblitzel:
It looks like there were two independent pull requests around the same time on the same topic, this and https://github.com/verus-lang/verus/pull/2744 .  Through happenstance, I got asked to review the other one first and that got merged first, even though this was submitted first.  Sorry about that!  I'll go ahead and close this since it seems essentially the same.

PR #2696, opened 2026-07-23
vstd: add specification for integer `checked_next_multiple_of`

arxgy:
## Problem

The `num_specs!` macro provides specifications for `checked_add`, `checked_sub`, `checked_mul`, `checked_div`, `checked_rem`, and similar methods, but not for `checked_next_multiple_of`. As a result, a call to `x.checked_next_multiple_of(rhs)` cannot be verified:

```rust
fn f() {
    let r = 10u8.checked_next_multiple_of(3);
    assert(r == Some(12u8)); // fails: checked_next_multiple_of is uninterpreted
}
```

## Fix

Add an `assume_specification` for `<uN>::checked_next_multiple_of`, following the existing `checked_*` specifications. This mirrors the standard library definition (a `checked_rem` followed by a `checked_add`):

- when `rhs == 0`: returns `None`;
- when `x` is already a multiple of `rhs` (`x % rhs == 0`): returns `Some(x)`;
- otherwise the next multiple is `x + (rhs - x % rhs)`, returning `None` if it overflows and `Some(...)` if not.

A `checked_next_multiple_of` regression test is included.

### DIFF
--- a/source/vstd/std_specs/num.rs
+++ b/source/vstd/std_specs/num.rs
@@ -5,6 +5,22 @@ use super::super::wrapping::*;
 
 use core::cmp::Ordering;
 
+verus! {
+
+/// The smallest multiple of `y` that is `>= x` (for `y > 0`), matching the value
+/// std's `next_multiple_of` / `checked_next_multiple_of` compute.
+pub open spec fn next_multiple_of(x: int, y: int) -> int
+    recommends
+        y > 0,
+{
+    if x % y == 0 {
+        x
+    } else {
+        x + (y - x % y)
+    }
+}
+
+} // verus!
 macro_rules! num_specs {
     ($uN: ty, $iN: ty, $mod_u_tmp:ident, $mod_i_tmp:ident, $mod_u:ident, $mod_i:ident, $range:expr) => {
         verus! {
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/num.rs
```diff
@@ -163,6 +163,21 @@ macro_rules! num_specs {
                     }
                 );
 
+            #[verifier::allow_in_spec]
+            #[cfg(not(verus_verify_core))]
+            pub assume_specification[<$uN>::checked_next_multiple_of](x: $uN, rhs: $uN) -> Option<$uN>
+                returns (
```
tjhance, line 185: I would recommend defining a `next_multiple_of(x: int, y: int)` function, and then do

```
if rhs == 0 {
    None
} else if next_multiple_of(lhs as int, rhs as int) > usize::MAX {
   None
} else {
   Some(next_multiple_of(lhs as int, rhs as int) as usize)
}
```

tjhance (changes requested):

source/vstd/std_specs/num.rs
arxgy: Defined the `next_multiple_of(x: int, y: int)` and rewrote the spec in terms of it, as suggested.

source/vstd/std_specs/num.rs
```diff
@@ -5,6 +5,19 @@ use super::super::wrapping::*;
 
 use core::cmp::Ordering;
 
+verus! {
+
+/// The smallest multiple of `y` that is `>= x` (for `y > 0`), matching the value
+/// std's `next_multiple_of` / `checked_next_multiple_of` compute.
+pub open spec fn next_multiple_of(x: int, y: int) -> int {
```
tjhance: add `recommends y > 0`, then we should be good to go

tjhance (changes requested):

source/vstd/std_specs/num.rs
arxgy: Resolved as suggested :)

tjhance (approved):

PR #2720, opened 2026-07-29
vstd/atomic: Add atomic load/store/swap via pointer.

ziqiaozhou:

### DIFF
--- a/source/vstd/atomic.rs
+++ b/source/vstd/atomic.rs
@@ -521,16 +521,10 @@ macro_rules! ptr_atomic_methods {
         /// The specification is similar to raw_ptr::ptr_mut_ref,
         /// but the implementation is atomic and so we can mark it as verifier::atomic,
         /// and so it can be used in open_atomic_invariant!.
-        ///
-        /// # Safety:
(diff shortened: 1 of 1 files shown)


source/vstd/atomic.rs
```diff
@@ -507,6 +508,99 @@ macro_rules! atomic_bool_methods {
...
+            unsafe { core::sync::atomic::$rust_ty::from_ptr(ptr).store(value, core::sync::atomic::Ordering::SeqCst) }
+        }
+
+        /// Create a copy of the value via atomic load.
+        ///
+        /// # Safety: see comments for store_from_ptr.
+        #[inline(always)]
+        #[verifier::atomic]
+        #[verifier::external_body]
+        pub unsafe fn from_ptr_load(ptr: *mut $value_ty, perm: Tracked<&PointsTo<$value_ty>>) -> (ret: $value_ty)
```
tjhance: We typically don't mark the ptr methods as 'unsafe' (which is arguably wrong, but we should keep it consistent for now)

tjhance (changes requested):

tjhance (approved):

Issue #2715, opened 2026-07-27
Support byte-string and byte-character literals in verified code


wood-ghost:
# Description
Verus currently rejects Rust byte-string and byte-character literals used in constants. For example

```
const IMAGE_X_ICON_CURSOR_PATTERN: &[u8] = b"\x00\x00\x02\x00";
d == b' ' || d == b'>' // same as d == 0x20u8 || d == 0x3eu8
```
with
```
error: Unsupported constant type
```
I encountered this while verifying code blocks in [Servo](https://github.com/servo/servo). The original Servo code contains structures initialized with byte-string literals representing binary file signatures:
```
fn image_x_icon() -> ByteMatcher {
    ByteMatcher {
        pattern: b"\x00\x00\x01\x00",
        mask: b"\xFF\xFF\xFF\xFF",
        content_type: "image/x-icon".parse().unwrap(),
        leading_ignore: &[],
    }
}
```
Marking the entire function as `#[verifier::external_body]` avoids the error. However, when the surrounding function needs to be verified, the current workaround is to extract every byte-string literal into a separate constant and mark that constant as external:
```
#[verifier::external_body] 
const IMAGE_X_ICON_CURSOR_PATTERN: &'static [u8] = b"\x00\x00\x02\x00";
#[verifier::external_body] 
const IMAGE_X_ICON_CURSOR_MASK: &'static [u8] = b"\xFF\xFF\xFF\xFF";
fn image_x_icon() -> ByteMatcher {
    ByteMatcher {
        pattern: IMAGE_X_ICON_CURSOR_PATTERN,
        mask: IMAGE_X_ICON_CURSOR_MASK,
        ...
    }
}
```
This requires changing the structure of the original source code and introduces an external trusted constant for every binary signature.

# Requested Feature

Verus should support byte-string literals such as:
```
b"\x00\x00\x02\x00"
```
in constants and executable expressions without requiring them to be marked as external.

For example, this should be supported:
```
const PATTERN: &'static [u8] = b"\x00\x00\x02\x00";
```
It would also be useful for Verus to expose the contents of the byte string to specifications, for example as a sequence of `u8` values equivalent to:
```
seq![0u8, 0u8, 2u8, 0u8]
```

PR #2723, opened 2026-07-29
Support byte-string and byte-character literals in verified code

wood-ghost:
#2715 

### DIFF
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -823,7 +823,7 @@ pub enum Constant {
     Real(String),
     /// Hold generated string slices in here
     StrSlice(Arc<String>),
-    /// Hode byte-string literal here
+    /// Hold byte-string literal here
     ByteStr(Arc<Vec<u8>>),
     // Hold unicode values here
     Char(char),
--- a/source/vir/src/ast.rs
+++ b/source/vir/src/ast.rs
@@ -823,6 +823,8 @@ pub enum Constant {
     Real(String),
     /// Hold generated string slices in here
     StrSlice(Arc<String>),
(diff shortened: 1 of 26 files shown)


source/vir/src/ast.rs
```diff
@@ -823,6 +823,8 @@ pub enum Constant {
     Real(String),
     /// Hold generated string slices in here
     StrSlice(Arc<String>),
+    /// Hode byte-string literal here
```
Chris-Hawblitzel: Typo: "Hode"

source/vir/src/datatype_to_air.rs
```diff
@@ -835,6 +835,19 @@ pub fn datatypes_and_primitives_to_air(ctx: &Ctx, datatypes: &crate::ast::Dataty
         vec![]
     };
 
+    let bytestr_commands = if ctx.used_builtins.uses_array {
```
Chris-Hawblitzel: Instead of `uses_array`, can you add a `uses_bytestr` to used_builtins?

Chris-Hawblitzel (changes requested):
Overall, this looks good.  My only request is to replace uses_array with uses_bytestr.

source/vir/src/ast.rs
wood-ghost: fixed to "Hold"

source/vir/src/datatype_to_air.rs
wood-ghost: Thank you for the advice! I also add `uses_bytestr` to prune.rs

Chris-Hawblitzel (approved):

PR #2735, opened 2026-07-31
vstd/resource: Loc is int

ziqiaozhou:
This allows Loc to be represented as an int, eliminating the need to prove a one-to-one mapping between Loc and int. For example, a unique token's instance_id can be used as the namespace, making it straightforward to prove that AtomicInvariant::namespace() differs from other namespaces when opening invariants in a nested manner.

### DIFF
--- a/source/vstd/resource/mod.rs
+++ b/source/vstd/resource/mod.rs
@@ -31,4 +31,16 @@ verus! {
 #[derive(Eq, PartialEq)]
 pub struct Loc(int);
 
+impl Loc {
+    pub closed spec fn to_int(&self) -> int {
+        self.0
+    }
+
+    pub proof fn lemma_to_int_injective(&self, other: Loc)
+        ensures
+            (self.to_int() == other.to_int()) <==> (self == other),
+    {
+    }
+}
+
 } // verus!
--- a/source/vstd/resource/mod.rs
+++ b/source/vstd/resource/mod.rs
@@ -28,6 +28,7 @@ pub use lib::*;
 
 verus! {
 
-pub type Loc = int;
+#[derive(Eq, PartialEq)]
+pub struct Loc(int);
 
 } // verus!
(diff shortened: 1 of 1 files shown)


Chris-Hawblitzel:
I'll defer to @tjhance and @bsdinis on this.

bsdinis:
Hi!

This was a design choice, where it avoids type confusion between an int and a ghost location of a resource (before, there could be multiple things that could be resource locations or atomic invariant ids or some other int identifier, hence the confusion). 

Can you speak more to the usecase this would help you with?

ziqiaozhou:
It seems that the benefit of the Loc type is fairly limited since it is essentially just an int without any additional bounds. 
I can keep it as-is if there is a real need for it, though. In that case, I would need to add a spec function for Loc, along with a proof function, so that I can use a Loc to represent a namespace (which is an int). Then I can prove that two namespaces are different because they are associated with two unique tokens that must have different Locs, so that one can nestly open multiple AtomicInvariant. This is useful when the code needs to work with a dynamically determined number of AtomicInvariant which I cannot assign then constant value as namespace.

For example,

```
struct UniqueAtomicInvariant {
  a: AtomicInvariant<A, B>,
  unique: FracGhost<()>,
}

impl UniqueAtomic
{
  #[type_invariant]
  spec fn inv(&self) {
     &&& self.a.namespace() == self.unique.id().to_int()
     &&& self.unique.frac() == 1.0
  }

   proof fn lemma_unique(&self, &mut other: Self)
   ensures
        self.a.namespace() != other.a.namespace()
  {
     ...
   }
}
```

To prove lemma_unique, I need the Loc::lemma_to_int_injective.

bsdinis:
Interesting. 

I don't see how the uniqueness of the locations is guaranteed though. Wouldn't using a `GhostMapAuth` be more appropriate?

bsdinis:
That being said, I have no particularly strong feelings about the change as it is right now (adding the spec fn and the proof fn).

PR #2738, opened 2026-08-01
vstd: specs for multiple common String methods

Marti2203:
Adds:

- String::push/String::pop - mirrors the existing pattern already used for Vec::push/Vec::pop.
- String::push_str
- String::is_empty
- String::clear

Verified: 
* vstd still verifies fully (2044 verified, 0 errors); 
* strings.rs regression suite passes (48/48), 
* including 6 new tests covering each addition and a fails-case for each.

### DIFF
--- a/source/rust_verify_test/tests/strings.rs
+++ b/source/rust_verify_test/tests/strings.rs
@@ -846,3 +846,38 @@ test_verify_one_file! {
         }
     } => Ok(())
 }
+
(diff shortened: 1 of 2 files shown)


tjhance (approved):

zero-to-nat (approved):

Marti2203:
Thanks for the approval! 😅 but I think only you can merge it

PR #2740, opened 2026-08-01
Fix pruning of imported private datatypes that cause Verus to panic

hiroki-chen:

The problematic logic assumes every referenced datatype belongs to the current module. Imported datatypes need a guarded lookup, e.g. treating absent entries as non-transparent, or separate imported-datatype handling.

Rust permits this "private_interfaces" warning:

```rs
pub(crate) struct Hidden;
pub struct Exposed {
    pub hidden: Hidden,
}
```

Verus’s module pruning retained `Exposed` but discarded `Hidden` before constructing the datatype map. `datatypes_invs` then followed the field reference and indexed the missing entry. Treating the lookup as non-transparent merely moved the failure to an undeclared AIR type, so opaque declaration handling was also required. This would trigger a panic during cross-crate verification otherwise.

### DIFF
--- a/source/rust_verify/src/verifier.rs
+++ b/source/rust_verify/src/verifier.rs
@@ -41,7 +41,7 @@ use vir::context::{FuncCallGraphLogFiles, GlobalCtx};
 use crate::buckets::{Bucket, BucketId};
 use crate::expand_errors_driver::ExpandErrorsResult;
 use vir::ast::{CrateId, Fun, Krate, VirErr};
-use vir::ast_util::{fun_as_friendly_rust_name, is_visible_to};
+use vir::ast_util::fun_as_friendly_rust_name;
 use vir::def::{CommandContext, CommandsWithContext, CommandsWithContextX, SnapPos};
 use vir::prelude::PreludeConfig;
 
@@ -1373,8 +1373,6 @@ impl Verifier {
             SmtSolver::Cvc5 => None,
         };
 
-        let module = &ctx.module_path();
-
         // Bucket context.
         //
         // Inserted into the main context, but also stored so the identical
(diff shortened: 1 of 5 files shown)


tjhance:
IMO, it would be better to enforce the consistency of privacy & transparency levels in the krate definition, rather than working around the inconsistency in prune.rs. That means are options are either:

 * disallow this code (i.e. effectively preventing users from enabling `private_interfaces`)
 * detect this case and downgrade the transparency of `Exposed` from `pub` to `pub(crate)`
 * detect this case and upgrade the visibility of `Hidden` from `pub(crate)` to `pub`

This probably needs some discussion to determine the right fix.

hiroki-chen:
My two cents:

Option 1 (reject the code) makes Verus stricter than rustc on a warn-by-default pattern. For a tool whose pitch is verifying existing systems Rust, that's a bad trade — and per the earlier discussion in Rust community about effective visibility, the rejection would be non-local: adding a `pub use` in one module could break an unrelated file.

Option 2 (downgrade `Exposed` to `pub(crate)` transparency) is sound and monotone — it only removes information from clients, so it can't introduce unsoundness, and the failure mode is a proof that stops going through with an obvious fix. But since transparency is per-datatype, not per-field, it's coarse: one private field turns off structural reasoning about all the public fields, including plain `==`. That will silently break downstream proofs that had nothing to do with hidden.

Option 3 (upgrade `Hidden` to `pub`) leaks. You'd export constructor and field axioms for a type the author deliberately kept internal, which means downstream proofs can come to depend on `Hidden`'s representation — a semver hazard the author has no way to see. It's also transitive: upgrading `Hidden` drags in whatever `Hidden` mentions, and so on, which inflates exactly the context pruning exists to shrink.

tjhance:
> (upgrade Hidden to pub) leaks. You'd export constructor and field axioms for a type the author deliberately kept internal, which means downstream proofs can come to depend on Hidden's representation — a semver hazard the author has no way to see.

Upgrading the visibility of `Hidden` doesn't necessarily imply making it transparent.

In fact Option 3 basically seems to be what your PR does, just by modifying the pruning logic rather than updating the VIR 'visibility' itself.

Chris-Hawblitzel:
Some relevant Rust history here (why are private types allowed in public interfaces): https://internals.rust-lang.org/t/lang-team-minutes-private-in-public-rules/4504

PR #2741, opened 2026-08-01
vstd: add string search patterns

Marti2203:
`str::starts_with`/`ends_with`/`contains`/`find`/`rfind` are all generic over `P: Pattern`, and `assume_specification` can't target just one instantiation of a generic function - so real Rust's own `&str`, `char`, `&[char]`, and closure-predicate Pattern implementors were all completely unspecified. 
Adds non-generic wrapper functions covering the most common uses of each:

- `str_starts_with_str`/`str_ends_with_str`/`str_contains_str`
- `str_ends_with_char` 
- `str_starts_with_chars`/`str_ends_with_chars`/`str_contains_chars` - `&[char] needle`
- `str_starts_with_pred`/`str_ends_with_pred`/`str_contains_pred`
- `str_find_str`/`str_rfind_str`/`str_find_char`/`str_rfind_char` - `&str/char` needle, returning the `first/last` matching byte offset (confirmed directly against rustc that find/rfind return byte offsets, not char indices - "héllo world".find("world") == Some(7), not Some(6), since 'é' is 2 bytes) - specced entirely via spec_bytes(), no char-index machinery needed.
- `str_find_pred`/`str_rfind_pred`

Verified: 
* vstd still verifies fully (2053 verified, 0 errors); 
* strings.rs regression suite passes
* including 4 new tests (a pass/fails pair per family) covering every wrapper across its Some/None and true/false outcomes.


parno:
Instead of wrapping all of these, I think it would be better to provide a [trait specification](https://verus-lang.github.io/verus/guide/traits.html) for the Pattern trait, which would then allow direct (assume) specifications for the `str` functions.

Marti2203:
Updated per the suggestion above: 
* `starts_with`/`ends_with`/`contains` now go through a real `#[verifier::external_trait_specification]` for `std::str::pattern::Pattern` (+ `external_trait_extension` for a `matches_at`/`obeys_pattern_spec` helper, following the same `obeys_*` pattern vstd already uses for `PartialEq`/`Ord`), instead of one non-generic wrapper per concrete pattern type. 

This removes:
*  `str_starts_with_str`, 
* `str_ends_with_str`, 
* `str_ends_with_char`, 
* `str_contains_str`, 
* `str_starts_with_chars`, 
* `str_ends_with_chars`, 
* `str_contains_chars`,
*  `spec_str_contains`
callers now use the real `.starts_with()`/`.ends_with()`/`.contains()` syntax directly.

`find`/`rfind` are being tackled after this comment.

- **The closure/predicate `Pattern` impl (`FnMut(char) -> bool`) is not covered by the trait spec** and keeps using the hand-written `str_starts_with_pred`/`str_ends_with_pred`/`str_contains_pred`
wrappers. Verus can only derive facts about a closure's `.ensures()` from an actual traced call in verified code, and `str::starts_with`'s real implementation is external/assumed, so Verus never sees it call the predicate internally.

Also found a Verus-internal issue while testing this against the `--is-core` build mode (verifies vstd's source is embeddable directly into libcore): the `Pattern` external_trait_specification declaration needs to always be registered, since Verus's own processing of the real `str` methods that reference `Pattern` needs it even when the rest of the spec (which needs `str`'s `View`/`Seq<char>`, unavailable under that mode) is excluded via `#[cfg(not(verus_verify_core))]`. Splitting the trait declaration (always registered) from its spec body (conditionally gated) fixed it.

Verified: 
* vstd still verifies fully (2052 verified, 0 errors), 
* full `rust_verify_test` suite passes including `core_special_setup`'s `--is-core` test, and `strings.rs` now covers real `.starts_with()`/`.ends_with()`/`.contains()` calls for `char`, `&str`, and `&[char]` patterns (both matching and non-matching cases).

parno:
> The closure/predicate Pattern impl (FnMut(char) -> bool) is not covered by the trait spec and keeps using the hand-written str_starts_with_pred/str_ends_with_pred/str_contains_pred wrappers. Verus can only derive facts about a closure's .ensures() from an actual traced call in verified code, and str::starts_with's real implementation is external/assumed, so Verus never sees it call the predicate internally.

I'm not sure what to make of this comment.  Can you please expand on what you mean?

Marti2203:
Hi @parno,
Apologies for the delayed response! The initial implementation didn't spec the negative ("didn't match") direction in the trait at all, so end users couldn't recover that guarantee generically. I spent some time fixing it.

source/vstd/utf8.rs
```diff
@@ -1053,6 +1053,173 @@ pub broadcast proof fn is_char_boundary_iff_is_leading_byte(bytes: Seq<u8>, inde
     }
 }
 
+proof fn leading_byte_width_1_is_width_1(scalar: u32)
```
parno: Since all of these proofs succeed without any proof annotations, that's often a sign that they don't need to be factored out, especially if they're only used at a single call site.

source/vstd/utf8.rs
```diff
@@ -1053,6 +1053,173 @@ pub broadcast proof fn is_char_boundary_iff_is_leading_byte(bytes: Seq<u8>, inde
...
+        encode_utf8(chars).subrange(byte_i, byte_j) =~= encode_scalar(c as u32),
+    ensures
+        0 <= char_i < chars.len(),
+        byte_i == encode_utf8(chars.subrange(0, char_i)).len(),
+        byte_j == encode_utf8(chars.subrange(0, char_i + 1)).len(),
+        chars[char_i] == c,
+{
+    encode_utf8_valid_utf8(chars);
+    scalar_leading_byte_class(c as u32);
+    assert(encode_utf8(chars)[byte_i] == encode_scalar(c as u32)[0]);
```
parno: I'm curious if this assertion, or the one on line 1211, is necessary for the proof

source/vstd/string.rs
```diff
@@ -132,6 +132,270 @@ pub assume_specification[ str::split_at ](s: &str, mid: usize) -> (res: (&str, &
         res.1.spec_bytes() =~= s.spec_bytes().subrange(mid as int, s.spec_bytes().len() as int),
 ;
 
+/// Specifies `Pattern` for `str::starts_with`/`ends_with`/`contains`/`find`/`rfind`.
```
parno: This AI comment seems unnecessary.

source/vstd/string.rs
```diff
@@ -132,6 +132,270 @@ pub assume_specification[ str::split_at ](s: &str, mid: usize) -> (res: (&str, &
...
+#[verifier::external_trait_extension(PatternSpec via PatternSpecImpl)]
+pub trait ExPattern: Sized {
+    type ExternalTraitSpecificationFor: core::str::pattern::Pattern;
+
+    spec fn obeys_pattern_spec(&self) -> bool;
+
+    /// True iff this pattern instance matches exactly haystack[start..end).
+    spec fn matches_at(&self, haystack: Seq<char>, start: int, end: int) -> bool;
+
+    /// Byte-offset sibling of `matches_at`: `find`/`rfind` return byte
```
parno: I think this can be simplified to "Byte-offset version of `matches_at`"

parno:
> The initial implementation didn't spec the negative ("didn't match") direction in the trait at all, so end users couldn't recover that guarantee generically.

Can you say more about why we need a spec function for both `matches_at` and `not_matches_at_witness`?  I would have expected that for something like `starts_with`'s postcondition of `r == exists |len: int| 0 <= len <= s@.len() && pat.matches_at(s@, 0, len)`, when `r` is `false`, you'd learn `forall |len: int| 0 <= len <= s@.len() ==> !pat.matches_at(s@, 0, len)`.  Is that not the right fact to learn when `starts_with` fails?

Marti2203:
Hi @parno, I agree that your proposed spec is the intuitive one and what one would like to have. The problem I saw is that when `pat` is instantiated for closures, this does not work. `matches_at` defines `self.ensures((haystack[start],), true) `. Verus does not provide the tooling to infer ` !self.ensures(args,true) ==> self.ensures(args,false)` . 

Replacing `not_matches_at_witness` with a direct negation of `matches_at` still lets vstd itself verify, but breaks proving the negative direction for a generic `F: Fn(char) -> bool` caller (added as an example `generic_contains_matches_wrapper` in this PR) - it holds for deterministic closures like the ones in the existing tests. Hope this clarifies!

PR #2743, opened 2026-08-01
rust_verify: fix allow_in_spec autospec name collision across types i…

Marti2203:
#[verifier::allow_in_spec] (with a returns clause) auto-generates a synthetic "autospec" function so the exec function can be called directly in spec position. Its name was derived from just the target's bare last path segment (e.g. `is_empty`), discarding which type it belonged to - so two different types' identically-named methods, specced in the same module, silently collided on one shared `<name>%returns_clause_autospec` identifier.

Found by hitting it directly: adding `String::is_empty` (allow_in_spec+returns) next to the pre-existing `str::is_empty` in `vstd/string.rs` (#2738  failed with "duplicate specification for `vstd::string::is_empty%returns_clause_autospec`", even though the two declarations target genuinely different real functions.

Fixed by joining the target's entire path (not just its last segment) when constructing the autospec name - as unique as the real function itself, since two different functions can never share a full path. Purely permissive: can only turn a previously-erroring combination into a working one, never changes an already-passing program's outcome.

Verified:
- vstd still verifies fully (2044 verified, 0 errors)
- new regression test (test_allow_in_spec_same_target_name_different_targets_no_collision) reproduces the exact collision with two synthetic functions sharing a bare name across modules - confirmed it fails on the unfixed compiler and passes on the fix
- full regression sweep passes: external_fn_specification.rs (79/79), char.rs, hash.rs, strings.rs, when_used_as_spec.rs, returns_postcondition.rs, contrib.rs

### DIFF
--- a/source/rust_verify/src/rust_to_vir_func.rs
+++ b/source/rust_verify/src/rust_to_vir_func.rs
@@ -202,9 +202,18 @@ fn handle_autospec<'tcx>(
         );
 
         if functionx.proxy.is_some() {
-            // Naively, this_path may contain a mangled _verus_external_fn_specification_ name.
-            // Clean this up before sending it to autospec_return_clause_spec_fn_name:
-            this_path = this_path.pop_segment().push_segment(functionx.name.path.last_segment());
+            // Use the target's full path, not just its last segment, so two
+            // types' identically-named methods (e.g. str::is_empty and
+            // String::is_empty) don't collide on the same autospec name.
+            let joined_target_name = functionx
+                .name
+                .path
+                .segments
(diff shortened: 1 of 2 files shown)


tjhance:
@Chris-Hawblitzel do you remember why it was necessary to avoid the "mangled" name `this_path`?

I am wondering if we should just use `this_path`, or perhaps `functionx.name`.

I think a problem with the proposed fix is that `_` is a valid identifier character, so joining a path by `_` isn't sound for avoiding collisions. But it will look a lot like the original mangled path anyway, so we might as well just use the mangled path?

PR #2744, opened 2026-08-01
vstd: slice range index mut writeback

Marti2203:
Following up on #2742: a write through a range-indexed mutable sub-slice reborrow (&mut s[range]) couldn't be related back to the original slice's own view once the reborrow's last use had passed, even though real Rust genuinely performs the mutation (confirmed directly: sub[0] = 99; after let sub = &mut s[1..3]; really does mutate s).

Root cause: the previous spec described what changed outside the written range but never gave Z3 a way to derive the whole final view of the original slice - so proving something like final(s)@ == old(s)@.update(1, 99) after a write through the sub-slice failed, even though the sub-slice's own view was already fully provable.

Fixed by solving for final(slice)@ directly, as a concatenation of the untouched leading/trailing parts plus final(r)@ - the same pattern split_at_mut's spec already uses - for all 6 range types (Range, RangeTo, RangeFrom, RangeToInclusive, RangeFull, RangeInclusive).

Verified:
- vstd still verifies fully (2044 verified, 0 errors)
- slices.rs regression suite passes, including a new test writing through each of the 6 range types' mutable sub-slice and checking the original slice's own view afterward
- a soundness check (asserting a wrong value) still correctly fails

Based on #2742 since 5 of the 6 range types only exist there.

### DIFF
--- a/source/rust_verify_test/tests/slices.rs
+++ b/source/rust_verify_test/tests/slices.rs
@@ -237,8 +237,6 @@ test_verify_one_file! {
     } => Err(err) => assert_fails(err, 8)
 }
 
-// verus-lang/verus#2657: SliceIndex had no spec for RangeTo, so `s[..3]`
-// failed with a confusing precondition error even though `s[0..3]` worked.
 test_verify_one_file! {
     #[test] test_slice_index_range_to verus_code! {
         use std::ops::Index;
@@ -270,9 +268,6 @@ test_verify_one_file! {
     } => Err(err) => assert_fails(err, 4)
 }
 
-// The same gap existed for RangeFrom, RangeToInclusive, RangeFull, and
-// RangeInclusive - `s[5..]`, `s[..=3]`, `s[..]`, `s[1..=3]` all failed the
-// same way as `s[..3]` did before the RangeTo fix above.
 test_verify_one_file! {
     #[test] test_slice_index_range_from verus_code! {
         use vstd::prelude::*;
@@ -340,9 +335,6 @@ test_verify_one_file! {
     } => Err(err) => assert_fails(err, 2)
 }
 
-// The non-panicking `.get()` form went through `spec_slice_get`, which was
-// only ever given meaning for a `usize` index - for every range type, the
-// result was completely opaque, unprovable in either direction.
 test_verify_one_file! {
     #[test] test_slice_get_ranges verus_code! {
         use vstd::prelude::*;
--- a/source/rust_verify_test/tests/slices.rs
+++ b/source/rust_verify_test/tests/slices.rs
@@ -407,6 +407,111 @@ test_verify_one_file! {
     } => Err(err) => assert_one_fails(err)
 }
 
+// Checks the mutable-indexing form (`&mut s[range]`) via the returned
+// sub-slice's own view, both before and after writing through it. Verus
+// currently can't relate a range-indexed sub-slice's mutation back to the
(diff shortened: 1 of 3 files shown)


source/rust_verify_test/tests/slices.rs
```diff
@@ -237,6 +237,327 @@ test_verify_one_file! {
...
+        fn range_inclusive_bounds(s: &[u8]) {
+            assume(s.len() == 5);
+            let x = &s[1..=5]; // FAILS
+        }
+    } => Err(err) => assert_fails(err, 2)
+}
+
+// The non-panicking `.get()` form went through `spec_slice_get`, which was
+// only ever given meaning for a `usize` index - for every range type, the
+// result was completely opaque, unprovable in either direction.
```
tjhance: these comments about what the code used to do aren't really necessary, not unless the fix is particularly unintuitive even in retrospect (which isn't the case here)

source/vstd/std_specs/slice.rs
```diff
@@ -44,7 +46,204 @@ pub assume_specification<T>[ <Range<usize> as SliceIndex<[T]>>::index_mut ](i: R
...
+    ensures
+        r@ == old(slice)@.subrange(
+            slice_range_start(&i),
+            slice_range_end(&i, old(slice)@.len() as nat),
+        ),
+        final(r)@ == final(slice)@.subrange(
+            slice_range_start(&i),
+            slice_range_end(&i, old(slice)@.len() as nat),
+        ),
+        final(slice)@ == final(r)@,
```
tjhance, line 123: It looks to me like this could just be simplified to `r@ == slice@, final(slice)@ == final(r)@`

source/rust_verify_test/tests/slices.rs
Marti2203: Sorry! I missed that one, will clean it up!

source/vstd/std_specs/slice.rs
Marti2203: Sure, will do!

source/vstd/std_specs/slice.rs
```diff
@@ -44,7 +46,190 @@ pub assume_specification<T>[ <Range<usize> as SliceIndex<[T]>>::index_mut ](i: R
...
+            + old(slice)@.subrange(
+                slice_range_end(&i, old(slice)@.len() as nat),
+                old(slice)@.len() as int,
+            ),
+;
+
+pub broadcast axiom fn axiom_slice_get_range<T>(v: &[T], i: Range<usize>)
+    ensures
+        slice_range_valid(&i, v@.len()) ==> {
+            &&& #[trigger] spec_slice_get(v, i).is_some()
```
Chris-Hawblitzel: This should probably be `(#[trigger] spec_slice_get(v, i)).is_some()` (same for other cases below).

Chris-Hawblitzel:
I think it makes sense to use `slice_range_start` and `slice_range_end` for RangeInclusive, since RangeInclusive has private fields.  But for the other range types, I think it would be clearer to use more direct fields or methods like `.start`, `.end`, and `.len()`.

source/vstd/std_specs/slice.rs
Chris-Hawblitzel: RangeFull `index` can also be simplified in the same way.

Marti2203:
Hi @Chris-Hawblitzel Thanks for the input! Will update the PR with the feedback

Issue #754, opened 2023-08-16
Verifier treats non-atomic function as atomic


jaylorch:
Soundness issue: The code below verifies with Verus, even though there's a bug. The implementation of the trait function isn't atomic but the trait requires it to be atomic. Nevertheless, Verus verifies it.

```
    trait AtomicOperation
    {
        #[verifier(atomic)]
        fn do_atomic_op(x: &PAtomicU64, p: &mut Tracked<PermissionU64>)
            requires
                old(p)@.is_for(*x)
            ensures
                p@.is_for(*x);
    }

    struct NotActuallyAtomicOperation
    {
    }

    impl AtomicOperation for NotActuallyAtomicOperation
    {
        // Bug: The verifier raises no error for this function even
        // though it's not atomic but the trait demands that it be
        // atomic. The verifier should fail to verify this function.
        fn do_atomic_op(x: &PAtomicU64, p: &mut Tracked<PermissionU64>)
        {
            x.fetch_add_wrapping(Tracked(p.borrow_mut()), 1);
            x.fetch_add_wrapping(Tracked(p.borrow_mut()), 1);
        }
    }

    // Because of the bug above, the verifier trusts that do_atomic_op
    // is atomic. This is a reasonable thing to trust, because the
    // trait marks it as atomic.

    #[verifier(atomic)]
    fn add_two_atomically(x: &PAtomicU64, p: &mut Tracked<PermissionU64>)
        requires
            old(p)@.is_for(*x)
        ensures
            p@.is_for(*x)
    {
        NotActuallyAtomicOperation::do_atomic_op(x, p);
    }

    pub fn main() {
    }
```

PR #2752, opened 2026-08-04
Support recursive function as atomic

ziqiaozhou:
I was trying to define a recursive type that carries an `AtomicInvariant` and needed to open the atomic invariant recursively. This requires a recursive function that can itself be marked as atomic, allowing it to recursively open the invariant while ensuring that only a single atomic function is invoked at the base case where the recursion terminates. Therefore, Verus needs to support precise branch analysis for atomic recursive functions.

### DIFF
--- a/source/rust_verify_test/tests/atomics.rs
+++ b/source/rust_verify_test/tests/atomics.rs
@@ -329,7 +329,7 @@ test_verify_one_file! {
         fn stuff() {
             stuff();
         }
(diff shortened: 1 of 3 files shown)


ziqiaozhou:
The recursive fn was banned due to #754 reported by @jaylorch in 2023. But I don't think recursion itself was the root cause of that issue, since just preventing atomic on trait should be enough? I suspect there were other reasons it was explicitly disallowed? Maybe missing termination checks in old verus?

PR #2754, opened 2026-08-04
[cargo-verus] fix: match abbreviated git revisions

### LINKED ISSUE #2753: cargo-verus: `--check-toolchain` cannot succeed for any Git-paired toolchain
### Summary

`cargo verus verify --check-toolchain` rejects every toolchain whose `vstd` is paired by Git commit, including pairings `cargo verus toolchain list` itself prints.

[`is_matching_known_and_used`](https://github.com/verus-lang/verus/blob/f6efce9c7fbbdf6b789e5446779c7f93d34ecacc/source/cargo-verus/src/toolchains.rs#L11-L14) compares revisions for string equality:

```rust
(
    Crate::GitCommit { git: known_url, rev: known_rev },
    PackageSource::Git { url: ref used_url, rev: Some(ref used_rev) },
) => known_url == used_url && known_rev == used_rev,
```

A toolchain manifest names an *abbreviated* revision — `create-manifest` writes one, and the round-trip test in `cargo-verus-toolchains` uses `rev = "e6a6d4f"` — but Cargo always resolves a Git dependency to the full 40-character commit hash. So the comparison is between an abbreviation and a full hash:

```
"f6efce9" == "f6efce9c7fbbdf6b789e5446779c7f93d34ecacc"   →   false, always
```

No manifest can satisfy the check, because Cargo cannot be made to report an abbreviated revision. Writing the full hash in the manifest is not a workaround for a released `cargo-verus`, since the manifest list is baked in at build time by `build.rs`.

### Impact

`.github/workflows/build.yml` runs `create-manifest --rolling` into `cargo-verus/toolchain-manifests` immediately before `vargo build --release`:

```yaml
- name: create temporary toolchain manifest for a rolling release
  if: inputs.is_rolling
  working-directory: ./source
  run: |
    cargo run -p cargo-verus-toolchains --bin create-manifest -- --rolling --write-to-dir cargo-verus/toolchain-manifests
```

So every rolling release ships a `cargo-verus` whose only matching toolchain is Git-paired with an abbreviated revision, and `--check-toolchain` cannot succeed for any project on that release. Verified against `0.2026.07.27.31579f0`; a Registry-paired release such as `0.2026.06.07.cd03505` is unaffected.

### The control: it is the comparison, not the versions or the Git source

Both arms below use one locally built toolchain, one scratch package, and the same manifests. Only `toolchains.rs` differs, which rules out a version mismatch or Git sources being unsupported.

Built at `f6efce9c7fbbdf6b789e5446779c7f93d34ecacc`, where the driver reports `0.2026.08.03.f6efce9`. A manifest naming exactly that version, of the shape `create-manifest --rolling` produces:

```toml
# source/cargo-verus/toolchain-manifests/0.2026.08.03.f6efce9.toml
verus = "0.2026.08.03.f6efce9"
vstd = { git = "https://github.com/verus-lang/verus.git", rev = "f6efce9" }
z3 = "4.16.0"
singular = "4.3.2"
```

A scratch package pinned to the same commit, which is what Cargo resolves the dependency to anyway:

```toml
[package.metadata.verus]
verify = true

[dependencies]
vstd = { git = "https://github.com/verus-lang/verus.git", rev = "f6efce9c7fbbdf6b789e5446779c7f93d34ecacc" }
```

**Before** — the URLs are identical and the manifest names the driver's exact version, and it still fails:

```console
$ cargo verus verify --check-toolchain
Error: Components are incompatible:
* verus = 0.2026.08.03.f6efce9
* vstd = PackageMetadata { version: ..., source: Git { url: "https://github.com/verus-lang/verus.git", rev: Some("f6efce9c7fbbdf6b789e5446779c7f93d34ecacc") } }
$ echo $?
1
```

**After**, comparing the revisions as Git compares abbreviated object names:

```console
$ cargo verus verify --check-toolchain
    Checking vstd v0.0.0-2026-08-02-0125 (https://github.com/verus-lang/verus.git?rev=f6efce9c...)
    Checking e2e-check-toolchain v0.1.0 (/tmp/e2e-check-toolchain)
verification results:: 1 verified, 0 errors
$ echo $?
0
```

### A second, related bug in the manifest generator

`get_git_rev` in `cargo-verus-toolchains/src/versions.rs` takes `None` to mean "do not abbreviate", but still passes `--short`:

```rust
let short_flag =
    if let Some(len) = limit { format!("--short={len}") } else { "--short".into() };
```

So `get_vstd_version(is_rolling = true)`, whose comment reads "pin to the Git commit", gets Git's automatic abbreviation instead of the full hash:

```console
$ cargo run -p cargo-verus-toolchains --bin create-manifest -- --rolling
vstd = { git = "https://github.com/verus-lang/verus.git", rev = "a0f5131f4" }
```

Dropping the flag when no length is requested makes the generated manifest name the full hash. That alone would also make `--check-toolchain` pass, but only for generated manifests — a hand-written one may still abbreviate, and the round-trip test asserts that shape, so the comparison is the load-bearing fix.

### Suggested fix

Compare revisions the way Git compares abbreviated object names: the shorter revision must be a prefix of the longer one, ignoring case, with a floor on the abbreviation length. A revision that cannot be a commit hash, such as a tag or a branch name, keeps matching exactly.

Happy to send a PR for both, if that is the direction you want. Given the feature is recent, I would rather check first which side you would like fixed.

### Notes

- Likely unhit so far because the Cargo-native path is the least-trodden route: the guide teaches `verus foo.rs`, and `examples/cargo-verus` uses relative path dependencies from inside the repo, so it never resolves a published `vstd`.
- Only two commits have touched `toolchains.rs`: 868e8b8c (#2550) and 49b8806c (#2584).

lavaman131:
Fixes #2753.

`is_matching_known_and_used` compared Git revisions for string equality. A toolchain manifest names an abbreviated revision, but Cargo always resolves a Git dependency to the full 40-character commit hash, so the comparison could never hold:

```
"f6efce9" == "f6efce9c7fbbdf6b789e5446779c7f93d34ecacc"   →   false, always
```

`--check-toolchain` therefore rejected every Git-paired toolchain, and no manifest could satisfy it, since Cargo cannot be made to report an abbreviated revision.

This compares revisions the way Git compares abbreviated object names instead: the shorter revision has to be a prefix of the longer one, ignoring case, and be at least 7 digits long. A revision that cannot be a commit hash, such as a tag or a branch name, still has to match exactly, so the previous behaviour is preserved wherever it was already correct.

### Verification

The tests are unit tests in `toolchains.rs` rather than in `tests/test_toolchain.rs`, because `mod toolchains` is private and exposing it only for the test seemed worse than following the existing `#[cfg(test)] mod tests` in `metadata.rs`. Three of them fail on `main` and pass here.

End to end, against a local build at `f6efce9c7fbbdf6b789e5446779c7f93d34ecacc` whose driver reports `0.2026.08.03.f6efce9`, with a manifest of the shape `create-manifest --rolling` produces:

```toml
verus = "0.2026.08.03.f6efce9"
vstd = { git = "https://github.com/verus-lang/verus.git", rev = "f6efce9" }
```

and a scratch package pinned to that same commit. Before, with identical URLs and a manifest naming the driver's exact version:

```console
$ cargo verus verify --check-toolchain
Error: Components are incompatible:
* verus = 0.2026.08.03.f6efce9
* vstd = ... rev: Some("f6efce9c7fbbdf6b789e5446779c7f93d34ecacc")
$ echo $?
1
```

After:

```console
$ cargo verus verify --check-toolchain
verification results:: 1 verified, 0 errors
$ echo $?
0
```

Only `toolchains.rs` differs between the two runs, which rules out a version mismatch or Git sources being unsupported.

### Note

The manifest generator has a related bug that #2753 describes and #2755 fixes: `get_git_rev(None)` still passes `--short`, so a generated rolling manifest also names an abbreviated revision. That fix is worth having on its own, but it does not replace this one, because a hand-written manifest may still abbreviate, and the round-trip test in `cargo-verus-toolchains` asserts that shape.


mkovaxx:
Hi @lavaman131, thanks for proposing this change!

I believe these changes are no longer necessary, because your other fix (https://github.com/verus-lang/verus/pull/2755) ensures that a `vstd` pinned to a specific git commit in a toolchain manifest `.toml` file now always uses the full (40-char) hash.

I think we can close this PR as not necessary.

lavaman131:
> Hi @lavaman131, thanks for proposing this change!
> 
> I believe these changes are no longer necessary, because your other fix (#2755) ensures that a `vstd` pinned to a specific git commit in a toolchain manifest `.toml` file now always uses the full (40-char) hash.
> 
> I think we can close this PR as not necessary.

Thanks for reviewing the changes! This was meant to handle hand-rolled configs with the abbreviated revision, but I'm ok with closing this.

PR #2756, opened 2026-08-04
line_count: refactor and improve CLI

bsdinis:
This PR tries to bring line_count into a better state, with hopes of (in the future) improving its reporting itself.

## Summary of Changes

- tests were added
- CLI uses `clap`
- default mode works like a regular line_counter (i.e., takes in a set of paths and counts those); legacy mode (looking at DEPS_FILE.d) is available through `--deps`

Moreover, the one file was broken into multiple for readability.


ahuoguo:
@bsdinis can you also update the readme here: https://github.com/verus-lang/verus/tree/main/source/tools/line_count

It will also be interesting to run this on all veritas projects and see if the numbers look way off from what the authors expect

bsdinis:
> @bsdinis can you also update the readme here: https://github.com/verus-lang/verus/tree/main/source/tools/line_count
> 
> It will also be interesting to run this on all veritas projects and see if the numbers look way off from what the authors expect

Good catch! I updated the readme. I don't have the bandwidth to check this against the veritas projects right now but can do it later. Also note that this PR does not change any of the accounting. That remains the same

.github/workflows/ci.yml
```diff
@@ -133,6 +133,26 @@ jobs:
         working-directory: ./source
         run: cargo test -p cargo-verus
 
+  # run line_count tests
```
parno, line 136: If we're adding this, can we remove the two other spots (`build line-count tool`) in the ci where the test if the line count can be built?  That should be subsumed by this test, right?

source/tools/line_count/src/config.rs
```diff
@@ -0,0 +1,27 @@
+#[derive(Clone, Copy, Debug, clap::Parser)]
+pub struct Config {
+    /// Print all the annotated files
+    #[arg(short, long)]
+    pub print_all: bool,
+
+    /// Do not ignore items outside of `verus!` by default
```
parno, line 7: This phrasing is a bit confusing.  Can we phrase it positively, e.g., "Also count items outside `verus!`"?

source/tools/line_count/src/config.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// Output as machine-readable json
+    #[arg(long)]
+    pub json: bool,
+
+    /// Consider delimiter-only lines as layout
+    #[arg(long)]
+    pub delimiters_are_layout: bool,
+
+    /// Do not apply _trusted_ to proofs
```
parno, line 19: I'm surprised they would be trusted.  Why's that?

source/tools/line_count/README.md
```diff
@@ -7,15 +7,38 @@ This accounts Verus lines in a project by whether they are
...
+  -p, --print-all               Print all the annotated files
+      --no-external-by-default  Do not ignore items outside of `verus!` by default
+      --json                    Output as machine-readable json
+      --delimiters-are-layout   Consider delimiter-only lines as layout
+      --proofs-arent-trusted    Do not apply _trusted_ to proofs
+      --deps                    Parse the DEPS_FILE.d file produced by rustc, finding the crate's paths from that source
+  -h, --help                    Print help
+```
+
+To use the line counter, you can point it at a set of paths. `line_count` searches recursively any directories passed on:
```
parno: I would suggest replacing
> you can point it at a set of paths. `line_count` searches recursively any directories passed on:

with
> you can provide a list of directories as arguments. `line_count` will then recursively search those directories for Rust files:

parno (approved):
Thanks for working on this.  Having read everything, I realize that some of my comments are w.r.t. issues that were there before the refactoring, so we can potentially address them separately from this PR.

bsdinis:
@parno I've addressed the review on the topics that weren't ported over from the existing CLI flags. I'll merge once the tests pass

Issue #2765, opened 2026-08-05
missing vstd specification for executable str equality


wood-ghost:
There seems to be a missing specification connecting executable str equality with equality of  `Seq<char>` .

The following example fails to verify: 

[equality with ==](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=4e851aeb0efa023c550bf36522264b47)

```
use vstd::prelude::*;

verus! {

struct TextSource;

uninterp spec fn modeled_text(source: &TextSource) -> Seq<char>;

#[verifier::external_body]
fn get_text<'a>(source: &'a TextSource) -> (result: &'a str)
    ensures
        result@ == modeled_text(source),
{
    "hello world"
}

spec fn is_hello(source: &TextSource) -> bool {
    modeled_text(source) == "hello world"@
}

fn check(source: &TextSource) -> (result: bool)
    ensures
        result == is_hello(source),
{
    get_text(source) == "hello world"
}

}
```

The postcondition of check cannot be proved, even though `get_text` guarantees `result@ == modeled_text(source)` and the executable comparison is `get_text(source) == "hello world"`

## Workaround

Currently it works by rewriting the original codes using a helper function 

```
#[verifier::external_body]
fn str_equal(a: &str, b: &str) -> (result: bool)
    ensures
        result == (a@ == b@),
{
    a == b
}
```

and replace the `check` body with 
```
str_equal(get_text(source), "hello world")
```

Complete working version:

```
use vstd::prelude::*;

verus! {

struct TextSource;

uninterp spec fn modeled_text(source: &TextSource) -> Seq<char>;

#[verifier::external_body]
fn get_text<'a>(source: &'a TextSource) -> (result: &'a str)
    ensures
        result@ == modeled_text(source),
{
    "hello world"
}

#[verifier::external_body]
fn str_equal(a: &str, b: &str) -> (result: bool)
    ensures
        result == (a@ == b@),
{
    a == b
}

spec fn is_hello(source: &TextSource) -> bool {
    modeled_text(source) == "hello world"@
}

fn check(source: &TextSource) -> (result: bool)
    ensures
        result == is_hello(source),
{
    str_equal(get_text(source), "hello world")
}

}
```

## Expected behavior

The first version should verify without requiring a trusted helper. 

The Playground share interface appears to preserve only one version at a time, so I included only the failing version as a Playground link and included the working workaround inline.

wood-ghost:
This also fails:

```
fn test() {
    assert(
[...]

PR #2773, opened 2026-08-06
feat: add PartialEq specification for str

wood-ghost:
#2765 

### DIFF
--- a/source/rust_verify_test/tests/strings.rs
+++ b/source/rust_verify_test/tests/strings.rs
@@ -945,3 +945,33 @@ test_verify_one_file! {
         }
     } => Err(e) => assert_one_fails(e)
 }
+
+test_verify_one_file! {
+    #[test]
+    str_equality_uses_view verus_code! {
+        use vstd::prelude::*;
+
+        struct TextSource;
+
+        uninterp spec fn modeled_text(source: &TextSource) -> Seq<char>;
+
+        #[verifier::external_body]
+        fn get_text<'a>(source: &'a TextSource) -> (result: &'a str)
+            ensures
+                result@ == modeled_text(source),
+        {
+            "hello world"
+        }
+
+        spec fn is_hello(source: &TextSource) -> bool {
+            modeled_text(source) == "hello world"@
+        }
+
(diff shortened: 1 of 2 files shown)


parno:
This expresses equality of `str` by comparing the `char`s.  The implementation almost certainly compares the bytes, so correctness here depends on Verus maintaining the invariant that all `str` are valid UTF8.  Rust expects this as well, but it's possible to violate that via `unsafe`.  OTOH, the existing str specs are all mostly written in terms of `chars`.  So this might be okay, but we might want to revisit all of these specs to see if the byte-representation should play a larger role.

zero-to-nat:
> This expresses equality of `str` by comparing the `char`s. The implementation almost certainly compares the bytes, so correctness here depends on Verus maintaining the invariant that all `str` are valid UTF8. Rust expects this as well, but it's possible to violate that via `unsafe`. OTOH, the existing str specs are all mostly written in terms of `chars`. So this might be okay, but we might want to revisit all of these specs to see if the byte-representation should play a larger role.

For the purposes of vstd, I don't believe that it makes a difference whether we define equality in terms of the view (`Seq<char>`) or in terms of the corresponding `Seq<u8>`. This is because vstd axiomatizes the view only -- the byte representation is defined in vstd as `encode_utf8(self@)`. And since we could derive an invertibility lemma of the UTF8 encoding (`encode_utf8(s1@) == encode_utf8(s2@) <==> s1@ == s2@`), I think this means that there would be no difference between the definitions in terms of what they express. So, if people prefer to work with `char`s, I am okay to keep using that in vstd.

In terms of the potential to violate the UTF8 invariant with `unsafe` code, I think this means that we would need to take care about any trusted specifications that we add to vstd for Rust standard library functions that could violate this invariant. I think the safety conditions are pretty easy to express for the existing trusted specs in vstd. Maybe this will be more of a problem when we eventually consider e.g., transmutes or `str::as_bytes_mut`.

zero-to-nat (dismissed):

zero-to-nat (approved):
Everything looks good, per offline discussion with Bryan. Thank you for adding this spec!

Issue #1670, opened 2025-05-21
Array equality


y1ca1:
One would expect the following implications to hold in both directions:
```rust
use vstd::prelude::*;

verus! {

proof fn array_eq<const N: usize>(a: [u8; N], b: [u8; N])
    ensures (forall|i: int| 0 <= i < N ==> a[i] == b[i]) <==> a == b
{}

}
```
Verus easily verifies the `<==`half, but balks at the `==>` half. Is this a limitation or is the other direction unsound?

[Link to playground](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=84129e78172c7e9957efea53045fd29f)

tjhance:
This works:

```
ensures (forall|i: int| 0 <= i < N ==> a[i] == b[i]) ==> a@ =~= b@
```

There are two differences here:

 * You need to use extensional equality.
 * You need to compare the views `a@` and `b@` instead of `a` and `b`.

The second is probably an unnecessary limitation. But the first is by design.

Chris-Hawblitzel:
> The second is probably an unnecessary limitation.

If we wanted to remove this limitation, we could add:

```
pub broadcast axiom fn axiom_array_ext_equal<T, const N: usize>(a1: [T; N], a2: [T; N])
    ensures
        #[trigger] (a1 =~= a2) <==> (forall|i: int| 0 <= i < N ==> a1[i] == a2[i]),
;
``` 

Arguably, an array [T; 4] is essentially the same as struct S(T, T, T, T), and we already have extensional equality for structs.  By contrast, we deliberately don't have extensional equality for abstract types like Vec (since Vec can have additional internal structure, like capacity).

tjhance:
Yes, I think arrays, and slices for that matter, should have extensional equality.

PR #2776, opened 2026-08-06
feat: add PartialEq specification for array and slice

wood-ghost:
# Summary

This PR adds `PartialEq` specifications for slices and arrays in `vstd`.

It connects executable Rust equality such as:

```
left == right
```

with the corresponding elementwise specification through `PartialEqSpecImpl`.

# Motivation

While working on byte-string support analogous to string support in #2773,  I found that Verus did not provide `PartialEqSpecImpl` implementations for arrays or slices.

For example, the example below could not previously be verified.
```
fn check(left: &[u8], right: &[u8]) -> (result: bool) 
    ensures result == (left@ =~= right@) 
{ 
    left == right
}
```

Rust defines slice and array equality by comparing their lengths and then comparing each pair of corresponding elements. Without the corresponding `PartialEq` specifications, Verus could not connect executable equality with equality of the modeled sequences.

# Changes

This PR adds specifications for the following Rust PartialEq implementations:

- slice compared with slice;
- array compared with array;
- slice compared with array;
- array compared with slice;
- the reference variants required for comparisons such as `&[T] == &[U; N]`

The specifications follow Rust’s equality semantics:

- the lengths must be equal;
- every pair of corresponding elements must satisfy the element type’s `PartialEqSpec`.

# Test

The PR adds regression tests covering:

```
&[T] == &[U]
[T; N] == [U; N]
&[T] == &[U; N]
[T; N] == &[U]
```

It also adds a byte-string regression test. The specification uses `=~=` to invoke extensional equality, for the same reason discussed in #1670.


source/vstd/std_specs/array.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    open spec fn eq_spec(&self, other: &&'a [U]) -> bool {
+        &&& self@.len() == (*other)@.len()
+        &&& forall|i: int|
+            #![auto]
+            0 <= i < self@.len() ==> <T as super::cmp::PartialEqSpec<U>>::eq_spec(&self@[i], &(*other)@[i])
+    }
+}
+
+// slice ref == array ref
```
Chris-Hawblitzel: The comment says "slice ref == array ref", but I think it's "slice == array".  Also, we could write a spec for "array == slice".

Chris-Hawblitzel (approved):

source/vstd/std_specs/array.rs
wood-ghost: Thank you for the comment. 'array == slice" has been added as well as more test cases.

PR #2784, opened 2026-08-07
Allow credit generation in limited proof-mode contexts

tjhance:
* Make `create_open_invariant_credit` a proof-mode function
 * Restrict the `create_open_invariant_credit` call to contexts where we can statically check that it cannot be called unboundedly, i.e., where it could clearly be hoisted to exec context
 * Remove the `spend_open_invariant_credit` function; instead, use the credit as an argument to the `open_atomic_invariant_begin` call

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -607,7 +607,11 @@ mod typing {
         }
 
         #[must_use]
-        pub(super) fn push_unbounded_ghost_context<'a>(&'a mut self, kind: UnboundedGhostContext, span: &Span) -> Typing<'a> {
+        pub(super) fn push_unbounded_ghost_context<'a>(
+            &'a mut self,
+            kind: UnboundedGhostContext,
+            span: &Span,
+        ) -> Typing<'a> {
             if self.unbounded_ghost_context.is_none() {
                 self.internal_state.unbounded_ghost_context = Some((kind, span.clone()));
                 Typing {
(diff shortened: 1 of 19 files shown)


source/vir/src/modes.rs
```diff
@@ -3191,16 +3258,22 @@ fn check_expr(
...
-                    record,
-                    &mut typing,
-                    Mode::Proof,
-                    body,
-                    Mode::Proof,
-                    outer_proph,
-                )?;
+                {
+                    let mut typing = typing.push_block_ghostness(Ghost::Ghost);
+                    let mut typing = typing.push_unbounded_ghost_context(
```
jaylorch, line 3263: Is the second `let mut` necessary here?

tjhance: yes, otherwise there's a lifetime error from overwriting the same variable

jaylorch (commented):
Looks good to me. But note that I don't have an encyclopedic enough knowledge of the code base to know if this covers all the cases that might lead to infinite descent.

I do notice, though, that `examples/atomic_increment.rs` also needs to be updated.

jaylorch:
Other files that seem to need changing include `examples/guide/logatom.rs`, `examples/helping.rs`, and `examples/logatom_lib.rs`.

Chris-Hawblitzel (approved):

tjhance:
i guess this needs an announcement since it's a breaking change

PR #2786, opened 2026-08-07
vstd/raw_ptr: make allocate work on no_std

lschuermann:
This removes the `std` dependency of `raw_ptr::allocate` by switching it away from `std::process::abort` towards `handle_alloc_error`. In contrast to the former, the latter may potentially unwind instead of aborting the process. This should be fine, however, as it unwinds before minting any capability that could allow accessing the failed allocation.


tjhance:
I think a better option for no_std environments is just a strict wrapper around `alloc` without any extra logic, i.e., it just returns `ptr::null()` sometimes and it's up to the client to handle it. Then you can panic, abort, whatever you want.

lschuermann:
@tjhance This makes sense to me. To make sure I understand: I'd change `allocate` such that the current `ensures` would be conditioned on the returned pointer being non-null, with no guarantees for a null-pointer? And then callers can use those current post-conditions after a simple runtime check in their own code? I'm happy to implement that (breaking) change.

PR #2790, opened 2026-08-09
vstd: add recommends check for get_variant/is_variant enum accessors

Marti2203:
Fixes the `get-variant`/`is-variant` half of #408. 
The `choose` half was already fixed by #807 (2023) - confirmed the exact repro from the issue produces the expected recommends warning today.

Calling a generated enum accessor like `f.get_A_0()` or `f->A_0` on a value that might not actually be the `A` variant silently produced zero errors and zero warnings, even under `#[verifier::spec(checked)]`. These accessors compile to a small `#[verifier::inline]` spec function (built in builtin_macros/src/is_variant.rs and enum_synthesize.rs) that calls `get_variant_field(...)` directly with no contract at all - so the existing call-site recommends-checking machinery had nothing to check.

Fixed by synthesizing `recommends is_variant(&self, "Variant")` onto the generated accessor functions themselves, so ordinary recommends-checking now catches this automatically. Also added the analogous check directly in `vir/src/ast_to_sst.rs` for a raw, unwrapped field access with `get_variant: true`, since that's a real, separately reachable path (confirmed hit during vstd's own build) that doesn't go through the named accessor.

Verified:
- New regression test in rust_verify_test/tests/recommends.rs, fails without the fix and passes with it
- No false positives on properly-guarded usage (e.g. `requires f is A`)
- recommends, adts, adts_generics, adts_opaque, unions, match, mutable_params all pass with no regressions
- vstd still verifies fully (2044/0 errors) before and after

### DIFF
--- a/source/rust_verify_test/tests/recommends.rs
+++ b/source/rust_verify_test/tests/recommends.rs
@@ -115,3 +115,21 @@ test_verify_one_file! {
         }
     } => Err(e) => assert_has_recommends_failure(e)
 }
+
+// Same as above, but via the named `get_A_0()` accessor rather than `->` - these compile
+// to different call shapes (a synthesized #[verifier::inline] fn with its own
+// `recommends` clause, vs. a raw field access), so both need their own coverage.
+test_verify_one_file! {
+    #[test] get_variant_field_method_call_recommends_issue408 verus_code! {
+        #[is_variant]
+        pub enum Foo {
+            A(u32),
+            B(bool),
+        }
+
+        proof fn test_ens(f: Foo)
+            ensures f.get_A_0() == 10  // FAILS: nothing establishes f is the A variant
+        {
+        }
+    } => Err(e) => assert_has_recommends_failure(e)
+}
--- a/source/rust_verify_test/tests/recommends.rs
+++ b/source/rust_verify_test/tests/recommends.rs
@@ -116,9 +116,7 @@ test_verify_one_file! {
     } => Err(e) => assert_has_recommends_failure(e)
 }
 
-// Same as above, but via the named `get_A_0()` accessor rather than `->` - these compile
-// to different call shapes (a synthesized #[verifier::inline] fn with its own
-// `recommends` clause, vs. a raw field access), so both need their own coverage.
+// Same as above, but via the named `get_A_0()` accessor rather than `->`.
 test_verify_one_file! {
     #[test] get_variant_field_method_call_recommends_issue408 verus_code! {
         #[is_variant]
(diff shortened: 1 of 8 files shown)


source/vir/src/ast_to_sst.rs
```diff
@@ -1932,21 +1932,36 @@ pub(crate) fn expr_to_stm_opt(
                             &expr.span, &exp, field_opr,
                         );
                         let field_opr = FieldOpr { check: VariantCheck::None, ..field_opr.clone() };
-                        (Some((condition, msg)), UnaryOpr::Field(field_opr))
+                        (Some((condition, msg, true)), UnaryOpr::Field(field_opr))
+                    }
+                    // `get_variant` accessors are total - just give a recommends-only hint.
+                    VariantCheck::None if field_opr.get_variant => {
```
tjhance: We should use a separate `VariantCheck::Recommends` kind for those that need the recommends check. `None` should be left as-is.

Though, your 'if' condition is probably correct (i.e., the nodes that will use `VariantCheck::Recommends` are probably exactly the same as those that use `get_variant`)

source/rust_verify_test/tests/recommends.rs
```diff
@@ -97,3 +97,21 @@ test_verify_one_file! {
         assert_one_fails(e);
     }
 }
+
+// https://github.com/verus-lang/verus/issues/408
+// A `get_variant` accessor (here via `->`) is a total function, so it's not unsound to
+// call it on a value that might not be the given variant, but it's a recommends-worthy
+// mistake, so it should be flagged just like an explicit `recommends` clause would be.
+test_verify_one_file! {
```
tjhance, line 105: Since the PR adds the recommendation check for both `->` and `get_variant_field`, there should be a test for the latter as well.

tjhance (changes requested):
thanks, this has been a missing feature for a while. only the two comments.

source/rust_verify_test/tests/recommends.rs
Marti2203: Sure, will prepare a test!

source/vir/src/ast_to_sst.rs
Marti2203: Sure, will fix it!

source/rust_verify_test/tests/recommends.rs
Marti2203: Added

source/vir/src/ast_to_sst.rs
Marti2203: Updated/Added

tjhance (approved):

Marti2203:
Hi,
I don't think I can merge it as I am not a contributor to the project 😅

PR #2791, opened 2026-08-09
Remove the interactive debugger (-V debug)

Marti2203:
Fixes verus-lang/verus's `-V debug` mode hard-panicking on three common constructs:
* reassigning an already-declared variable,
* `by(nonlinear_arith)` blocks,
* `by(bit_vector)` blocks.
All three previously crashed the worker thread with `unimplemented!(...)`. Also fixes the debugger's expression evaluator, which was either unsound or wrong for any query beyond a bare variable.

Root causes and fixes:

1. sst_to_air.rs's handler for reassigning an already-declared variable unconditionally panicked under ctx.debug instead of emitting the assignment. Fixed by emitting it normally, then creating a fresh AIR snapshot right after (mirroring the existing pattern already used for a mutating call destination), so a later debugger query at that line resolves to the just-written value instead of a stale one.

2. The debugger's own value lookup issued a live Z3 query after the model was already returned - unsound, since obtaining a model queues a command that invalidates it before that follow-up query reaches Z3 (confirmed directly: `(error "model is not available")`). Fixed by extending `air::model::Model` with `raw_values`, captured once at the point the model is first parsed, and reading a bare variable's value straight from there instead.

3. `by(nonlinear_arith)` and `by(bit_vector)` hit the identical panic at two sibling call sites. Unlike (1), neither needed new snapshot logic - each spins off a fully self-contained sub-query rather than mutating the ongoing session - so the fix there is just removing the panic guard. `nonlinear_arith` gets full debugger support after this. `bit_vector`'s crash is fixed and its result is reported correctly, but the interactive shell doesn't attach to a `bit_vector` failure specifically, since `by(bit_vector)` deliberately runs in single_check_query mode (no incremental Z3 state) - a separate, pre-existing limitation this doesn't try to paper over.

4. The debugger's evaluation of a compound expression (e.g. `(add_one x)`) had two bugs of its own: it fell back to the same kind of unsound live Z3 query as (2), and separately its AIR name construction for the callee was naive (missing the crate/module prefix), failing independently with `unknown constant`. Fixed the same way as (2): `Model` gained `raw_defs`/`find_def_by_suffix`, so a function application is answered directly from the already-captured model dump instead of a live round-trip. Only a flat (non-case-split) recorded body is returned; anything with case-split structure reports it can't be evaluated this way rather than attempting a real interpreter. The suffix match requires a `.`/`!` boundary (so a query for `add_one` can't accidentally match an unrelated `xadd_one`) and returns nothing rather than an arbitrary pick if the match is ambiguous. `eval_expr` no longer needs a live `Context` at all, so `start_shell` dropped that parameter.

Also replaced the debugger's hardcoded two-query demo with a real REPL (`line <N>`, an expression, or `quit` from stdin).

Verified:
- vstd still verifies fully (2044/0 errors)
- bitvector.rs (87/87), assert_bitvector_by.rs (6/6), nonlinear.rs (18/18) all pass unaffected
- A broad regression sweep (loops, mut_refs\*, mutable_params, closures, atomics - 700+ tests) passes unaffected, confirming the change only takes effect under `-V debug`
- 8 new regression tests: reassignment/nonlinear_arith/bit_vector each get a pass and fail case under -V debug, a dedicated test driving the real REPL via stdin confirming it reports the correct, distinct value on each side of a reassignment, and a test confirming a function-application query against a real counterexample evaluates correctly (no live round-trip, no naming bug)


Marti2203:
Possible follow-ups, none attempted here:
* Hand-evaluate simple case-split function bodies. Right now a function whose model body has ite structure just reports "can't evaluate." But for the common case — an ite chain branching only on the function's own (now-concrete) arguments or other already-captured constants — the debugger could substitute the known values into the conditions and walk the chain itself, entirely in Rust, no Z3 involved. That covers a real class of currently-unsupported queries (e.g. `iLo/uHi`-style width-dispatch functions) without the bigger live-reload undertaking below.
* General live expression evaluation (arbitrary nesting, not just one flat function call) would need a real live Z3 session again. I looked at "reload the captured model as fresh asserts, then eval live" as a way to get this back soundly — it doesn't work as a small addition: Z3 rejects redefining an already-declared symbol outright (and worse, silently falls back to evaluating against the original, under-constrained state instead of erroring), and most of a real AIR model's values are Z3-internal opaque tokens (`Dcr!val!0`, `FuelId!val!54`, ...) that aren't valid to reassert at all. A working version would mean replaying the entire original prelude into a fresh solver process, not just the model — a separate, larger piece of work.
* `by(bit_vector)` still doesn't attach the interactive shell to a `bit_vector` failure specifically, since that mode runs in a separate, prelude-free, one-shot Z3 context by design (single_check_query, predates this PR) — the crash is fixed and the result reports correctly, just without live/model-backed shell support for that one prover mode.

parno:
Unfortunately, [as discussed](https://verus-lang.zulipchat.com/#narrow/channel/399078-help/topic/exporting.20a.20counterexample) some time back on Zulip, the counterexamples we get from the SMT solvers are sometimes misleading, so we've shelved the debugger prototype until that can be resolved.  I think it would probably be better to remove the debugger option from the help menu for now, and revisit the code implementation if we come up with a solution to the more fundamental issue with the SMT solvers.

Marti2203:
Thank you for the response! I did not know that Verus's encoding also suffers from quantifiers :( I have the same issue with some work on Frama-C's WP plugin, so I fully understand and should not have been surprised personally.
If you think it is useful, I can probably update the PR to remove the debugger...
 
I personally think that too many axioms are being loaded, which is causing the difficulty, but maybe I have not been in the weeds enough with the VC encodings to know better.

parno:
> If you think it is useful, I can probably update the PR to remove the debugger...

Yes, I think that might be the best course, unfortunately.  Another contributor is planning a PR that will put in some generic hooks that should support a variety of proof debugging features, so if we're eventually able to resolve the SMT issue, we should be able to use the new infrastructure to pretty simply build a fresh debugger.

Marti2203:
Hi @parno , sure! Will rework the PR in such a case

parno:
Cool, thanks very much!

PR #2792, opened 2026-08-10
Stop Verus-macro-reprocessing the bodies of external_body/external functions

### LINKED ISSUE #657: Functions with `#[verifier(external_body)]` should not be transformed by the syntax macro
This currently causes issues with using e.g. `assert!` in bodies of trusted `#[verifier(external_body)]` functions inside the verus macro.

Discovered by @jialin-li and @hayley-leblanc.

Marti2203:
Fixes #657.

`assert`/`assume`/`reveal`/`hide`/`reveal_with_fuel` aren't reserved words, so a real, user-defined function with one of these names, called without a ! inside an external/external_body body, was silently rewritten into Verus's ghost builtin instead - which is erased entirely on `--compile`.

The first commit here fixes `assert`/`assume` (the case reported in the issue). This PR extends the same fix to `reveal`/`hide`/`reveal_with_fuel`, which had the identical bug - confirmed via a real repro before fixing. These five are the complete set of non-reserved-word forms at risk (checked both of `verus_syn`'s expression dispatch sites directly).

Verified: two regression tests that actually compile and run the produced binary (fail without the fix, pass with it); vstd still verifies fully; opaque_reveal.rs (real reveal/hide/reveal_with_fuel usage) still passes; full test suite clean.

### DIFF
--- a/dependencies/syn/src/item.rs
+++ b/dependencies/syn/src/item.rs
@@ -1001,12 +1001,13 @@ pub(crate) mod parsing {
     use crate::path::Path;
     use crate::punctuated::Punctuated;
     use crate::restriction::Visibility;
-    use crate::stmt::Block;
+    use crate::stmt::{Block, Stmt};
     use crate::token;
     use crate::ty::{Abi, ReturnType, Type, TypePath, TypeReference};
     use crate::verbatim;
     use crate::verus::{Context, DataMode, Ensures, FnMode, Publish};
     use alloc::boxed::Box;
+    use alloc::string::ToString;
     use alloc::vec::Vec;
     use proc_macro2::TokenStream;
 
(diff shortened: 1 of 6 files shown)


tjhance:
Thanks for identifying this issue! 

@Chris-Hawblitzel, @parno, and I discussed this PR today. We concluded it would probably be best if the verus macro avoids parsing `external_body` bodies entirely; that would be more foolproof since we then wouldn't need to handle every keyword on a case-by-case basis. Both syn and verus_syn have an `Expr::Verbatim` node that allows one to embed unparsed token streams into the AST; thus, verus_syn could be modified to recognize the `external_body` attribute and return an `Expr::Verbatim` for the body rather than parsing it.

Marti2203:
I have investigated whether I can implement external_body bodies as `Expr::Verbatim`.

The mechanism works cleanly for the general case but it broke vstd's own build at two call sites. Both used `Tracked(x)` construction sugar inside an external_body exec function:

```
pub fn replace(&self, Tracked(perm): Tracked<&mut PointsTo<T>>, in_v: T) -> (out_v: T) {
    let mut v = in_v;
    core::mem::swap(&mut v, self.borrow_mut(Tracked(perm)));
    v
}
```

Only 2 real call sites in the whole codebase hit this (`PCell::replace` in `vstd/cell/pcell.rs`, `PCell::write` in the deprecated `vstd/cell.rs`), and in both cases `external_body` was not needed — `replace`'s own doc comment even says "Trusted core ends here" right above it. Verus verifies both cleanly with `external_body` simply removed, since their requires/ensures are directly provable from the callee they delegate to (vstd goes from 2043 → 2045 verified functions, 0 errors).

Marti2203:
I also had to rework a couple of tests that were still using external_body for some reason

PR #2794, opened 2026-08-10
[cargo-verus] fix: mitigate known caching issue

### LINKED ISSUE #2737: Result caching by `cargo` may improperly skip verification
Sometimes `cargo` caches results improperly, leading it to skip verification even when it's warranted. This can lead to it returning a 0 error code even though there are bugs.

Here's one way to get this to happen:
1) Create a project with two modules `a.rs` and `b.rs`.
2) Also create an incorrect version `b.breaking.rs` but don't use it in the project.
3) Verify successfully with `cargo verus verify`.
4) Run `mv b.breaking.rs b.rs`.
5) Run `cargo verus verify` again. It will exit with error code 0 even though it shouldn't.

This type of scenario can occur naturally when an editor or coding agent atomically renames a previously created file into place. The replacement retains its older modification time, so Cargo considers it fresh.

Here are examples of such files:
```
// a.rs
use vstd::prelude::*;

verus! {

pub fn call_identity() -> u64 {
    crate::b::identity(0)
}

}
```

```
// b.rs
use vstd::prelude::*;

verus! {

pub fn identity(value: u64) -> (result: u64)
    requires value == value,
    ensures result == value,
{
    value
}

}
```

```
// b.breaking.rs
use vstd::prelude::*;

verus! {

pub fn identity(value: u64) -> (result: u64)
    requires value != value,
    ensures result == value,
{
    value
}

}
```

mkovaxx:


source/.cargo/config.toml
```diff
@@ -8,6 +8,7 @@ rustflags = [
 ]
 
 [env]
+CARGO_UNSTABLE_CHECKSUM_FRESHNESS = "true"
```
mkovaxx, line 12: Adding this so that the Cargo-based (a.k.a. Vargo-free) build flows mixing plain `cargo` and `cargo-verus` invocations can share the same cache.

jaylorch (commented):
Works for me, thanks!

jaylorch:
I've since learned even more details about the issue I was having. It seems it was deeper than just `mtime` issues.

What was actually going on was that verus-mode in emacs was using flycheck. This was making it so that whenever I saved a file, it would run a partial verification on just the module in that file (`cargo verus verify -- --verify-module <m>`). That module-only run succeeded and poisoned Cargo’s freshness cache. The subsequent full `cargo verus verify` reused it and reported success. After `cargo clean`, full verification failed correctly.

So the issue is that a `--verify-module` run poisons Cargo's freshness cache, making it think that every module is OK when only the verified module is.

In other words, this PR fixes the `mtime` issue (which I'm not sure was ever the actual underlying cause of the issues I was facing). But be aware that the PR doesn't fix what the *real* issue turns out to be, which is partial verification polluting Cargo's freshness cache.

Sorry it took me this long to figure out what the problem really was!

mkovaxx:
> I've since learned even more details about the issue I was having. It seems it was deeper than just `mtime` issues.

@jaylorch Thanks for adding those details! I've opened a new issue to track that problem separately: https://github.com/verus-lang/verus/issues/2806

PR #2801, opened 2026-08-11
Fix RangeInclusive::end_bound, switch from SpecBound to Bound

Chris-Hawblitzel:
The definition of RangeInclusive::end_bound ( https://doc.rust-lang.org/src/core/ops/range.rs.html#1122 ) is this:

```
impl<T> const RangeBounds<T> for RangeInclusive<T> {
    fn start_bound(&self) -> Bound<&T> {
        Included(&self.start)
    }
    fn end_bound(&self) -> Bound<&T> {
        if self.exhausted {
            // When the iterator is exhausted, we usually have start == end,
            // but we want the range to appear empty, containing nothing.
            Excluded(&self.end)
        } else {
            Included(&self.end)
        }
    }
}
```

Therefore, our assume_specification for this needs to account for `exhausted`:

```
pub assume_specification<'s, T>[ <RangeInclusive<T> as RangeBounds<T>>::end_bound ](
    range: &'s RangeInclusive<T>,
) -> (result: Bound<&'s T>)
    ensures
        range@.exhausted ==> result == Bound::Excluded(&range@.end),
        !range@.exhausted ==> result == Bound::Included(&range@.end),
;
```

Also, I noticed while fixing this that SpecBound had the same definition as Bound.  I don't think we need both, so I replaced SpecBound with Bound to simplify the code.


tjhance:
does this fix https://github.com/verus-lang/verus/issues/2674 ?

Chris-Hawblitzel:
> does this fix #2674 ?

Yes, good point. I'll merge in https://github.com/verus-lang/verus/pull/2687 first, since that precedes my pull request.

jaylorch (approved):
Thanks for fixing this!

PR #2805, opened 2026-08-11
Fixes #2799

### LINKED ISSUE #2799: Verifying with `--verify-module` makes `cargo`'s cache think all modules are OK
The following code, divided into a `Cargo.toml` and two `.rs` files, verifies with `cargo verus verify`. However, if you then comment out the `assume_specification` in the second file, and rerun `cargo verus verify`, the output indicates no error. Verus should realize that `main.rs` is dependent on that change, and try to re-verify it (and fail). But you have to do `cargo clean` to see the error.

Cargo.toml:
```
[package]
name = "test1"
version = "0.1.0"
edition = "2024"

[package.metadata.verus]
verify = true

[dependencies]
vstd = { version = "=0.0.0-2026-08-09-0044" }
num-bigint = { version = "0.5", default-features = false }
```

src/main.rs:
```
use vstd::prelude::*;
use num_bigint::BigInt;

mod b;

use crate::b::*;

verus! {

pub fn convert_integer(i: i16)
{
    let m = BigInt::from(i);
    assert(m@ == i);
}

}

fn main() {
    convert_integer(3i16);
    println!("Hello, world!");
}
```

src/b.rs:
```
use vstd::prelude::*;
use num_bigint::BigInt;

verus! {

#[verifier::external_body]
pub struct ExNumBigInt(num_bigint::BigInt);

pub trait BigIntAdditionalSpecFns {
    spec fn view(&self) -> int;
}

impl BigIntAdditionalSpecFns for BigInt {
    uninterp spec fn view(&self) -> int;
}

pub assume_specification[ <BigInt as core::convert::From<i16>>::from ](i: i16) -> (res: BigInt)
    ensures
        res@ == i,
;

}
```

### LINKED ISSUE #2806: bug: `--verify-module` interferes with full `verus verify`
Originally reported by @jaylorch on #2794.

> I've since learned even more details about the issue I was having. It seems it was deeper than just `mtime` issues.
>
> What was actually going on was that verus-mode in emacs was using flycheck. This was making it so that whenever I saved a file, it would run a partial verification on just the module in that file (`cargo verus verify -- --verify-module <m>`). That module-only run succeeded and poisoned Cargo’s freshness cache. The subsequent full `cargo verus verify` reused it and reported success. After `cargo clean`, full verification failed correctly.
>
> So the issue is that a `--verify-module` run poisons Cargo's freshness cache, making it think that every module is OK when only the verified module is.
>
> In other words, this PR fixes the `mtime` issue (which I'm not sure was ever the actual underlying cause of the issues I was facing). But be aware that the PR doesn't fix what the *real* issue turns out to be, which is partial verification polluting Cargo's freshness cache.

jaylorch:
Also fixes #2806 

### DIFF
--- a/source/rust_verify/src/cargo_verus_dep_tracker.rs
+++ b/source/rust_verify/src/cargo_verus_dep_tracker.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 
 use rustc_span::symbol::Symbol;
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct DepTracker {
     env: BTreeMap<String, Option<String>>,
     files: BTreeSet<PathBuf>,
--- a/source/rust_verify/src/driver.rs
+++ b/source/rust_verify/src/driver.rs
@@ -117,12 +117,17 @@ for all functions (it would only be needed for functions with tracked data in pr
 struct CompilerCallbacksEraseMacro {
     pub do_compile: bool,
     pub override_stability: bool,
+    pub dep_tracker: Option<crate::cargo_verus_dep_tracker::DepTracker>,
 }
 
 impl rustc_driver::Callbacks for CompilerCallbacksEraseMacro {
     // Adding `override_stability` and `stable_attr` functions is a hacky solution specifically for verifying core,
     // to fix an issue with stability attributes.
     fn config(&mut self, config: &mut rustc_interface::interface::Config) {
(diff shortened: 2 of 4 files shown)


jaylorch:
At the Verus meeting today, we decided to instead just forbid the use of `--verify-module` with `cargo verus verify`. So this PR is no longer needed.

PR #2820, opened 2026-08-15
Fix interpreter missing cross-module opaque fn bodies (#944)

### LINKED ISSUE #944: Access to spec function bodies defined in vstd
This program
```
use vstd::prelude::*;

verus!{

fn main() { }

pub proof fn bar() {
    assert(vstd::std_specs::bits::u64_leading_zeros(0) == 64) by (compute_only);
}

}
```
uses `u64_leading_zeros`, which is defined as `public` and `open`:
https://github.com/verus-lang/verus/blob/2b607f74e57f0010a26c3e8d9f3b047a752f6c10/source/pervasive/std_specs/bits.rs#L655

Oddly, when the interpreter tries to run, when it reaches the point of trying to apply `u64_leading_zeros`, it does a lookup for the function's body and the lookup fails, specifically we get `None` here:
https://github.com/verus-lang/verus/blob/2b607f74e57f0010a26c3e8d9f3b047a752f6c10/source/vir/src/interpreter.rs#L1365
where `fun_ssts` is passed into the interpreter from the `state` object in the `ast_to_sst` conversion.

If I inline the definition in the same file, there's no trouble finding and using the definition. 
```
use vstd::prelude::*;

verus!{

fn main() { }

#[verifier::opaque]
pub open spec fn u64_leading_zeros(i: u64) -> int
    decreases i
{
    if i == 0 {
        64
    } else {
        u64_leading_zeros(i / 2) - 1
    }
}

pub proof fn bar() {
    assert(u64_leading_zeros(0) == 64) by (compute_only);
}

}
}
```
 Is this expected behavior?  Or is something going wrong in the collection of `fun_ssts`?

Marti2203:
Fixes #944.

`prune_krate_for_module_or_krate` strips the body of any spec fn outside the root module unless it's non-opaque or explicitly revealed. `by(compute)`/`by(compute_only)` already special-cased this for builtin `Seq` functions (`assert_by_compute_seq_funs`), but not for user-defined functions — so an opaque fn compiled fine when inlined into the same file as the assertion (the root module keeps full bodies unconditionally), but failed when defined elsewhere, even in vstd, even though the interpreter is supposed to ignore reveal/fuel entirely.

Fix: extend the existing `assert_by_compute` exemption in `prune.rs` to all spec functions, not just builtin `Seq` ones.

Verified:
- vstd still verifies fully (2043/0)
- New regression test fn_calls_cross_module_opaque in assert_by_compute.rs fails without the fix and passes with it
- Full assert_by_compute.rs suite passes (44/44), no regressions

By submitting this pull request, I confirm that my contribution is made under the terms of the [MIT license](https://github.com/verus-lang/verus/blob/main/LICENSE).

### DIFF
--- a/source/rust_verify/src/buckets.rs
+++ b/source/rust_verify/src/buckets.rs
@@ -1,7 +1,8 @@
 use std::collections::{HashMap, HashSet};
 use vir::{
-    ast::{Fun, Krate, Path},
+    ast::{Fun, Function, Krate, Path},
     ast_util::fun_as_friendly_rust_name,
+    prune::contains_assert_by_compute,
 };
 
 // A "bucket" is a group of functions that are processed together
@@ -11,7 +12,12 @@ use vir::{
 // to be coherent for a single bucket, so a bucket cannot be cross-module).
 //
 // More precisely, we determine the buckets based off the spinoff_prover attribute
-// (see get_buckets).
+// (see get_buckets and needs_own_bucket).
+/// A function gets its own bucket if it's spinoff_prover, or contains
+/// `assert(..) by (compute)` (see vir::prune::functions_reachable_for_compute).
+fn needs_own_bucket(func: &Function) -> bool {
+    func.x.attrs.spinoff_prover || func.x.body.as_ref().is_some_and(contains_assert_by_compute)
+}
 
 #[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
 pub enum BucketId {
@@ -84,7 +90,7 @@ impl Bucket {
 
 /// Arrange the given modules into buckets.
 /// Typically, this means 1 bucket per module;
-/// However, any functions marked 'spinoff_prover' get their own bucket.
+/// However, any function flagged by needs_own_bucket gets its own bucket.
 pub fn get_buckets(
     krate: &Krate,
     modules_to_verify: &Vec<vir::ast::Module>,
@@ -94,7 +100,7 @@ pub fn get_buckets(
     for func in &krate.functions {
         if let Some(owning_module) = &func.x.owning_module {
             if module_set.contains(owning_module) {
-                let bucket_id = if func.x.attrs.spinoff_prover {
+                let bucket_id = if needs_own_bucket(func) {
                     BucketId::Fun(owning_module.clone(), func.x.name.clone())
                 } else {
                     BucketId::Module(owning_module.clone())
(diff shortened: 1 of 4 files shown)


parno:
Thanks for taking a look at this!  It looks like this change would have the effect that if there's any use of `assert(e) by (compute)` in a module, then all spec function definitions in that module are retained (rather than pruned).  If I'm understanding it correctly, I think that might be more revelation than we want.  I think we should either (1) only reveal functions that are mentioned (transitively) in the asserted expression `e`, or (2) take @Chris-Hawblitzel's suggestion from #944's discussion that we should treat assert-by-compute in the same way as the bit-vector and nonlinear queries that get their own top-level proof context in rust_verify, so that revealing all of these spec functions won't harm the performance of the normal SMT queries generated by other proof obligations in the module.

Marti2203:
Hi @parno, thank you for the feedback! I agree; the initial version revealed across the whole module was a bit too much. Thank you for catching that performance risk! I have decided to rewrite this to use your first proposed solution, as it will be lighter compared to the second.

parno:
Thanks for making those updates.  We talked about this at a recent Verus meeting, and the consensus was that we should take an approach that treats any function that contains an assert-by-compute as though the function had the `spinoff_prover` attribute, so that it gets its own pruning context, and only such contexts get this altered pruning treatment.  That way, we know the performance impact is limited to only that function, and since assert-by-compute is not a common case, we can afford to spinoff a fresh context when we see it.

Marti2203:
I updated per the meeting consensus: any function containing `assert(..) by (compute)` now gets its own bucket, the same way `#[verifier::spinoff_prover]` does, instead of the separate SST elaboration pass from the last version of this PR. 
`functions_reachable_for_compute`'s forced reveal is now bounded to that function's own pruning context via the existing bucket mechanism (`buckets::needs_own_bucket`), so the change is much smaller: a one-line predicate in `get_buckets` plus a small detection helper in `prune.rs`.

One thing worth calling out from testing this: giving a function its own bucket doesn't give it *complete* isolation from its sibling functions in the same module.
I confirmed this is pre-existing behavior of `spinoff_prover` itself (not something new here) by testing an explicit `#[verifier::spinoff_prover]` function with a plain `reveal()` and observing the identical characteristic via `--log-all` .smt2 output. 
Since the request was specifically to match `spinoff_prover`'s existing behavior, I think this is fine, but flagging it in case it matters.

Issue #1764, opened 2025-06-20
empty enums / never type: unintuitive behavior


tjhance:
Verus supports the never type `!` as well as empty enums (when they are opaque). However, they have really unintuitive behaviors. ~~As far as I'm aware, there is no soundness issue~~ (EDIT: see below), only strange behaviors.

Consider, for example:

```rust
enum Foo {
    Bar(Option<!>),
    Qux,
}

fn test(f: Foo) {
    match f {
        Foo::Bar(None) => {
        }
        Foo::Qux => {
            assert(f is Qux); // this fails
        }
    }
}
```

The assertion fails as a consequence of our encoding which looks something like:

```
if (f is Bar) && f->Bar_0 is None {
    // first match arm
} else {
    // second match arm
    assert(f is Qux);
}
```

In other words, the `Foo::Qux =>` branch becomes an `else`, but Verus doesn't know that the `Bar(Some(_))` case is impossible. This is made worse by the fact that `Bar(Some(_))` _is_ possible in spec mode, where `!` is inhabited by a fake element.

You can see the same thing with an empty enum (due to https://github.com/verus-lang/verus/issues/1398 this requires using external_type_specification):

```rust
#[verifier::external]
enum EmptyEnum { }

#[verifier::external_type_specification]
#[verifier::external_body]
struct ExEmptyEnum(EmptyEnum);

enum Foo {
    Bar(Option<EmptyEnum>),
    Qux,
}

proof fn test(f: Foo) {
    match f {
        Foo::Bar(None) => {
        }
        Foo::Qux => {
            assert(f is Qux);
        }
    }
}
```

(The latter also results in an unrelated lifetime_generate error.)

tjhance:
Turns out this IS a soundness issue.

```rust
#[verifier::external_body]
tracked struct False {
}

axiom fn false_from_False(tracked f: False)
    ensures false;

tracked enum Enum {
    GhostNvr { ghost ghost_never: std::convert::Infallible },
    TrackedNvr { tracked tracked_never: False },
}

proof fn test()
    ensures false
{
    let tracked e = Enum::GhostNvr { ghost_never: arbitrary() };
    let tracked Enum::TrackedNvr { tracked_never } = e;
    false_from_False(tracked_never);
}
```

tjhance:
with #2824 we can remove the soundness label, but there are still weird behaviors for match expressions in spec fns

PR #2824, opened 2026-08-18
fix the pattern soundness issue

tjhance:
fixes the soundness issue in https://github.com/verus-lang/verus/issues/1764

Attempting to patch the rustc exhaustiveness checker to accommodate verus ghost fields seems difficult. Thus a general fix for every oddity under the #1764 umbrella remains elusive. This fixes the soundness issue by injecting assertions which are then checked by the SMT solver.

So this code for example:

```
proof fn test()
    ensures false
{
    let tracked e = Enum::GhostNvr { ghost_never: arbitrary() };
    let tracked Enum::TrackedNvr { tracked_never } = e;
    false_from_False(tracked_never);
}
```

will get an `assert(e matches TrackedNvr)` in it (which then fails).

### DIFF
--- a/source/vir/src/modes.rs
+++ b/source/vir/src/modes.rs
@@ -3529,7 +3529,7 @@ fn check_stmt(
             let _ = check_expr(ctxt, record, typing, outer_mode, expect, e, outer_proph)?;
             Ok(())
         }
-        StmtX::Decl { pattern, mode: None, init, els: _ } => {
+        StmtX::Decl { pattern, mode: None, init, els: _, assert_irrefutable: _ } => {
             // Special case mode inference just for our encoding of "let tracked pat = ..."
             // in Rust as "let xl; ... { let pat ... xl = xr; }".
             match (&pattern.x, init) {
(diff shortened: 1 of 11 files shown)


source/vir/src/modes.rs
```diff
@@ -3520,6 +3526,15 @@ fn check_stmt(
     stmt: &Stmt,
     outer_proph: &Proph,
 ) -> Result<(), VirErr> {
+    if matches!(typing.block_ghostness, Ghost::Ghost)
+        && !typing.in_pure
+        && let StmtX::Decl { pattern, mode: _, init: Some(_), els: None, assert_irrefutable: _ } =
+            &stmt.x
+        && !matches!(&pattern.x, PatternX::Var(_))
```
Chris-Hawblitzel: Rather than just the `PatternX::Var` case, can we define the following patterns to be complete:
- Wildcard
- Var
- MutRef(p) or ImmutRef(p) where p is complete
- Constructor(dt, _, ...p...) where dt is a single-variant datatype and all p are complete

Chris-Hawblitzel (approved):

PR #2826, opened 2026-08-19
[rust_verify] fix: `rerun-if-changed` paths in `build.rs`

mkovaxx:


mkovaxx:
@Chris-Hawblitzel @parno This fixes the staleness-detection issue that Chris noticed. The problem was that the original `println!("cargo::rerun-if-changed=../.git/HEAD");` line had the wrong path. Cargo considers a missing file as perpetual change, so it always performed a clean build of `rust_verify`.

PR #2831, opened 2026-08-20
vstd: add BTreeMap specifications

Marsman1996:
While [verifying `util::range_alloc` from Asterinas](https://github.com/asterinas/vostd/pull/699), I found that some necessary `BTreeMap` specifications were missing from vstd. 
I added the required specifications to [complete the verification](https://github.com/asterinas/vostd/pull/704). 
This PR adds the relevant missing specifications directly to vstd.

Assisted-by: Codex: gpt-5.6-sol


source/vstd/std_specs/btree.rs
```diff
@@ -475,6 +600,8 @@ pub assume_specification<
 >[ BTreeMap::<Key, Value, A>::get::<Q> ](m: &'a BTreeMap<Key, Value, A>, k: &Q) -> (result: Option<
     &'a Value,
 >)
+    requires
+        borrowed_key_ordering_matches::<Key, Q>(),
```
jaylorch, line 618: Does this need to be here? If one is never relying on the ordering properties, one might still want to call `get`.

source/vstd/std_specs/btree.rs
```diff
@@ -541,13 +668,192 @@ pub assume_specification<
...
+pub open spec fn borrowed_key_mutated<Key, Value, Q: ?Sized>(
+    old_map: Map<Key, Value>,
+    new_map: Map<Key, Value>,
+    key: &Q,
+    old_value: Value,
+    new_value: Value,
+) -> bool {
+    &&& maps_borrowed_key_to_value(old_map, key, old_value)
+    &&& maps_borrowed_key_to_value(new_map, key, new_value)
+    &&& exists|remainder: Map<Key, Value>|
```
jaylorch, line 695: A simpler way to express this, which doesn't involve an existential, would be `sets_differ_by_borrowed_key(old_map.keys(), new_map.keys(), key)`.

source/vstd/std_specs/btree.rs
```diff
@@ -541,13 +668,192 @@ pub assume_specification<
...
+        {
+            &&& borrowed_key_removed(old_map, remainder, key)
+            &&& borrowed_key_removed(new_map, remainder, key)
+        }
+}
+
+/// Simplifies [`borrowed_key_mutated`] when the borrowed key has the map's key type.
+pub broadcast proof fn lemma_borrowed_key_mutated_deref<Key, Value>(
```
jaylorch, line 703: This lemma may be substantially simplified by my above suggestion. (It may even be obviated?)

source/vstd/std_specs/btree.rs
```diff
@@ -194,6 +199,126 @@ pub assume_specification<'a, Key, Value, A: Allocator + Clone>[ BTreeMap::<Key,
...
+///
+/// A cursor points at the gap immediately before `keys[position]`. Therefore, `peek_next`
+/// accesses `keys[position]`, while `peek_prev` accesses `keys[position - 1]`.
+pub ghost struct CursorMutModel<Key, Value> {
+    /// All keys in the underlying map, in strictly increasing order.
+    pub keys: Seq<Key>,
+    /// The index of the element immediately after the cursor.
+    pub position: int,
+    /// The current contents of the complete map borrowed by the cursor.
+    pub map: Map<Key, Value>,
```
jaylorch, line 225: I wonder if it would be simpler to have an `items: Seq<(Key, Value)>` field instead of both a `keys: Seq<Key>` and a `map: Map<Key, Value>` that have to be kept in sync (i.e., for which we maintain the invariant that the former is the domain of the latter).

jaylorch (commented):
Cool!

source/vstd/std_specs/btree.rs
jaylorch: On second thought, I don't think that would be simpler. It's pretty nice to be able to compactly express that the `keys` field stays the same when you do a mutable get or peek.

source/vstd/std_specs/btree.rs
Marsman1996: Well, this spec is based on the [std doc](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.get), which says
> The key may be any borrowed form of the map’s key type, but the ordering on the borrowed form must match the ordering on the key type.

jaylorch: OK, that's pretty definitive!

source/vstd/std_specs/btree.rs
Marsman1996: But this is not equivalent. 

The proposed relation only compares key sets, so it cannot ensure that other values remain unchanged. Also, the current `sets_differ_by_borrowed_key` implies removal of the key, whereas `get_mut` preserves it.

jaylorch: Ah, yes, good point.

source/vstd/std_specs/btree.rs
jaylorch: Never mind, based on above conversation.

jaylorch (approved):
Thanks for this!

Marsman1996:
Shall I add some tests for this?

jaylorch:
I'm currently asking GPT-5.6 Sol to try to find an unsoundness (i.e., prove `false`) with these added specifications. So far it hasn't. It goes for a while and eventually stops because a Responsible AI check thinks I'm trying to use it for a hack. I'll try Opus 5 next.

jaylorch:
Opus didn't find an unsoundness. But it did find an issue:

`CursorMut::next/prev/insert_after/remove_next` are unspecified, so cursors can't be advanced. This is safe, but limits the feature to peeking at one gap.

jaylorch:
Opus is also not happy that the spec doesn't mean much unless you assume `borrowed_key_ordering_matches::<Key, Q>()`, and there's no way to establish that. I'm personally OK with that, at least for now, but it is something we should eventually fix.

Marsman1996:
> `CursorMut::next/prev/insert_after/remove_next` are unspecified, so cursors can't be advanced.

Have added the specs for these functions and the corresponding tests.

These were missed since the verification for `util::range_alloc` only involved the peek operations.

Marsman1996:
> eventually stops because a Responsible AI check thinks I'm trying to use it for a hack.

Weird, verification shouldn't trigger the cybersecurity warning.

jaylorch:
Thanks for the additional specs and tests for them! They look great to me, so I hereby re-affirm my approval of this PR. :-) It would be good to wait for another reviewer before merging, though.

jaylorch:
I guess no one else wants to review the PR, so my approval will just have to do. Though, first it seems you have some conflicts to resolve.

PR #2832, opened 2026-08-20
vstd: PartialEq specs for str/String (fixes #2765)

### LINKED ISSUE #2765: missing vstd specification for executable str equality

LibertychaserUS:
## Summary

Fixes missing vstd link between executable `str` / `String` `PartialEq` and `Seq<char>` views (`@`), so issue #2765 verifies without a trusted helper.

- Add `PartialEqSpecImpl` + bare `assume_specification` for `str` (same pattern as `bool` / integers / `ExPartialEq`; no separate `ne` assume_spec).
- Unify `String == String` to the same SpecImpl style (replacing the older ensures-only shortcut).
- Add cross-type specs matching Rust `alloc::string::impl_eq!`: `String`↔`str` and `String`↔`&str` (both directions), so comparisons like `"a" == some_string` verify (C12).
- Gate SpecImpl / assume_specs with `verus_keep_ghost` (std_specs is absent otherwise; caught by cargo `unverified_*` builds).
- `String`↔`&str` assume_specs live under `verus_!` in `std_specs/string.rs` because `verusfmt` cannot parse `'_0` early-binder lifetimes inside `verus!` (same binder pattern as `std_specs/cmp.rs`).

## Test plan

- [x] `vargo fmt -- --check`
- [x] `vargo build --release` → 2043 verified, 0 errors
- [x] `rust_verify_test` strings regression (issue repro, empty/unicode/`!=`/`.eq`, C12 cross equal/unequal + fail control)
- [x] Full `vargo test -p rust_verify_test` on Mac → `EXIT:0`
- [x] Full `vargo test -p rust_verify_test` on Linux VPS tip including C12 → `EXIT:0`
- [x] `cargo` / `unverified_*` path after `verus_keep_ghost` gate

## Notes for reviewers

- Related open work: #2773 (`wood-ghost:str_eq`) also targets #2765 with a `str`-only SpecImpl; this PR is broader (assume_spec + String unify + C12). Happy to coordinate.
- Style reference: #2776 (slice/array PartialEq SpecImpl + bare assume_spec).

Assisted-by: Cursor:Composer

Fixes #2765


parno:
Also related to #2773

source/vstd/string.rs
```diff
@@ -35,6 +39,30 @@ impl DeepView for str {
...
+    }
+
+    open spec fn eq_spec(&self, other: &str) -> bool {
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(not(verus_verify_core))]
+pub assume_specification[ <str as PartialEq>::eq ](s: &str, other: &str) -> bool
```
zero-to-nat: This `assume_specification` is not necessary, given the `PartialEqSpecImpl` above.

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+    }
+
+    open spec fn eq_spec(&self, other: &String) -> bool {
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ <String as PartialEq>::eq ](s: &String, other: &String) -> bool
```
zero-to-nat: same as above

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+    }
+
+    open spec fn eq_spec(&self, other: &str) -> bool {
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ <String as PartialEq<str>>::eq ](s: &String, other: &str) -> bool
```
zero-to-nat: same as above

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+    }
+
+    open spec fn eq_spec(&self, other: &String) -> bool {
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ <str as PartialEq<String>>::eq ](s: &str, other: &String) -> bool
```
zero-to-nat: same as above

source/vstd/std_specs/string.rs
```diff
@@ -0,0 +1,25 @@
...
+use verus as verus_;
+
+#[cfg(feature = "alloc")]
+use alloc::string::String;
+
+verus_! {
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification<'_0>[ <String as PartialEq<&str>>::eq ](s: &String, other: &&str) -> bool
```
zero-to-nat: same as in vstd/string.rs - these additions are not needed

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ <String as PartialEq>::eq ](s: &String, other: &String) -> bool
+;
+
+// String == str  /  str == String
```
zero-to-nat: My understanding is that this specific impl will only apply for the `String == str` case, is that correct? I think it might be more clear if this comment only stated `// String == str`, and if there was an analogous comment `// str == String` for the corresponding impl below.

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+    open spec fn obeys_eq_spec() -> bool {
+        true
+    }
+
+    open spec fn eq_spec(&self, other: &&str) -> bool {
+        self@ == (*other)@
+    }
+}
+
+// assume_specification for String↔&str lives in std_specs/string.rs (verus_!) —
```
zero-to-nat: Can remove this comment when these assume_specifications are deleted

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
...
+        self@ == other@
+    }
+}
+
+#[cfg(verus_keep_ghost)]
+#[cfg(all(feature = "alloc", not(verus_verify_core)))]
+pub assume_specification[ <str as PartialEq<String>>::eq ](s: &str, other: &String) -> bool
+;
+
+// String == &str  /  &str == String  (typical `"lit" == s` / `s == "lit"`)
```
zero-to-nat: Same as above for the different types on LHS/RHS

source/vstd/string.rs
```diff
@@ -317,6 +345,86 @@ impl DeepView for String {
     }
 }
 
+// String == String (replaces older ensures-only assume_specification; same SpecImpl style as str)
```
zero-to-nat: Could you please clean up this comment as well? I don't think that the reference to the old state of the repository is needed, and the note about "SpecImpl style" feels like it should be obvious to someone reading this file

source/vstd/string.rs
```diff
@@ -35,6 +39,30 @@ impl DeepView for str {
     }
 }
 
+// PartialEq for str / String: SpecImpl + bare assume_specification (same pattern as
```
zero-to-nat: Could you please clean up this comment? IMO, it is good to explain the overall approach and reason for providing these specifications, but I think that many of the details here will add clutter: references to open PRs (they may no longer be open in the future); that `ne` comes from the same external trait specification (this should be self-documenting); referencing `alloc::string::impl_eq!` (Rust could change this name in the future).

I think that a fine comment would be:
```
// PartialEq for str / String / &str
// Equality is defined here at the char-level, as vstd axiomatizes the View of a str/String as a Seq<char>.
// The below impls allow the comparison of str, String, and &str (the type of string literals in Rust).
```

source/rust_verify_test/tests/strings.rs
```diff
@@ -945,3 +945,159 @@ test_verify_one_file! {
         }
     } => Err(e) => assert_one_fails(e)
 }
+
```
zero-to-nat: Thank you for adding tests for the new specs! I think it would be helpful to clean up the tests a bit to ensure that they are systematically testing the new code.

Based on the changes in this PR, I would expect the tests to be of this format:
- For each `PartialEqSpecImpl` that was added:
  - Have a test that exercises `PartialEq::eq` for that impl
  - Have a test that exercises `PartialEq::neq` for that impl

It's up to you whether you want to create a separate `#[test]` for each impl, or include them as different functions under a single `#[test]`. In either case, for each piece of test code, it should be clear which impl it is testing.

If you want to test on both string literals and abstract strings for which you know something about the view (e.g. `a@ != b@`), that is fine, but I would again ask that it is applied systematically.

source/rust_verify_test/tests/strings.rs
```diff
@@ -945,3 +945,159 @@ test_verify_one_file! {
...
+// Regression for https://github.com/verus-lang/verus/issues/2765 —
```
zero-to-nat: This comment is not 100% accurate as this test covers more than the original issue. If desired, you could leave the regression test for the issue, and add your tests separately

PR #2834, opened 2026-08-21
Give bare `ensures` loop clauses full context for the zero-iteration 

### LINKED ISSUE #925: Incompleteness in loop `ensures` when there's an opaque constant in the loop condition
[Edited from the original report to replace with a smaller repro by @utaal]

```rust
use vstd::prelude::*;

verus!{

fn main() { }

#[verifier::opaque]
const X: usize = 1;

fn foo() {
    let mut i: usize = 0;
    reveal(X);
    while i < X
        ensures i >= 1
    {
        assume(false);
        break;
        assume(false);
    }
}
}
```

```
ξ verus-release /st/verus/test.rs --multiple-errors 5
warning: unreachable statement
  --> /st/verus/test.rs:18:9
   |
17 |         break;
   |         ----- any code following this expression is unreachable
18 |         assume(false);
   |         ^^^^^^^^^^^^^^ unreachable statement
   |
   = note: `#[warn(unreachable_code)]` on by default

error: loop invariant not satisfied
  --> /st/verus/test.rs:13:11
   |
13 |     while i < X
   |           ^^^^^ at this loop exit
14 |         ensures i >= 1
   |                 ------ failed this invariant

error: aborting due to previous error; 1 warning emitted

verification results:: 3 verified, 1 errors
```

---

* [ ] un-ignore https://github.com/verus-lang/verus/commit/21ad99b3c31c0fcae6df787576a425c98fd436b5

Marti2203:
Fixes #925.

Concrete repro (from the issue), before this PR:

```rust
#[verifier::opaque]
const X: usize = 1;

fn foo() {
    let mut i: usize = 0;
    reveal(X);
    while i < X
        ensures i >= 1
    {
        i += 1;
    }
}
```

fails with "loop invariant not satisfied at this loop exit", even though the loop body never runs anything that could violate it - the failure is purely that the zero-iteration case can't see the outer `reveal(X)`. After this PR, it verifies.

A `while cond { body } ensures P` loop with a bare `ensures` (no matching `invariant`) always got desugared into `loop { if !cond { break }; body }`, moving the exit check for P entirely inside the isolated, spun-off loop-body query - which has no access to anything established before the loop (like the `reveal()` above). So even the simplest case - the loop never running at all - couldn't use outer context to prove P.

Root cause: a combined `invariant X ensures X` clause gets its exit check "for free" by induction (assert X once before the loop, re-prove it's maintained at each loop-back) - no separate exit check needed. A bare `ensures` has no such induction to piggyback on, since nothing maintains it between iterations.

Fix: for while-loops whose invariants are *all* bare `ensures` and have no `break`, keep `cond` in the outer query and split the exit obligation into two checks:
- a new assert in the outer query, before any havoc, using full pre-loop context (the actual fix, covers zero iterations)
- a new assert at the existing loop-back point inside the spun-off query, using inner-query context only (unchanged power, covers >=1 iterations)

Scoped to exclude for-loops, atomic-update loops, and any loop with an explicit `break` (real breaks are unavoidably inner-query-only, since they can occur anywhere in the body). This does *not* extend to loops that run more than zero iterations and still depend on an outer opaque fact - that remains addressable via `loop_isolation_boundary` or an inline reveal, unchanged.

Also updates `while_to_loop_ensures` in loops.rs, which was asserting a previously-real limitation of the same kind - now fixed, so `Err` → `Ok`.

The doc comment above `loop_to_stmts` in `sst_to_air.rs` (which already sketches the exact outer/inner AIR shape for the `loop` and `while`-with-`invariant` cases) now has a matching paragraph for this bare-`ensures` case, so the split-assert's soundness can be checked by reading it directly.

Verified: vstd still verifies fully (2043/0); the issue's own repro now verifies without the inline-reveal workaround; a deliberately false ensures clause still correctly fails; full rust_verify_test suite passes.

### DIFF
--- a/source/rust_verify_test/tests/loops.rs
+++ b/source/rust_verify_test/tests/loops.rs
@@ -666,15 +666,16 @@ test_verify_one_file_with_options! {
 }
 
 test_verify_one_file_with_options! {
+    // bare `ensures`, no `break`: keeps `cond` in the outer query
     #[test] while_to_loop_ensures ["exec_allows_no_decreases_clause"] => verus_code! {
         fn test(b: bool) {
             while b
                 ensures true
             {
             }
-            assert(!b); // FAILS (while converted to loop due to ensures)
+            assert(!b);
         }
-    } => Err(e) => assert_one_fails(e)
+    } => Ok(())
 }
 
 test_verify_one_file_with_options! {
--- a/source/vir/src/ast_to_sst.rs
+++ b/source/vir/src/ast_to_sst.rs
@@ -2674,7 +2674,14 @@ pub(crate) fn expr_to_stm_opt(
             };
             let simple_invs =
                 invs.iter().all(|inv| inv.kind == LoopInvariantKind::InvariantAndEnsures);
-            let simple_while = !has_user_break && simple_invs && cond.is_some() && loop_isolation;
+            // bare-ensures-only loops also keep `cond` outer; see loop_to_stmts
+            let simple_ensures_only = !is_for_loop
+                && !*atomic_call
+                && invs.iter().all(|inv| inv.kind == LoopInvariantKind::Ensures);
+            let simple_while = !has_user_break
+                && (simple_invs || simple_ensures_only)
+                && cond.is_some()
+                && loop_isolation;
 
             if allow_complex_invariants && loop_isolation && !is_for_loop {
                 return Err(error(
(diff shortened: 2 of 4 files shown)


tjhance:
I _think_ this seems fine ... I've long thought it's a bit weird how the presence or absence of certain signatures changes the fundamentals of the encoding, though there are a lot of dimensions to the problem, and off the top of my head, I'm not sure if this specific change makes it better or worse. Definitely should get @Chris-Hawblitzel's opinion, since he probably knows why it was done this way to begin with.

tjhance:
By the way, in the future, I'd recommend including in your PR description a specific code snippet that illustrates the changed behavior. (I had to look at the test file in your commit to understand what this PR was really about.)

Marti2203:
Hi @tjhance , will keep it in mind 😅 thanks! This PR was not ready for review; that is why I put it as a draft as I was looking at old comments, but as somebody coming as a user, I am unsure whether this is a real issue.

Marti2203:
Okay, I see when somebody would use the 3 different "loop invariant"-like forms (`ensures`, `invariant`, `invariant_except_break`  and I do think that this issue needs some review attention.

PR #2854, opened 2026-08-25
Specs for slice-indexing strings

jaylorch:
This PR:
* adds specifications for slice-indexing operations on strings;
* adds `index_postcondition` and `index_mut_postcondition` specification functions to `SliceIndexSpec` to simplify specification of `index` and `index_mut`; and
* strengthens each existing mutable slice-indexing specification by specifying that the sliced object's length is unchanged.

The specifications were written by @jaylorch with some help from GPT-5.6 Sol and some advice from @Chris-Hawblitzel and @zero-to-nat. The test code was written mostly by GPT-5.6 Sol. 


source/vstd/string.rs
```diff
@@ -490,6 +491,69 @@ impl<'a> super::std_specs::iter::IteratorSpecImpl for Chars<'a> {
...
+// There are various types you can use to index into a `str` to get a
+// slice, i.e., to implement `SliceIndex<str>`. Here we indicate, for
+// each such type (e.g., `Range<usize>`), whether an index of that
+// type is valid when applied to a given `&str`.
+
+pub open spec fn str_slice_index_req<R: RangeBoundsSpec<usize>>(range: &R, s: &str) -> bool {
+    &&& slice_range_valid(range, s.spec_bytes().len() as nat)
+    &&& is_char_boundary(s.spec_bytes(), slice_range_start(range))
+    &&& is_char_boundary(s.spec_bytes(), slice_range_end(range, s.spec_bytes().len() as nat))
+}
```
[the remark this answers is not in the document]
jaylorch: Thanks, fixed.

source/vstd/string.rs
```diff
@@ -490,6 +494,77 @@ impl<'a> super::std_specs::iter::IteratorSpecImpl for Chars<'a> {
     }
 }
 
+// There are various types you can use to index into a `str` to get a
+// slice, i.e., to implement `SliceIndex<str>`. Here we indicate, for
+// each such type (e.g., `Range<usize>`), whether an index of that
+// type is valid when applied to a given `&str`.
+#[cfg(not(verus_verify_core))]
```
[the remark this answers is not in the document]
jaylorch: Thanks, done.

parno:
@zero-to-nat Is there any overlap here with the verified specifications you put together for the Rust stdlib effort?

zero-to-nat:
> @zero-to-nat Is there any overlap here with the verified specifications you put together for the Rust stdlib effort?

Yes, there is overlap. However, I think this version is better, because it expresses the range of indices using the `RangeBounds` trait, making it more reusable. We did an equivalent thing by adding spec fns on the `SliceIndex` trait directly.

Perhaps this should not be in scope for this PR, but is there a plan to replace the `assume_specification`s in `vstd/std_specs/slice.rs` for individual impls of `SliceIndex::<[T]>::{index, index_mut}` with a generic postcondition on the `SliceIndexSpec` trait extension? (Also, are there equivalent `assume_specification`s for `str`? I couldn't find them, and I am actually not sure how the test cases in this PR are passing.)

jaylorch:
The reason the new test cases in this PR are passing is that they just perform the indexing operation but don't depend on anything about it. So the fact that there are no specifications about the result of such slice indexing doesn't affect the test cases. Should I add those, or do you already have those as part of the Rust stdlib effort?

jaylorch:
I started writing specs for `index` and `index_mut`, and ran into a question about the way this works in the current `std_specs/slice.rs`. One of the specs says the following, which doesn't seem sufficient to me. Suppose we start with `old(slice) == [10, 20]`, `i.start == 0`, and `i.end == 2`. Then this spec allows `final(slice)` to be `[]` and `final(r)` to be `[]` (since it's not well defined what happens when you take a subrange with an invalid range). Right? I think we need another postcondition like `final(slice).len() == old(slice).len()`.
```
pub assume_specification<T>[ <Range<usize> as SliceIndex<[T]>>::index_mut ](i: Range<usize>, slice: &mut [T]) -> (r: &mut [T])
    ensures
        r@ == old(slice)@.subrange(i.start as int, i.end as int),
        final(r)@ == final(slice)@.subrange(i.start as int, i.end as int),
        final(slice)@ == old(slice)@.subrange(0, i.start as int) + final(r)@ + old(slice)@.subrange(
            i.end as int,
            old(slice)@.len() as int,
        ),
;
```

jaylorch:
I agree with @zero-to-nat that it's not much use to have specs for `index_req` without specifications for `index` and `index_mut`, so I've added the latter (and tests for them).

source/vstd/string.rs
```diff
@@ -505,55 +509,232 @@ pub open spec fn str_slice_index_req<R: RangeBoundsSpec<usize>>(range: &R, s: &s
...
+    old_s: Seq<u8>,
+    final_s: Seq<u8>,
+    r: Seq<u8>,
+    final_r: Seq<u8>,
+) -> bool {
+    let start = slice_range_start(range);
+    let end = slice_range_end(range, old_s.len());
+    &&& r == old_s.subrange(start, end)
+    &&& final_s.len() == old_s.len()
+    &&& final_s.len() == old_s.len()
```
zero-to-nat: It looks like this clause is duplicated

source/vstd/string.rs
```diff
@@ -505,55 +509,232 @@ pub open spec fn str_slice_index_req<R: RangeBoundsSpec<usize>>(range: &R, s: &s
...
-impl super::slice::SliceIndexSpecImpl<str> for core::ops::Range<usize> {
+impl super::slice::SliceIndexSpecImpl<str> for Range<usize> {
     open spec fn index_req(&self, s: &str) -> bool {
         str_slice_index_req(self, s)
     }
 }
 
 #[cfg(all(verus_keep_ghost, not(verus_verify_core)))]
-impl super::slice::SliceIndexSpecImpl<str> for core::ops::RangeFrom<usize> {
+pub assume_specification[ <Range<usize> as SliceIndex<str>>::index ](
```
zero-to-nat, line 634: Is it also possible to make these postconditions part of the `SliceIndexSpec` trait extension, rather than separate `assume_specification`s?

source/rust_verify_test/tests/strings.rs
```diff
@@ -132,15 +132,87 @@ test_verify_one_file! {
...
         use core::ops::{Bound, Index};
+        use core::slice::SliceIndex;
         use vstd::prelude::*;
+        use vstd::seq::lemma_seq_subrange_len;
         use vstd::string::StringSliceAdditionalSpecFns;
         use vstd::utf8::*;
 
         broadcast use group_utf8_lib;
 
+        pub assume_specification[ str::as_bytes_mut ](s: &mut str) -> (b: &mut [u8])
```
zero-to-nat, line 160: Unfortunately, I don't think we can support `str::as_bytes_mut` right now. This is because we do not have a way of ensuring the UTF8 invariant is enforced on `final(b)`, so uses of this function could easily create a `str` which violates the invariant. In vstd, we assume that the UTF8 invariant always holds, because the `View` of a `str` is `Seq<char>` instead of `Seq<u8>`. 

We discussed how we might support functions like this, and it seems one possible avenue would be type invariants with special support for breaking the invariants. We kind of need this feature for the std lib verification effort too (currently we have to use a lot of workarounds). However, I think this feature is still in the design phase. Cc @tjhance in case he has anything to add.

zero-to-nat: Also, I realize that without support for something like `as_bytes_mut`, this makes the postcondition on `index_mut` for `str` less useful. I wonder whether we should still add that postcondition (on top of that for `index`) for completeness. Do you have thoughts? Also cc @parno ?

For adding test cases for `index` and `index_mut`, what would you think about either or both of these:
- add test cases for `index` pre & postcondition (similar to the test cases you added for `index_mut`)
- also add test cases for `index_mut`, but use `#[verifier::external_body]` on the functions like `overwrite_2` and leave a note for why we need to trust those functions right now. I think I would prefer this approach to adding an `assume_specification` for `as_bytes` within the test case, as it could be confusing as to why this trusted spec is not included in vstd itself.

zero-to-nat (commented):
Thank you for your work on adding these specs! Overall the PR looks fine to me, but I think there is some more discussion needed about how to handle mutable references, so I will hold off on approving until that is resolved.

zero-to-nat:
> I started writing specs for `index` and `index_mut`, and ran into a question about the way this works in the current `std_specs/slice.rs`. One of the specs says the following, which doesn't seem sufficient to me. Suppose we start with `old(slice) == [10, 20]`, `i.start == 0`, and `i.end == 2`. Then this spec allows `final(slice)` to be `[]` and `final(r)` to be `[]` (since it's not well defined what happens when you take a subrange with an invalid range). Right? I think we need another postcondition like `final(slice).len() == old(slice).len()`.
> 
> ```
> pub assume_specification<T>[ <Range<usize> as SliceIndex<[T]>>::index_mut ](i: Range<usize>, slice: &mut [T]) -> (r: &mut [T])
>     ensures
>         r@ == old(slice)@.subrange(i.start as int, i.end as int),
>         final(r)@ == final(slice)@.subrange(i.start as int, i.end as int),
>         final(slice)@ == old(slice)@.subrange(0, i.start as int) + final(r)@ + old(slice)@.subrange(
>             i.end as int,
>             old(slice)@.len() as int,
>         ),
> ;
> ```

This is a good question. In our std lib verification project, we also did not have the postcondition about the length of the final sub-slice. @tjhance, would you be able to help answer whether the postcondition  `final(slice).len() == old(slice).len()` is necessary?

source/vstd/string.rs
jaylorch: Thanks for catching that. I wrote that function by hand, so I can't blame AI. :-) I've fixed it.

source/vstd/string.rs
jaylorch: I don't know how to do that, but perhaps someone else on this thread does.

source/rust_verify_test/tests/strings.rs
jaylorch: OK, for now I've just added a comment explaining why this assumption isn't justified in general and is only intended for testing here.

tjhance:
Yes, I think it should be included; in principle, this is true of all slice mut refs, but there's no easy way of exporting such a fact<sup>1</sup>, so it'd be a good idea to just add it in the ensures clause here

<sup>1</sup> Because such an axiom is currently unsound with the prelude axioms if applied to spec mut refs

source/vstd/string.rs
zero-to-nat: My thinking is to add another spec fn(s) [to the SliceIndexSpec trait extension](https://github.com/verus-lang/verus/blob/main/source/vstd/slice.rs#L120) similar to `index_req`, e.g. two new spec fns called `index_ens` and `index_mut_ens`, which would then be added in `ensures` clauses for `index` and `index_mut`. Then, in each `SliceIndexSpecImpl`, you would provide definitions for `index_ens`, and `index_mut_ens`, which would be the same as the postconditions currently stated with `assume_specification`. To me, this structure seems preferred over the current one, because both the pre and postconditions are located in the same place. Note that I believe this change would require changing the `SliceIndexSpecImpl` for `SliceIndex<[T]>` as well.

I hope this makes sense -- let me know if you would like me to take a stab at it.

jaylorch:
> Yes, I think it should be included; in principle, this is true of all slice mut refs, but there's no easy way of exporting such a fact1, so it'd be a good idea to just add it in the ensures clause here

OK, I've updated all the `index_mut` specs to include that.

source/vstd/string.rs
jaylorch: Sure, please be my guest!

jaylorch: Let me know if you need me to give you permission to edit this branch.

jaylorch:
I've added `index_postcondition` and `index_mut_postcondition` to `SliceIndexSpec` as @zero-to-nat suggested. If CI passes, please take a look and let me know if this is what you had in mind.

source/vstd/string.rs
jaylorch: OK, I've made that change.

zero-to-nat (approved):
Hi Jay, thank you for making the changes to the postconditions! I think that everything looks good.

PR #2858, opened 2026-08-27
Add specs for Iterator::zip

parno:
Uses the new `#[verifier::impls_cannot_extend_spec]` attribute to break some cyclic dependencies.


wood-ghost:
Hi, while testing this PR’s `Iterator::zip` specification together with `Iterator::all`, I found that manual trigger is needed for indexing property.

Here is the testcase:

```rust
test_verify_one_file! {
    #[test] zip_all_component_index verus_code! {
        use vstd::prelude::*;

        fn test(left: &[u8], right: &[u8])
            requires
                left.len() == right.len(),
        {
            let all_equal = left
                .iter()
                .zip(right.iter())
                .all(
                    |x: (&u8, &u8)| -> (r: bool)
                        ensures
                            r == (*x.0 == *x.1),
                    {
                        *x.0 == *x.1
                    },
                );

            if all_equal {
                assert forall |i: int| 0 <= i < left@.len() 
                    implies 
                    left@[i] == right@[i]
                by {
                    let zipped = left@.as_ref().zip_truncate(right@.as_ref());

                    // necessary for passing the test
                    let x = zipped[i];
                }
            }
        }
    } => Ok(())
}
```
Currently I work around it by by ensuring indexing equality to `zip_postcondition` and to `<[T]>::iter` in slice,  allowing the assertion body to remain empty.

I wonder whether relying on an unused let binding solely to trigger quantifier instantiation considered an acceptable or idiomatic proof pattern in Verus. Alternatively, should the iterator specifications expose enough indexed relationships for this proof to work automatically?

source/vstd/std_specs/iter.rs
```diff
@@ -373,6 +387,77 @@ impl <I> IteratorSpecImpl for &mut I
...
+            I: Sized + IteratorSpec,
+            U: IntoIterator, //Spec,
+    requires
+        i.obeys_prophetic_iter_laws(),
+        #[trigger] zip_post(i, other, r),
+    ensures
+        call_ensures(U::into_iter, (other,), zip_iter_snd(r)),
+        zip_iter_fst(r) == i,
+        IteratorSpec::remaining(&r) == i.remaining().zip_truncate(zip_iter_snd(r).remaining()),
+        IteratorSpec::will_return_none(&r) ==> i.will_return_none() && zip_iter_snd(r).will_return_none(),
```
tjhance, line 456: The will_return_none line should be `||`

tjhance (changes requested):

source/vstd/std_specs/iter.rs
parno: Good point.  Done.

tjhance (approved):

PR #2860, opened 2026-08-28
fix full-test error for zip

wood-ghost:
I had encountered the same full-test failure while working on my experimental Iterator::zip implementation. Since #2858 now covers the main zip support, I extracted just my fix and opened a small PR against iter-zip4.

 


parno:
Thanks very much for the PR!  I took the `#[verifier::impls_cannot_extend_spec]` attribute for cmp, but I think we can just delete the test's definition of IntoIter, since I don't think it's actually necessary for the aspects of the trait system the test is trying to exercise.  I went ahead and pushed that change to my PR.  Also, apologies that #2858 collided with your PR adding zip and position.  I'll try to take a look at the `position` and `find_map` portions soon.

PR #2862, opened 2026-08-29
Fix `global size_of`/`global layout` facts not usable outside their declaring module

### LINKED ISSUE #1114: global size_of/layout does not work across multiple modules

Marti2203:
Fixes #1114.

The literal usize repro in the issue no longer reproduces, since `arch_word_bits` is already crate-global; the underlying bug still reproduces for `global size_of`/`global layout` on any other type.

Root cause: `global size_of T == N` / `global layout T is ...` expands (`builtin_macros/src/syntax.rs`) into a `#[verus::internal(size_of_broadcast_proof)] broadcast proof fn VERUS_layout_of_T()` whose axiom is emitted unconditionally (`sst_to_air_func.rs` bypasses the usual fuel/reveal gate for these). But `vir/src/prune.rs`'s per-module reachability pass never reaches this lemma for a module other than the one where it's textually declared: it's not called, not in a reveal_group, and nothing auto-broadcasts it, so it's pruned away before other modules' queries ever see the axiom.

Fix: in `prune_krate_for_module_or_krate`, unconditionally reach every function flagged `size_of_broadcast_proof`, regardless of module, matching the feature's own stated intent that these sizes are set once per crate.

Verification: 
* added issue_1114_size_of_cross_module and issue_1114_layout_cross_module to `rust_verify_test/tests/layout.rs` (a type's size/layout declared in one module, used from an unrelated module). Both fail without the fix (repro's own "recommendation not met" error) and pass with it. 
* Full layout.rs (28/28),
*  adts_opaque.rs (10/10),
*  opaque_reveal.rs (24/24), 
* assert_by_compute.rs (43/43), and traits.rs (214/214) all still pass. 
* vstd still verifies 2043/0, unchanged.

### DIFF
--- a/source/rust_verify/src/rust_to_vir.rs
+++ b/source/rust_verify/src/rust_to_vir.rs
@@ -617,6 +617,7 @@ pub fn crate_to_vir<'a, 'tcx>(
     .map_err(|e| vec![e])?;
 
     crate::rust_to_vir_adts::setup_type_invariants(&mut vir).map_err(|e| vec![e])?;
+    crate::rust_to_vir_adts::fix_size_of_broadcast_lemma_visibility(&mut vir);
     vir::traits::set_krate_dyn_compatibility(imported, &mut vir);
 
     let mut fun_warn_configs: HashMap<Fun, Option<WarningConfig>> = HashMap::new();
--- a/source/rust_verify/src/rust_to_vir_adts.rs
+++ b/source/rust_verify/src/rust_to_vir_adts.rs
@@ -14,7 +14,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use vir::ast::{
     CrateId, CtorPrintStyle, Datatype, DatatypeTransparency, DatatypeX, Dt, Fun, Function, Ident,
-    KrateX, Mode, Path, TypX, Variant, VirErr,
+    KrateX, Mode, Path, TypX, Variant, VirErr, Visibility,
 };
 use vir::ast_util::ident_binder;
 use vir::def::field_ident_from_rust;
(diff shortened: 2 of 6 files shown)


tjhance:
What happens when S is private?

```rust
mod m1 {
  struct S { }
  global ...
}

mod m2 {
  fn test() {
    // Is the broadcast axiom exported here? If so, it could cause a crash from ill-typed AIR
  }
}
```

Please check that this works (and add it as a test)

Marti2203:
Thanks for checking! 

Added your exact scenario as a test (`issue_1114_sibling_module_size_of`).

And did the following fixes:

1. `ast_to_sst_crate.rs` applied its own module-visibility check when building a function's SST, independent of `prune.rs`'s reachability. The generated `size_of_broadcast_proof` lemma has no `pub`, so it should be private to its declaring module by default. Fixed by exempting this attribute from that check too.

2. Naively reaching the lemma unconditionally (regardless of whether its target type is reachable) does crash with ill-typed AIR when the type genuinely isn't used anywhere the lemma is now visible - as expected. I fixed it: the lemma is now reached only when its target type is reached, piggybacking on type reachability, which already correctly respects visibility.

tjhance (changes requested):
IMO, this is a bit too much special-casing for my taste. I think the cleanest solution would be to set the visibility to the desired value; then everything else should 'just work'.

That is, some time early, we identify every `global`-based broadcast lemma and set its visibility to match the maximum visibility needed to describe the type. The best place for this would be at the end of `crate_to_vir` (rust_to_vir.rs). All the visibility information will be available at that point; it's relatively early, and there are already a few similar "patch-up" routines there.

Marti2203:
Hi @tjhance , thanks for the feedback! Will prepare a patch in such a case

PR #2864, opened 2026-08-29
builtin_macros: don't force `mut` on exec-level tracked/ghost let bindings

### LINKED ISSUE #1169: Verus doesn't complain about missing 'mut'
Was making a basic lock example, and after I got done I realized I forgot a 'mut' somewhere and Verus didn't complain.

```rust
use vstd::prelude::*;
use vstd::atomic_ghost::*;
use vstd::cell;
use vstd::cell::*;
use vstd::modes::*;

verus!{

struct_with_invariants!{
    struct Lock<T> {
        pub atomic: AtomicBool<_, Option<cell::PointsTo<T>>, _>,
        pub cell: PCell<T>,
    }

    spec fn wf(self) -> bool {
        invariant on atomic with (cell) is (v: bool, g: Option<cell::PointsTo<T>>) {
            match g {
                None => {
                    // When there's no PointsTo, the lock must be taken, thus
                    // the boolean value is 'true'.
                    v == true
                }
                Some(points_to) => {
                    points_to.id() == cell.id()
                      && points_to.is_init()
                      && v == false
                }
            }
        }
    }
}

impl<T> Lock<T> {
    fn new(t: T) -> (lock: Self)
        ensures lock.wf()
    {
        let (cell, Tracked(cell_perm)) = PCell::new(t);
        let atomic = AtomicBool::new(Ghost(cell), false, Tracked(Some(cell_perm)));
        Lock { atomic, cell }
    }

    fn acquire(&self) -> (points_to: Tracked<cell::PointsTo<T>>)
        requires self.wf(),
        ensures points_to@.id() == self.cell.id(), points_to@.is_init()
    {
        loop
            invariant self.wf(),
        {
            let tracked points_to_opt = None;     // Verus/Rust should be complaining there's no 'mut' here.
            let res = atomic_with_ghost!(&self.atomic => compare_exchange(false, true);
                ghost points_to_inv => {
                    tracked_swap(&mut points_to_opt, &mut points_to_inv);
                }
            );
            if res.is_ok() {
                return Tracked(points_to_opt.tracked_unwrap());
            }
        }
    }

    fn release(&self, points_to: Tracked<cell::PointsTo<T>>)
        requires
            self.wf(),
            points_to@.id() == self.cell.id(), points_to@.is_init()
    {
        atomic_with_ghost!(&self.atomic => store(false);
            ghost points_to_inv => {
                points_to_inv = Some(points_to.get());
            }
        );
    }
}

}

fn main() { }
```

Marti2203:
Fixes #1169.

Confirmed with @tjhance on the issue (https://github.com/verus-lang/verus/issues/1169#issuecomment-5462711422) that the current behavior is not intended.

`let tracked x`/`let ghost x`/`Tracked(x)`/`Ghost(x)` bindings written directly in an exec function's body never got real rustc mutability check — `ExecGhostPatVisitor::visit_pat_mut` (`builtin_macros/src/syntax.rs`) always synthesized `let mut #x;` for these regardless of what the user actually wrote, so the borrow checker never flagged a missing `mut`. The equivalent path inside `proof { }` blocks was already handlign it correctly.

Fix: only emit `mut` in the synthesized decl when the user's pattern was declared `mut`.

**Compatibility note**: code relying on the missing `mut` to compile (i.e., mutably borrowing one of these bindings without declaring it `mut` as in the example of #1169 ) will now get a standard borrow checker error. 
Within Verus's own codebase this only affected 2 existing tests (both incidental, not testing this behavior) — vstd itself is unaffected.

Added a regression test, `mut_refs_modes.rs::mut_borrow_of_tracked_local_missing_mut_issue1169`, covering the exact issue repro.

Verified: 
* vstd verifies fully (2045 verified, 0 errors); 
* full `mut_refs_modes`/`lifetime`/`atomics`/`regression`/`struct_with_invariants`/`mutable_params`/`proph`/`prophecy`/`modes` suites all pass.

### DIFF
--- a/source/builtin_macros/src/syntax.rs
+++ b/source/builtin_macros/src/syntax.rs
@@ -1575,7 +1575,11 @@ impl VisitMut for ExecGhostPatVisitor {
                     let span = id.span();
                     let decl = if path_is_ident(&pts.path, "Tracked") {
                         if self.inside_ghost == 0 {
-                            parse_quote_spanned!(span => #[verus::internal(proof)] let mut #x;)
+                            if id.mutability.is_some() {
+                                parse_quote_spanned!(span => #[verus::internal(proof)] let mut #x;)
+                            } else {
+                                parse_quote_spanned!(span => #[verus::internal(proof)] let #x;)
+                            }
                         } else if id.mutability.is_some() {
                             parse_quote_spanned!(span => #[verus::internal(proof)] let mut #x = #tmp_x.get();)
                         } else {
@@ -1583,7 +1587,11 @@ impl VisitMut for ExecGhostPatVisitor {
                         }
                     } else {
                         if self.inside_ghost == 0 {
-                            parse_quote_spanned!(span => #[verus::internal(spec)] #[verus::internal(infer_proph)] let mut #x;)
+                            if id.mutability.is_some() {
+                                parse_quote_spanned!(span => #[verus::internal(spec)] #[verus::internal(infer_proph)] let mut #x;)
+                            } else {
+                                parse_quote_spanned!(span => #[verus::internal(spec)] #[verus::internal(infer_proph)] let #x;)
+                            }
                         } else if id.mutability.is_some() {
                             parse_quote_spanned!(span => #[verus::internal(spec)] let mut #x = #tmp_x.view();)
                         } else {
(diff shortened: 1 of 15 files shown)


source/builtin_macros/src/syntax.rs
```diff
@@ -1575,15 +1575,23 @@ impl VisitMut for ExecGhostPatVisitor {
                     let span = id.span();
                     let decl = if path_is_ident(&pts.path, "Tracked") {
                         if self.inside_ghost == 0 {
-                            parse_quote_spanned!(span => #[verus::internal(proof)] let mut #x;)
+                            if id.mutability.is_some() {
```
tjhance: For simplicity I think you can just do:

```
let mutability = id.mutability;
parse_quote_spanned!(span => #[verus::internal(proof)] let #mutability #x;)
```

This should work since `id.mutability` is an Option which implements ToTokens

tjhance (changes requested):
This looks good to me, except for the minor comment.

Note that this change doesn't work for ghost variables instead of tracked:

```rust
    struct X { } 
    fn test() {
        let ghost x = X { };
        proof {
            x = X { };
        }   
    }   
```

This is probably much more complicated to fix, though, so I'd rather merge the PR as-is. Please make sure to leave the issue open for this (or file a new issue).

Marti2203:
Thanks for the feedback, @tjhance ! Applied the simplification you suggested to all four spots with this pattern.

Confirmed the ghost case and filed #2865 to track it separately, since #1169's own repro (which uses `tracked`) is genuinely fixed by this PR. I will prepare a seperate fix for that as it will probably be more invasive

source/builtin_macros/src/syntax.rs
Marti2203: oh, cool, thanks!

tjhance (approved):
This looks good! I'm going to run our large test suite before merging to see if there's any breakage. Some of the projects might inadvertently be relying on the buggy behavior.

tjhance (changes requested):
(well, it looks like there are also some tests failing for the same reason, so those need to be fixed too)

Marti2203:
@tjhance  😅 yep, taking care of them now!

tjhance:
verita run: https://github.com/verus-lang/verus/actions/runs/33284459515

Marti2203:
Alas :\ the verita run failed

tjhance:
yeah, I figured this would happen. We need to open PRs on the relevant projects to add the missing 'mut' keywords

PR #2883, opened 2026-09-02
[cargo-verus] feat: `cargo verus fmt`

### LINKED ISSUE #2758: [cargo-verus] feat: `cargo verus fmt`
This would behave the same way as `cargo fmt`, finding all files to format based on the Cargo metadata, module info, etc.

mkovaxx:
- Extend `cargo-verus` with new `fmt` subcommand.
- Support the `--check` and `-v` flags as well as usual Cargo options such as `-p` and `--manifest-path`.
- Determine whether a package is Rust or Verus based on `[package.metadata.verus]` in `Cargo.toml`.
- A Verus package can opt out via `fmt-as-rust = true`. Currently used by `builtin_macros`.
- Format Rust projects using `cargo fmt` and Verus projects using `verusfmt`.
- Allow overriding `verusfmt` via the `VERUSFMT` env var.
- Add planning-level unit tests.
- Move `--override-verus-version` to be specific to `new` and the `verify` family of subcommands.
- Update CI to use `cargo verus fmt --check`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,10 +41,8 @@ jobs:
       - name: check rustfmt/verusfmt
         working-directory: ./source
         run: |
-          cargo fmt -- --check
-          # TODO: Should use `cargo verus fmt` when it's implemented:
-          #   https://github.com/verus-lang/verus/issues/2758
-          find ./vstd -name '*.rs' -type f -exec verusfmt --check {} +
+          cargo run -p cargo verus -- fmt --check
+          cargo run -p cargo verus -- fmt --check --manifest-path vstd/Cargo.toml
 
       - name: check cargo fmt for vargo
         working-directory: ./tools/vargo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,8 +41,8 @@ jobs:
       - name: check rustfmt/verusfmt
         working-directory: ./source
         run: |
-          cargo run -p cargo verus -- fmt --check
-          cargo run -p cargo verus -- fmt --check --manifest-path vstd/Cargo.toml
+          cargo run -p cargo-verus -- fmt --check
+          cargo run -p cargo-verus -- fmt --check --manifest-path vstd/Cargo.toml
 
       - name: check cargo fmt for vargo
         working-directory: ./tools/vargo
--- a/source/builtin_macros/Cargo.toml
+++ b/source/builtin_macros/Cargo.toml
@@ -23,6 +23,7 @@ verus_syn = { workspace = true }
 verus_prettyplease = { workspace = true }
 
 [package.metadata.verus]
+fmt-as-rust = true
 is-builtin-macros = true
 
 [features]
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -302,6 +302,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "tempfile",
+ "walkdir",
 ]
 
 [[package]]
--- a/source/cargo-verus/Cargo.toml
+++ b/source/cargo-verus/Cargo.toml
@@ -18,6 +18,7 @@ serde_json = { workspace = true }
 serde = { workspace = true }
 sha2 = { workspace = true }
 tempfile = { workspace = true }
+walkdir = { workspace = true }
 
 [build-dependencies]
 cargo-verus-toolchains = { workspace = true }
--- a/source/cargo-verus/src/cli.rs
+++ b/source/cargo-verus/src/cli.rs
@@ -12,10 +12,6 @@ use clap::{ArgAction, Args, Parser, Subcommand, ValueEnum};
     styles = clap_cargo::style::CLAP_STYLING,
 )]
 pub struct CargoVerusCli {
-    /// Override the version reported by `verus --version`.
-    #[arg(long, global = true)]
-    pub override_verus_version: Option<String>,
-
     #[command(subcommand)]
     pub command: VerusSubcommand,
 }
@@ -57,8 +53,19 @@ pub enum ToolchainSubcommand {
 }
 
 #[derive(Clone, Debug, Args)]
-#[group(required = true, multiple = false)]
+#[group(skip)]
 pub struct NewCommand {
+    #[command(flatten)]
+    pub project_kind: NewProjectKind,
+
+    /// Override the version reported by `verus --version`
+    #[arg(long)]
+    pub override_verus_version: Option<String>,
+}
+
+#[derive(Clone, Debug, Args)]
+#[group(required = true, multiple = false)]
+pub struct NewProjectKind {
     /// Create a binary
     #[arg(short, long)]
     pub bin: Option<String>,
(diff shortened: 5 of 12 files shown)


source/builtin_macros/Cargo.toml
```diff
@@ -23,6 +23,7 @@ verus_syn = { workspace = true }
 verus_prettyplease = { workspace = true }
 
 [package.metadata.verus]
+fmt-as-rust = true
```
mkovaxx, line 26: This is needed because otherwise `cargo verus fmt` assumes that a package with `[package.metadata.verus]` is to be formatted as Verus code.

.github/workflows/ci.yml
```diff
@@ -41,10 +41,8 @@ jobs:
       - name: check rustfmt/verusfmt
         working-directory: ./source
         run: |
-          cargo fmt -- --check
-          # TODO: Should use `cargo verus fmt` when it's implemented:
-          #   https://github.com/verus-lang/verus/issues/2758
-          find ./vstd -name '*.rs' -type f -exec verusfmt --check {} +
+          cargo run -p cargo-verus -- fmt --check
+          cargo run -p cargo-verus -- fmt --check --manifest-path vstd/Cargo.toml
```
mkovaxx, line 45: Currently `vstd` is not included in the Cargo workspace, so it needs a separate invocation.

mkovaxx:
Help output for reference:

```
> cargo verus fmt --help
Format Verus and Rust source files

Usage: cargo verus fmt [OPTIONS] [-- [VERUSFMT_ARGS]...]

Arguments:
  [VERUSFMT_ARGS]...  

Options:
      --check                 Check whether formatting is needed without modifying files
  -v, --verbosity...          Increase verbosity (use -vv for more output)
      --manifest-path <PATH>  Path to Cargo.toml
  -p, --package <PACKAGE>     Specify packages to format
      --all                   Format all workspace packages
  -h, --help                  Print help
```

source/builtin_macros/Cargo.toml
parno: In theory, if verusfmt doesn't see a `verus!`, it should just run `rustfmt` on the file without making other changes, so I might have expected this would work without a special case.  Did you run into an instance where it didn't work?

parno (approved):
Looks good, modulo the question about builtin.

source/builtin_macros/Cargo.toml
mkovaxx: @parno Thanks for the review and digging into the details here! When `fmt-as-rust = true` is set, `cargo verus fmt` uses `cargo fmt` to format the project instead of `verusfmt`. There are benefits to this, such as relying on `rustfmt`'s own module discovery instead of the quick-and-dirty `**/*.rs` glob.

mkovaxx: @parno It seems that `rustfmt` has parsing issues when run on `builtin_macros`, see below. Since we're aiming to reproduce the behavior of `vargo fmt` here, it makes sense to use `cargo fmt` internally, because that's what `vargo` does. (It first runs `cargo fmt` on the whole `source` workspace, then also runs `rustfmt` on `vstd/**/*.rs`.)

```
Error:   × Multiple errors found: [  × Failed to parse
  │       ╭─[~/prog/verus/source/builtin_macros/src/contrib/exec_spec.rs:3412:39]
  │  3411 │         Ok(ts) => quote_vstd! { vstd =>
  │  3412 │             #vstd::prelude::verus! { #(#ts)* }
  │       ·                                       ┬
  │       ·                                       ╰── here
  │  3413 │         }
  │       ╰────
  │   help: Expected one of: COMMENT, bang_str
  │ ,   × Failed to parse
  │     ╭─[~/prog/verus/source/builtin_macros/src/contrib/spec_derive.rs:98:14]
  │  97 │         #vstd::prelude::verus! {
  │  98 │             #input
  │     ·              ┬
  │     ·              ╰── here
  │  99 │             #spec_data
  │     ╰────
  │   help: Expected one of: COMMENT, bang_str
  │ ,   × Failed to parse
  │      ╭─[~/prog/verus/source/builtin_macros/src/is_variant.rs:134:22]
  │  133 │                     #[verifier::inline]
  │  134 │                     #publish
  │      ·                      ┬
  │      ·                      ╰── here
  │  135 │                     #vis fn #fun_ident(&self) -> bool {
  │      ╰────
  │   help: Expected one of: COMMENT, bang_str
  │ ,   × Failed to parse
  │     ╭─[~/prog/verus/source/builtin_macros/src/struct_decl_inv.rs:85:14]
  │  84 │         #vstd::prelude::verus!{
  │  85 │             #stream
  │     ·              ┬
  │     ·              ╰── here
  │  86 │         }
  │     ╰────
  │   help: Expected one of: COMMENT, bang_str
  │ ,   × Failed to parse
  │       ╭─[~/prog/verus/source/builtin_macros/src/syntax.rs:2030:56]
  │  2029 │                                     #[allow(non_snake_case)]
  │  2030 │                                     broadcast proof fn #lemma_ident()
  │       ·                                                        ┬
  │       ·                                                        ╰── here
  │  2031 │                                         ensures
  │       ╰────
  │   help: Expected one of: COMMENT, name
  │ ]
```

.github/workflows/ci.yml
```diff
@@ -38,13 +38,17 @@ jobs:
       - name: setup verusfmt
         run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/verus-lang/verusfmt/releases/latest/download/verusfmt-installer.sh | sh
 
+      - name: check formatting of `dependencies`
```
mkovaxx, line 41: Adding these back because `vargo fmt --check` used to cover them.

source/builtin_macros/Cargo.toml
parno: > It seems that rustfmt has parsing issues when run on builtin_macros,

I'm guessing you meant `verusfmt` here, right? I'm guessing the problem is that `verusfmt` thinks it saw a normal `verus! { .. }` when really we're in the midst of a quote macro, so we should be ignoring it.  

This is probably a rare case, but at least in this instance, it does say that we should be running only rustfmt on this particular crate.

jaybosamiya-ms: This is a bug in the verusfmt parser (it got confused because the macro that should be _outside_ `verus!` has a `verus!` inside it, which it then started to try to parse). I've made a PR for the fix https://github.com/verus-lang/verusfmt/pull/227

parno: Ironically, right after typing that, I see that @jaybosamiya-ms opened a [verusfmt PR](https://github.com/verus-lang/verusfmt/pull/227) that looks like it might address this particular bug.

parno: Sorry for the GitHub-induced causality violation in my message above.

jaybosamiya-ms: I plan to cut a new release by EoW, so hopefully with that in latest verusfmt release, we don't need `fmt-as-rust` as a workaround.

PR #2886, opened 2026-09-02
refactor(air): Use solver-neutral option API names

planethunter86:
The set_z3_param* family and log_set_z3_param predate the cvc5 backend. Both solvers now flow through them, so the Z3-specific names misdescribe what the code does: set_z3_param_bool is what configures cvc5's incremental mode, and the sole path by which any option reaches any solver is log_set_z3_param.

Rename to set_solver_option*/log_set_solver_option, and rename spinoff_z3_context to spinoff_context in the verifier.

Also de-specialize the operator-facing strings on the touched lines. Comments and panic/expect messages in smt_process.rs referred to "Z3 process" when the process may be cvc5, and one message read "receiving data to Z3 process" where it meant "from".

No behavioral change. Only identifiers and message text differ: every SMT option-name literal sent over the wire is byte-identical, and the visibility of each renamed function is unchanged.

Assisted-by: Kiro:claude-opus-5

 


parno (approved):
Looks reasonable to me.  Thanks!

PR #2887, opened 2026-09-02
fix(air): Name the actual solver in version errors

planethunter86:
The version-mismatch diagnostic hardcoded "z3". The expected version it quotes comes from the configured solver, so running the cvc5 backend against an unexpected cvc5 build produced a message that named the wrong solver alongside a version z3 has never had:

    error: expected z3 version 1.1.2, found 1.3.4

Add SmtSolver::name() and thread it through MessageInterface so the message names whichever solver is in use. That same run now reports:

    error: expected cvc5 version 1.1.2, found 1.3.4

and the z3 path is unaffected:

    error: expected z3 version 4.16.0, found 4.15.4

This widens the MessageInterface::unexpected_z3_version trait method to unexpected_solver_version, which takes the solver name. Both implementations in the tree, AirMessageInterface and VirMessageInterface, are updated; there are no others.

Assisted-by: Kiro:claude-opus-5

 

### DIFF
--- a/source/air/src/smt_verify.rs
+++ b/source/air/src/smt_verify.rs
@@ -186,14 +186,15 @@ pub(crate) fn smt_check_assertion<'ctx>(
                 let value: &str = &line[GET_VERSION_RESPONSE_PREFIX.len()..line.len() - 1];
                 let version = value.trim_matches(&[' ', '"'][..]);
                 if version != expected_version.as_str() {
-                    diagnostics.report(
-                        &context
-                            .message_interface
-                            .unexpected_z3_version(&expected_version, version),
-                    );
+                    let solver = context.solver.name();
+                    diagnostics.report(&context.message_interface.unexpected_solver_version(
+                        solver,
+                        &expected_version,
+                        version,
+                    ));
                     panic!(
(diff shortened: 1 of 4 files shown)


source/air/src/smt_verify.rs
```diff
@@ -186,14 +186,17 @@ pub(crate) fn smt_check_assertion<'ctx>(
                 let value: &str = &line[GET_VERSION_RESPONSE_PREFIX.len()..line.len() - 1];
                 let version = value.trim_matches(&[' ', '"'][..]);
                 if version != expected_version.as_str() {
-                    diagnostics.report(
-                        &context
-                            .message_interface
-                            .unexpected_z3_version(&expected_version, version),
-                    );
+                    // Name the solver actually in use: this check runs for
```
parno: I think we can remove this (AI-generated) comment.  It refers to the old code, so it's not useful to future readers.

parno:
I think this looks good, modulo the comment above and a rebase.

source/air/src/smt_verify.rs
planethunter86: done

PR #2889, opened 2026-09-02
fork rustc_hir_analysis and delay certain queries

### LINKED ISSUE #2521: Panic when a trait returns `impl Future` and a `verus!` block contains an `async fn`
`cargo verus verify` panics on async functions if outside of the `verus!` block, there's a trait method returning an `impl Future`:
```rust
use std::future::Future;
use vstd::prelude::*;

trait T {
    fn f() -> impl Future;
}

struct S;

impl T for S {
  fn f() -> impl Future { async {} }
}

verus! {
  async fn foo();
}
```
~~Error:~~
```
thread panicked at air/src/parser.rs:890:50:
  called `Result::unwrap()` on an `Err` value: IllegalChrInString { pos: Pos
  { line: 0, column: 68 }, chr: 10 }
```

UPDATE: the correct error is
```
thread 'rustc' (2294115) panicked at rustc_mir_build/src/../../rustc_mir_build_additional_files/verus.rs:176:21:
Internal Verus Error: The thir_body query is running for item DefId(0:14 ~ verus_async_repro[324a]::foo) which may require erasure, but the VerusErasureCtxt has not been initialized.
```

tjhance:
fixes https://github.com/verus-lang/verus/issues/2521

The `rustc_hir_analysis::check_crate` can depend on borrowck results in 2 cases (that I'm aware of):

 * DefKind::OpaqueType
 * DefKind::Impl (when an opaque type is involved)

This causes a problem for us as `rustc_hir_analysis::check_crate` runs before rust_to_vir.

This PR forks `rustc_hir_analysis` in order to delay the problematic checks until Verus's borrowck phase.

I think forking the crate is not entirely necessary, but the alternatives are either:
  * Lots of copy/pasted code
  * Delaying more checks that necessary

(Also, I kinda wanted to fork this crate anyway, as it's one of the crates needed to removed 'Sized' bounds in ghost code.)

 

### DIFF
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -2336,6 +2336,7 @@ dependencies = [
  "num-format",
  "regex",
  "rust_verify_test_macros",
+ "rustc_hir_analysis",
  "rustc_mir_build",
  "serde",
  "serde_json",
@@ -2367,6 +2368,14 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.119",
+]
+
+[[package]]
+name = "rustc_hir_analysis"
+version = "0.0.0"
+dependencies = [
+ "itertools 0.12.1",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


Chris-Hawblitzel (approved):

tjhance:
need to update version before merging

PR #2894, opened 2026-09-03
Add spec for slice `get_mut`

jaylorch:
This PR adds a spec for the `get_mut` operation on slices.

This code was largely written by AI, specifically GPT-5.6 Sol.

### DIFF
--- a/source/rust_verify/src/rust_to_vir_impl.rs
+++ b/source/rust_verify/src/rust_to_vir_impl.rs
@@ -687,6 +687,8 @@ pub(crate) fn collect_external_trait_impls<'tcx>(
     }
 
     // Process only the new implementations that could be visible to Verus:
+    let auto_import_impl_paths: HashSet<Path> =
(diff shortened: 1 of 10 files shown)


jaylorch:
I found an issue with this. It doesn't work with the following example. I'm asking the AI to fix it.
```
use vstd::prelude::*;

pub struct Error(
    Box<dyn std::error::Error + Send + Sync + 'static>,
);

impl core::ops::Deref for Error {
    type Target =
        dyn std::error::Error + Send + Sync + 'static;

    fn deref(&self) -> &Self::Target {
        &*self.0
    }
}

impl core::ops::DerefMut for Error {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut *self.0
    }
}

verus! {

#[verifier::external_body]
pub struct ExError(Error);

fn vec_get_mut(
    values: &mut Vec<u8>,
    index: usize,
) -> Option<&mut u8> {
    values.get_mut(index)
}

}

fn main() {}
```

jaylorch:
Superceded by #2895.

PR #2899, opened 2026-09-04
support pattern matching in closure parameters

wood-ghost:
Follow discussion #1540 

The PR supports  pattern matching in closure parameters, including reference, tuple, struct, and tuple-struct patterns (all covered by regression tests).

- Create `pattern_stmts` to store destructuring declarations
- Add `closure_pat_to_mut_var`
    - Preserve the existing path for simple bindings `pat_to_mut_var`
    - Create a hidden parameter and add its destructuring declaration to `pattern_stmts`.
- Modify `PatKind::Ref` branch in `pattern_to_vir_unadjusted` to handle ordinary reference patterns, respecting rustc’s `skipped_ref_pats` and continuing to reject pinned reference patterns.
- Wrap each `requires` and `ensures` expression, and the closure body, with the generated destructuring declarations so pattern-bound variables are available in each scope.

The implementation preserves VIR’s existing representation of closure parameters as individual variables, along with the original argument types and closure arity.

Assisted-by: GPT-5.6 Sol, GPT-6 Astra
 

### DIFF
--- a/source/rust_verify/src/rust_to_vir_expr.rs
+++ b/source/rust_verify/src/rust_to_vir_expr.rs
@@ -309,6 +309,38 @@ pub(crate) fn pat_to_mut_var<'tcx>(pat: &Pat) -> Result<(bool, VarIdent), VirErr
     }
 }
 
+fn closure_pat_to_mut_var<'tcx>(
+    bctx: &BodyCtxt<'tcx>,
+    pat: &Pat<'tcx>,
+    typ: &Typ,
+    pattern_stmts: &mut Vec<vir::ast::Stmt>,
+) -> Result<(bool, VarIdent), VirErr> {
+    // Preserve the existing path for single bindings
+    if matches!(pat.kind, PatKind::Binding(_, _, _, None)) {
+        return pat_to_mut_var(pat);
+    }
+
+    let name = str_unique_var(
+        "%closure_param",
+        vir::ast::VarIdentDisambiguate::RustcId(pat.hir_id.local_id.index()),
+    );
+    let pattern = pattern_to_vir(bctx, pat)?;
+    let init = SpannedTyped::new(&pattern.span, typ, PlaceX::Local(name.clone()));
+
+    pattern_stmts.push(bctx.spanned_new(
+        pat.span,
+        StmtX::Decl {
+            pattern,
+            mode: Some((Mode::Exec, vir::ast::DeclProph::Default)),
+            init: Some(init),
+            els: None,
+            assert_irrefutable: false,
+        },
+    ));
+
+    Ok((false, name))
+}
+
 pub(crate) fn pat_to_var<'tcx>(pat: &Pat) -> Result<VarIdent, VirErr> {
     let (_, name) = pat_to_mut_var(pat)?;
     Ok(name)
@@ -857,10 +889,27 @@ pub(crate) fn pattern_to_vir_unadjusted<'tcx>(
             PatternX::Range(e1, e2)
         }
         PatKind::Guard(..) => unsupported_err!(pat.span, "pattern guards", pat),
-        PatKind::Ref(..) => {
-            // note: to handle this, you need to check skipped_ref_pats
-            // see rustc_mir_build/src/thir/pattern/mod.rs
-            unsupported_err!(pat.span, "ref patterns", pat);
+        PatKind::Ref(subpat, pinnedness, mutability) => {
+            // pinned patterns such as `&pin const x` unsupported
+            unsupported_err_unless!(
+                matches!(pinnedness, rustc_hir::Pinnedness::Not),
+                pat.span,
+                "pinned ref patterns"
+            );
+
+            // Follow rustc_mir_build/src/thir/pattern/mod.rs: when rustc marks this reference
+            // pattern as skipped, lower only its inner pattern.
+            // Adding a reference wrapper here would introduce an extra layer.
+            if bctx.types.skipped_ref_pats().contains(pat.hir_id) {
+                return pattern_to_vir(bctx, subpat);
+            }
+
+            let subpattern = pattern_to_vir(bctx, subpat)?;
+
+            match mutability {
+                Mutability::Not => PatternX::ImmutRef(subpattern), // '&' pattern
+                Mutability::Mut => PatternX::MutRef(subpattern),   // '&mut' pattern
+            }
         }
         PatKind::Slice(..) => unsupported_err!(pat.span, "slice patterns", pat),
         PatKind::Never => unsupported_err!(pat.span, "never patterns", pat),
@@ -4090,6 +4139,9 @@ pub(crate) fn closure_to_vir<'tcx>(
 
         let typs = closure_param_typs(bctx, closure_expr)?;
         assert!(typs.len() == body.params.len());
+
+        let mut pattern_stmts: Vec<vir::ast::Stmt> = Vec::new();
+
         let params: Vec<VarBinder<Typ>> = body
             .params
             .iter()
@@ -4101,7 +4153,7 @@ pub(crate) fn closure_to_vir<'tcx>(
                     return err_span(x.span, "closures only accept exec-mode parameters");
                 }
 
-                let (_is_mut, name) = pat_to_mut_var(x.pat)?;
+                let (_is_mut, name) = closure_pat_to_mut_var(bctx, x.pat, &t, &mut pattern_stmts)?;
                 Ok(Arc::new(VarBinderX { name, a: t }))
             })
             .collect::<Result<Vec<_>, _>>()?;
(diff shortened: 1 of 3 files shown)


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -309,6 +309,38 @@ pub(crate) fn pat_to_mut_var<'tcx>(pat: &Pat) -> Result<(bool, VarIdent), VirErr
     }
 }
 
+fn closure_pat_to_mut_var<'tcx>(
+    bctx: &BodyCtxt<'tcx>,
+    pat: &Pat<'tcx>,
+    typ: &Typ,
+    pattern_stmts: &mut Vec<vir::ast::Stmt>,
+) -> Result<(bool, VarIdent), VirErr> {
+    // Preserve the existing path for single bindings
```
tjhance: this comment is hard to follow, what "existing path"?

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -309,6 +309,38 @@ pub(crate) fn pat_to_mut_var<'tcx>(pat: &Pat) -> Result<(bool, VarIdent), VirErr
...
+    if matches!(pat.kind, PatKind::Binding(_, _, _, None)) {
```
tjhance: also check that the BindingMode's ByRef is ByRef::No

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -309,6 +309,38 @@ pub(crate) fn pat_to_mut_var<'tcx>(pat: &Pat) -> Result<(bool, VarIdent), VirErr
     }
 }
 
+fn closure_pat_to_mut_var<'tcx>(
```
tjhance: Add documentation that this only works for exec-mode closures (or name it `exec_closure_pat_to_mut_var`, or something)

tjhance (changes requested):
Thanks for working on this!

It looks like there are some problems if the pattern contains any ref mut bindings (vir::ast::MutRef). For one thing, you won't be able to lower such patterns in spec code (as you'd need to for the requires and ensures). For another, it needs support in the rustc_mir_build fork.

Specifically, if I try:

```
fn test() {
    let y = |(x, y): &mut (u64, u64)| {
    };
}
```

I get a `error: Verus Internal Error: setup_verus_ctxt_for_thir_erasure failed, var lookup failed`.

IMO, you should just disable this case for now, i.e., check if there are any ByRef::Mut in the Pattern and if so, return a 'not supported' error.

source/rust_verify_test/tests/exec_closures.rs
```diff
@@ -158,6 +158,53 @@ test_verify_one_file_with_options! {
...
+        fn testfn() {
+            let is_seven = |ref x: u8| -> (result: bool)
+                ensures result == (*x == 7)
+            { *x == 7 };
+            let result = is_seven(7u8);
+            assert(result);
+        }
+    } => Ok(())
+}
+
```
tjhance, line 207: I would also like to see some failure cases (e.g. precondition failed) in the test cases.

source/rust_verify/src/rust_to_vir_expr.rs
wood-ghost: Thank you for the comment! Sorry that I should not write comments like commits., and have changed it with the help of ai.

source/rust_verify/src/rust_to_vir_expr.rs
wood-ghost: I changed the name to `exec_closure_pat_to_mut_var`

source/rust_verify/src/rust_to_vir_expr.rs
wood-ghost: I want to reuse `pattern_find_mut_binding` in source/vir/src/patterns.rs, but it is not visible outside, so I create a helper function `check_exec_closure_param_pattern` in source/vir/src/ast_util.rs.

source/rust_verify_test/tests/exec_closures.rs
wood-ghost: Thank you! I have added more failure cases:
```Rust
|(x, y): &mut (u64, u64)| {}                 // Implicit ref mut
|ref mut x: u64| {}                         // Explicit ref mut
|Pair { values: (x, ref mut y) }: Pair| {}   // Nested ref mut
```
and some contract related testcases.

wood-ghost:
> Thanks for working on this!
> 
> It looks like there are some problems if the pattern contains any ref mut bindings (vir::ast::MutRef). For one thing, you won't be able to lower such patterns in spec code (as you'd need to for the requires and ensures). For another, it needs support in the rustc_mir_build fork.
> 
> Specifically, if I try:
> 
> ```
> fn test() {
>     let y = |(x, y): &mut (u64, u64)| {
>     };
> }
> ```
> 
> I get a `error: Verus Internal Error: setup_verus_ctxt_for_thir_erasure failed, var lookup failed`.
> 
> IMO, you should just disable this case for now, i.e., check if there are any ByRef::Mut in the Pattern and if so, return a 'not supported' error.

Thank you so much for the review! Hope that everything is fixed now. I'll be more considerable next time :)

source/rust_verify/src/rust_to_vir_expr.rs
tjhance: You can just mark the existing function public.

source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -328,7 +328,20 @@ fn exec_closure_pat_to_mut_var<'tcx>(
...
+            &span,
+            "mutable-reference bindings in closure parameters are not supported",
+        ));
+    }
+
+    // Register the adjusted pattern's span for erasure of the generated initializer.
+    let mut erasure_info = bctx.ctxt.erasure_info.borrow_mut();
+    let mapping = (pat.hir_id, pattern.span.id);
+    if !erasure_info.hir_vir_ids.contains(&mapping) {
+        erasure_info.hir_vir_ids.push(mapping);
```
tjhance: Why is this necessary? Is there a specific test case that failed if it isn't included?

tjhance:
OK, I pulled your commit to experiment locally and see what's going on with the span ids. I think I found the circumstances that led you add this line, but it's a real mess.

First of all, you shouldn't need to add any mapping ids for "adjusted patterns". This isn't the case for any other patterns (in match expressions or let decls) so it shouldn't be the case here.

The reason it _seems_ like you do is that there are duplicate AstIds. The `init` expression had the same ID as the pattern, causing the two nodes to get confused.

You need to use `spanned_typed_new` to get a fresh ID:

```rust
-    let init = SpannedTyped::new(&pattern.span, typ, PlaceX::Local(name.clone()));
+    let init = bctx.spanned_typed_new(pat.span, typ, PlaceX::Local(name.clone()));
```

In rust_to_vir, always use the `spanned_typed_new` function. 

Okay, now there's still the issue that `erase.rs` expects the `id_to_hir` mapping to have an entry for every Local node, including the one in your `init` expression. This is kind of a problem, because there's no HirId that corresponds to this node. IMO, `hir_vir_ids` really ought to take an Option<HirId> for this situation, though you could also use a junk HirId (like the HirId of the closure) which will cause it to be ignored later. This is kind of jank, though, and obfuscates the purpose of the mapping.

I think if you make these fixes, everything should be working again. However, there are still a couple of other things I'm worried about:

 * `pattern_stmts` gets cloned (for the requires and ensures), which duplicates the pattern IDs. This seems to work correctly now, but I think this should be fixed. It could result in the modes getting mixed up.
   * Unfortunately, I don't think we have a good function for "clone a node and create fresh ids". There is a similar function, `cleanup_span_ids`, but it doesn't support patterns. 
 * There's another `SpannedTyped::new` when you create the `ExprX::Block`, which should again be `spanned_typed_new`.

source/rust_verify/src/rust_to_vir_expr.rs
wood-ghost: Sorry for the late reply due to time difference.  The failing testcases are 
```Rust
    #[test] test_pattern_implicit_shared_ref ["vstd"] => verus_code! {
        fn test() {
            let f = |(x, y): &(u64, u64)| -> (result: u64)
                requires *x == *y
                ensures result == *x
            { *y };
            let pair = (7, 7);
            let result = f(&pair);
            assert(result == 7);
        }
    } => Ok(())
}

    #[test] test_pattern_implicit_shared_ref_requires_fail ["vstd"] => verus_code! {
        fn test() {
            let f = |(x, y): &(u64, u64)|
                requires *x == *y
            {};
            let pair = (7, 8);
            f(&pair); // FAILS
        }
    } => Err(err) => assert_one_fails(err)
}

**Comment by wood-ghost** [comment]
Thank you for the review! I’ll address the remaining issues as follows:
- [x] Merge the latest main into my working branch.
- [x] Represent generated nodes explicitly
    - Change `hir_vir_ids` to `Vec<(Option<HirId>, AstId)>`, using None for generated VIR nodes with no corresponding Rust HIR node
    - Update erasure setup to handle these entries while retaining errors for genuinely missing mappings
- [x]  Give the hidden-parameter initializer a fresh ID and register it with None
- [x] Give contract copies independent IDs
    - Support `Pattern` as `fp` in `source/vir/src/ast_visitor.rs`
    - Register the generated contract copies with `None` so their mode information does not affect erasure of the executable source bindings
    - Use fresh IDs for the generated block expressions
- [x] Testing

File: source/rust_verify/src/spans.rs [hunk 6]
```diff
@@ -318,6 +318,23 @@ impl SpanContextX {
 }
 
 impl<'tcx> crate::context::ContextX<'tcx> {
+    /// Clones generated statements with fresh IDs, while preserving diagnostic locations. 
+    pub(crate) fn clone_stmts_with_fresh_ids(&self, stmts: &[vir::ast::Stmt]) -> vir::ast::Stmts {
```
wood-ghost, line 322: I'm not sure where to put this helper function. I put it here just because I want to call it conveniently :-(

source/rust_verify/src/spans.rs
```diff
@@ -318,6 +318,23 @@ impl SpanContextX {
...
+        Arc::new(
+            stmts
+                .iter()
+                .map(|stmt| {
+                    vir::ast_visitor::map_stmt_spans(stmt, &mut |span| {
+                        let mut span = span.clone();
+                        span.id = self.spans.get_next_span_id();
+                        self.erasure_info.borrow_mut().hir_vir_ids.push((None, span.id));
+                        span
```
tjhance, line 331: This only applies to statements, but not expressions or patterns. (I'm confused how the temporary test passes, since the test seems to be checking the pattern IDs.)

wood-ghost: > This only applies to statements, but not expressions or patterns. (I'm confused how the temporary test passes, since the test seems to be checking the pattern IDs.)

Thanks for the comment! I think `map_stmt_spans` only uses a statement as the entry point and it recursively maps its child nodes. It takes a statement and a span-transforming closure, `f_span`, which handles new ID here.
The call chain in `source/vir/src/ast_visitor.rs` is shown below and expressions or patterns are visited.
```Rust
map_stmt_spans
    ->map_stmt_visitor_env
        -> visit_stmt
            -> visit_stmt_rec
                -> visit_pattern / visit_place / visit_expr
```
`FP` is added to `MapExprStmtTypVisitor` following other type like `FE`. Its `visit_pattern` first calls `visit_pattern_rec` to process nested patterns, then applies `fp` to the resulting pattern. In `map_stmt_spans`, the `fp` closure calls `f_span` on the pattern’s span, so each visited pattern receives a fresh ID. The expression and place callbacks apply `f_span` similarly.
That's why I think it passes the test checking pattern IDs.

tjhance (approved):
Thanks for working on this, and for dealing with the id madness!

PR #2900, opened 2026-09-04
Update Verita for Vest 2.0 layout

y1ca1:
Vest 2.0 moved the verified combinator library to `vest_lib` and the DSL generated files to `vest_tests`. The old `vest` target is now the DSL compiler, while `vest-examples/tls_dsl` no longer exists. Generated Rust code are checked into `vest_tests/src`, so `Verita` no longer needs to install the compiler before verification.

There is a separate compatibility issue: Vest's currently pinned `vstd` predates the newly-added standard slice `PartialEq` specification. Vest carries a duplicate local spec and nightly Verus complains about it. 

Vest will remove that local spec when Verus publish the next release that contains the slice `PartialEq` spec. This PR corrects Verita's stale project layout and the Vest verification run should become green again once the next Verus public release comes out.


parno:
Thanks for the update.  It looks like we previously only verified the TLS example, and now the `vest_tests` target will verify all of the examples.  Do you know about how long that will take?

A new Verus release also came out over the weekend, so we should be set to point Vest at the newer release.

y1ca1:
According to the [latest Vest CI](https://github.com/secure-foundations/vest/actions/runs/34074847874/job/101598932098), it looks like the entire `vest_tests` can finish within 10 mins. But maybe we can omit that in Verita if Vest's CI already covers it? If a Verus update triggers a regression in Verita-`vest_lib`, it's not going to pass `vest_tests` anyway.

The new Verus release has [a separate issue](https://github.com/verus-lang/verus/issues/2914) that's blocking Vest's current CI. But I think I can work around it temporarily.

parno:
@y1ca1 What if we split the existing Vest project entry into two entries, one for the DSL, and one for an example of the emitted code?  Then the latter project can pass the `--verify-module` argument to focus on one example of the generated code, rather than all of the examples.

PR #2904, opened 2026-09-04
vstd: add IteratorSpec for core::iter::Enumerate

arxgy:
[`vstd/std_specs/iter.rs`](https://github.com/verus-lang/verus/blob/main/source/vstd/std_specs/iter.rs) provides `IteratorSpec` implementations for iterator adapters such as `Rev`, but not for `core::iter::Enumerate`. A function that consumes an `enumerate` therefore fails to verify:

```rust
fn tag_indices(v: Vec<u32>) -> (out: Vec<(usize, u32)>)
    ensures out@ == v@.map(|k: int, x: u32| ((k as usize), x)),
{
    let mut out: Vec<(usize, u32)> = Vec::new();
    let it = v.into_iter();
    let en = it.enumerate();
    for (i, x) in en {
        out.push((i, x));
    }
    out                     // fails: `Enumerate` is not supported (Enumerate has no IteratorSpec)
}
```

## Fix

This PR adds an `IteratorSpec` for `core::iter::Enumerate`. It contributes:

- `remaining()`: the items the iterator will yield. It is the inner iterator's remaining items, each paired with the index it will be reached at.
- `enumerate_postcondition`: a broadcast axiom giving `enumerate_iter(r) == i` and `enumerate_count(r) == 0`, the inner iterator and the counter the result starts from.
- `peek()`: the inner iterator's guess, paired with the same index, so that a `for`-loop's `(i, x)` pattern is usable in its invariant.
- an `open` `decrease()`, `will_return_none()` and `obeys_prophetic_iter_laws()`, each forwarding to the inner iterator.

After introducing `Enumerate` in this PR, we can reason the `Enumerate` iterator in Verus. And the problem code above can be proved as below:
```diff
 fn tag_indices(v: Vec<u32>) -> (out: Vec<(usize, u32)>)
     ensures out@ == v@.map(|k: int, x: u32| ((k as usize), x)),
 {
     let mut out: Vec<(usize, u32)> = Vec::new();
     let it = v.into_iter();
     let en = it.enumerate();
-    for (i, x) in en {
+    for (i, x) in iter: en
+        invariant out@ == iter.seq().take(iter.index()),
+    {
         out.push((i, x));
     }
     out
 }
```

A `for`-loop-over-`enumerate` regression test is included.
 

### DIFF
--- a/source/rust_verify_test/tests/iterators.rs
+++ b/source/rust_verify_test/tests/iterators.rs
@@ -481,3 +481,87 @@ test_verify_one_file! {
 
     } => Ok(())
 }
+
+test_verify_one_file! {
+    #[test] enumerate_works verus_code! {
+        use vstd::prelude::*;
+
+        // The index the loop yields is the loop index, and the item is the
+        // corresponding element of the source.
+        fn copy_it(v: &Vec<u32>) -> (w: Vec<u32>)
+            ensures
+                w.len() == v.len(),
+                forall |k: int| 0 <= k < w.len() ==> w[k] == v[k],
+        {
+            let mut w: Vec<u32> = Vec::new();
+            for (i, x) in iter: v.iter().enumerate()
+                invariant
+                    w.len() == iter.index(),
+                    forall |k: int| 0 <= k < w.len() ==> w[k] == v[k],
+            {
+                // `w.len() == iter.index()` bounds the loop index by `usize::MAX`,
+                // which is what relating it to `Enumerate`'s `usize` index needs.
+                assert(i == iter.index());
+                w.push(*x);
+            }
+            w
+        }
+
+        // The `(i, x)` pattern is usable in the invariant (via `peek`).
+        // `Enumerate`'s index is a `usize`, so relating it to the (unbounded) loop
+        // index needs `v@.len() <= usize::MAX`, which the `v.len()` call surfaces.
+        #[verifier::loop_isolation(false)]
+        fn peek_binding(v: &Vec<u32>) {
+            let n = v.len();
+            for (i, x) in iter: v.iter().enumerate()
+                invariant
+                    iter.index() < iter.seq().len() ==>
+                        i == iter.index() && x == &v[iter.index()],
+            {
+            }
+        }
+
+        fn over_range(n: usize) {
+            for (i, x) in iter: (0..n).enumerate()
+                invariant
+                    iter.index() < iter.seq().len() ==> i == x,
+            {
+                assert(i == x);
+            }
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] enumerate_spec_level verus_code! {
+        use vstd::prelude::*;
+        use vstd::std_specs::iter::IteratorSpec;
+
+        fn remaining_of_fresh_enumerate(v: &Vec<u32>) {
+            let it = v.iter().enumerate();
+            assert(IteratorSpec::remaining(&it).len() == v.len());
+            assert(forall |k: int| 0 <= k < v.len() ==>
+                #[trigger] IteratorSpec::remaining(&it)[k] == ((k as usize), &v[k]));
+        }
+
+        fn collect_it(v: &Vec<u32>) {
+            let w: Vec<(usize, &u32)> = v.iter().enumerate().collect();
+            assert(w.len() == v.len());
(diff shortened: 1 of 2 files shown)


source/vstd/std_specs/iter.rs
```diff
@@ -350,6 +356,71 @@ impl <I> DoubleEndedIteratorSpecImpl for Rev<I>
...
+    where I: Iterator + IteratorSpec {
+
+    open spec fn obeys_prophetic_iter_laws(&self) -> bool {
+        enumerate_iter(*self).obeys_prophetic_iter_laws()
+    }
+
+    // Item `k` of what's left to yield is the inner iterator's item `k`, tagged with
+    // the index it will be reached at: `count` items have been consumed already.
+    #[verifier::prophetic]
+    open spec fn remaining(&self) -> Seq<(usize, I::Item)> {
```
parno, line 402: In our other iterators, we have marked `remaining`, `will_return_none`, and `decrease` as closed, `uninterp` functions, and provided the necessary knowledge via the `*_postcondition` conditions.  See, for example, [the way `Map` is handled](https://github.com/verus-lang/verus/blob/382a9921e515abf26bfcd0d219623d252051e04e/source/vstd/std_specs/iter.rs#L416-L430).  For consistency, can we update this iterator accordingly?

source/rust_verify_test/tests/iterators.rs
```diff
@@ -481,3 +481,87 @@ test_verify_one_file! {
...
+        {
+            let mut w: Vec<u32> = Vec::new();
+            for (i, x) in iter: v.iter().enumerate()
+                invariant
+                    w.len() == iter.index(),
+                    forall |k: int| 0 <= k < w.len() ==> w[k] == v[k],
+            {
+                // `w.len() == iter.index()` bounds the loop index by `usize::MAX`,
+                // which is what relating it to `Enumerate`'s `usize` index needs.
+                assert(i == iter.index());
```
parno, line 504: Is this assert necessary for the proof or is it illustrative?  Either way, the comment is a bit unclear.

source/rust_verify_test/tests/iterators.rs
```diff
@@ -481,3 +481,87 @@ test_verify_one_file! {
...
+                w.push(*x);
+            }
+            w
+        }
+
+        // The `(i, x)` pattern is usable in the invariant (via `peek`).
+        // `Enumerate`'s index is a `usize`, so relating it to the (unbounded) loop
```
parno, line 511: I'm not sure this line or the next is needed; it seems more confusing than clarifying.

source/rust_verify_test/tests/iterators.rs
```diff
@@ -481,3 +481,87 @@ test_verify_one_file! {
...
+                    iter.index() < iter.seq().len() ==> i == x,
+            {
+                assert(i == x);
+            }
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
+    #[test] enumerate_spec_level verus_code! {
```
parno, line 536: This test seems tautological, since it's just repeating the relevant definitions in `remaining_of_fresh_enumerate`, and the second one seems more about `collect`'s functionality.  Let's cut the test.

source/rust_verify_test/tests/iterators.rs
```diff
@@ -481,3 +481,87 @@ test_verify_one_file! {
...
+        fn collect_it(v: &Vec<u32>) {
+            let w: Vec<(usize, &u32)> = v.iter().enumerate().collect();
+            assert(w.len() == v.len());
+            assert(forall |k: int| 0 <= k < v.len() ==>
+                #[trigger] w[k] == ((k as usize), &v[k]));
+        }
+    } => Ok(())
+}
+
+test_verify_one_file! {
```
parno, line 556: I'm not convinced this one is useful either.

parno (commented):
Thanks for the contribution!  I added some comments/suggestions.

parno:
I also notice that the [Rust docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#overflow-behavior-1) say:
> The method does no guarding against overflows, so enumerating more than [usize::MAX](https://doc.rust-lang.org/std/primitive.usize.html#associatedconstant.MAX) elements either produces the wrong result or panics. 
We might need to adjust your specs to account for that.

Issue #2914, opened 2026-09-07
Verus binary release omit Z3 after Vargo --> Cargo migration


y1ca1:
[The most recent binary release](https://github.com/verus-lang/verus/releases/tag/release%2F0.2026.09.06.8dea4a2) does not contain `z3` (or `z3.exe`).

On arm64 macOS, for example:

```console
$ unzip -Z1 verus-0.2026.09.06.8dea4a2-arm64-macos.zip | grep '/z3$'
# nothing
```

When user's `PATH` contains an incompatible solver (version != Z3 4.16.0), Verus would panicked in `air/src/parser.rs` with `IllegalChrInString`. This would also affect the CIs of various Verus projects that rely on Verus's binary release to contain z3.

### Likely cause

#2881 changed the release build from `vargo build` to ordinary Cargo commands. Vargo explicitly copied the selected solver into `target-verus/release`; the Cargo and `cargo-verus` build paths do not. The release workflow also archives `target-verus/release` unchanged.

The Z3 version metadata is also inconsistent:

- `source/tools/get-z3.sh` downloads 4.16.0;
- `tools/common/consts.rs` requires 4.16.0;
- `source/external-deps.toml` and the manifests for the August 23, August 30, and September 6 releases say 4.12.5.

At each of those release commits, both `get-z3.sh` and `EXPECTED_Z3_VERSION` were 4.16.0.

parno:
Fixed by #2915.  Thanks!

PR #2915, opened 2026-09-07
Fix Z3 packaging in binary release

y1ca1:
#2881 changed release builds from Vargo to Cargo. Vargo copied Z3 into `target-verus/release` as part of its build process, but ordinary Cargo and `cargo-verus` builds omit this step and the `0.2026.09.06.8dea4a2` binary release was published without Z3.

This PR changes the release workflow to package `z3`/`z3.exe` after the Cargo build. It keeps external dependency packaging in the release workflow rather than adding a side effect to ordinary Cargo builds. But there might be better ideas.

Also corrected stable manifests' corresponding z3 versions.

Fixing #2914 .

 


mkovaxx:
@y1ca1 Thanks for this fix!

mkovaxx (approved):
Looks good, thanks again! :)

mkovaxx:
Follow-up PR to consolidate versions of Z3 and other external deps: https://github.com/verus-lang/verus/pull/2916

PR #2917, opened 2026-09-08
fix the other known opaque types issue

### LINKED ISSUE #2323: Crash when passing Future to function with additional bound

tjhance:
(depends on https://github.com/verus-lang/verus/pull/2889)

fixes https://github.com/verus-lang/verus/issues/2323

In this case, the problem is that the call to `opaque_type` from the `rustc_trait_selection` crate depends on mir_borrowck. Replacing this with a call to `opaque_type_hir_typeck` (which is okay during the hir_typeck phase) solves the issue.

Unfortunately, forking `rustc_trait_selection` also requires us to fork several crates that depend on it, particularly `rustc_hir_typeck` and `rustc_traits`. So this adds a total of 3 crates. (But we wanted to fork rustc_hir_typeck anyway.)

TODO:
- [ ] tests
- [ ] update the updater script


tjhance:
Also TODO: see if there are any problems with opaque types from imported crates

PR #2919, opened 2026-09-10
vstd: allow char::is_whitespace in spec

Marti2203:
Prompted by a post-merge comment (https://github.com/verus-lang/verus/pull/2683#issuecomment-...) on #2683 from @wood-ghost: char::len_utf8's assume_specification carries #[verifier::allow_in_spec], but no test actually exercises calling it directly in spec position (assert('a'.len_utf8() == 1)) — every existing test only calls it in exec code and asserts on the resulting exec-bound value. Removing allow_in_spec entirely still passed all 4 pre-existing char.rs tests.

Fix:
- Added #[verifier::allow_in_spec] to char::is_whitespace too (wood-ghost's second suggestion) — it didn't have it before, even though its spec equivalent (is_white_space) is the same shape as len_utf8's.
- Added two regression tests calling both directly inside assert(...) in a proof fn, confirmed to fail to compile without the respective allow_in_spec attribute and pass with it.

Verified: vstd still verifies fully (2058/0); full char.rs suite (7/7, including the two new tests) passes.

### DIFF
--- a/source/rust_verify_test/tests/char.rs
+++ b/source/rust_verify_test/tests/char.rs
@@ -59,34 +59,6 @@ test_verify_one_file! {
     } => Err(err) => assert_one_fails(err)
 }
 
-// #[verifier::allow_in_spec] lets these be called directly in spec position, not just
-// through an exec `let` first - the tests above don't exercise that.
-test_verify_one_file! {
-    #[test] test_char_len_utf8_allow_in_spec verus_code! {
-        use vstd::prelude::*;
-
-        proof fn test() {
-            assert('a'.len_utf8() == 1); // ASCII, U+0061
-            assert('\u{a3}'.len_utf8() == 2); // £, U+00A3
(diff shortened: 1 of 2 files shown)


parno:
> no test actually exercises calling it directly in spec position (assert('a'.len_utf8() == 1)) — every existing test only calls it in exec code and asserts on the resulting exec-bound value.

Thanks for trying to improve the test suite, but I don't think we need this style of test, since it seems to just be testing whether `allow_in_spec` is working, which is already covered in other tests.

Adding `#[verifier::allow_in_spec]` to `char::is_whitespace` seems reasonable though.

Marti2203:
Hi @parno ,
I'll remove the test for `allow_in_spec` usage for `len_utf8`. I made a small tool to check how many specs are not covered, so I'd be happy to trim the PR.

parno (approved):
Thanks for the update!

PR #2927, opened 2026-09-11
Take mutability into account in boundary_suggestions

### LINKED ISSUE #2925: Suggested `assume_specification`s don't take mutability into account
In the following example:

```rust
use vstd::prelude::*;

verus! {

fn verified_caller() {
    let mut x = 42;
    unverified_function(&mut x);
}

}

fn unverified_function(x: &mut u64) {
    *x += 42;
}

```
[playground](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=b1cb2b2a593d9bfff6f27f1b86d5c91a)

verus suggests the following specification:

```
error: cannot use function `main::unverified_function` which is ignored because it is either declared outside the verus! macro or it is marked as `external`.
 --> /playground/src/main.rs:7:5
  |
7 |     unverified_function(&mut x);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
          assume_specification [crate::unverified_function] (_0: &u64);
```

This is missing a `mut`, resulting in another error when using the suggested specification.

dschoepe:
Fixes #2925.

Assisted-By: codex:openai.gpt-5.6-sol

 

### DIFF
--- a/source/rust_verify_test/tests/boundary_suggestions.rs
+++ b/source/rust_verify_test/tests/boundary_suggestions.rs
@@ -232,22 +232,7 @@ test_verify_one_file! {
            where
            'c: 'a + 'b,;")
 }
-test_verify_one_file! {
-    #[test] test_assume_specification_region_outlives_correct code! {
-        fn foo<'a, 'b, 'c, A, B>(a: &'a A, b: &'b B) -> &'c A
-        where 'c: 'a + 'b {
-            panic!()
-        }
-        verus! {
-            assume_specification<'a, 'b, 'c, A, B> [crate::foo] (_0: &'a A, _1: &'b B) -> &'c A
-            where
-            'c: 'a + 'b,;
-            pub fn bar<'a, 'b, 'c, A, B>(a: &'a A, b: &'b B) -> &'c A {
-                foo(a, b)
-            }
-        }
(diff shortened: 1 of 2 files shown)


source/rust_verify_test/tests/boundary_suggestions.rs
```diff
@@ -123,6 +123,41 @@ test_verify_one_file! {
...
+        verus! {
+            fn verified_caller() {
+                let mut x = 42;
+                unverified_function(&mut x);
+            }
+        }
+    } => Err(err) => assert_help_error_msg(err, "assume_specification [crate::unverified_function] (_0: &mut u64);")
+}
+test_verify_one_file! {
+    #[test] test_assume_specification_mut_ref_suggestion_correct code! {
```
tjhance, line 143: do we really need this test? (honestly this is niche enough i'm not sure we need any tests at all, but i definitely don't think we need this one)

tjhance (changes requested):

source/rust_verify_test/tests/boundary_suggestions.rs
dschoepe: Good point, removed.

tjhance (approved):

PR #2929, opened 2026-09-11
Use `ExactSizeIterator::exact_len` instead of `iter_len` 

jaylorch:
Before this PR, it was impossible to implement `ExactSizeIterator`. @parno suggested fixing that by making `iter_len` a trait function, and I suggested renaming it to be `exact_len` to reflect this.

This PR makes that change, and adds a test example showing how to implement `ExactSizeIterator` for a simple `Vec` iterator.

It was written with the aid of GPT-5.6 Sol; most notably, Sol wrote the entire test case.


examples/iterators/vec_iterator.rs
```diff
@@ -0,0 +1,148 @@
...
+        self.front <= self.back <= self.values.len()
+    }
+}
+
+impl<'a, T> MyVecIterator<'a, T> {
+    pub closed spec fn exact_len_spec(&self) -> usize {
+        (self.back - self.front) as usize
+    }
+
+    pub closed spec fn peek_front(&self, index: int) -> Option<&'a T> {
```
parno: Typically `peek_{front,back}` should be `open`, since they're used to reason about the particular value you're operating on inside a `for` loop.  If they're closed, they don't provide any useful info.

parno: Similarly, I'm not sure it's useful to make `exact_len_spec` closed.

source/vstd/std_specs/iter.rs
```diff
@@ -284,23 +284,35 @@ pub trait ExDoubleEndedIterator : Iterator {
...
 #[verifier::external_trait_specification]
+#[verifier::external_trait_extension(ExactSizeIteratorSpec via ExactSizeIteratorSpecImpl)]
 pub trait ExExactSizeIterator: Iterator {
     type ExternalTraitSpecificationFor: ExactSizeIterator;
 
+    // An `ExactSizeIterator` is an iterator that knows its exact
+    // length. This lets it have a non-prophetic spec function that
+    // produces a value equal to the prophetic
+    // `self.remaining().len()`. That non-prophetic spec function is
+    // `exact_len`.
```
parno: Can we condense this to the following?
> An`ExactSizeIterator` can specify its length non-prophetically (without using `self.remaining().len()`)

examples/iterators/vec_iterator.rs
parno, line 1: This looks very similar to the VecIterator in `examples/guide/iterators.rs`.  Maybe we should just move the new `ExactSizeIterator` material over there.  Even though it won't show up in the guide, it will still test that the new spec can be implemented.

source/vstd/std_specs/iter.rs
```diff
@@ -284,23 +284,35 @@ pub trait ExDoubleEndedIterator : Iterator {
...
-            self.obeys_prophetic_iter_laws() ==> len as int == iter_len(self) == self.remaining().len();
+            self.obeys_prophetic_iter_laws() ==> len == self.exact_len() == self.remaining().len();
 }
 
+// A type implementing the `ExactSizeIteratorSpec` trait must
+// implement a `len` function that returns something that's
+// simultaneously equal to both `self.exact_len()` and
+// `self.remaining().len()`. Thus, it's reasonable to have this axiom
+// saying that those two expressions are always equal.
+pub broadcast axiom fn axiom_exact_len_exact<I: ExactSizeIteratorSpec>(i: &I)
```
parno: Do we need this axiom?  Or is it enough to reason from the postcondition for `len`?  I.e., do you need this info if you haven't called (and won't call) `len`?

examples/iterators/vec_iterator.rs
jaylorch: I agree that `peek_{front,back}` should be `open`. So I've made that change. We do need to have it refer to a closed spec in it somewhere, to let us hide the details of the implementation (`front` and `back`). I've rewritten `peek_{front,back}` to be open and yet still hide the internals. I now have a (closed) `exact_remaining` spec that gives a non-prophetic sequence that's provably equal to `remaining()`.

source/vstd/std_specs/iter.rs
jaylorch: Sure, done.

source/vstd/std_specs/iter.rs
jaylorch: Maybe. Let's give it a try. I've removed the axiom. In my example, I now have a custom proof that users of the iterator can use to learn the correspondence between exact values and prophetic values.

examples/iterators/vec_iterator.rs
jaylorch: Seems like a good idea. Are you happy with the way the example is now, before I merge it into the guide version?

examples/iterators/vec_iterator.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+impl<'a, T> MyVecIterator<'a, T> {
+    pub closed spec fn exact_remaining(&self) -> Seq<&'a T>
+    {
+        self.values@.subrange(self.front as int, self.back as int).as_ref()
+    }
+
+    pub open spec fn peek_front(&self, index: int) -> Option<&'a T> {
+        if self.exact_remaining().len() > index {
```
parno: Since `index` is an `int`, I think we want `0 <= index < self.exact_remaining().len()`, here and in `peek_back`.

parno (commented):
One more minor comment on the example, but otherwise it looks good.

source/vstd/std_specs/iter.rs
parno: Thanks!

examples/iterators/vec_iterator.rs
parno: Seems reasonable to me.

examples/iterators/vec_iterator.rs
jaylorch: Ah, right, good point. Done.

examples/iterators/vec_iterator.rs
jaylorch: OK, I've integrated it into the guide version.

jaylorch: Oh, hmm, upon reading the guide I see that `peek_next` and `peek_back` are supposed to index from the very front and very back, not from the current position. That's kind of weird, right? Shouldn't `peek_next(0)` tell you the next one you'll get from `next`, not the first one you ever got from `next`?

jaylorch: My versions of `peek_x` index from the current position, so I should change them to be consistent with the rest of the guide and the comments in `iter.rs`. Or, alternately, we should change the rest of the guide and the comments in `iter.rs` to have the `peek_x` be relative to the current position. I'm inclined to do the latter; what do you think?

jaylorch: Oh, never mind, it was Sol who invented `peek_next`; that's not in `vstd` or the existing example. I'll get rid of that.

parno (approved):

PR #2935, opened 2026-09-14
fix ICE when projection normalization leaves inference variables

### LINKED ISSUE #2350: Verus panic when implementing traits with associated types for `Fn` types
Verus crashes when implementing a trait with an associated type for a function-like type `F: Fn(u8) -> T`.

A very similar version using a generic trait parameter compiles successfully, so this seems specific to vir lowering for associated-types.

### Minimal reproducer

```rust

use vstd::prelude::*;

verus! {

// compiles

trait Bar<T> {

    fn b(&self, i: u8) -> T;
}

impl<T, F> Bar<T> for F
where
    F: Fn(u8) -> T,
{
    fn b(&self, i: u8) -> T {
        (self)(i)
    }
}

// panics

trait Foo {

    type T;

    fn f(&self, i: u8) -> Self::T;
}

impl<T, F> Foo for F
where
    F: Fn(u8) -> T,
{
    type T = T;

    fn f(&self, i: u8) -> Self::T {
        (self)(i)
    }
}

}
```

```
thread 'rustc' (26) panicked at rust_verify/src/rust_to_vir_base.rs:1256:25:
assertion failed: !matches!(t.kind(), TyKind::Infer(..))
```

Verus version: 2026.04.16 4bdefc3

[Playground link](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=a683243ecaaf5b9391cc349c4ed061ab).

Marti2203:
Fixes #2350 
`rust_to_vir_base`'s `TyKind::Alias` arm normalizes an associated-type projection and then asserts the result contains no inference variables:

```rust
if norm.value != *ty {
    for arg in norm.value.walk().into_iter() {
        if let Some(t) = arg.as_type() {
            assert!(!matches!(t.kind(), TyKind::Infer(..)));
        }
    }
    return t_rec(&norm.value);
}
```

When an associated type is fixed only through a where-clause bound, normalization succeeds but returns a bare inference variable, and Verus panics:

```
thread 'rustc' panicked at rust_verify/src/rust_to_vir_base.rs:1266:33:
assertion failed: !matches!(t.kind(), TyKind::Infer(..))
```

### Minimal example

`Adapt<I>`'s `Item` is pinned only by the `I: Src<Item = &'a X>` bound, so `<Adapt<I> as Src>::Item` cannot be normalized to a concrete type. This is the added regression test; on `main` it ICEs with the assertion above.

```rust
use vstd::prelude::*;

#[verifier::external]
pub trait Src { type Item; }

#[verifier::external]
pub struct Adapt<I>(I);

#[verifier::external]
impl<'a, I, X: 'a> Src for Adapt<I>
    where I: Src<Item = &'a X>, X: Copy,
{
    type Item = X;
}

#[verifier::external_trait_extension(SrcSpec via SrcSpecImpl)]
pub trait ExSrc {
    type ExternalTraitSpecificationFor: Src;
    type Item;
    spec fn items(&self) -> Seq<Self::Item>;
}

#[verifier::external_body]
#[verifier::reject_recursive_types(I)]
pub struct ExAdapt<I>(Adapt<I>);

impl<'a, I, X: 'a> SrcSpecImpl for Adapt<I>
    where I: Src<Item = &'a X> + SrcSpec, X: Copy,
{
    closed spec fn items(&self) -> Seq<Self::Item> { Seq::empty() }
}
```

Instrumenting the arm on this input shows the assertion's premise is simply false — normalizing `<Adapt<I> as Src>::Item` yields `?1t`, and it is still `?1t` after `resolve_vars_if_possible`:

```
ty = <Adapt<I> as Src>::Item   norm = ?1t   resolved = ?1t   has_infer = true
```

I hit this writing a vstd spec for `Iterator::copied`, whose impl has to match std's `impl<'a, I, T: 'a> Iterator for Copied<I> where I: Iterator<Item = &'a T>`, putting `Self::Item` in exactly this position. A simpler version of the example that doesn't go through `external_trait_extension` doesn't reproduce it.

If anything is unresolved at the end, it falls through to the projection representation immediately below instead of asserting.

With the fix, the example above reports a normal rustc error (`E0275`, overflow evaluating `<I as Src>::Item == &<Adapt<I> as Src>::Item`) instead of crashing.

### Testing

- The regression test uses local stand-ins for `Iterator`/`Copied` rather than depending on vstd. It fails on unpatched Verus with the ICE above and passes with the fix.
- vstd verifies unchanged: 2059 verified, 0 errors.
- No regressions across 514 tests in `external_traits`, `assoc_type_impls`, `external_type_specification`, `traits`, `generics`, `iterators`, `regression`, `std` and `recursive_types`.

Assisted-by: Claude Code:Opus-5.


tjhance:
does this also fix https://github.com/verus-lang/verus/issues/2350 ?

Marti2203:
Hi @tjhance ,
I tried the repro from the referenced issue, and it no longer panics, so this should be the same fix.

tjhance (approved):

tjhance:
thanks!

PR #2937, opened 2026-09-15
Fix crate version bumping to account for lack of `vstd/src`

jaylorch:
I noticed that the `vstd` crate wasn't getting updated even when there were changes to it. GPT 5-6 Sol diagnosed the issue:

In `source/tools/bump_crate_versions/src/main.rs`, in `src_modified`, it looks for changes to `src`. However, unlike say, `builtin` and `builtin_macros`, `vstd` doesn't have a `src` subdirectory. So it always detects `vstd` as unchanged. This means that the `vstd` crate will only get bumped when one of its dependencies changes, not when it itself changes. This has been going on a while; recent history reveals that every `vstd/Cargo.toml` bump has coincided with a change to at least one of `builtin`, `builtin_macros`, or `state_machines_macros`.

This PR fixes that issue. It was written by GPT 5-6 Sol with light guidance from me.


parno (approved):
Looks reasonable to me.  Thanks for the fix!

PR #2939, opened 2026-09-15
Make distinct string literals provably distinct

julesmt:
Distinct string literals cannot be proved to have distinct views:
`"hello"@ != "world"@` is not derivable, even with `reveal_strlit`, which
makes string literals unusable as spec `Map` keys.

Design sketched by @Chris-Hawblitzel in [this Zulip thread](https://verus-lang.zulipchat.com/#narrow/channel/399078-help/topic/proving.20.22a.22.40.20.21.3D.20.22b.22.40.20for.20string.20literals/with/624433658):
every literal already has a unique integer id in the SMT encoding. This PR exposes that constructor as a builtin and adds a vstd broadcast axiom mapping each literal's view back to its id. Distinct ids then force distinct views by congruence. 

---

*Written with the help of Fable5 (AI).*


dschoepe:
Isn't this a side effect of sequence reasoning more generally and can be discharged by giving a witness of an index where they differ?

```rust
use vstd::prelude::*;

verus! {

proof fn foo() {
    reveal_strlit("hello");
    reveal_strlit("world");
    assert("hello"@ != "world"@) by {
        assert("hello"@[0] != "world"@[0]);
    }
}
}
```
https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=d4f7686930904d8cfd4b7cc6e20a3704

julesmt:
> Isn't this a side effect of sequence reasoning more generally and can be discharged by giving a witness of an index where they differ?
> 
> ```rust
> use vstd::prelude::*;
> 
> verus! {
> 
> proof fn foo() {
>     reveal_strlit("hello");
>     reveal_strlit("world");
>     assert("hello"@ != "world"@) by {
>         assert("hello"@[0] != "world"@[0]);
>     }
> }
> }
> ```
> 
> https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=d4f7686930904d8cfd4b7cc6e20a3704

It's what I was doing before this PR but it doesn't scale: it's a quadratic (in nb of literals) proof text, and the index witnesses break when names change. The axiom replaces all of it with one `broadcast use` making the inequalities ambiant facts. More generally I find it cleaner.

PR #2940, opened 2026-09-15
Various fixes for more complex logically atomic calls

parno:
In one of my projects, I ran into four different issues using the new logically atomic framework.  To keep things simpler, each commit in this PR includes a (previously failing) test to illustrate the problem, along with a small fix for it.  Hence it may be simpler to review commit-by-commit.

Assisted-by: Codex:Sol


slerpyyy:
You really made this as comfortable as possible for me to review; I appreciate it!

Though I do have one remark, I believe it's not necessary to add lifetimes to the atomic update predicate. The predicate type does not contain the actual function arguments; it only holds a ghost copy of them for the atomic pre- and postcondition to use, which does not require lifetime tracking.

Even if it did, if a logically atomic function takes a `&'a T` as an argument, then `'a` is guaranteed to be valid for the duration of the function, and we already force users to prove that the atomic update resolves before the function returns, meaning the AU can never outlive `'a` anyway.

I'm bringing this up because adding lifetimes to the AU can make it impossible to put them into an atomic invariant that outlives the function, as is commonly done in wait-free data structures (see the [section on helping](https://verus-lang.github.io/verus/guide/logatom.html#notes-on-helping)).

source/rust_verify_test/tests/logatom.rs
```diff
@@ -71,6 +71,97 @@ test_verify_one_file! {
...
+        use vstd::prelude::*;
+        use vstd::atomic::*;
+
+        tracked struct Token;
+
+        fn atomic_array<const N: usize>(_value: [u8; N])
+            atomically (atomic_update) {
+                (old: Token) -> (new: Commit<Token>),
+                ensures new@ == old,
+            },
```
slerpyyy: This test would be slightly more interesting if it used `N` in the atomic spec
```suggestion
        pub tracked struct Token { pub len: usize }

        pub fn atomic_array<const N: usize>(_value: [u8; N])
            atomically (atomic_update) {
                (old: Token) -> (new: Commit<Token>),
                requires old.len == N,
                ensures new@ == old,
            },
```

parno: Good point!

parno:
> I believe it's not necessary to add lifetimes to the atomic update predicate. 

Hmm, interesting.  I think you're right.  Although, we still need the lifetime in scope, since the implementations mention `&'a T`.  I'll try pushing a fix that keeps the scope without adding it to the predicate.

parno:
I think [524eda8](https://github.com/verus-lang/verus/pull/2940/commits/524eda84a3b76a2ab6cd94d0d65b5536d6f7fc0e) should keep the parameter in scope without actually being treated as a borrow.

slerpyyy:
Ah, I see the problem. I think ideally, we'd remove all lifetimes from the argument types so that the predicate type never needs any lifetimes, but I can imagine this would be a rather complicated transformation, especially with user-defined types that have lifetime parameters. That can be a future PR.

parno:
@slerpyyy CI is finally green, but I'd appreciate it if you can double check the final commit (b8ad72128c4603576959c0c62c34492)

slerpyyy:
This works, but I think it makes the API a bit more surprising: this change doesn't make Rust ignore the lifetimes; it makes the predicate type contravariant in all its lifetime parameters, so the constraints are still there, just in the other direction. This does mean you can convert any `Pred<'a>` to a `Pred<'static>` to "remove" the lifetime parameter, but this is a "trick" users must discover and use consciously, and I don't think it's very obvious.

parno:
> but this is a "trick" users must discover and use consciously, and I don't think it's very obvious.

Is there another approach we can take?

slerpyyy:
I'm still holding onto the idea of removing all lifetime parameters from the predicate type, but I need to work through that and figure out if it's feasible. Either way, that would be a separate change; no need to hold up the PR over it.

slerpyyy (approved):

parno:
Okay, that makes sense.  Thanks for taking a look at it.

PR #2945, opened 2026-09-16
Fix panic for #[verifier::type_invariant] on a type with a lifetime and a type param

### LINKED ISSUE #2898: Verus/rustc panic for `#[type_invariant]` on type with lifetime and generic
Trying to verify the following:
```rust
use vstd::prelude::*;

verus! {
    pub struct X<'a, T> {
        r: &'a T,
    }

    #[verifier::type_invariant]
    pub closed spec fn inv<'a, T>(x: X<'a, T>) -> bool {
        true
    }
}
```
with e.g. `verus --crate-type=lib mwe.rs` will crash Verus (tested on latest `main` and on the [playground](https://play.verus-lang.org/?version=stable&mode=basic&edition=2021&gist=3033acccf21545983826542d25fffe60)).
I get the following panic message:
```
expected type for `T/#0` ... but found Lifetime(...::inv::'a)
when instantiating, args=['a, T]
```
The following works fine, however:
```rust
use vstd::prelude::*;

verus! {
    pub struct X<'a, T> {
        r: &'a T,
    }

    impl<'a, T> X<'a, T> {
        #[verifier::type_invariant]
        pub closed spec fn inv(self) -> bool {
            true
        }
    }
}
```
The crashing form also works fine if the struct is parametrized only by a lifetime, or only by a generic type.

Marti2203:
Fixes #2898.

`check_generics_for_invariant_fn` reused the datatype's own generic args (`substs`, taken from the invariant function's parameter type, e.g. `x: X<'a, T>`) positionally as the substitution for the invariant function's own predicates — which only works when both lists have the same arity and order.

They can differ: a lifetime that appears only in a function's signature (never constrained by a where-clause) is "late-bound" and excluded from that function's own generics_of, while a datatype's own lifetime parameters are always early-bound. So `fn inv<'a, T>(x: X<'a, T>)` gets a 1-element generics_of ([T]), but `substs` (from X's own arity) has 2 elements (['a, T]) — feeding the mismatched list into `func_predicates.instantiate` panics rustc trying to substitute a lifetime into a type parameter's slot.

Confirmed via instrumentation: dumping `tcx.generics_of` for both the struct and the function shows exactly this — the struct's own_params is `['a: Lifetime, T: Type]`, the function's is only `[T: Type]`.

Fix: `extracted substs_for_own_generics`, which re-keys substs by each parameter's own index (each entry is already expressed in terms of the function's own `'a/T identities`, via `ParamTy::index`/`EarlyParamRegion::index`) rather than assuming position N in one list means position N in the other — and walks the function's parent chain (e.g. an enclosing impl block) too, since a method's own generics_of excludes its impl's, even though its predicates can still reference the impl's parameters (found this the hard way — an initial version handling only the function's own generics_of, with no parent-chain walk, broke real vstd invariant methods with a different instantiate panic).

Verified:
- vstd verifies unchanged (2059 verified, 0 errors)
- The issue's own repro no longer panics; reverting just this fix reproduces the exact original panic
- The issue's own already-working variants (impl-based form, lifetime-only struct, type-only struct) still work
- New regression test in user_defined_type_invariants.rs (86/86); confirmed to fail with the original panic without the fix
- No regressions across exec_closures, expand_errors, shr_ref_struct_wrap, iterators, proof_closures, consts (the type-invariant-touching test files), plus traits, generics, recursive_types, assoc_type_impls (475 tests total)


source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -2376,6 +2376,66 @@ fn param_names_for_async_func<'tcx>(
...
+        chain.push(g);
+        cur = g.parent;
+    }
+    chain.reverse();
+
+    let mut result = Vec::new();
+    for g in chain {
+        for param in &g.own_params {
+            let arg = arg_by_index.get(&param.index).unwrap_or_else(|| {
+                panic!(
```
tjhance: let's use `internal_error!` for this

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -2376,6 +2376,66 @@ fn param_names_for_async_func<'tcx>(
...
+/// `x: X<'a, T>`) to `target_id`'s own generics_of, which can differ in length/order: an
+/// unconstrained lifetime is "late-bound" and excluded from a function's generics_of,
+/// but never from a datatype's. Matches by each parameter's own index, and walks the
+/// parent chain (e.g. an enclosing impl) too, since a method's generics_of excludes its
+/// impl's own params.
+///
+/// Panics if a parameter isn't found in `substs` - should be unreachable, since the
+/// caller's earlier check_item_external_generics call already requires `substs` to be
+/// bare references to `target_id`'s own params (rejecting e.g. `X<Vec<T>>`).
+fn substs_for_own_generics<'tcx>(
```
tjhance: can we come up with a better name? like `args_for_early_binder_instantiation`

source/rust_verify/src/rust_to_vir_func.rs
```diff
@@ -2376,6 +2376,66 @@ fn param_names_for_async_func<'tcx>(
...
+    tcx: TyCtxt<'tcx>,
+    target_id: DefId,
+    substs: GenericArgsRef<'tcx>,
+) -> GenericArgsRef<'tcx> {
+    let mut arg_by_index: HashMap<u32, GenericArg> = HashMap::new();
+    for arg in substs.iter() {
+        match arg.kind() {
+            GenericArgKind::Lifetime(r) => {
+                if let RegionKind::ReEarlyParam(early) = r.kind() {
+                    arg_by_index.insert(early.index, arg);
```
tjhance: can you assert that these map inserts don't overwrite anything

tjhance (approved):

tjhance (changes requested):
whoops meant to hit 'request changes'. See the above comments.

source/rust_verify/src/rust_to_vir_func.rs
Marti2203: Updated

source/rust_verify/src/rust_to_vir_func.rs
Marti2203: Sure, sounds good!

source/rust_verify/src/rust_to_vir_func.rs
Marti2203: added an assert!

PR #2951, opened 2026-09-18
Treat identical VIR casts as identity

tangruize:
## Summary

Preserve the translated source expression when rustc has already applied an implicit coercion to the source of an explicit cast and the resulting source and destination VIR types are equal. Non-identical casts continue through the existing cast validation and lowering paths.

## Motivation

Rust may represent an explicit trait-object cast with an implicit coercion attached to its source expression. Verus translates that source adjustment before processing the explicit cast, so the source has already become the requested trait-object type by the time the cast is classified. The remaining cast is therefore an identity operation in VIR, but the current explicit-cast path still rejects it as an unsupported cast.

## Minimal reproduction

```rust
use vstd::prelude::*;

verus! {

trait T {
    fn f(&self);
}

impl T for u8 {
    fn f(&self) {}
}

fn test(value: &u8) {
    let value: &dyn T = value as _;
    value.f();
}

}
```

Verus reports:

```text
error: Verus does not support this cast: `&'{erased} dyn [Binder { value: Trait(T), bound_vars: [] }] + '{erased}` to `&'{erased} dyn [Binder { value: Trait(T), bound_vars: [] }] + '{erased}`
  |
  |     let value: &dyn T = value as _;
  |                         ^^^^^^^^^^
```

The diagnostic shows that both sides have already lowered to the same trait-object reference type.

## Change

The explicit `ExprKind::Cast` translation now compares the complete translated source and destination types using the existing VIR type-equivalence operation after source adjustments have been applied. If those types are equal, the frontend returns the translated source expression directly because no cast operation remains to encode.

All casts with non-equal VIR types retain the previous behavior and continue through pointer-cast handling, numeric conversion, enum conversion, float conversion, or the existing unsupported-cast diagnostic. This applies only when the complete VIR source and destination types are equal.

Assisted-by: GitHub Copilot:gpt-5.6-sol


source/rust_verify/src/rust_to_vir_expr.rs
```diff
@@ -2486,6 +2486,11 @@ pub(crate) fn expr_to_vir_innermost<'tcx>(
 
             let source_vir_ty = &source_vir_expr.typ;
             let to_vir_ty = expr_typ()?;
+            // rustc may apply an implicit coercion to the source of an explicit cast,
+            // leaving the cast itself with identical source and destination VIR types.
+            if types_equal(source_vir_ty, &to_vir_ty) {
+                return Ok(ExprOrPlace::Expr(source_vir_expr));
+            }
```
tjhance, line 2492: can you move this check up a bit, before the `maybe_do_ptr_cast`? I think it makes sense for this to just be the first check we do.

For the comment, write something like:

```
// If the source and dest type are the same, we don't need to do anything.
// Also note this case can show up unexpectedly;
// for example, in: `fn test(value: &u8) { let value = value as &dyn T; ... }`
// this appears to be a non-trivial cast, but rustc actually inserts an
// explicit coercion to the source, making the explicit as-coercion a trivial coercion.
```

tjhance (changes requested):

source/rust_verify/src/rust_to_vir_expr.rs
tangruize: Thanks! Fixed in https://github.com/verus-lang/verus/pull/2951/commits/5cffc0c892f96738f54a9266d39ede3fbef12cb3

tjhance (approved):

PR #2952, opened 2026-09-18
Support value-bearing break expressions

tangruize:
## Summary

Support Rust loop expressions that return a value through `break`, including labeled exits, nested loops, non-`Copy` results, and break operands with side effects.

## Motivation

Rust permits a `loop` expression to produce a value from its terminating `break`, but Verus currently rejects every break carrying an operand.

## Minimal reproduction

```rust
use vstd::prelude::*;

verus! {

fn test() {
    let value = loop {
        break 5u64;
    };
    assert(value == 5);
}

}
```

Verus reports:

```text
error: The verifier does not yet support the following Rust feature: complex break expressions
  |
  |         break 5u64;
  |         ^^^^^^^^^^
```

## Change

The Rust-to-VIR translation now carries the optional break operand through `ExprX::BreakOrContinue`. Continue expressions and operand-free breaks retain their existing representation and behavior.

When a loop produces a value, VIR-to-SST lowering creates a hidden destination for the loop result. A value-bearing break evaluates its operand exactly once, initializes that destination immediately before the targeted exit, and the loop expression returns the resulting value after control leaves the loop.

Result flow across labeled and nested exits remains in one AIR control-flow scope so the value assignment is visible after the target loop. Mode checking, AST traversal, and resolution inference now also visit the optional operand.

Assisted-by: GitHub Copilot:gpt-5.6-sol


source/vir/src/ast_to_sst.rs
```diff
@@ -2617,7 +2633,12 @@ pub(crate) fn expr_to_stm_opt(
             atomic_call,
         } => {
             let is_for_loop = *is_for_loop;
-            let loop_isolation = *loop_isolation;
+            let produces_value =
+                !crate::ast_util::is_unit(&expr.typ) && !crate::ast_util::is_never(&expr.typ);
+            let carries_outer_result = has_break_to_any_label(&state.result_flow_labels, expr);
+            let requires_result_flow = produces_value || carries_outer_result;
+            // Keep the result assignment and every intervening labeled break in one AIR query.
+            let loop_isolation = *loop_isolation && !requires_result_flow;
```
tjhance: we shouldn't be silently changing the loop_isolation behavior on users. If we don't support a value-bearing `break` statement with loop_isolation=true, then we should halt with an error in that case.

Then I think you can also cut `carries_outer_result`; you shouldn't need any special consideration for in-between loops, as it's already the case that you can't break to a loop_isolation=false loop with a loop_isolation=true loop in between.

source/vir/src/ast_to_sst.rs
```diff
@@ -2617,7 +2633,12 @@ pub(crate) fn expr_to_stm_opt(
             atomic_call,
         } => {
             let is_for_loop = *is_for_loop;
-            let loop_isolation = *loop_isolation;
+            let produces_value =
+                !crate::ast_util::is_unit(&expr.typ) && !crate::ast_util::is_never(&expr.typ);
+            let carries_outer_result = has_break_to_any_label(&state.result_flow_labels, expr);
+            let requires_result_flow = produces_value || carries_outer_result;
```
tjhance: put the call to `has_break_to_any_label` to the right of the `||` so we can skip it most of the time.

source/vir/src/ast_to_sst.rs
```diff
@@ -109,6 +109,8 @@ pub(crate) struct State<'a> {
     statics: IndexSet<Fun>,
     pub assert_id_counter: u64,
     loop_id_counter: u64,
+    loop_result_dests: HashMap<Label, VarIdent>,
+    result_flow_labels: HashSet<Label>,
```
tjhance: Why do we need both of these?

source/vir/src/ast_to_sst.rs
```diff
@@ -3580,10 +3625,24 @@ pub(crate) fn expr_to_stm_opt(
...
+            let mut stms = Vec::new();
+            if let Some(value) = value {
+                assert!(*is_break);
+                let (mut value_stms, value) = expr_to_stm_opt(ctx, state, value)?;
+                stms.append(&mut value_stms);
+                let value = match value {
+                    Maybe::Some(value) => value.to_exp(),
+                    Maybe::Never => return Ok((stms, Maybe::Never)),
+                };
+                if let Some(dest) = state.loop_result_dests.get(label) {
```
tjhance, line 3621: raise an `internal_error` if `get` returns `None` here

tjhance (changes requested):

source/vir/src/ast_to_sst.rs
tjhance: or actually I guess it might be None in the unusual case that it's typed unit? In that case it's fine as it is, maybe just add a comment to that effect

PR #2965, opened 2026-09-21
verification observer: interface to capture information about VC lowering

matthewbdwyer:
This PR provides infrastructure to observe the VC lowering process that applications can use to drive analysis and reporting.  Three consumers have been built against this interface: a VC lifting library (which is a forthcoming PR), a capability that renders proof-states on verification failure (another forthcoming PR), and a proof coverage collector which was built independently against the published interface.  None of these required additional hooks beyond what is defined by this PR and without this PR each such application would have to separately integrate with the core Verus VC lowering code.

The PR adds three small observer traits each with default method callbacks that fire at different points during lowering and receive relevant lowering data structures as arguments.  All three observers are registered when the VIR lowering context is created.  The two AIR related observers are additionally installed in `air::Context` through a setter, following Context's existing configuration methods.  Like AIR's existing `MessageInterface` and `Diagnostics`, they are trait objects through which AIR calls back into code it does not otherwise know about.

The production code in the PR consists of the observer traits and the callback wiring into core Verus.  The `VirObserver`, `AirObserver`, and `QueryResultObserver` define 19 callbacks and collectively comprise approximately 300 lines of code.  Just over 400 lines of callback wiring are spread over 13 files primarily in VIR and `AIR`, but also including support for a command line option to select an observer, used by the test suite and consumers described above.  

The only change to an existing API is that `block_to_assert::lower_query` is re-exported as `air::lower_query`.  Observers receive each structured AIR query before this lowering step and the re-export lets an observer produce the flattened AIR that the solver sees.  The proof-state consumer uses this.

The PR introduces substantial testing infrastructure (2000 lines) to enable validation of this framework independent of consuming applications.  To that end, a set of examples covering combinations of VC shapes (`source/rust_verify_test/tests/vc_shapes/mod.rs`) drives testing.  Independent test observers are defined for each of the observer traits as well as a composite `TestObserver` that jointly exercises them.  For all 17 VC-shape programs, a differential test runs Verus with and without observers and asserts identical verification results.  In addition, shape-specific tests run with the test observer enabled and assert the expected payload content for each callback.

An early version of this PR was written by hand. Agentic AI (Kiro CLI, Claude) was then used to rebase it onto current main, to implement most of the testing infrastructure, and to apply review-driven revisions.  Every line of AI modified code was read and reviewed by me.  The commit has the required Assisted-by: trailer.

 


source/air/src/air_observer.rs
```diff
@@ -0,0 +1,58 @@
+//! AIR-level verification pipeline observer.
+//!
+//! Defines `AirObserver` — the interface for observing the AIR-level
+//! verification pipeline: query lowering, WP version creation, and lambda /
+//! choose / axiom declarations.
+//!
+//! This trait is independent: it does **not** require any other observer
+//! trait. Query *results* are observed separately via `QueryResultObserver`
+//! (`query_result_observer.rs`), and VIR→AIR lowering via `VirObserver`
+//! (`vir/src/vir_observer.rs`). A consumer implements only the traits it needs.
```
tjhance, line 10: This doc comment has a bit of an "llm flavor" to them, in the way it address odd concerns that the reader probably doesn't have. (Why would I expect it to require another observer in the first place?)

Maybe a bit more about what an air observer does and why you'd use it.

source/air/src/closure.rs
```diff
@@ -418,6 +422,14 @@ fn simplify_choose(
             let closure_fun = Arc::new(name);
             ctxt.choose_count += 1;
             let _ = ctxt.choose_map.insert(closure, closure_fun.clone());
+            if let Some(obs) = &ctxt.air_observer {
+                let binder_name = binders
+                    .iter()
+                    .next()
+                    .map(|b| b.name.clone())
+                    .unwrap_or_else(|| Arc::new("_".to_string()));
+                obs.borrow_mut().on_choose_decl(&closure_fun, &binder_name);
```
tjhance, line 431: this seems to drop all but the first binder; is that intentional?

source/air/src/query_result_observer.rs
```diff
@@ -0,0 +1,66 @@
...
+//! `VirObserver`. A consumer that only cares about verification outcomes
+//! implements this trait alone.
+//!
+//! **`Invalid` is an interactive, live-window callback** — the one deliberate
+//! exception to "observers are passive." Alongside the counterexample model
+//! (`model_defs`), it hands the observer `eval_bool_expr`, a closure that
+//! evaluates a boolean AIR expression against the *live* solver model. The model is
+//! only valid for the duration of the callback: the core discovers the failing
+//! assertion, invokes the observer, and only then disables the assertion label
+//! (which invalidates the model); see `smt_check_assertion` in `smt_verify.rs`.
```
tjhance, line 17: same

source/vir/src/context.rs
```diff
@@ -135,10 +135,44 @@ pub struct Ctx {
     pub(crate) byte_string_hashes: RefCell<HashMap<BigUint, Arc<Vec<u8>>>>,
     // proof debug purposes
     pub debug: bool,
+    /// Observer for the verification pipeline. Three layers of indirection:
+    /// - `Rc`: shared between VIR `Ctx` (for lowering callbacks) and AIR
+    ///   `Context` (for query/result callbacks) which have independent lifetimes.
+    /// - `RefCell`: callbacks require `&mut` through shared `Rc` references.
```
tjhance, line 141: I think the explanation of how Rc and RefCell work aren't necessary

source/vir/src/context.rs
```diff
@@ -135,10 +135,44 @@ pub struct Ctx {
...
+    /// Ask the observer for an assert ID, or use the default (clone parent).
+    pub(crate) fn make_assert_id(
+        &self,
+        kind: &crate::vir_observer::AssertIdKind,
+        index: usize,
+        parent: &Option<air::ast::AssertId>,
+    ) -> Option<air::ast::AssertId> {
+        self.observer.as_ref().map_or_else(
+            || parent.clone(),
+            |obs| obs.borrow_mut().make_assert_id(kind, index, parent),
```
tjhance, line 172: Is this a workaround for the fact that Verus currently doesn't emit an AssertId for the loop invariant assertions?

I think it probably makes more sense to just always have every assert get an AssertID, rather than making it conditional on the existence of an observer (In fact, there's another PR (https://github.com/verus-lang/verus/pull/2682) for this)

source/vir/src/vir_observer.rs
```diff
@@ -0,0 +1,176 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AssertIdKind {
+    Ensures,
+    LoopInvariant,
+    DecreasesCheck,
+}
+
+/// Observer for the VIR-level verification pipeline.
+///
+/// Callbacks fire during VIR→AIR lowering. All methods have default no-op
```
tjhance, line 30: Specifically SST→AIR, right?

tjhance (commented):
This looks great! I'm looking forward to Verus having proof state.

PR #2966, opened 2026-09-21
Add Option contracts and reusable proof lemmas

nickgian:
This PR adds 27 new Option contracts and 28 proof lemmas, mined via an Agentic workflow and (lightly) validated using PBT. The current goal is to review and collect feedback on the generated contracts/lemmas rather than merging, hence marking this as draft.

Assisted-by: Codex:openai.gpt-5.6-sol


source/vstd/std_specs/option.rs
```diff
@@ -409,4 +578,771 @@ pub assume_specification<T>[ Option::get_or_insert ](option: &mut Option<T>, val
...
+
+// or_else
+pub assume_specification<T, F: FnOnce() -> Option<T>>[ Option::<T>::or_else ](
+    option: Option<T>,
+    f: F,
+) -> (res: Option<T>)
+    requires
+        option is None ==> f.requires(()),
+    ensures
+        option is Some ==> res is Some,
```
wood-ghost, line 628: It seems that checking only `res is Some` is not enough. According to the official [docs](https://doc.rust-lang.org/std/option/enum.Option.html#method.or_else), we should also check that `res` preserve the element when `option is Some`.

source/vstd/std_specs/option.rs
```diff
@@ -409,4 +578,771 @@ pub assume_specification<T>[ Option::get_or_insert ](option: &mut Option<T>, val
...
+        (option is None || other is None) ==> res is None,
+;
+
+// unzip
+pub assume_specification<T, U>[ Option::<(T, U)>::unzip ](option: Option<(T, U)>) -> (res: (
+    Option<T>,
+    Option<U>,
+))
+    ensures
+        option is Some ==> res == (Some((option->0).0), Some((option->0).1)),
```
wood-ghost: [Unzips an option containing a tuple of two options.](https://doc.rust-lang.org/std/option/enum.Option.html#method.unzip) Whether `option` contains a tuple is not checked here. IMHO using pattern matching here will be better.

wood-ghost (commented):
Hi! I have also implemented some option contracts in my pr #2817 so I quickly compared them. Since nobody has reviewed my pr yet, I can not guarantee the correctness of my suggestions. Overall, I'm a little bit curious on why ai model chose to check the type manually instead of using pattern matching in many method contracts, as I saw it using pattern matching in lemmas. 
btw testcases can be added to `source/rust_verify_test/tests/option.rs`, which will make it easier for human to review I think.

source/vstd/std_specs/option.rs
nickgian: Hey! Thanks for taking the time to review this, this was meant to communicate the current state rather than present a finished work, but your feedback is very welcome and useful, so thanks again!

To get back to your points:
* I suspect because some of the current contracts do not use pattern matching, AI doesn't feel obliged to stick to that paradigm, but it's probably something we should nudge it to do.
* On the `or_else`, indeed this postcondition is weak. The original contract did have the stronger postcondition you suggest, but subsequent refinements dropped it because it was deemed unsound. The (annoyingly contrived) example constructed was one where the destructor of the unused/dropped closure could mutate shared state with the value of the option, see: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=063ba16f9be4809c1b847559a0e6ed44

source/vstd/std_specs/option.rs
nickgian, line 688: Isn't that guaranteed by the types?

source/vstd/std_specs/option.rs
tjhance: I think this isn't a real counter-example, because Verus spec equality on cells does not compare their contents. (Otherwise, this kind of issue would be everywhere and Verus would be wildly unsound.)

tjhance: See: https://verus-lang.github.io/verus/guide/interior_mutability.html

source/vstd/std_specs/option.rs
wood-ghost: Yes, you’re right—I overlooked the Option<(T, U)> parameter type. Thanks!

source/vstd/std_specs/option.rs
```diff
@@ -215,6 +292,58 @@ pub assume_specification<T, U, F: FnOnce(T) -> U>[ Option::<T>::map ](a: Option<
...
 ;
 
+// inspect
+pub assume_specification<T, F: FnOnce(&T)>[ Option::<T>::inspect ](option: Option<T>, f: F) -> (res:
+    Option<T>)
+    requires
+        option is Some ==> f.requires((&option->0,)),
+    ensures
+        res is Some <==> option is Some,
+        res is None <==> option is None,
```
tjhance, line 302: these are redundant

source/vstd/std_specs/option.rs
```diff
@@ -215,6 +292,58 @@ pub assume_specification<T, U, F: FnOnce(T) -> U>[ Option::<T>::map ](a: Option<
...
+// map_or
+pub assume_specification<T, U, F: FnOnce(T) -> U>[ Option::<T>::map_or ](
+    option: Option<T>,
+    default: U,
+    f: F,
+) -> (res: U)
+    requires
+        option is Some ==> f.requires((option->0,)),
+    ensures
+        option is Some ==> exists|output: U| f.ensures((option->0,), output),
```
tjhance, line 315: Should probably be:

```
option is Some ==> f.ensures((option->0,), res)
option is None ==> res == default
```

source/vstd/std_specs/option.rs
```diff
@@ -215,6 +292,58 @@ pub assume_specification<T, U, F: FnOnce(T) -> U>[ Option::<T>::map ](a: Option<
...
+    option: Option<T>,
+    default: D,
+    f: F,
+) -> (res: U)
+    requires
+        option is None ==> default.requires(()),
+        option is Some ==> f.requires((option->0,)),
+    ensures
+        option is None ==> exists|output: U| default.ensures((), output),
+        option is Some ==> exists|output: U| f.ensures((option->0,), output),
```
tjhance, line 329: `res` instead of `output`

source/vstd/std_specs/option.rs
```diff
@@ -223,6 +352,34 @@ pub assume_specification<'a, T: Clone>[ Option::<&'a T>::cloned ](opt: Option<&'
...
+    no_unwind
+;
+
+// cloned (Option<&mut T>)
+pub assume_specification<'a, T: Clone>[ Option::<&'a mut T>::cloned ](
+    opt: Option<&'a mut T>,
+) -> (res: Option<T>)
+    ensures
+        opt.is_none() ==> res.is_none(),
+        opt.is_some() ==> res.is_some() && cloned::<T>(*opt.unwrap(), res.unwrap()),
```
tjhance, line 380: missing `*final(opt.unwrap()) == *opt.unwrap()`

source/vstd/std_specs/option.rs
```diff
@@ -409,4 +578,771 @@ pub assume_specification<T>[ Option::get_or_insert ](option: &mut Option<T>, val
...
 
+#[doc(hidden)]
+pub assume_specification<T, F: FnOnce() -> T>[ Option::<T>::get_or_insert_with ](
+    option: &mut Option<T>,
+    f: F,
+) -> (res: &mut T)
+    requires
+        *old(option) is None ==> f.requires(()),
+    ensures
+        *old(option) is None ==> f.ensures((), *res),
```
tjhance, line 589: `*old(option) is Some ==> *res == old(option).unwrap()`

(alternatively, `*old(option) matches Some(x) ==> *res == x`)

source/vstd/std_specs/option.rs
nickgian: > I think this isn't a real counter-example, because Verus spec equality on cells does not compare their contents. (Otherwise, this kind of issue would be everywhere and Verus would be wildly unsound.)

Yep, you are right, after looking at this closer the counterexample is asserting the wrong thing. Confidently wrong LLMs...

source/vstd/std_specs/option.rs
```diff
@@ -409,4 +578,771 @@ pub assume_specification<T>[ Option::get_or_insert ](option: &mut Option<T>, val
...
+    option: Option<Result<T, E>>,
+) -> (res: Result<Option<T>, E>)
+    ensures
+        option is None ==> res == Ok(None::<T>),
+        (option is Some && option->0 is Ok) ==> res == Ok(Some(option->0->Ok_0)),
+        (option is Some && option->0 is Err) ==> res == Err(option->0->Err_0),
+    no_unwind
+;
+
+////// Lemmas
```
tjhance, line 713: These lemmas feel like bloat to me, though they're a nice sanity check. Maybe suitable for tests?

nickgian: I agree, I don't see the point for lemmas since (so far) they seem trivial to prove anyway if you need them.

